PR #670, opened 2014-11-21
Add try except to is_admin check. Fix #645

### LINKED ISSUE #645: write_error excepts, and doesn't write an error if the User is invalid
See [here](https://github.com/biocore/qiita/blob/master/qiita_pet/handlers/base_handlers.py#L24), if `self.current_user` doesn't exist then the method will except

squirrelo:


squirrelo:
Can someone please merge this?

qiita_pet/handlers/base_handlers.py
```diff
@@ -25,10 +25,13 @@ def write_error(self, status_code, **kwargs):
...
             return
 
+        is_admin = False
         if self.current_user:
-            is_admin = User(self.current_user).level == 'admin'
-        else:
-            is_admin = False
+            try:
+                is_admin = User(self.current_user).level == 'admin'
+            except:
```
antgonza: Is this conversation closed? Seems that is not. @josenavas @squirrelo

squirrelo: It was closed outside of this thread. Essentially it's needed only because you guys wanted a link to the error page for admins, so this bulletproofs that,

antgonza: Thanks!

PR #732, opened 2014-12-18
Pick otus

josenavas:
The qiita_ware part is done. I'm only missing connecting it with the interface. https://github.com/biocore/qiita/pull/731 should be merged before this PR is merged (it is build on top of it)


qiita_pet/handlers/processing_handlers.py
```diff
@@ -0,0 +1,27 @@
...
+from qiita_ware.context import submit
+
+
+class ProcessHandler(BaseHandler):
+    @authenticated
+    def post(self):
+        study_id = int(self.get_argument('study_id'))
+        preprocessed_data_id = int(self.get_argument('preprocessed_data_id'))
+
+        # We currently do not support defining the parameters
```
antgonza: @josenavas is the solution ready? Can you add?

josenavas: No, the solution is not ready. I prefer to add it in another PR.

2015-01-01 10:48 GMT-07:00 Antonio Gonzalez <notifications@github.com>:

> In qiita_pet/handlers/processing_handlers.py
> <https://github.com/biocore/qiita/pull/732#discussion-diff-22400795>:
>
> > @@ -0,0 +1,27 @@
> > +from tornado.web import authenticated
> > +
> > +from .base_handlers import BaseHandler
> > +from qiita_ware.dispatchable import processor
> > +from qiita_db.parameters import ProcessedSortmernaParams
> > +from qiita_ware.context import submit
> > +
> > +
> > +class ProcessHandler(BaseHandler):
> > +    @authenticated
> > +    def post(self):
> > +        study_id = int(self.get_argument('study_id'))
> > +        preprocessed_data_id = int(self.get_argument('preprocessed_data_id'))
> > +
> > +        # We currently do not support defining the parameters
>
> @josenavas <https://github.com/josenavas> is the solution ready? Can you
> add?
>
> —
> <https://github.com/biocore/qiita/pull/732/files#r22400795>.
>



-- 
Jose Navas

antgonza: OK, do you have an ETA for that? Is this ready for another review?

josenavas: Yes this is ready for another review. ETA 1-2 days.

2015-01-01 21:27 GMT+01:00 Antonio Gonzalez <notifications@github.com>:

> In qiita_pet/handlers/processing_handlers.py
> <https://github.com/biocore/qiita/pull/732#discussion-diff-22402078>:
>
> > @@ -0,0 +1,27 @@
> > +from tornado.web import authenticated
> > +
> > +from .base_handlers import BaseHandler
> > +from qiita_ware.dispatchable import processor
> > +from qiita_db.parameters import ProcessedSortmernaParams
> > +from qiita_ware.context import submit
> > +
> > +
> > +class ProcessHandler(BaseHandler):
> > +    @authenticated
> > +    def post(self):
> > +        study_id = int(self.get_argument('study_id'))
> > +        preprocessed_data_id = int(self.get_argument('preprocessed_data_id'))
> > +
> > +        # We currently do not support defining the parameters
>
> OK, do you have an ETA for that? Is this ready for another review?
>
> —
> <https://github.com/biocore/qiita/pull/732/files#r22402078>.
>



-- 
Jose Navas

antgonza:
A few comments:
- [ ] In the popup window, can you replace the phrase "Running with default parameters" with a list the parameters being used (in a readable format)?
- [ ] You are missing a patch for those already existing databases to add processed_params_sortmerna. Note that 10.sql adds the structure but is missing the insert with default parameters.

josenavas:
@antgonza The second point is a bit tricky. If the database was created with `--download-reference`, the parameters already exist, and the 10.sql patch only changes the database columns, However, if the database was not created using that flag, the parameters, neither the reference exist. Thus, I'm unsure on how to move forward on this situation... The 2 options that I can think of are the following:
 - Leave the DB as it is and with the CLI commands add-reference (already merged) and add-parameters (currently working on it) the user has to add those missing entries.
 - Create a new patch that forces the downloading of the reference if it does not exist in first place. I'm against this option because it will ignore the flag passed by the user and it will download the reference files in all our tests.

The main problem is that in order to add the default parameters for SortMeRNA, the reference database should exist. What do you think? The first option (executing 2 commands) seems the more reasonable, and we will suggest to download always the references.

Related with your first point, I wanted to work in allowing to change the parameters, particularly the reference database, which is needed for the workshop (that is why I was working on adding the 2 previously mentioned commands). In that case, will it make sense to list the parameters being used, a part from the reference db? Note that this is not the approach that we used for split libraries (i.e we are not showing a list of parameters) so I just wanted to make sure that everything is consistent.

antgonza:
For 1, are you sure you will be able to have those changes ready
before the end of the weekend? Also, remember that we really only need
pick against unite and GG (remember that the default will depend on
the prep template type), not sure if the other options are urgent.
However, it all depends on how/when can we have this ready. Remember
we are already extremely late.

For 2, http://stackoverflow.com/questions/4069718/postgres-insert-if-does-not-exist-already

rob-knight:
Other options are not urgent but would be desirable later.

On Jan 2, 2015, at 7:53 AM, Antonio Gonzalez <notifications@github.com<mailto:notifications@github.com>> wrote:

For 1, are you sure you will be able to have those changes ready
before the end of the weekend? Also, remember that we really only need
pick against unite and GG (remember that the default will depend on
the prep template type), not sure if the other options are urgent.
However, it all depends on how/when can we have this ready. Remember
we are already extremely late.

For 2, http://stackoverflow.com/questions/4069718/postgres-insert-if-does-not-exist-already

antgonza:
There are conflicts, could you fix?

josenavas:
Should be fixed.

PR #745, opened 2014-12-23
Return user object with self.current_user

squirrelo:
This cleans up the codebase by returning a user object for self.current_user instead of just the email. Almost every instance of the call was just instantiating a user object anyway, so this makes that boilerplate. It also allows the base template to automatically get user info and level for rendering.


squirrelo:
This is acting extremely odd. When running  as a normal webserver 'get_current_user()' functions correctly, but during tests it returns a string instead of the User object and all hell breaks loose. I'm trying to figure out why but if anyone please has ideas please let me know.

antgonza:
The issue is that you haven't changed the tests. For example, in https://github.com/biocore/qiita/blob/master/qiita_db/test/test_search.py#L144. Other option is that you did in your local machine but didn't push those changes.

squirrelo:
Good catch, but that's not what the main failure point is. The problem is the `current_user` in the sitebase template is a string in tests, and a User object in practice, so when the template checks `user.level` it errors out, since strings don't have a .level parameter.

antgonza:
Strange cause the first error in Travis doesn't point to that:
ERROR: test_call (qiita_db.test.test_search.DecoratedClass)
  File "/home/travis/build/biocore/qiita/qiita_db/test/test_search.py",
line 144, in test_call
    "test@foo.bar")
  File "/home/travis/build/biocore/qiita/qiita_db/search.py", line
204, in __call__
    user.user_studies +
AttributeError: 'str' object has no attribute 'user_studies'

Perhaps solving those issues first will make clearer the others ...

squirrelo:
Yes, but every other error below it is the level one.

antgonza:
If you want to chat, just let me know.

Looking at the files that were changed, neither
qiita_pet/test/test_analysis_handlers.py nor
qiita_db/test/test_search.py have changed. Thus, leading me to believe
that you haven't changed the tests from str to objects. Again, solving
first the easy errors and then focusing on the hard is the easiest
path forward.

antgonza:
@squirrelo ping!

squirrelo:
I'm still at a loss as to why this is erroring. Anyone have a guess?

squirrelo:
This is all coming down to the issue stated above, where it works fine in production but in tests it returns a string.

squirrelo:
Alright, now it is entirely the following error:

File "error_html.generated.py", line 93, in _tt_execute
    user_level = user.level  # sitebase.html:149
AttributeError: 'str' object has no attribute 'level'

Which means yes, it is actually returning a string instead of the objects when testing, as I have been saying before.

josenavas:
@squirrelo I will take a look to this next week.

antgonza:
Just tested on my laptop and I'm getting the same errors/failures that
Travis is having. If you are not getting this in your local version,
then there is a difference between what's on your local and uploaded
branches. Could you verify this? If that's not the case I can take a
look at why it is failing.

squirrelo:
Are you doing tests or actually using the site? As I've said multiple times now, the tests fail but the site itself works.

antgonza:
Got it, believe it or not, that wasn't clear from your previous emails. 

Anyway, try making these modifications:
```
$ git diff
diff --git a/qiita_db/test/test_meta_util.py b/qiita_db/test/test_meta_util.py
index 91ff572..93157e1 100644
--- a/qiita_db/test/test_meta_util.py
+++ b/qiita_db/test/test_meta_util.py
@@ -77,7 +77,7 @@ class MetaUtilTests(TestCase):
         count = self.conn_handler.execute_fetchone("SELECT count(*) FROM "
                                                    "qiita.filepath")[0]
         exp = set(range(1, count + 1))
-        obs = get_accessible_filepath_ids('admin@foo.bar')
+        obs = get_accessible_filepath_ids(User('admin@foo.bar'))
         self.assertEqual(obs, exp)
 
 
diff --git a/qiita_pet/test/tornado_test_base.py b/qiita_pet/test/tornado_test_base.py
index 0faca9f..bda355f 100644
--- a/qiita_pet/test/tornado_test_base.py
+++ b/qiita_pet/test/tornado_test_base.py
@@ -10,6 +10,7 @@ from qiita_pet.webserver import Application
 from qiita_pet.handlers.base_handlers import BaseHandler
 from qiita_db.sql_connection import SQLConnectionHandler
 from qiita_db.environment_manager import drop_and_rebuild_tst_database
 
 
 class TestHandlerBase(AsyncHTTPTestCase):
@@ -18,7 +19,7 @@ class TestHandlerBase(AsyncHTTPTestCase):
     app = Application()
 
     def get_app(self):
-        BaseHandler.get_current_user = Mock(return_value="test@foo.bar")
+        BaseHandler.get_current_user = Mock(return_value=User("test@foo.bar"))
         return self.app
 
     def setUp(self): 
```

squirrelo:
Fuuuuu yup, it was Mock not being changed. Thanks for finding that.

squirrelo:
Ready for review

antgonza:
:+1: Seems like @josenavas and @wasade have already done a pass, could one of you do another one, comment back or merge? Thanks.

qiita_pet/handlers/analysis_handlers.py
```diff
@@ -267,12 +263,12 @@ def post(self, analysis_id):
...
             rarefaction_depth = None
         analysis = Analysis(analysis_id)
-        check_analysis_access(User(user), analysis)
+        check_analysis_access(self.current_user, analysis)
 
         command_args = self.get_arguments("commands")
         split = [x.split("#") for x in command_args]
 
-        moi_user_id = get_id_from_user(user)
+        moi_user_id = get_id_from_user(str(self.current_user))
```
josenavas: Doesn't the user has the property `id`? I saw that you added the `__str__` method to the user object, but the `QiitaUser` class extends the `QiitaBase` clase which already has the `id` property define. I would say that this is duplicating information and we can potentially use `__str__` in the future to output some other more important information rather than just the id, and we already have the `id` property to retrieve that information.

qiita_pet/handlers/base_handlers.py
```diff
@@ -4,38 +4,34 @@
...
-            return User(self.current_user)
-        return None
-
     def get_current_user(self):
         '''Overrides default method of returning user curently connected'''
-        user = self.get_secure_cookie("user")
-        if user is None:
+        username = self.get_secure_cookie("user")
+        if username is not None:
+            username = username.strip("\"' ")
```
josenavas: can you add a comment on why is this needed?

qiita_pet/handlers/compute.py
```diff
@@ -69,10 +68,10 @@ def post(self):
                     if exists(ft):
                         filepaths.append((ft, filetype))
 
-        job_id = submit(self.current_user, add_files_to_raw_data, raw_data_id,
-                        filepaths)
+        job_id = submit(str(self.current_user), add_files_to_raw_data,
```
josenavas: same comment as above. It looks like `self.current_user.id` will work here.

qiita_pet/handlers/compute.py
```diff
@@ -94,12 +93,12 @@ def post(self):
             # Study not in database so fail nicely
             raise HTTPError(404, "Study %d does not exist" % study_id)
         else:
-            check_access(User(self.current_user), study, raise_error=True)
+            check_access(self.current_user, study, raise_error=True)
 
-        job_id = submit(self.current_user, unlink_all_files, raw_data_id)
+        job_id = submit(str(self.current_user), unlink_all_files, raw_data_id)
```
josenavas: Same

qiita_pet/handlers/preprocessing_handlers.py
```diff
@@ -60,10 +60,10 @@ def post(self):
         else:
             raise ValueError('Unknown filetype')
 
-        job_id = submit(self.current_user, preprocessor, study_id,
+        job_id = submit(str(self.current_user), preprocessor, study_id,
```
josenavas: Same here

qiita_pet/handlers/study_handlers/ebi_handlers.py
```diff
@@ -71,15 +70,16 @@ def get(self, preprocessed_data_id):
 
     @authenticated
     def post(self, preprocessed_data_id):
+        user = self.current_user
         # make sure user is admin and can therefore actually submit to EBI
-        if User(self.current_user).level != 'admin':
+        if self.current_user.level != 'admin':
             raise HTTPError(403, "User %s cannot submit to EBI!" %
-                            self.current_user)
+                            str(user))
```
josenavas: And here. Last time that I comment, although there are more instances.

qiita_pet/templates/sitebase.html
```diff
@@ -145,8 +146,7 @@
                 <li><a href="/study/public/">View Public Studies</a></li>
               </ul>
             </li>
-            {% set user_tmp = handler.user() %}
-            {% set user_level = user_tmp.level if user_tmp is not None else None %}
+            {% set user_level = user.level %}
```
josenavas: We try to minimize database accesses from the template. Is there a way that you can have this information from the handler rather than accessing it from the template?

scripts/qiita
```diff
@@ -328,7 +328,7 @@ def start(port):
 
     http_server = tornado.httpserver.HTTPServer(Application())
     http_server.listen(port)
-    click.echo("Qiita started on port %s" % port)
+    click.echo("Qiita started on port " + str(port))
```
josenavas: Why this change? if port is an integer just use:
`click.echo("Qiita started on port %d" % port)`

josenavas:
@squirrelo few comments. I haven't tested on my end. Will do once those comments are solved.
Thanks!

scripts/qiita
squirrelo: Tired and forgot %d existed. Fixed

qiita_pet/templates/sitebase.html
squirrelo: This is needed to show the admin or dev menu at the top so it's needed in every page. Easier to put it in the template because of that.

qiita_pet/handlers/analysis_handlers.py
squirrelo: yeah, thats fair. I'll change it to that.

qiita_pet/handlers/base_handlers.py
squirrelo: It returns the username with quotes around it when you use the cookie function, so I strip them off here.

qiita_pet/templates/sitebase.html
squirrelo: Also to add to this, by making the self.current_user return a User object, tornado actually caches the user object on first instantiation, so we save all the "does the id exist?" calls made when we kept instantiating a User object in each function. This should hopefully save a lot of unneeded calls as well.

squirrelo:
This is good to go

antgonza:
@josenavas have you tested? Is this ready for merge?

squirrelo:
Anyone? Bueller?

josenavas:
@squirrelo Few issues that I found so far:
- The name of the user does not appear on the main page:
screen shot 2015-02-03 at 11 30 06 am
- When doing any kind of processing, the page does not update: i.e. it is blocked on "queue" although the  processing is done (e.g. linking raw files, preprocessing them...). The meta-analysis processing, however, it does work.
- If I try to process a preprocessed data, this error is raise:
```
  File "/Users/jose/.virtualenvs/qiita/lib/python2.7/site-packages/tornado/web.py", line 1141, in _when_complete
    callback()
  File "/Users/jose/.virtualenvs/qiita/lib/python2.7/site-packages/tornado/web.py", line 1162, in _execute_method
  File "/Users/jose/.virtualenvs/qiita/lib/python2.7/site-packages/tornado/web.py", line 2297, in wrapper
  File "/Users/jose/qiime_software/QiiTa/qiita_pet/handlers/processing_handlers.py", line 21, in post
    param_id, param_constructor)
  File "/Users/jose/qiime_software/QiiTa/qiita_ware/context.py", line 256, in _submit
    'messages': channel + ':messages'}
TypeError: unsupported operand type(s) for +: 'User' and 'str'
```

wasade:
I'll reraise my initial concern here: is it actually safe to have the
`current_user` method return a custom object?

squirrelo:
It's safe as long as when you do 'if user:` it acts as true when the user is logged in and false when not, which this does.

squirrelo:
This should be ready for final review now.

josenavas:
I've tested this locally and this work. @antgonza there has been more commits after your :+1: so I'm not sure if you agree with the current version of the code. You can merge it if you think the code is :+1:

PR #758, opened 2015-01-02
Add parameter

josenavas:
Adds a command to add a parameter set to the database.

@antgonza needed for the workshop and testing #732

### DIFF
--- a/qiita_db/commands.py
+++ b/qiita_db/commands.py
@@ -267,7 +267,7 @@ def load_parameters_from_cmd(fp, table):
     ------
     ValueError
         If the table does not exists on the DB
-        If the fp is not well formatted
+        If the fp is not correctly formatted
     """
     if table not in SUPPORTED_PARAMS:
         raise ValueError("Table %s not supported. Choose from: %s"
@@ -293,11 +293,5 @@ def load_parameters_from_cmd(fp, table):
                         "The format of the parameters files is not correct. "
                         "The format is PARAMETER_NAME<tab>VALUE")
                 params[values[0]] = values[1]
-
-    db_cols = set(get_table_cols(table))
-    missing = db_cols.difference(params)
-
-    if missing:
-        raise ValueError("Missing parameters: %s" % ', '.join(missing))
 
     return constructor.create(**params)
--- a/qiita_db/commands.py
+++ b/qiita_db/commands.py
@@ -268,6 +268,13 @@ def load_parameters_from_cmd(name, fp, table):
     ValueError
         If the table does not exists on the DB
         If the fp is not correctly formatted
+
+    Notes
+    -----
+    `fp` should be a tab-delimited text file following this format:
+        parameter_1<TAB>value
+        parameter_2<TAB>value
+        ...
     """
     if table not in SUPPORTED_PARAMS:
         raise ValueError("Table %s not supported. Choose from: %s"
@@ -282,16 +289,10 @@ def load_parameters_from_cmd(name, fp, table):
 
     constructor = constructor_dict[table]
 
-    # Parse the parameters file
-    params = {}
-    with open(fp, 'U') as f:
-        for line in f:
-            values = line.strip().split('\t')
-            if values:
-                if len(values) != 2:
-                    raise ValueError(
-                        "The format of the parameters files is not correct. "
-                        "The format is PARAMETER_NAME<tab>VALUE")
-                params[values[0]] = values[1]
+    try:
+        params = dict(tuple(l.strip().split('\t')) for l in open(fp, 'U'))
+    except ValueError:
+        raise ValueError("The format of the parameters files is not correct. "
+                         "The format is PARAMETER_NAME<tab>VALUE")
 
     return constructor.create(name, **params)
--- a/qiita_db/commands.py
+++ b/qiita_db/commands.py
@@ -20,6 +20,10 @@ from .util import get_filetypes, get_filepath_types
 from .data import RawData, PreprocessedData, ProcessedData
 from .metadata_template import (SampleTemplate, PrepTemplate,
                                 load_template_to_dataframe)
+from .parameters import PreprocessedIlluminaParams, Preprocessed454Params
+
+SUPPORTED_PARAMS = ['preprocessed_sequence_illumina_params',
+                    'preprocessed_sequence_454_params']
 
 
 def load_study_from_cmd(owner, title, info):
@@ -242,3 +246,58 @@ def load_processed_data_cmd(fps, fp_types, processed_params_table_name,
     return ProcessedData.create(processed_params_table_name,
                                 processed_params_id, list(zip(fps, fp_types)),
                                 preprocessed_data, study, processed_date)
+
+
+def load_parameters_from_cmd(fp, table):
+    """Add a new parameters entry on `table`
+
+    Parameters
+    ----------
+    fp : str
+        The filepath to the parameters file
+    table : str
+        The name of the table to add the parameters
+
+    Returns
+    -------
+    qiita_db.BaseParameters
+        The newly `qiita_db.BaseParameters` object
+
+    Raises
+    ------
+    ValueError
+        If the table does not exists on the DB
+        If the fp is not well formatted
+    """
+    if table not in SUPPORTED_PARAMS:
+        raise ValueError("Table %s not supported. Choose from: %s"
+                         % (table, ', '.join(SUPPORTED_PARAMS)))
+
+    # Build the dictionary to get the parameter constructor
+    constructor_dict = {}
+    constructor_dict[
+        'preprocessed_sequence_illumina_params'] = PreprocessedIlluminaParams
+    constructor_dict[
+        'preprocessed_sequence_454_params'] = Preprocessed454Params
+
+    constructor = constructor_dict[table]
+
+    # Parse the parameters file
+    params = {}
+    with open(fp, 'U') as f:
+        for line in f:
+            values = line.strip().split('\t')
+            if values:
+                if len(values) != 2:
+                    raise ValueError(
+                        "The format of the parameters files is not correct. "
+                        "The format is PARAMETER_NAME<tab>VALUE")
+                params[values[0]] = values[1]
+
+    db_cols = set(get_table_cols(table))
+    missing = db_cols.difference(params)
+
+    if missing:
+        raise ValueError("Missing parameters: %s" % ', '.join(missing))
+
+    return constructor.create(**params)
--- a/qiita_db/commands.py
+++ b/qiita_db/commands.py
@@ -248,7 +248,7 @@ def load_processed_data_cmd(fps, fp_types, processed_params_table_name,
                                 preprocessed_data, study, processed_date)
 
 
-def load_parameters_from_cmd(fp, table):
+def load_parameters_from_cmd(name, fp, table):
     """Add a new parameters entry on `table`
 
     Parameters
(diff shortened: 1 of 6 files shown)


antgonza:
Will this not be easier to add this as a patch?

josenavas:
No, this is more general. This allows to add new parameters to the database without patching it (it is only adding data, and it should not be added in a patch). Right now, we are limiting the parameters, but, ideally, we are going to allow to change some of the parameters. This will also allow to create parameters sets through the interface, once we modify it.

antgonza:
How many parameter sets do you envision users actually adding? Adding
required info via a patch seems fine to me ...

squirrelo:
I agree the more general setup is better. Eventually we will allow users to edit options on the frontend, as Jose said, so better to have the functionality now then hard code and change later.

qiita_db/commands.py
```diff
@@ -242,3 +246,52 @@ def load_processed_data_cmd(fps, fp_types, processed_params_table_name,
...
+    Returns
+    -------
+    qiita_db.BaseParameters
+        The newly `qiita_db.BaseParameters` object
+
+    Raises
+    ------
+    ValueError
+        If the table does not exists on the DB
+        If the fp is not correctly formatted
```
squirrelo: Would be good to have a notes section explaining the format of the parameters file.

qiita_db/commands.py
```diff
@@ -242,3 +246,52 @@ def load_processed_data_cmd(fps, fp_types, processed_params_table_name,
...
+
+    constructor = constructor_dict[table]
+
+    # Parse the parameters file
+    params = {}
+    with open(fp, 'U') as f:
+        for line in f:
+            values = line.strip().split('\t')
+            if values:
+                if len(values) != 2:
```
squirrelo: why two ifs instead of 
```python
if values and len(values) != 2
```
Python does short-circuit.

squirrelo: wait, my bad, got tabbing mixed up.

qiita_db/parameters.py
```diff
@@ -9,16 +9,36 @@
...
+
+        if missing:
+            raise ValueError("Missing columns: %s" % ', '.join(missing))
+
+        conn_handler = SQLConnectionHandler()
+
+        vals = kwargs.values()
+        vals.insert(0, param_set_name)
+
+        id_ = conn_handler.execute_fetchone(
```
squirrelo: needs an exists check before adding to make sure we don't duplicate things.

qiita_db/commands.py
```diff
@@ -242,3 +246,52 @@ def load_processed_data_cmd(fps, fp_types, processed_params_table_name,
...
+                    raise ValueError(
+                        "The format of the parameters files is not correct. "
+                        "The format is PARAMETER_NAME<tab>VALUE")
+                params[values[0]] = values[1]
+
```
squirrelo: Should there be an else here in case a blank file is passed in?

qiita_db/commands.py
```diff
@@ -242,3 +246,52 @@ def load_processed_data_cmd(fps, fp_types, processed_params_table_name,
...
+    constructor_dict[
+        'preprocessed_sequence_454_params'] = Preprocessed454Params
+
+    constructor = constructor_dict[table]
+
+    # Parse the parameters file
+    params = {}
+    with open(fp, 'U') as f:
+        for line in f:
+            values = line.strip().split('\t')
```
wasade: more robust to do `values = [l.strip() for l in line.split('\t')]` as it catches accidental whitespace in cells

qiita_db/commands.py
```diff
@@ -242,3 +246,52 @@ def load_processed_data_cmd(fps, fp_types, processed_params_table_name,
...
+    constructor_dict = {}
+    constructor_dict[
+        'preprocessed_sequence_illumina_params'] = PreprocessedIlluminaParams
+    constructor_dict[
+        'preprocessed_sequence_454_params'] = Preprocessed454Params
+
+    constructor = constructor_dict[table]
+
+    # Parse the parameters file
+    params = {}
```
wasade: this could all just be done with `return {tuple(l.strip().split('\t')) for l in open(fp, 'U')}`. The method could then be wrapped to throw a format error. This doesn't strip the parameter_name (see next comment) but it benefits from being succinct

wasade: should have read further... would still need to pass to the constructor of course

scripts/qiita
```diff
@@ -224,6 +225,18 @@ def load_reference_db(name, version, seq_fp, tax_fp, tree_fp):
...
 
+@db.command()
+@click.argument('fp', required=True,
+                type=click.Path(resolve_path=True, readable=True, exists=True))
+@click.option('--table', required=True, type=click.Choice(SUPPORTED_PARAMS),
+              help='The table to add the parameter set')
+def load_parameters(fp, table):
+    """Loads a parameter set to the database"""
+    param = load_parameters_from_cmd(fp, table)
+    print("Parameters successfully added to table %s with id %s"
```
wasade: `click.echo` please

scripts/qiita
```diff
@@ -224,6 +225,18 @@ def load_reference_db(name, version, seq_fp, tax_fp, tree_fp):
...
 
 
+@db.command()
+@click.argument('fp', required=True,
+                type=click.Path(resolve_path=True, readable=True, exists=True))
+@click.option('--table', required=True, type=click.Choice(SUPPORTED_PARAMS),
+              help='The table to add the parameter set')
+def load_parameters(fp, table):
+    """Loads a parameter set to the database"""
+    param = load_parameters_from_cmd(fp, table)
```
wasade: The function signature for `load_parameters_from_cmd` indicates a non-optional `name` parameter...

qiita_db/test/test_parameters.py
```diff
@@ -15,6 +15,29 @@
 @qiita_test_checker()
 class PreprocessedIlluminaParamsTests(TestCase):
 
+    def test_create(self):
+        obs_obj = PreprocessedIlluminaParams.create(
```
wasade: The actual code isn't casting the types and is only passing in strings, can the test be updated to reflect that?

wasade:
Few comments, thanks for this @josenavas! Regarding adding parameters via cmd line or patch, I do not have strong opinions. I suspect this is not going to be all that common of a thing to do though given that a precondition for meta analysis is normalized parameters (to the best extent possible).

antgonza:
If we go this route, I think we need to have a folder with all the
valid/supported parameters so people can use them + the suggested
example.

wasade:
I agree, the downside to not having these as patches is that we cannot ensure that every deployment of qiita has the same initial parameter sets.

josenavas:
@wasade we currently cannot ensure that every deployment of qiita has the same initial parameters sets. We added an optional flag `--download-reference` to qiita make env which is the one that is generating this whole situation.

I don't have any problem in adding the parameter as a patch, but the first problem is that I cannot ensure that the first row in the reference table has been added, so I cannot add the parameter. Also, I cannot enforce adding that row, because to add it I need the reference files.

That is why I'm creating this command, so the user is in charge of uploading the references and the parameters.

I think the problem actually resides in the `qiita_env make` command, that depending on the flags passed leaves the database with different information. We can modify that command so it always adds the reference to the database; and, by extension, the first parameters set. E.g. if the user does not provides `--download-reference`; it has to provide the paths to the reference files that we are downloading. However, this still does not solve the issue on the currently deployed databases (which can be addressed using the commands...)

qiita_db/commands.py

qiita_db/parameters.py
josenavas: Done

qiita_db/commands.py
josenavas: Not needed. `params` will be an empty dictionary and the constructor will raise an error with a list of missing columns (in this case, all)

qiita_db/commands.py
josenavas: Good catch

qiita_db/commands.py
josenavas: Thanks, done!

scripts/qiita

scripts/qiita
josenavas: oops...

qiita_db/test/test_parameters.py
josenavas: Done

josenavas:
This is ready for another review.

antgonza:
i believe you can assure the order of the entries using conditional
inserting, like the example I sent you, if you use patches.
@josenavas, why will this not work?

antgonza:
@josenavas you are missing the folder with the current acceptable params info and the example, can you add?

josenavas:
@antgonza the issue is downloading the reference, not the insertion of the parameters. The reference should be inserted before. If it was inserted on creation time (using `qiita_env make --download-ontologies`) both entries are created (references and parameters). Otherwise, none of them exist. In order to insert the reference one, I need input from the user, so that is why we have the add-reference command.

antgonza:
It makes sense to have it as a command but it also makes sense to add
all the defaults via patches. We are the ones that define the
references, post them online, and set the parameters. In a fresh
install, if we use patches, we can define order. In previous installs,
if the user has added something and the order changes, we can trigger
either a protocol to move parameters or simply error out saying that
we can not apply the patch and that the user must do it. I think the
latter is the best solution as at this point, AFAIK, nobody had added
anything in their dbs.

BTW why are we focusing on order vs. names or codes for the parameters?

squirrelo:
Just a warning: from a biologist or anyone not computer science perspective, applying the patch manually is a daunting - and probably unrealistic - task.

antgonza:
Agree and is not much better generating parameters files of something
they do not care or understand. Again, I do not believe that can of
user will have something in their database ...

squirrelo:
We are going to hook this up to the frontend though, which will make changing parameters easy in the near future. Then this will be an issue.

antgonza:
Define near future? IMOO there are more important issues that allowing
GUI interaction of parameters. Again, even with allowing that, we
should enforce the qiita supported, official set of parameters.

wasade:
Order assumes we care somewhere about meaningless IDs (which is currently
the case). We should not force ourselves to remain in this risky situation.

josenavas:
The qiita supported, official set of parameters are created using `qiita_env make --download-ontologies`. If the current qiita deployment does not include this option was a mistake during installation. That is why the commands allow us to fix the problem, while leveraging the correct installations with the correct data.

josenavas:
I'm not worried about IDs, I'm worried about the data that does not exist on the database and I cannot add in an easy way using a patch. Again, I'm talking about the reference databases (i.e. Greengenes) not the prameters for picking OTUs...

antgonza:
If it's an option, is not a mistake,  it's optional. All current live
systems were created without that option so that's the reality

josenavas:
The point that I'm seeing right now is that we then are forcing the user to download the references if it does not exist. The only solution that I can think right now is to create a patch that, in case the reference has not been added, it will ask the user for input, since the user will need to point to the right filepaths. Is that a solution that everybody agrees?

josenavas:
Also note that, such solution is independent of this PR, as this PR can still be merged cause it can be useful in the future as it is adding the ability to create parameters (it implements the `create` classmethod of BaseParameters). So this PR will need to be merged before that weird patch that I'm proposing.

antgonza:
Agree on solution. Note that you are still missing the example
configuration and the default configuration files and folder before
this can be merged.

antgonza:
@josenavas ping on this, can you solve this today?!

josenavas:
@antgonza what is missing on this PR? The patch is coming in another PR. My last commit (a day ago) added the default parameters file...

PR #759, opened 2015-01-02
BUG:Remove rows without sample name & improve docs

### LINKED ISSUE #746: Should scrub empty rows before parsing templates
Excel has a habit of leaving empty rows resulting in full rows of `Nan` in the `DataFrame`s

ElDeveloper:
Rows with empty sample names are now removed from the dataframe, note that
these rows could include relevant data in other columns but since we have no
sample name, they are useless.

The documentation for read_csv was lacking a bit, so I expanded on the reason
why the arguments are being passed the way they are.

Fixes #746


antgonza:
:+1:

PR #760, opened 2015-01-04
Adding patch for default processing parameters set

josenavas:
@antgonza this adds the patch for adding the default parameters set.

### DIFF
--- a/qiita_db/support_files/patches/13.sql
+++ b/qiita_db/support_files/patches/13.sql
@@ -1,0 +1,4 @@
+-- This file should have been empty as we only need 16.py but it will fail if
+-- it's actually empty so doing a simple query
+
+SELECT * FROM settings;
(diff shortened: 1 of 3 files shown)


antgonza:
Failures but seems that is something with Travis, restarted the build.

antgonza:
I think the problem is that the patch is waiting for user input.

josenavas:
@antgonza it should be fixed now

antgonza:
No luck ... :weary: seems like you are not testing this locally.

josenavas:
I did tested locally, but there is a specific test case that I did not tested and it is the one happening on Travis. At this point I run out of ideas on how to solve this. Any ideas?? Note that at this point you can ad the default parameters set using the command, so if there are no ideas on how to solve this, use the command to add the default parameter, so this does not block anything...

qiita_db/support_files/patches/13.sql
```diff
@@ -0,0 +1,4 @@
+-- This file should have been empty as we only need 16.py but it will fail if
```
adamrp: Should this say "13.py"?

antgonza:
@josenavas Should we close this?

PR #769, opened 2015-01-06
Multiple process parameters interface

josenavas:
Updates the interface to allow multiple processing parameters. The interface is based in the parameter set name and shows the values to the user.

@antgonza this should be the last needed bit to support ITS.

### DIFF
--- a/qiita_db/parameters.py
+++ b/qiita_db/parameters.py
@@ -122,6 +122,21 @@ class BaseParameters(QiitaObject):
 
         return " ".join(result)
 
+    @classmethod
+    def iter(cls):
+        """Iterates over all parameters
+
+        Returns
+        -------
+        generator
+            Yields a parameter instance
+        """
+        conn_handler = SQLConnectionHandler()
+        sql = "SELECT {0} FROM qiita.{1}".format(cls._column_id, cls._table)
+
+        for result in conn_handler.execute_fetchall(sql):
+            yield cls(result[0])
+
 
 class PreprocessedIlluminaParams(BaseParameters):
     r"""Gives access to the preprocessed parameters of illumina data"""
--- a/qiita_db/parameters.py
+++ b/qiita_db/parameters.py
@@ -66,6 +66,29 @@ class BaseParameters(QiitaObject):
 
         return cls(id_)
 
+    @classmethod
+    def iter(cls):
+        """Iterates over all parameters
+
+        Returns
+        -------
+        generator
+            Yields a parameter instance
+        """
+        conn_handler = SQLConnectionHandler()
+        sql = "SELECT {0} FROM qiita.{1}".format(cls._column_id, cls._table)
+
+        for result in conn_handler.execute_fetchall(sql):
+            yield cls(result[0])
+
+    @property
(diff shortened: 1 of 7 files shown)


antgonza:
Errors!! Could you fix?

qiita_db/parameters.py
```diff
@@ -66,6 +66,56 @@ def create(cls, param_set_name, **kwargs):
...
+        """The values of the parameter set
+
+        Returns
+        -------
+        dict
+            Dictionary with the parameter values keyed by parameter name
+        """
+        conn_handler = SQLConnectionHandler()
+        result = dict(conn_handler.execute_fetchone(
+            "SELECT * FROM qiita.{0} WHERE {1} = %s".format(
```
antgonza: Will it be easier to simply retrieve the info you need vs. having to delete the ones that you do not need?

qiita_pet/templates/study_description_templates/preprocessed_data_info_tab.html
```diff
@@ -1,5 +1,19 @@
 {% from qiita_pet.handlers.util import download_link_or_path %}
 
+<script>
+function parameters_changed(obj, ppd_id){
+  data = "foo"
```
antgonza: Do you really need to initialize this variable? If so, does it have to be "foo" or can it be simply ""?

antgonza:
Works great but some comments on the code.

qiita_db/parameters.py
josenavas: This method belongs to a base class, so I don't know which are those columns. However, I know that the two columns that are below exist, and they are not needed, so I just remove them.

antgonza: Got it, now why do you remove them vs. leaving? I guess they are always there, right?

qiita_pet/templates/study_description_templates/preprocessed_data_info_tab.html
josenavas: Fixed

qiita_db/parameters.py
josenavas: Those values are not technically part of the parameters. They're id's and name's used for internal use, and we can access them using properties

antgonza: OK thanks.

qiita_pet/util.py
```diff
@@ -50,3 +50,21 @@ def clean_str(item):
...
+    ----------
+    param : BaseParameters
+        The parameter to generate the str
+
+    Returns
+    -------
+    str
+        The html string with the parameter set values
+    """
+    result = ["<b>%s:</b> %s" % (name, value)
```
ElDeveloper: IMHO using a table would make more sense in this case, not sure that this should block this PR though.

qiita_db/parameters.py
```diff
@@ -66,6 +66,56 @@ def create(cls, param_set_name, **kwargs):
 
         return cls(id_)
 
+    @classmethod
+    def iter(cls):
```
ElDeveloper: Should this be `__iter__` instead?

PR #770, opened 2015-01-06
@squirrelo and @ElDeveloper were complaining

adamrp:


squirrelo:
:+1:  to OCD!

josenavas:
Why is that space being removed? There are a lot of instances of that space and the reason is that the interface looks way better with that space. Otherwise, the text looks too close to the icon...

ElDeveloper:
I disagree, the interface looks misaligned, I think that's why @adamrp is removing it.

Yoshiki Vázquez-Baeza

adamrp:
@josenavas look at the page currently with the space and you'll see.
@eldeveloper that's exactly right.

PR #776, opened 2015-01-07
Cut body to 1024 chars. Fix #775

### LINKED ISSUE #775: Error messages are being logged with a large number of characters
There have been a few error messages showing up in the logging table that are so large that they cannot be displayed on the user interface and psql is having trouble printing this to the screen.

Try something like to see what I mean:

```
select time, length(msg) from logging order by time desc limit 50;
````

Do not try to get the full msg string, otherwise you will lock your psql session.

### PR DESCRIPTION
**Pull request by squirrelo**

### CODE REVIEW

**Comment by antgonza** [comment]
Not sure this is the solution as we might miss something in the log or perhaps I'm missing something.

**Comment by squirrelo** [comment]
It's only cutting down the body information, which is entire files or entire pages worth of html. It doesn't cut out any of the debugging info.

**Comment by antgonza** [comment]
cool, thanks.

File: qiita_pet/handlers/base_handlers.py [hunk 1]
```diff
@@ -39,9 +39,12 @@ def write_error(self, status_code, **kwargs):
         exc_info = kwargs["exc_info"]
         trace_info = ''.join(["%s\n" % line for line in
                              format_exception(*exc_info)])
+        req_dict = self.request.__dict__
+        # must trim body to 1024 chars to prevent huge error messages
+        req_dict['body'] = req_dict.get('body', '')[:1024]
```
antgonza: This assumes that there is always a body, is this the case?

qiita_pet/handlers/base_handlers.py
```diff
@@ -39,9 +39,12 @@ def write_error(self, status_code, **kwargs):
...
         request_info = ''.join(["<strong>%s</strong>: %s\n" %
-                               (k, self.request.__dict__[k]) for k in
-                                self.request.__dict__.keys()])
+                               (k, req_dict[k]) for k in
```
antgonza: Will it be better to do iteritems?

antgonza:
I think this is fine :+1:

qiita_pet/handlers/base_handlers.py
squirrelo: There should always be, but thats what the get is for.

qiita_pet/handlers/base_handlers.py
squirrelo: maybe, but it's a short dict.

ElDeveloper:
:+1:

wasade:
I realize this is now closed, but often the first part and the last part of an error are the most useful, but this approach removes all of the tail end of the message

antgonza:
This only cuts the body of the file that caused the error, not the error log.

wasade:
ah, okay, thanks

PR #778, opened 2015-01-07
Fix trying to render without dropped samples

squirrelo:
This makes sure we don't loop over a None when doing viewitems.


antgonza:
:+1:

PR #786, opened 2015-01-07
Fix only finding missing samples for one processed data id

squirrelo:
This was originally only looking at the processed data id of the first sample in the table. Now it looks at all processed data ids.


antgonza:
@squirrelo ping!

squirrelo:
MOI compilation issues prevent anything from happening. Same for every other ping.

qiita_db/analysis.py
```diff
@@ -241,9 +241,11 @@ def dropped_samples(self):
         for biom, filepath in viewitems(bioms):
             table = load_table(filepath)
             # remove the samples from the sets as they are found in the table
-            proc_data_id = table.metadata()[0]['Processed_id']
+            proc_data_ids = set(sample['Processed_id']
+                                for sample in table.metadata())
             ids = set(table.ids())
-            all_samples[proc_data_id] = all_samples[proc_data_id] - ids
+            for proc_data_id in proc_data_ids:
```
antgonza: All the sample ids are unique, right? So why do this in a loop vs. getting all of the samples and doing the "difference" in the full set?

squirrelo: Because it's a dictionary of sets, so you need to do the difference for each key in the dictionary.

antgonza: Got it, could you add a test for the specific case that raised this bug?

squirrelo: Not without building an entire new biom table and adding a ton of new stuff to populate_test_db.sql

antgonza: Sounds like a plan.

squirrelo: OK, I can't do this sanely or quickly. I need to add an entirely new study with all data to test this, AKA almost the entire populate_test_db.py needs to be duplicated for another study. This also means changing every test for everything else to reflect that new study existing. This seems like a huge time sink given the other more pressing things we need to do.

josenavas: @squirrelo you can use the function `Study.create` on the setUp for this test and you don't need to modify anything else....

squirrelo: It needs processed data as well, which I guess can be added directly to the study through the command line. I'll take a look at this route.

josenavas: you don't need to go to the command line. On the setUp for the test you can use all the `*.create` functions that you need to update all the data.

squirrelo: oh right ,forgot that functionality was added.

antgonza:
@squirrelo is this ready for another pass?

squirrelo:
yup

qiita_db/test/test_analysis.py
```diff
@@ -117,15 +123,98 @@ def test_retrieve_samples(self):
 
     def test_retrieve_dropped_samples(self):
         biom_fp = join(self.fp, "1_analysis_18S.biom")
+
+        # Create and poppulate second study to do test with
```
antgonza: populate

qiita_db/test/test_analysis.py
```diff
@@ -117,15 +123,98 @@ def test_retrieve_samples(self):
...
+                "(1,2,'2.SKB8.640193'), (1,2,'2.SKD8.640184'), "
+                "(1,2,'2.SKB7.640196')")
+        except Exception as e:
+            sec_fp = join(mp, "2_2_study_1001_closed_reference_otu_table.biom")
+            if exists(sec_fp):
+                move(sec_fp,
+                     join(mp, "2_study_1001_closed_reference_otu_table.biom"))
+            raise RuntimeError(e)
+
+        try:
```
antgonza: This try doesn't have an except ...

qiita_db/test/test_analysis.py
```diff
@@ -117,15 +123,98 @@ def test_retrieve_samples(self):
...
+                            'longitude': 4.41},
+            }
+        metadata = pd.DataFrame.from_dict(metadata_dict, orient='index')
+
+        Study.create(User("test@foo.bar"), "Test study 2", [1], info)
+
+        SampleTemplate.create(metadata, Study(2))
+
+        mp = get_mountpoint("processed_data")[0][1]
         try:
```
antgonza: This seems weird, is there a better way to test this than having a try/except?

squirrelo: it's not try-except, it's try-finally. This way if the test fails or something screws up the files are still cleaned up.

qiita_db/test/test_analysis.py
squirrelo: doesn't need one, it's a try-finally for cleanup.

qiita_db/test/test_analysis.py
antgonza: This one is an actual exception (`except Exception as e`) but I think it will be better to catch the errors that will be raised as this is a test and it should test the specific possibilities. Same for the other one. Other option is to simply test what we expect to go wrong vs. catching all. IMOO it's fine to leave a file or two if the test fail, they will be overwritten and cleaned once the test pass, right?

squirrelo: Nope, this does permanent file changes that must be cleaned up even on failure or else the tests will always fail. Hence the catching exceptions, cleaning up, then raising the exception so you can see what the error really is. This is necessary to have the test exist, which is what you asked for.

antgonza: After going over the code with @squirrelo we agreed is fine but I think we need a small block of text, comment, to explain why we have these try/finally so when a new dev needs to change this code it reads an explanation vs. needing to figure it out by some intraline one liner comments. Note that this test only became clear to me after an online explanation and I think is worth to have it documented so this doesn't get lost.

josenavas: I think this is not the correct way to go. There is a reason why the function `tearDown` exists on the tests and it is to do any clean up needed

antgonza: Excellent point!

squirrelo: Problem is the teardown is different for every test, so to me it makes more sense to just do it in each test.

antgonza: An option will be to create another class just for these tests and
have the special teardown in there.

josenavas: I agree with @antgonza 
If a test is too complex, having it in it's own class makes sense. Also is not worth to re-invent the wheel. :smiley:

antgonza:
:+1:

josenavas:
Thanks @squirrelo

PR #796, opened 2015-01-09
Make sure processed data exists before displaying datatype. Fix #784

### LINKED ISSUE #784: Searching studies with multiple datatypes can error
If you have a study with processed data of one data type (say 16S) and just raw data/preprocessed data  of another type (say ITS), searching for that study will give you an error. This comes from the fact that the study says it has ITS data, but it is not completely processed yet.

squirrelo:
this adds an if statement to make sure we don't index a non-existant list.

### DIFF
--- a/qiita_pet/templates/search_studies.html
+++ b/qiita_pet/templates/search_studies.html
@@ -98,17 +98,18 @@ function add_metacat(metacat) {
 
     <table class='table table-hover' style="width:100%">
       <tr><th>Data Types</th><th>Name</th><th>Selected</th><th>Description</th><th>Status</th></tr>
-      {% for sid, samples in viewitems(results) %}
-        {% set study = Study(sid) %}
+{% for sid, samples in viewitems(results) %}
+  {% set study = Study(sid) %}
         <tr id="study{{sid}}"><td>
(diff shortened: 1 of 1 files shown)


antgonza:
This works as expected but why solve this after the query is retrieved vs. solving within the query?

squirrelo:
Because I'm trying to leave the search functionality as generic as possible so it can be used for other things besides just the analysis pipeline.

antgonza:
Not sure that's the way to go. In previous experiences, we have seen that this kind of generalization causes performance bottlenecks. We can discuss during today's meeting.

squirrelo:
This is literally just adding an if statement. Not sure how that causes a performance bottleneck.

antgonza:
Got it, thought the loop was on the studies, which can be a large
number, vs. the datatypes that is minimal and my concern is looping on
all the studies and discarding those that do not satisfy the if
statement.

squirrelo:
Ok. I'm adding another fix in here really fast and then it should be ready to go.

antgonza:
Thanks.

squirrelo:
Good to go once tests pass

antgonza:
Tests not passing.

PR #797, opened 2015-01-09
More informative reset password messages. Fix #788

### LINKED ISSUE #788: Reset password should be more informative
* Add a note in the reset message and email that the link is only valid for 30 mins
* more informative error when can't reset than "COULD NOT RESET"

squirrelo:
This adds the 30 minute timeout information to the email sent and a more informative message when the user's password is not reset.

### DIFF
--- a/qiita_pet/handlers/user_handlers.py
+++ b/qiita_pet/handlers/user_handlers.py
@@ -7,6 +7,7 @@ from qiita_db.user import User
 from qiita_db.logger import LogEntry
 from qiita_db.exceptions import QiitaDBUnknownIDError
 from qiita_core.util import send_email
+from qiita_core.qiita_settings import qiita_config
 
 
 class UserProfile(Form):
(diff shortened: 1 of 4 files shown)


wasade:
Maybe 24hr? I've never seen a site do 30 min for this before

wasade:
Oh, I see, just expands the message

qiita_pet/handlers/user_handlers.py
```diff
@@ -93,8 +93,11 @@ def post(self):
             info = user.info
             try:
                 send_email(user.id, "Qiita: Password Reset", "Please go to "
-                           "the following URL to reset your password: "
-                           "http://qiita.colorado.edu/auth/reset/%s" %
+                           "the following URL to reset your password: \n"
+                           "http://qiita.colorado.edu/auth/reset/%s  \nYou "
```
wasade: The hostname here should be sourced from the configuration file as we cannot assume qiita is running from under colorado.edu

antgonza: Agree, @squirrelo could you fix?

squirrelo:
Good to go once tests pass

Issue #599, opened 2014-11-13
make upload file merge run as a job


antgonza:
Currently the upload uploads chunks of data, which is fine, but the merge happens in tornado, which should be blocking, which is not fine.

PR #801, opened 2015-01-16
Upload fixes

antgonza:
This solves #599 and partially solves #795. Will finish #795 in another PR.


antgonza:
This is ready for review: @ElDeveloper @squirrelo @adamrp, thanks!

ElDeveloper:
hehehe 
screen shot 2015-01-18 at 9 53 14 am

qiita_pet/handlers/upload.py
```diff
@@ -110,19 +112,14 @@ def get(self):
...
-        _, fp = get_mountpoint("uploads")[0]
-        tfp = join(fp, study_id,
-                   resumable_filename + '.part.' + resumable_chunk_number)
-
-        if exists(tfp):
-            self.set_status(200)
-        else:
-            self.set_status(400)
+        # in the original version we used to check if a chunk was already
+        # uploaded and if it was we will send self.set_status(200). Now, as
```
ElDeveloper: will -> would

qiita_pet/handlers/upload.py
```diff
@@ -110,19 +112,14 @@ def get(self):
...
+        # we are not chuncking by file we can simply pass the no exists
```
ElDeveloper: chuncking -> chunking ... I realize these are my misspellings :P

qiita_db/test/test_util.py
```diff
@@ -470,11 +470,19 @@ def test_get_mountpoint(self):
...
         exp = ['uploaded_file.txt']
         obs = get_files_from_uploads_folders("1")
         self.assertEqual(obs, exp)
 
+        # something has been uploaded and not ignoring hidden files/folders
+        # and folders
+        exp = ['.hidden_file.txt', 'a_folder', 'uploaded_file.txt']
+        obs = get_files_from_uploads_folders("1",
+                                             ignore_hidden_and_folders=False)
+        self.assertEqual(sorted(obs), sorted(exp))
```
ElDeveloper: You should use [`assertItemsEqual`](https://docs.python.org/2/library/unittest.html#unittest.TestCase.assertItemsEqual).

ElDeveloper:
:+1:

antgonza:
Ready for final review/merge!

qiita_db/util.py
```diff
@@ -506,13 +506,15 @@ def compute_checksum(path):
...
 
-def get_files_from_uploads_folders(study_id):
+def get_files_from_uploads_folders(study_id, ignore_hidden_and_folders=True):
     """Retrive files in upload folders
 
     Parameters
     ----------
     study_id : str
         The study id of which to retrive all upload folders
+    ignore_hidden_and_folders : bool
```
squirrelo: Please update as optional with default True.

qiita_pet/static/vendor/js/resumable-uploader.js
```diff
@@ -50,10 +50,10 @@
        this.resumable = new Resumable({
            chunkSize:3*1024*1024,
            maxFileSize:this.maxFileSize*1024*1024*1024,
-           simultaneousUploads: 6,
+           simultaneousUploads: 1,
```
squirrelo: Why the drop in uploads?

qiita_db/util.py
```diff
@@ -506,13 +506,15 @@ def compute_checksum(path):
     return crc & 0xffffffff
 
 
-def get_files_from_uploads_folders(study_id):
+def get_files_from_uploads_folders(study_id, ignore_hidden_and_folders=True):
```
squirrelo: Would it be better to have this as two bools, one for hidden and one for folders? Would be more useful in the long term, especially if we start allowing folder uploads.

antgonza: Tried to do that but the logic was a bit odd and decided to leave it as a single one. Note that we are not using anywhere this functionality and if we actually need it we can figure it out then.

qiita_db/util.py
antgonza: done

qiita_pet/static/vendor/js/resumable-uploader.js
antgonza: In the past we were saving all chunks which allowed us to do parallel uploads. Now, we save everything in the same file and we need the chucks in a defined order, thus 1 by 1.

qiita_db/util.py
squirrelo: Thought we were moving away from that type of logic for programming, but if everyone else agrees, fine.

antgonza: Not sure what you mean, wanna clarify? We could also discuss during
meeting, if easier.

antgonza:
Ready for merge.

PR #804, opened 2015-01-19
fix #703

### LINKED ISSUE #703: Create Study: Study Title should have text on right side with suggested format
One possible option:

Recommended: PI last name followed by brief project description.
e.g., Darwin Galapagos Finch Feces

antgonza:


qiita_pet/templates/edit_study.html
```diff
@@ -237,12 +237,15 @@
     {% set kwargs = {'tabindex': idx, 'class_': 'form-control', 'required': form_item.flags.required} %}
     {% set additional_info = '' %}
     {# form-item-specific rendering #}
+    {% if form_item.label.text == 'Study Title' %}
+      {% set additional_info = '<u>Recommended format:</u><br/>PI last name brief project description<br/>For example: <i>Darwin Galapagos Finch Feces</i><br/><br/>' %}
```
squirrelo: Do we ever show study titles without the PI attached? Is the PI necessary in the title?

antgonza: Remember this is a suggestion (not enforced) which we agreed on during
a lab meeting so searching is easier.

squirrelo: I can add the PI to the search though and that will make it easier on everyone. To me it makes more sense to do `PI = Rob Knight` than 'study_title includes Knight`.

antgonza: That will be great but separate to this _simple_ suggestion that we
all agreed on.

ElDeveloper:
Given that we cannot enforce this naming convention, we risk having inconsistent names across the system.

PR #809, opened 2015-01-22
Add reference name and version to selection for processing

squirrelo:
The current setup is pretty much already set for reference changes, so this adds that information to the params displayed.

To be clear: If the reference is added to the reference table, then used in a params row, it will show up as a new parameter because it is in that row.

### DIFF
--- a/qiita_pet/util.py
+++ b/qiita_pet/util.py
@@ -23,6 +23,7 @@ Methods
 # -----------------------------------------------------------------------------
 from qiita_db.reference import Reference
 
+
 def linkify(link_template, item):
     """Formats a strings into a URL using string replacement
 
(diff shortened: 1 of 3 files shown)


squirrelo:
Good for review

antgonza:
:+1: can someone else review?

qiita_pet/util.py
```diff
@@ -65,6 +66,9 @@ def generate_param_str(param):
     str
         The html string with the parameter set values
     """
-    result = ["<b>%s:</b> %s" % (name, value)
-              for name, value in param.values.items()]
+    ref = Reference(param.reference)
+    result = ["<b>Reference:</b> %s %s" % (ref.name, ref.version)]
+    result.extend("<b>%s:</b> %s" % (name, value)
+                  for name, value in param.values.items()
```
ElDeveloper: Should this be using `viewitems`?

squirrelo: yup, fixed

ElDeveloper:
:+1:

PR #815, opened 2015-02-01
Erasing upload files

antgonza:


qiita_db/util.py
```diff
@@ -520,15 +521,53 @@ def get_files_from_uploads_folders(study_id):
         List of the filepaths for upload for that study
     """
     fp = []
-    for _, p in get_mountpoint("uploads", retrive_all=True):
+    for pid, p in get_mountpoint("uploads", retrive_all=True):
         t = join(p, study_id)
         if exists(t):
-            fp.extend([f for f in listdir(t) if not f.startswith('.') and
-                      not isdir(join(t, f))])
+            fp.extend([(pid, f) for f in listdir(t) if not f.startswith('.')
```
wasade: This could be simplified with `glob.glob`. non-blocking comment

qiita_db/util.py
```diff
@@ -520,15 +521,53 @@ def get_files_from_uploads_folders(study_id):
...
+                      and not isdir(join(t, f))])
 
     return fp
 
 
+def move_upload_files_to_trash(study_id, files_to_move):
+    """Retrive files in upload folders
```
wasade: Retrive -> Retrieve

qiita_db/util.py
```diff
@@ -520,15 +521,53 @@ def get_files_from_uploads_folders(study_id):
...
+
+    Parameters
+    ----------
+    study_id : int
+        The study id of which to retrive all upload folders
```
wasade: same

qiita_db/util.py
```diff
@@ -520,15 +521,53 @@ def get_files_from_uploads_folders(study_id):
...
+
+    for fid, filename in files_to_move:
+        if fid not in folders:
+            raise QiitaDBError("Filepath not stored in the database")
+
+        foldername = join(folders[fid], str(study_id))
+        if not exists(foldername):
+            raise QiitaDBError("Filepath not stored in the database")
+
+        trashpath = join(foldername, 'trash')
```
wasade: Is it possible for negative consequences if I upload a file named `trash`? Or some variant of that pathological case?

qiita_db/util.py
```diff
@@ -520,15 +521,53 @@ def get_files_from_uploads_folders(study_id):
...
+    Raises
+    ------
+    QiitaDBError
+        If folder_id or the study folder don't exist
+    """
+    folders = {k: v for k, v in get_mountpoint("uploads", retrive_all=True)}
+
+    for fid, filename in files_to_move:
+        if fid not in folders:
+            raise QiitaDBError("Filepath not stored in the database")
```
wasade: The error messages are not all accurate. In the first case (this one), it looks like a failure can arise if the `fid` is not actually stored in the database. The second error that can arise is if the file path to the study appear to exist on the file system. The third error is if the file path to the actual file doesn't appear to exist. Possible to adjust the error messages?

qiita_pet/handlers/upload.py
```diff
@@ -44,6 +44,24 @@ def get(self, study_id):
...
+    def post(self, study_id):
+        try:
+            study = Study(int(study_id))
+        except QiitaDBUnknownIDError:
+            raise HTTPError(404, "Study %s does not exist" % study_id)
+        check_access(User(self.current_user), study, no_public=True,
+                     raise_error=True)
+
+        files_to_move = []
+        for v in self.get_arguments('files_to_erase', strip=True):
```
wasade: Is there a risk here of a malicious user injecting filepaths?

wasade: I think we're protected, but just want to make sure this point has definitely been thought of

qiita_db/util.py
josenavas: Following @wasade's comment: can we use a hidden folder name (e.g. `.trash` or `.qiita_trash`)?

wasade:
Few comments, nothing major

josenavas:
The upload page is not working for me. If I drag and drop any file, I receive this error:
screen shot 2015-02-03 at 12 00 00 pm
and the uploader no longer works. If I drag and drop any other file it doesn't do anything...

The file that I'm using is a sample template (.txt) that I always use to test the interface.

qiita_db/util.py
ElDeveloper: This string refers to folders but the function name refers to files.

antgonza:
I think this is because you are in a Mac and it has some hidden
folders that are not valid, can you check?

qiita_db/util.py
ElDeveloper: @josenavas I agree that this would be a better name but and we need to document whichever name we end up using just so others know where these files will be put.

qiita_pet/static/vendor/js/resumable-uploader.js
```diff
@@ -117,11 +117,8 @@
      // Add a new file (or rather: glue between newly added resumable files and the UI)
      this.addFile = function(resumableFile){
        // A list and and edit item for the UI
-       if (typeof resumableFile == 'object') {
-         name = resumableFile.fileName
-       } else {
-         name = resumableFile
-       }
+       fileName = resumableFile.fileName
```
ElDeveloper: Out of curiosity why was this check removed?

qiita_db/util.py
antgonza: On a qiita-dev list we agreed on not using hidden folders as they sometimes play nasty with NSF, for reference search in your email for: `deleting files and automatic jobs`

qiita_pet/handlers/upload.py
antgonza: We are cause, you can not upload a file without a specific extension.

qiita_pet/static/vendor/js/resumable-uploader.js
antgonza: In the past we were passing str (already uploaded) or object (just uploaded) and now we are always passing an object.

qiita_db/util.py
antgonza: Done

qiita_db/util.py
antgonza: Changed all occurrences.

qiita_db/util.py
antgonza: Fixing.

qiita_db/util.py
antgonza: Fixing.

qiita_db/util.py
antgonza: After testing locally I realized that glob returns fullpaths which will make a bunch of the logic on other sections to change, so decided to return to listdir.

qiita_pet/handlers/upload.py
antgonza: Oopps, this answer was about the `trash` name, not about injections. Injections could happen but if you see it only can erase files within the study_id folder to which your user have access to.

antgonza:
I have addressed comments, except @josenavas cause not sure what's going on.

qiita_db/util.py
wasade: Issue isn't with respect to files or directories prefixed with '.' but if a user uploads a file that is the name "trash"

Issue with NFS is not whether files are prefixed with '.', but that it is not safe to assume files that are prefixed with '.' can be deleted.

antgonza: @wasade You shouldn't be able to upload a file named trash as the GUI
doesn't allow it as it needs a valid extension to be uploaded. Thanks
for the clarification. In your opinion, is there something else
missing in this PR?

adamrp:
I see the same thing that @josenavas sees. Note that the warning popup appears *before* I try to upload anything (it pops up as soon as I go to the upload page). Investigating further

adamrp:
On the second try, I didn't see any errors (neither when I navigated to the upload page nor when I uploaded a test directory), but got a 500 error when I tried to delete the files:

> MissingArgumentError: HTTP 400: Bad Request (Missing argument resumableIdentifier)

josenavas:
The error that I was seeing disappeared. I tried to remove a file and I received this error:
```
  File "/Users/jose/.virtualenvs/qiita/lib/python2.7/site-packages/tornado/web.py", line 1141, in _when_complete
    callback()
  File "/Users/jose/.virtualenvs/qiita/lib/python2.7/site-packages/tornado/web.py", line 1162, in _execute_method
  File "/Users/jose/.virtualenvs/qiita/lib/python2.7/site-packages/tornado/web.py", line 2297, in wrapper
  File "/Users/jose/qiime_software/QiiTa/qiita_pet/handlers/upload.py", line 59, in post
    files_to_move.append((int(v[0]), v[1]))
ValueError: invalid literal for int() with base 10: 'undefined'
```

However, I'm not able to reproduce the error again, so probably it was something weird on my file system...

adamrp:
Nope, I'm seeing that, too, @josenavas

josenavas:
Weird @adamrp , I'm not able to reproduce any of the errors that I was having before...
I started both times with the same clean environment, so not sure what is going on...

adamrp:
Hmmm... I'm also seeing this lots of times (more than 100 times) from tornado:

> WARNING:tornado.general:Invalid multipart/form-data: no final boundary
WARNING:tornado.general:400 POST /upload/ (::1): Missing argument resumableIdentifier
WARNING:tornado.access:400 POST /upload/ (::1) 42.95ms

qiita_db/util.py
wasade: Conceptually this is a problem: the qiita_db code then requires GUI
validation.

antgonza:
I'm able to reproduce the extension error, I think the issue was a
cached js. Anyway, new push should solve it. Thanks for catching this.

qiita_db/util.py
antgonza: OK, suggestions?

qiita_pet/static/vendor/js/resumable-uploader.js
ElDeveloper: Got it thanks!

qiita_db/util.py
wasade: What about changing "trash" to a centralized variable as it is used in a
few places, and add an explicit check to make sure that `filename` is not
equal to that variable?

adamrp:
That helped, @antgonza, I don't *think* I'm getting the `Not a valid extension: ! please try again` error any more. However, when I try to delete an uploaded file, I still get a 500 server error with this message

>***************************************
undefined-test.fastq
***************************************
ERROR:tornado.application:Uncaught exception POST /study/upload/2 (::1)
HTTPRequest(protocol='http', host='localhost:8888', method='POST', uri='/study/upload/2', version='HTTP/1.1', remote_ip='::1', headers={'Origin': 'http://localhost:8888', 'Content-Length': '35', 'Accept-Language': 'en-US,en;q=0.8', 'Accept-Encoding': 'gzip, deflate', 'Host': 'localhost:8888', 'Accept': 'text/html,application/xhtml+xml,application/xml;q=0.9,image/webp,*/*;q=0.8', 'User-Agent': 'Mozilla/5.0 (Macintosh; Intel Mac OS X 10_8_5) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/39.0.2171.95 Safari/537.36', 'Connection': 'keep-alive', 'Referer': 'http://localhost:8888/study/upload/2', 'Cache-Control': 'max-age=0', 'Cookie': 'user="InRlc3RAZm9vLmJhciI=|1423000443|8b505be2ea82b03800a6ad3cb9d8956bab6b58c9"', 'Content-Type': 'application/x-www-form-urlencoded'})
  File "/Users/adam/.virtualenvs/python276/lib/python2.7/site-packages/tornado/web.py", line 1218, in _when_complete
    callback()
  File "/Users/adam/.virtualenvs/python276/lib/python2.7/site-packages/tornado/web.py", line 1239, in _execute_method
  File "/Users/adam/.virtualenvs/python276/lib/python2.7/site-packages/tornado/web.py", line 2395, in wrapper
  File "/Users/adam/git/qiita/qiita_pet/handlers/upload.py", line 62, in post
    files_to_move.append((int(v[0]), v[1]))
ValueError: invalid literal for int() with base 10: 'undefined'

Note that the value between the `****************` lines is the value of `v` from line 57 of the upload.py handler -- it looks like the `study_id` is not being transmitted correctly to the handler.

adamrp:
Note that when I try again to delete the file, it works:

```
***************************************
7-test.fastq
***************************************
```

adamrp:
Wait, no -- When I try to upload a directory, I get the `Not a valid extension` error, and the webserver continuously outputs lines like these until I close the tab (even if I stop the webserver and start it again):

>WARNING:tornado.access:400 POST /upload/ (::1) 48.86ms
WARNING:tornado.general:Invalid multipart/form-data: no final boundary
WARNING:tornado.general:400 POST /upload/ (::1): Missing argument resumableIdentifier
WARNING:tornado.access:400 POST /upload/ (::1) 42.53ms
WARNING:tornado.general:Invalid multipart/form-data: no final boundary
WARNING:tornado.general:400 POST /upload/ (::1): Missing argument resumableIdentifier
WARNING:tornado.access:400 POST /upload/ (::1) 43.21ms
WARNING:tornado.general:Invalid multipart/form-data: no final boundary
WARNING:tornado.general:400 POST /upload/ (::1): Missing argument resumableIdentifier
WARNING:tornado.access:400 POST /upload/ (::1) 41.58ms
WARNING:tornado.general:Invalid multipart/form-data: no final boundary
WARNING:tornado.general:400 POST /upload/ (::1): Missing argument resumableIdentifier

qiita_db/util.py
antgonza: Where do you want to have the centralized variable? And the name check
will only be for erase?

adamrp:
Okay, I there are several bugs here. Here is how to reproduce 2 of them (at least on my system...):

**invalid value for cast to int**
1) create a new sandbox study
2) upload a single file
3) try to delete the file *without reloading the page or navigating away*
The id for the file is `undefined`. If you do steps 1 and 2, then go back to the study description, then click upload again, then delete the file, it works correctly.

**continuous `Missing argument resumableIdentifier` output**
1) create new sandbox study
2) attempt to upload a directory -- should get the `Not a valid extension` message
3) attempt to upload a single file -- it hangs and I see the continuous messages from tornado until I close the tab

antgonza:
I think that's a problem with your configuration, where does your
BASE_DATA_DIR point to and do you have a folder named upload? I'm not
able to reproduce that error. Which browser are you using?

adamrp:
`BASE_DATA_DIR = /Users/adam/git/qiita/qiita_db/support_files/test_data/` And I do see the files being uploaded to that location (in the `uploads/<study_id>` subdirectory) when it is successful.

antgonza:
Got it, will see if can fix, the issue is how resumable works.

adamrp:
okay, thanks @antgonza! Let me know when you would like additional testing.

qiita_db/test/test_job.py
```diff
@@ -232,11 +232,11 @@ def test_create_exists_return_existing(self):
         self.assertEqual(new.id, 2)
 
     def test_retrieve_datatype(self):
-        """Makes sure datatype retriveal is correct"""
+        """Makes sure datatype retrieveal is correct"""
```
adamrp: minor: this is spelled `retrieval`

qiita_db/util.py
wasade: If its used just within the function then a local variable. If its used in
multiple functions, then at the top level of `util.py`. I think it makes
sense for only having the name check on erase?

antgonza: Cool, will do.

antgonza:
Ready for another pass. AFAIK have addressed all comments.

qiita_db/test/test_job.py
```diff
@@ -232,11 +232,11 @@ def test_create_exists_return_existing(self):
         self.assertEqual(new.id, 2)
 
     def test_retrieve_datatype(self):
-        """Makes sure datatype retriveal is correct"""
+        """Makes sure datatype retrieval is correct"""
         self.assertEqual(self.job.datatype, '18S')
 
     def test_retrieve_command(self):
-        """Makes sure command retriveal is correct"""
+        """Makes sure command retrieveal is correct"""
```
adamrp: This is also spelled `retrieval`

adamrp:
This is much more stable now. I cannot upload directories, though -- I still get the `Not a valid extension` error (although I no longer observe the second bug I mentioned previously, so that's good). Is there something in my settings that might be causing this?  I have

`VALID_UPLOAD_EXTENSION = fastq,fastq.gz,txt,sff,fna,qual`

antgonza:
You shouldn't be able to upload directories. However, you could select
several files and upload all of them.

adamrp:
Oh okay, that explains it!  I thought that was one of the goals of the PR, my mistake.  In that case, the code looks good, and it works. :+1:

antgonza:
@wasade @ElDeveloper @josenavas Anything else? Can we merge?

josenavas:
No issues on this end! Probably good if @wasade or @ElDeveloper can also check on their machines since in this PR we have seen weird behaviors on different machines. Thanks @antgonza

ElDeveloper:
Two minor comments that I don't think would block merging this pull request:

- Changing the glyphicon of the delete button to `glyphicon-trash`.
- The delete button should only be there if there are files to be deleted.
- The delete button should be red i. e. have a `danger` class.

Sorry for not getting these comments out before, I'm ok if we want to defer them.

antgonza:
1 and 3 are easy and will be done in the next push. 2 will require a
lot of JS so not gonna do.

On Thu, Feb 5, 2015 at 1:07 PM, Yoshiki Vázquez Baeza
<notifications@github.com> wrote:
> Two minor comments that I don't think would block merging this pull request:
>
> Changing the glyphicon of the delete button to glyphicon-trash.
> The delete button should only be there if there are files to be deleted.
> The delete button should be red i. e. have a danger class.
>
> Sorry for not getting these comments out before, I'm ok if we want to defer
> them.
>
> —
> Reply to this email directly or view it on GitHub.



-- 
Antonio

josenavas:
Is it worth to add an issue for @ElDeveloper's comment 2?

antgonza:
AFAIK no.

ElDeveloper:
The build genuinely failed, it seems like User is not imported :confused:

antgonza:
The issue was due to `pull upstream master`, should be fixed now.

ElDeveloper:
:+1: to merge after tests are done.

PR #821, opened 2015-02-04
fix #748

### LINKED ISSUE #748: when unlinking files do not remove but return to upload folder

antgonza:


antgonza:
This is ready for review.

qiita_db/util.py
```diff
@@ -679,6 +681,36 @@ def purge_filepaths(conn_handler=None):
                 remove(fp)
 
 
+def move_filepaths_to_upload_folder(study_id, filepaths, conn_handler=None):
+    r"""Goes over the filepaths list and moves all the filepaths that are not
+    used in any place to the upload folder of the study
```
wasade: I'm not seeing in the code anything that covers the "...not used in any place..." aspect of the doc string. Is this accurate?

antgonza: Left that text by mistake, it only moves specific paths.

qiita_db/data.py
```diff
@@ -416,6 +416,17 @@ def _remove_filepath(self, fp, conn_handler, queue):
             self._set_link_filepaths_status("failed: %s" % msg)
             raise QiitaDBError(msg)
 
+        # The filepath belongs to one or more studies
+        studies_linked = self.studies
```
josenavas: I don't think this check should be here. This is avoiding to unlink any files from the raw data if the raw data is being used by another study. Is this the desired behavior?
If so, there is no mechanism in place to remove a raw data from a study; so once a raw data has been shared, you can add as many files as you want, but you cannot unlink them... I think this is a weird behavior.

antgonza: I think is fine to have it here but I see your point about adding more files. IMOO the best behavior will be to not allow linking more file paths if some have been already added. If you really want to add more you will have to unlink and then link again.

josenavas: I don't think that behavior is desired. A user could link 3 files (because he forgot to add one of the barcodes e.g.) and he just needs to add one more. The point that I'm making is that if you're not able to unlink files because the raw data is used in another study, you should not be able to link them either. Does that make sense?

antgonza: @josenavas and I just discussed offline and I will remove the
possibility of linking more files via the GUI to a raw data if it has
been shared. One of my next PRs will be to erase a raw data from a
study. This will allow that if a raw data is shared between n studies,
n-1 studies could remove the raw data, and the last one can
link/unlink files.

antgonza:
Ready for another pass.

qiita_db/util.py
```diff
@@ -727,6 +729,36 @@ def purge_filepaths(conn_handler=None):
...
+    filepaths : list
+        List of filepaths to move to the upload folder
+    conn_handler : SQLConnectionHandler, optional
+        The connection handler object connected to the DB
+    """
+    conn_handler = conn_handler if conn_handler else SQLConnectionHandler()
+    uploads_fp = join(get_mountpoint("uploads")[0][1], str(study_id))
+
+    # We can now go over and remove all the filepaths
+    for fp_id, fp, fp_type in filepaths:
```
ElDeveloper: `fp_type` is not being used anywhere in this loop, for clarity it may make sense to put `_`. Not a blocking comment though.

ElDeveloper:
:+1:

ElDeveloper:
Works like a charm!

qiita_db/data.py
```diff
@@ -498,11 +512,11 @@ def remove_filepath(self, fp):
...
         # We can already update the status to done, as the files have been
-        # unlinked, the purge_filepaths call will not change the status
-        # of the raw data object
+        # unlinked, the move_filepaths_to_upload_folder call will not change
+        # the status of the raw data object
         self._set_link_filepaths_status("idle")
 
-        # Delete the file if it is not used anywhere
+        # Delete the files, if they are not used anywhere
         purge_filepaths(conn_handler)
```
josenavas: Sorry for not commenting this earlier, but this function and the one above are very similar, the only difference is that `remove_filepath` only removes one filepath, while `clear_filepaths` removes all of them. In that case, should this function use the `move_filepaths_to_upload_folder` instead of the `purge_filepaths` one?

josenavas:
Just one more comment @antgonza

qiita_db/data.py
antgonza: It could but decided not to touch that one because the test will
become trickier and it's not used anywhere in the actual code.

josenavas: We should change that at some point, since the code is not used anywhere, I think that adding an issue to keep track of this will be enough

ElDeveloper: Can one of you merge when all remaining comments in this issue are resolved?

josenavas: Yup, I created https://github.com/biocore/qiita/issues/836 for this

PR #823, opened 2015-02-05
BUG: Add warning when removing empty columns

### LINKED ISSUE #740: load_template_to_dataframe silently drops columns
If they are all `NA`, see [here](https://github.com/biocore/qiita/blob/master/qiita_db/metadata_template.py#L1748)

ElDeveloper:
A UserWarning is now raised whenever load_template_to_dataframe drops a column
if it is empty. This message is shown on the interface as well.

Fixes #740

### DIFF
--- a/qiita_db/test/test_metadata_template.py
+++ b/qiita_db/test/test_metadata_template.py
@@ -16,6 +16,7 @@ from os import close, remove
 from os.path import join, basename
 from collections import Iterable
 
+import numpy.testing as npt
 import pandas as pd
 from pandas.util.testing import assert_frame_equal
 
@@ -1445,8 +1446,8 @@ class TestUtilities(TestCase):
         assert_frame_equal(obs, exp)
 
     def test_load_template_to_dataframe_empty_columns(self):
-        obs = load_template_to_dataframe(
-            StringIO(EXP_SAMPLE_TEMPLATE_SPACES_EMPTY_COLUMN))
+        obs = npt.assert_warns(UserWarning, load_template_to_dataframe,
+                               StringIO(EXP_ST_SPACES_EMPTY_COLUMN))
         exp = pd.DataFrame.from_dict(SAMPLE_TEMPLATE_DICT_FORM)
         exp.index.name = 'sample_name'
         assert_frame_equal(obs, exp)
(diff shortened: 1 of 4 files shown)


qiita_db/test/test_metadata_template.py
```diff
@@ -1481,6 +1482,21 @@ def test_load_template_to_dataframe_empty_sample_names(self):
         exp.index.name = 'sample_name'
         assert_frame_equal(obs, exp)
 
+    def test_load_template_to_dataframe_empty_column(self):
+
+            obs = npt.assert_warns(UserWarning, load_template_to_dataframe,
```
antgonza: Block is over indented, interesting that flake8 didn't complain.

qiita_pet/handlers/study_handlers/description_handlers.py
```diff
@@ -280,9 +295,20 @@ def add_prep_template(self, study, user, callback):
...
+                # deleting previous uploads and inserting new one
+                pt_id = self.remove_add_prep_template(fp_rpt, raw_data_id,
+                                                      study, data_type_id,
+                                                      investigation_type)
+
+                # join all the warning messages into one. Note that this info
+                # will be ignored if an exception is raised
+                if warns:
+                    msg = '; '.join([str(w.message) for w in warns])
+                    msg_level = 'danger'
```
antgonza: If this is a warning, should it be warning level vs. error?

antgonza: While testing I realized that danger is actually the warning color so ignore.

antgonza:
One comment that is not blocking so :+1:

ElDeveloper:
I've addressed your comments, thanks! Note that the level was wrong,
should have been warning, not danger. Not sure why did it show you the
correct level. Maybe your browser cached the CSS value?

ElDeveloper:
If someone else can review this PR, it should be ready for merge.

josenavas:
@ElDeveloper which is the expected color of the message? From the code, it looks like it should be yellow (`warning`). However, when it shows up as `info` level (light blue):
screen shot 2015-02-05 at 5 44 27 pm

josenavas:
Looks like the problem is [here] (https://github.com/biocore/qiita/blob/master/qiita_pet/templates/sitebase.html#L101) in sitebase.html

ElDeveloper:
Thanks for catching that, I tested it but did not pay much attention to
the color and ended up blaming my browser. I'll fix this.

josenavas:
Thanks @ElDeveloper It works perfect now:
screen shot 2015-02-05 at 6 20 57 pm

ElDeveloper:
Thanks!

Issue #822, opened 2015-02-05
update/modify metadata


antgonza:
I think we need to be able to update metadata (update values) at any point or status of the samples. However, we should have some limitations based on adding/removing samples. Add allowed at any point but remove will only be allowed if not public. 

Some other pointers from @josenavas:
Correcting data of existing studies (i.e. host_common_name is unknown but it has to have a different value) and adding new samples (ie. more samples have been collected, they are added to the sample template and a new prep template is added). Also allow

wasade:
Will be adding in methods at the `qiita_db` level shortly. Have logic/tests in place but working on improving consistency with the existing API

antgonza:
Just so whoever address this knows: the main backend functionality was added by @wasade here https://github.com/biocore/qiita/pull/824

squirrelo:
I'm going to subsume this as also the ability to add samples to a pre-existing study, as that came up in the meeting today.

josenavas:
Re-opening. Currently it only applies to sample templates. This should be able in prep templates, specially for missing data in case of EBI submission (e.g. `center_name`)

squirrelo:
Can you add this to the alpha 0.1 or 0.2 zenhub queue, whichever is appropriate?

josenavas:
I'm not using zenhub sorry...

josenavas:
We are able to modify everything, so I think this can be safely closed. I open a new issue to keep track of the UI changes to allow this on prep templates (https://github.com/biocore/qiita/issues/1130) and another issue to keep track of the user notification when the template information changes (https://github.com/biocore/qiita/issues/1129)

Issue #790, opened 2015-01-08
Cannot download and re-upload a prep template


ChristineLowe:
Hi,

During the MoBE workshop I loaded a prep workbook that successfully demultiplexed my data.  I downloaded the Prep template and tried to reupload it as a test after having troubles with the demultiplexing.  When you try to re-add this prep template in the raw data tab you get the error message:

An error occurred parsing the prep template: 10079_prep_146_20150107-145042.txt
[...]

PR #824, opened 2015-02-05
ENH/API: methods to update sample metadata

wasade:
Partially addresses #822. These changes include:

* updating of existing sample metadata
* adding new columns
* removing columns

Right now, this is limited to the SampleTemplate, and support is not in
for the PrepTemplate. This would be easy to change, but put it in place
this way based on the email traffic I saw regarding the issue.

The API for get/set/del at the template level is not consistent yet. Pushing
in now to exercise full tests, and will issue follow up commits to resolve the
inconsistencies.

### DIFF
--- a/qiita_db/metadata_template.py
+++ b/qiita_db/metadata_template.py
@@ -1372,7 +1372,7 @@ class SampleTemplate(MetadataTemplate):
             NOT NULL DEFAULT '{3}'""".format(table_name, category, dtype,
                                              default))
 
-        self[column] = values
+        self[category] = values
 
 
 class PrepTemplate(MetadataTemplate):
--- a/qiita_db/metadata_template.py
+++ b/qiita_db/metadata_template.py
@@ -1287,13 +1287,13 @@ class SampleTemplate(MetadataTemplate):
         """
         return self._id
 
-    def __delitem__(self, column):
-        """Remove a column from the sample template
+    def remove_category(self, category):
+        """Remove a category from the sample template
 
         Parameters
         ----------
-        column : str
-            The column to remove
+        category : str
+            The category to remove
 
         Raises
         ------
@@ -1303,24 +1303,24 @@ class SampleTemplate(MetadataTemplate):
         table_name = self._table_name(self.study_id)
         conn_handler = SQLConnectionHandler()
 
-        if column not in self.categories():
+        if category not in self.categories():
             raise QiitaDBColumnError("Column %s does not exist in %s" %
-                                     (column, table_name))
+                                     (category, table_name))
 
         # This operation may invalidate another user's perspective on the
         # table
         conn_handler.execute("""
             ALTER TABLE qiita.{0} DROP COLUMN {1}""".format(table_name,
-                                                            column))
+                                                            category))
 
-    def __setitem__(self, column, values):
+    def update_category(self, category, samples_and_values):
         """Update an existing column
 
         Parameters
         ----------
-        column : str
-            The column to add or update
-        values : dict
+        category : str
+            The category to update
+        samples_and_values : dict
             A mapping of {sample_id: value}
 
         Raises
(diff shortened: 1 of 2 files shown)


antgonza:
Thanks! Are you planing to also add a front end (modify GUI) for these updates?

wasade:
Out of scope for this PR, and had not planned on working on the GUI

wasade:
Is that a stochastic test failure?

wasade:
Ready for review

qiita_db/metadata_template.py
```diff
@@ -1232,6 +1287,93 @@ def study_id(self):
...
+            If the column does not exist in the table
+        """
+        table_name = self._table_name(self.study_id)
+        conn_handler = SQLConnectionHandler()
+
+        if category not in self.categories():
+            raise QiitaDBColumnError("Column %s does not exist in %s" %
+                                     (category, table_name))
+
+        # This operation may invalidate another user's perspective on the
```
ElDeveloper: What is this comment referring to?

ElDeveloper:
:+1:

qiita_db/metadata_template.py
wasade: If user A is looking at study FOO, and user B deletes a column out of FOO, the user A's view is incorrect.

ElDeveloper: Got it, thanks for the explanation!

ElDeveloper:
If someone else wants to review/merge this should be ready.

qiita_db/metadata_template.py
```diff
@@ -579,6 +579,40 @@ def _check_template_class(self, md_template):
...
+        QiitaDBColumnError
+            If the column does not exist in the table
+        """
+        conn_handler = SQLConnectionHandler()
+
+        exists = conn_handler.execute_fetchone("""
+            SELECT column_name
+            FROM information_schema.columns
+            WHERE table_name='{0}'
+                AND column_name='{1}'""".format(self._dynamic_table, column))
```
adamrp: As unlikely as it might be for a (table, column) to exist in a schema other than `qiita`, I feel that this query should also verify that the table_catalog = qiita and the table_schema = qiita

qiita_db/metadata_template.py
```diff
@@ -1051,6 +1085,27 @@ def get_filepaths(self, conn_handler=None):
 
         return [(fpid, base_fp(fp)) for fpid, fp in filepath_ids]
 
+    def categories(self):
```
adamrp: This *could* be a property; should it be?

qiita_db/metadata_template.py
```diff
@@ -1051,6 +1085,27 @@ def get_filepaths(self, conn_handler=None):
...
+        """Get the categories associated with self
+
+        Returns
+        -------
+        set
```
adamrp: Is it conceivable that column order might matter in some contexts? `information_schema.columns` incldues an `ordinal_position` column that could potentially be used to provide the order in an output `list` instead of an unordered `set`

qiita_db/metadata_template.py
wasade: Was wondering how to restrict it like that...

qiita_db/metadata_template.py
wasade: Primary use is test of membership though

qiita_db/metadata_template.py
```diff
@@ -1051,6 +1085,27 @@ def get_filepaths(self, conn_handler=None):
...
+            The set of categories associated with self
+        """
+        conn_handler = SQLConnectionHandler()
+        table_name = self._table_name(self.study_id)
+
+        raw = conn_handler.execute_fetchall("""
+            SELECT column_name
+            FROM information_schema.columns
+            WHERE table_name='{0}'""".format(table_name))
```
adamrp: similar comment to above regarding verifying `table_schema` and `table_catalog` are as expected (equal to `'qiita'`)

qiita_db/metadata_template.py
wasade: ...at least in this PR

adamrp: One possible use case would be for round-tripping users' metadata files (see issue #790), so they can download exactly what they uploaded at a future date

adamrp:
Just a few comments/questions, but code looks great, and I agree that adding GUI elements can be a separate issue (although we should probably create a separate issue to track that).

qiita_db/metadata_template.py
wasade: Done. Note, it is just `table_schema`, restricting using that and `table_catalog` was resulting in no hits.

qiita_db/metadata_template.py
wasade: done

qiita_db/metadata_template.py
wasade: Syntactic sugar... its easy to forget that there is actual leg work going on when things are properties. The nice thing about keeping it a function is it encourages the developer to acknowledge that there will be overhead. We could get around this through memoization, but we have not started down that route and there are additional consistency challenges there if we pursue that.

qiita_db/metadata_template.py
wasade: The order of columns does not matter with the exception of the index (SampleID). The test for equality should respect this.

antgonza:
:+1:

wasade:
@antgonza, this got merged before Adam's comments had been merged in

PR #826, opened 2015-02-06
Adds scripts to modify the preprocessed data through CLI

josenavas:
This is needed to be able to update the preprocessed data of the EMP studies already present on the DB


qiita_db/commands.py
```diff
@@ -300,3 +304,101 @@ def load_parameters_from_cmd(name, fp, table):
...
+
+    Returns
+    -------
+    qiita_db.PreprocessedData
+        The updated preprocessed data
+
+    Raises
+    ------
+    ValueError
+        If sl_out_dir does not contain all the required files
```
ElDeveloper: I think this specific case should go under `IOError`.

ElDeveloper: Sorry, let me be more clear about this, I think this condition should raise an IOError instead of a ValueError.

qiita_db/commands.py
```diff
@@ -300,3 +304,101 @@ def load_parameters_from_cmd(name, fp, table):
...
+        If the study does not have any preprocessed data
+
+    Notes
+    -----
+    If the study has more than one preprocessed data, the one with lower id
```
ElDeveloper: with -> with the

qiita_db/commands.py
```diff
@@ -300,3 +304,101 @@ def load_parameters_from_cmd(name, fp, table):
...
+    is updated.
+    """
+    # Check that we have all the required files
+    path_builder = partial(join, sl_out_dir)
+    new_fps = {'preprocessed_fasta': path_builder('seqs.fna'),
+               'preprocessed_fastq': path_builder('seqs.fastq'),
+               'preprocessed_demux': path_builder('seqs.demux'),
+               'log': path_builder('split_library_log.txt')}
+
+    missing_files = [key for key, val in new_fps.items() if not exists(val)]
```
ElDeveloper: items -> viewitems?

qiita_db/test/test_commands.py
```diff
@@ -462,6 +465,164 @@ def test_python_patch(self):
...
+
+        # Checksum of the fastq file
+        obs_checksum = self.conn_handler.execute_fetchone(
+            sql, (obs_fps[1][0],))[0]
+        self.assertEqual(obs_checksum, '2958832064')
+
+        # Checksum of the demux file
+        # The checksum is generated dynamically, so the checksum changes
+        # We are going to test that the checksum is not the one that was
+        # before, which correspond to an empty file
```
ElDeveloper: correspond -> corresponds

qiita_db/test/test_commands.py
```diff
@@ -462,6 +465,164 @@ def test_python_patch(self):
...
+        obs_checksum = self.conn_handler.execute_fetchone(
+            sql, (obs_fps[2][0],))[0]
+        self.assertNotEqual(obs_checksum, '852952723')
```
ElDeveloper: I don't know how likely this would be but could we be getting an empty string? Or something that is not a checksum? if likely enough, then we might want to check for length and composition of the string.

scripts/qiita
```diff
@@ -331,5 +349,18 @@ def start(port):
...
+# WARE COMMANDS
+# #############################################################################
+
+@ware.command()
+@click.argument('sl_out_dir', required=True,
+                type=click.Path(resolve_path=True, readable=True, exists=True))
+def generate_demux(sl_out_dir):
+    """Generates the demux file for the split libraries output sl_out_dir"""
+    generate_demux_file(sl_out_dir)
+    click.echo("Demux file successfully generated")
```
ElDeveloper: Maaaaayyybe worth to print the full file path of the demux file? Not a blocking comment and more of an idea.

ElDeveloper:
:+1:

qiita_db/commands.py
adamrp: :+1: @eldeveloper

qiita_db/commands.py
adamrp: both appear throughout the codebase, so I don't think it's blocking, but I do agree that we should be using `viewitems` wherever possible

qiita_db/test/test_commands.py
adamrp: I'm not sure how that would happen, although stranger things have happened -- perhaps in the future if some functions change their return type or the API changes in some other way this could become a problem? I defer to others' suggestions here.

qiita_db/commands.py
adamrp: Also, won't this become a fairly dangerous assumption as we begin to acquire/generate studies with multiple preprocessed_data?

qiita_db/commands.py
```diff
@@ -300,3 +304,101 @@ def load_parameters_from_cmd(name, fp, table):
                          "The format is PARAMETER_NAME<tab>VALUE")
 
     return constructor.create(name, **params)
+
+
+def update_preprocessed_data_from_cmd(sl_out_dir, study_id):
```
adamrp: Is it possible to use the queue system in this function to ensure that the operation is atomic? Otherwise, I fear we may run into integrity issues

scripts/qiita
```diff
@@ -239,6 +246,17 @@ def load_parameters(fp, table, name):
                % (table, param.id))
 
 
+@db.command()
+@click.argument('sl_out_dir', required=True,
+                type=click.Path(resolve_path=True, readable=True, exists=True))
```
adamrp: there is also a `file_okay` option for `click.Path`s, which in this case would be `False`

scripts/qiita
```diff
@@ -331,5 +349,18 @@ def start(port):
...
     tornado.ioloop.IOLoop.instance().start()
 
+
+# #############################################################################
+# WARE COMMANDS
+# #############################################################################
+
+@ware.command()
+@click.argument('sl_out_dir', required=True,
+                type=click.Path(resolve_path=True, readable=True, exists=True))
```
adamrp: `file_okay=False`

scripts/qiita
adamrp: :+1: could be useful information, and should be easy enough to implement

qiita_db/commands.py
antgonza: Agree.

josenavas: Good point! Done

qiita_db/commands.py
josenavas: Done!

qiita_db/commands.py
josenavas: I see your point. This is quite useful for the current use case (updating the EMP studies where we are sure that the preprocessed data that we want to modify is the one with the lower id). However, for a more general use case, this can be dangerous. What about adding an optional parameter `preprocessed_data_id` that, if passed, it updates that one instead that the one with the lowest id?

qiita_db/commands.py

qiita_db/test/test_commands.py
josenavas: I added to extra tests: the first one checks that it is a string and the second one checks that the string is not empty.

scripts/qiita
josenavas: I did not know that! Thanks!

scripts/qiita
josenavas: Done!

scripts/qiita
josenavas: Done!

josenavas:
This should be ready for another review!

qiita_db/commands.py
```diff
@@ -300,3 +305,123 @@ def load_parameters_from_cmd(name, fp, table):
...
+def update_preprocessed_data_from_cmd(sl_out_dir, study_id, ppd_id=None):
+    """Updates the preprocessed data of the study 'study_id'
+
+    Parameters
+    ----------
+    sl_out_dir : str
+        The path to the split libraries output directory
+    study_id : int
+        The study_id of the study to be updated
+    ppd_id : int
```
ElDeveloper: I think this is missing the `, optional` part.

qiita_db/commands.py
```diff
@@ -300,3 +305,123 @@ def load_parameters_from_cmd(name, fp, table):
...
+    ------
+    IOError
+        If sl_out_dir does not contain all the required files
+    ValueError
+        If the study does not have any preprocessed data
+        If ppd_id is provided and it does not belong to the given study
+
+    Notes
+    -----
+    If the study has more than one preprocessed data, the one with the lowest
```
ElDeveloper: This note is outdated.

qiita_db/commands.py
```diff
@@ -300,3 +305,123 @@ def load_parameters_from_cmd(name, fp, table):
...
+    # Get the preprocessed data to be updated
+    study = Study(study_id)
+    ppds = study.preprocessed_data()
+    if not ppds:
+        raise ValueError("Study %s does not have any preprocessed data")
+
+    if ppd_id:
+        if ppd_id not in ppds:
+            raise ValueError("The preprocessed data %d does not exist in "
+                             "study %d." % (ppd_id, study_id))
```
ElDeveloper: Unless we foresee there being way too many preprocessed data identifiers, it might be useful to print the available ids as part of this error. I understand if we don't, but it would be nice *i think*.

ElDeveloper:
Just a few comments, but looks good overall.

qiita_db/commands.py

qiita_db/commands.py
josenavas: shame on me....

qiita_db/commands.py
josenavas: Good point, done

ElDeveloper:
@josenavas this build has legitimate failures.

josenavas:
Thanks for the heads up @ElDeveloper 
This should be ready to go!

ElDeveloper:
No problem, :+1:

josenavas:
Can I have more one reviewer here??

PR #827, opened 2015-02-07
BUG: Enforce QIIME sample name naming conventions

### LINKED ISSUE #792: QIIME sample id naming constraints are not enforced
Sample identifiers can include underscores and the system will not raise an error.

ElDeveloper:
Sample and prep templates are now checked to ensure there are not invalid
characters in any sample name.

Fixes #792

### DIFF
--- a/qiita_db/metadata_template.py
+++ b/qiita_db/metadata_template.py
@@ -68,7 +68,6 @@ else:
     from string import letters, digits
 
 
-
 TARGET_GENE_DATA_TYPES = ['16S', '18S', 'ITS']
 REQUIRED_TARGET_GENE_COLS = {'barcodesequence', 'linkerprimersequence',
                              'run_prefix', 'library_construction_protocol',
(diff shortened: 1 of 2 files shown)


antgonza:
This looks fine but should it have a python patch to fix the current files in the system? If not, how should we deal with them?

ElDeveloper:
We can't "patch" studies, if there's templates that have been uploaded
with bad sample names, then the best we can do is just print a list of
studies that don't pass this check so we can take care of them, right?

josenavas:
Yeah, I don't know how easy will be to "patch" this. My question is, if they did not pass the check, are we able to process them (i.e. running split libraries?).

ElDeveloper:
Not all violations to the naming conventions will make split_libraries
fail, underscores are really the main problem there, different tools
have different problems with different characters. That's why I don't
think we can really patch this problem.

adamrp:
Agreed, patching doesn't seem possible here. I think the best thing to do is, as @ElDeveloper suggests, see which studies' most up-to-date metadata fail these checks and contact the study owners directly. We should contact them only after this PR (and probably #829) is merged and deployed live.

antgonza:
OK, then this is ready for merge, right?

ElDeveloper:
Yes, AFAIK, everyone is ok with the code.

josenavas:
:+1:

PR #829, opened 2015-02-08
BUG: Stop forcing lowercasing of sample names

### LINKED ISSUE #799: sample name in map and biom files
if you upload a sample/prep template with upper case sample names the qiime mapping file that gets generated will keep them upper case but the biom table will have them lower case. This will block local analysis as the sample names do not match as all qiime/emperor/etc are case sensitive.

ElDeveloper:
I want to excercise the full test suite as I was getting some unexpected failures in my system.

Fixes #799


ElDeveloper:
The patch is a work in progress, please be careful.

qiita_db/support_files/patches/python_patches/14.py
```diff
@@ -0,0 +1,20 @@
...
+_id, fp_base = get_mountpoint('templates')[0]
+
+for prep_template_id in conn_handler.execute_fetchall(
+        "SELECT prep_template_id FROM qiita.prep_template"):
+    prep_template_id = prep_template_id[0]
+    pt = PrepTemplate(prep_template_id)
+    study_id = pt.study_id
+
+    for _, fpt in pt.get_filepaths():
+        pt.create_qiime_mapping_file(fpt)
```
ElDeveloper: Question for the Qiita devs, is it ok if I just do this or should I first find the previous mapping file, delete it and then re-create it?

josenavas: IMOO is fine as you're doing it, but will be useful to get more input from others @biocore/qiita-dev

ElDeveloper: Thanks for input!

qiita_db/support_files/patches/14.sql
```diff
@@ -0,0 +1,2 @@
+-- we don't really need an SQL patch, but having nothing here breaks the build
```
adamrp: Doesn't necessarily need to change, but no SQL statements are required to be in file, AFAIK (the file just needs to exist).

ElDeveloper: When I had an empty file, the build failed, that's why I added those
lines.

josenavas: Yeah. There are a couple of other patches like that. We need something that SQL executable...

qiita_db/support_files/patches/python_patches/14.py
adamrp: Good question. I think that theoretically, it should be okay to do it this way, but I'm slightly hesitant because I'm not sure if there are places in the code that unadvisable assumptions are made with regard to the number of templates associated with a study (e.g., is there code that assumes that there is only one prep template, and if so, does that code presume to use the prep template with the highest ID, or the lowest ID?)

ElDeveloper: So just to be sure, if I do that, the code will not be deleting or
forgetting the old/incorrect QIIME mapping file, it will instead just
add one more file to the prep template? If that's the case then I think
we need to remove the old mapping files because of this few lines of
code:
https://github.com/biocore/qiita/blob/7b0630dc7ddca95d38f9f5ec026221b6e6dc254e/qiita_pet/uimodules/raw_data_tab.py#L219-225

ElDeveloper: *pings the @biocore/qiita-dev team*

squirrelo: For safety's sake as we go forward, it's probably safer to get rid of the old versions in the DB but still have some way of caching the old ones. Im not sure how often the older SVN versions are gone back to but a similar setup should still exist in the new system as well.

ElDeveloper: If I understand correctly what we would want to do is to "unlink" (just
remove the reference in the DB) the previous QIIME mapping file and then
add this newly created mapping file. As far as I can tell, there's no
way to unlink a QIIME mapping file, @josenavas, can you confirm this? If
that's the case, then I think there are two options (1) add a
`unlink_qiime_mapping_file` method to the object and use that method as
part of the patch or (2) unlink the files with inline code within the
patch (could even be SQL code).

josenavas: AFAIK there is no unlink for the mapping file, but probably worth to double check with @antgonza as he added the code.

squirrelo: Yes, unlink is the word I was trying to find. @ElDeveloper that makes perfect sense. Option 1 is probably better, espeically since it also allows easy changes to what needs to be done for unlinking if needed down the line.

antgonza: Just to be clear, AFAIK the code only accepts 1 sample template per study but several prep templates. All sample and prep templates stay "forever" in the base_dir/templates/ folder. Once a prep template is processed a qiime mapping file is also generated from combining the sample and prep templates. I think the way to go will be to reprocess the file mapping files, simply because we can assure that both the raw files and the data in the db are the same. Note that you have to be careful cause if you drop and then add templates, you might be erasing other levels like process data, etc.

antgonza: One more pro about reprocessing is that the copied sample/prep template has the original casing. You should also be aware that most of our methods will prepend the name of the files with the study or prep ids so if you are using one of this scripts you should remove those prefixes.

ElDeveloper: If I understand correctly, you are suggesting that we remove all the
prep templates from every study, then re-create the same prep templates
(at this point the new QIIME mapping file will be re-created), then we
reprocess the raw data and then we reprocess the OTU table? I think this
is a lot of unneeded work as the problem is only present in the QIIME
mapping files which are not used for processing, the minimal mapping
file is used for processing and this problem is not propagated there:
https://github.com/biocore/qiita/blob/6614f27a540a6ee652b5e966979c06575056903c/qiita_ware/processing_pipeline.py#L23-73

@josenavas can you correct me if what I'm assuming about the mapping
files is wrong.

josenavas: Uhm... if the files have been already (pre)processed with the sample names in the database, if those sample names are modified, then all the split libraries and OTU tables that resulted from those (pre)processing are incorrect and need to be re-generated.

ElDeveloper: Hmmm I think I understand what you are saying, so this means that we
would have to reprocess everything? :L

josenavas: yes if we are changing the samples that are already on the database...

josenavas: And just note that are some studies that have been uploaded to EBI, which means that we potentially need to modify those studies on EBI as the sequence files and the mapping files are not sync...

What about leaving the current studies as they are and stop enforcing that on the new ones?

ElDeveloper: OK, thanks. I still think we are not modifying any of the sample names,
if you look at the code in `create_qiime_mapping_file`, you'll see it
only was changing the letter-case for the sample ids in this file,
everything else in the rest of the system should be done with the
correct sample identifiers. If not, this PR is not even aware of those
problems. Please let me know if I'm missing something terribly obvious
here.

antgonza: I think you are right about the mapping file, the issue is that the generated biom files will still not match the sample names, right? Which is also a problem.

josenavas: Oh, I see. I missed the goal of the PR completely... If no data in the database is changed, besides the paths to the new qiime mapping files, I'm not seeing any issue :smile:

josenavas: AFAIK, no code is using the qiime mapping file AFAIK, cause the split libraries command is using its own minimal qiime mapping file

ElDeveloper: I am not aware of that being the problem. In the original issue (#799)
the following is reported `The qiime mapping file will lower case
everything but the biom will keep the casing.`. And since all the
processing is done using the file produced by
`_get_minimal_mapping_file`, I think we should be good to go, because
that method does not change the letter casing of the samples.

antgonza: You are 100% right, just tested it. The biom has the right casing but the qiime map file doesn't. So simply regenerating the map files (sample/prep/qiime + md5s) should be fine.

ElDeveloper: I'm glad we could figure this out, I'll make the changes right now.

ElDeveloper:
Just pushed a new version of the python patch, if someone can have a look that would be awesome. I don't want to test it in my system just to find out that I missed something obvious :fearful:

ElDeveloper:
I'm debugging the patch right now ... and found an error.

ElDeveloper:
Ok, fixed the problem, I was able to successfully update all the QIIME mapping files in my system. Note that the patch is very aggressive, but I couldn't think of a "nicer" way to do what we need to do, please review carefully.

qiita_db/support_files/patches/python_patches/14.py
```diff
@@ -0,0 +1,66 @@
...
+    for mf in mapping_files:
+        q_name = 'queue-for-mapping-file-%d' % mf[0]
+        conn_handler.create_queue(q_name)
+
+        # (1) get the ids that we are going to delete.
+        # because of the FK restriction, we cannot just delete the ids
+        ids = conn_handler.execute_fetchall(
+            'SELECT filepath_id FROM qiita.{0} WHERE '
+            '{1}=%s and filepath_id=%s'.format(table, column), (pt.id, mf[0]))
+        ids = sum(ids, [])
```
josenavas: Really nice, I did not know that you can do this! Can you add a small comment about what is this line exactly doing? I had to test it by myself to understand what is going on in here

josenavas:
Code looks great. I have only one concern. In the current set up, if the patch fails at some point, part of the database will be updated, while the other one won't. This is because the queue are applied in a per prep template basis... I'm unsure if it will be better to apply all the commands at once (adding all of them at the queue and then execute the queue outside the for loop) so if any SQL query fails, it will rollback the entire patch. This will require a second for loop to create the new mapping files, but I would say is the safest way to proceed.

qiita_db/support_files/patches/14.sql
```diff
@@ -0,0 +1,2 @@
...
+select 1+1;
```
wasade: Obviously this should be `SELECT 42`

qiita_db/support_files/patches/python_patches/14.py
```diff
@@ -0,0 +1,66 @@
...
+for prep_template_id in conn_handler.execute_fetchall(sql):
+
+    prep_template_id = prep_template_id[0]
+    pt = PrepTemplate(prep_template_id)
+    fps = pt.get_filepaths()
+
+    # get the QIIME mapping file, note that the way to figure out what is and
+    # what's not a qiime mapping file is to check for the existance of the
+    # word qiime in the basename of the file path, hacky but that's the way
+    # it is being done in qiita_pet/uimodules/raw_data_tab.py
```
wasade: Is there perhaps a better way to do this though? It would be nice to have a magic number but that would require difficult-in-practice changes to qiime. What about extending the filename to "_qiime_mapping_file_" to be even more explicit?

qiita_db/support_files/patches/python_patches/14.py
wasade: It might be more explicit to do:

```
from operator import add
ids = reduce(add, ids)
```

Or perhaps a `list(flatten(ids)`

wasade: ...though that is a nifty use of `sum`...

qiita_db/support_files/patches/python_patches/14.py
```diff
@@ -0,0 +1,66 @@
...
+        conn_handler.add_to_queue(
+            q_name, "DELETE FROM qiita.{0} "
+            "WHERE {1}=%s and filepath_id=%s;".format(table, column),
+            (pt.id, mf[0]))
+
+        # (3) delete the entries from the filepath table
+        conn_handler.add_to_queue(
+            q_name,
+            "DELETE FROM qiita.filepath WHERE "
+            "filepath_id IN ({0});".format(', '.join(map(str, ids))))
```
wasade: `map` isn't throwing a flake8 message? I thought the recommendation was to do these via list comps?

wasade: The cast _could_ be done within Postgres too by modifying the select and adding a cast

qiita_db/support_files/patches/14.sql
ElDeveloper: Fixing in my next commit, thanks for catching this major oversight on my
end.

qiita_db/support_files/patches/python_patches/14.py
ElDeveloper: I think I'm just gonna import flatten from skbio.

qiita_db/support_files/patches/python_patches/14.py
ElDeveloper: flake8 raises an error, but we don't flake8 check our patches so it's
not showing in the build.

ElDeveloper: How do you do that?

wasade:
```sql
select cast(filepath_id as varchar) from qiita.filepath
```

ElDeveloper: How does that change what the code is doing? I guess I don't understand
what that code would do that the code is currently doing.

wasade: Same end effect, just that you wouldn't need to do the `map` of `str`

qiita_db/support_files/patches/python_patches/14.py
ElDeveloper: I don't have strong feelings about this, can change it if we reach a consensus as a group.

qiita_db/support_files/patches/python_patches/14.py
ElDeveloper: Ahhhhh, got it.

qiita_db/support_files/patches/14.sql
```diff
@@ -0,0 +1,2 @@
+-- we don't really need an SQL patch, but having nothing here breaks the build
+select 42;
```
ElDeveloper: @wasade OP deliverd :boom

wasade: :dolphin:

ElDeveloper:
Coveralls didn't post to GitHub but this should be ready for another review. Tested it in my system and it worked fine.

antgonza:
Pull from upstream master.

ElDeveloper:
There was a stochastic failure, I just restarted the build. AFAIK it is ready to be merged.

ElDeveloper:
:loudspeaker: tests passed :fireworks:

qiita_db/support_files/patches/python_patches/14.py
antgonza: I think the filename is fine + I like the idea of having the prefix of the study so we can search for all the files belonging to that study quickly + the GUI button says qiime mapping file.

Issue #834, opened 2015-02-09
qiita.colorado.edu is not showing the sample template info


josenavas:
It looks like is a db inconsistency issue:
```
In [1]: from qiita_db.study import Study

In [2]: s = Study(1694)

In [6]: from qiita_db.metadata_template import SampleTemplate

In [7]: st = SampleTemplate(s.sample_template)

In [9]: st.get_filepaths()
Out[9]: []
```

We need to figure out the source of the issue and a way to fix it.

josenavas:
Update on the issue:

It looks like all the rows in the filepath table that belong to the EMP studies have been removed. The files (luckily) at still present on the filesytem.

Does anybody know if postgres has a log file where we can check when these deletes where executed? It will be useful to find the issue.

In terms of restoring the integrity of the system, I would say that the only option that we have right know is to create a custom script that modifies the database directly and adds back those rows to the `qiita.filepath` table and the `qiita.sample_template_filepath` table.

Does anybody have a better idea? I can start working on that custom script if there are no better ideas.

josenavas:
The filepaths have been removed due to the `purge_filepaths` function, specifically [this](https://github.com/biocore/qiita/blob/master/qiita_db/util.py#L716) line. That line should be removed, although a refactoring of the function is needed. I'm currently working on it so PR soon. It will partially solve the issue, as we still need to fix the inconsistency in the DB.

Issue #814, opened 2015-02-01
create automatic jobs


antgonza:
This can be used to empty folders in the upload folder, check that the db and the file system are consistent, etc

From @squirrelo:
Looks like you want a combination of IOLoop.PeriodicCallback and IOLoop.add_timeout
http://stackoverflow.com/questions/24089086/unexpected-tornado-ioloop-periodiccallback-behavior

antgonza:
It also should remove unused file paths.

josenavas:
This issue is quite broad. We should better define which are the things that we want to do.

Also, IMOO we don't want to do this in the IOLoop, because that will be running on the server itself, so we are not freeing the main system of doing some work. Ideally this can be multiple cron jobs that start at `qiita webserver start` time.

squirrelo:
We may be able to have two IOloops going, one for web server and one as a cron-type thing.

josenavas:
Thanks @squirrelo 
But this means that the IOloop is still executed on the server machine right? I was thinking in something that is submitted to the queues and executed in a worker.

antgonza:
Good point so perhaps an IOloop that starts a job in the workers is the way to go.

wasade:
Yes, one of the motivations to make job submission as easy as possible with
moi, and without a need to tie it to a user

PR #837, opened 2015-02-09
Refactor purge function

josenavas:
Fixes (partially) https://github.com/biocore/qiita/issues/834

Some tables where missing on the `purge_filepaths` function. Now these tables are fetched programmatically, so this issue should not arise again.

Also, there was a line that should not be there at all, so I removed it. I'm actually surprise that the code was working with that line in there. The function `move_filepaths_to_upload_folder` also has that line added, so I removed it too.


antgonza:
Should we simply remove purge_filepaths? it's only used in a function that is not used any more

josenavas:
It can be useful for https://github.com/biocore/qiita/issues/814

ElDeveloper:
As long as it is correctly working and being used somewhere else, then
I'm all in for keeping it, otherwise there's no point.

qiita_db/util.py
```diff
@@ -697,26 +697,36 @@ def purge_filepaths(conn_handler=None):
...
-        WHERE filepath_id NOT IN (
-            SELECT filepath_id FROM qiita.raw_filepath UNION
-            SELECT filepath_id FROM qiita.preprocessed_filepath UNION
-            SELECT filepath_id FROM qiita.processed_filepath UNION
-            SELECT filepath_id FROM qiita.job_results_filepath UNION
-            SELECT filepath_id FROM qiita.analysis_filepath UNION
-            SELECT sequence_filepath FROM qiita.reference UNION
-            SELECT taxonomy_filepath FROM qiita.reference UNION
-            SELECT tree_filepath FROM qiita.reference)""")
+        WHERE filepath_id NOT IN (%s)""" % union_str)
```
squirrelo: Should this also make sure the files are legitimately deleted off the filesystem at the time of database removal?

josenavas: Lines 724-739

squirrelo: Ah, sorry, not part of request so didn't see it. Carry on.

qiita_db/util.py
```diff
@@ -697,26 +697,36 @@ def purge_filepaths(conn_handler=None):
     """
     conn_handler = conn_handler if conn_handler else SQLConnectionHandler()
 
+    # Get all the (table, column) pairs that reference to the filepath table
+    # Code adapted from http://stackoverflow.com/questions/5347050/
+    # sql-to-list-all-the-tables-that-reference-a-particular-column-in-a-table
+    table_cols_pairs = conn_handler.execute_fetchall(
+        """SELECT R.TABLE_NAME, R.column_name
```
wasade: ...at first i was afraid then I was petrified... 

This is pretty cool, good find! Following on the heels of one of @ElDeveloper prior comments on a different PR, would it be possible to cite the permanent URL for the SO post?

wasade: thanks!

antgonza:
Pull from upstream master.

antgonza:
Note that @josenavas and I are checking why these changes do _not_ work in my machine, we have some ideas but we are not sure yet. So please do not merge.

josenavas:
@antgonza this is ready for review. I fixed the issue.

For other @biocore/qiita-dev, the issue is that when you're using the SQL operator `NOT IN`, if there is a `NULL` value on the list to test, the `NOT IN` will also return a `NULL`. For more information about the issue, check this [link](http://www.depesz.com/2008/08/13/nulls-vs-not-in/) (thanks @antgonza for the useful link!)

antgonza:
This now _actually_ works well, thanks for fixing. :+1:

josenavas:
Thanks @antgonza !! Another reviewer?

qiita_db/test/test_util.py
```diff
@@ -421,6 +422,68 @@ def test_purge_filepaths(self):
         self.assertTrue(exists(fp1))
         self.assertFalse(exists(fp_exp_id))
 
+    def test_purge_filepaths_null_cols(self):
+        """Tests the nulls vs not in issue in purge_filepaths"""
```
ElDeveloper: What does this docstring mean? ...

qiita_db/test/test_util.py
```diff
@@ -421,6 +422,68 @@ def test_purge_filepaths(self):
...
+        exp_id = 1 + self.conn_handler.execute_fetchone(
+            "SELECT count(1) FROM qiita.filepath")[0]
+        fp_id = self.conn_handler.execute_fetchone(
+            "INSERT INTO qiita.filepath "
+            "(filepath, filepath_type_id, checksum, checksum_algorithm_id) "
+            "VALUES (%s, %s, %s, %s) RETURNING filepath_id", (fp, 1, "", 1))[0]
+        self.assertEqual(fp_id, exp_id)
+
+        # Connect the just added filepath to a raw data
+        self.conn_handler.execute(
```
ElDeveloper: Can't we do these with actual API calls or is this really the only way to do it?

qiita_db/test/test_util.py
```diff
@@ -421,6 +422,68 @@ def test_purge_filepaths(self):
...
+        close(fd)
+        exp_id = 1 + self.conn_handler.execute_fetchone(
+            "SELECT count(1) FROM qiita.filepath")[0]
+        fp_id = self.conn_handler.execute_fetchone(
+            "INSERT INTO qiita.filepath "
+            "(filepath, filepath_type_id, checksum, checksum_algorithm_id) "
+            "VALUES (%s, %s, %s, %s) RETURNING filepath_id", (fp, 1, "", 1))[0]
+        self.assertEqual(fp_id, exp_id)
+
+        # Connect the just added filepath to a raw data
```
ElDeveloper: connect -> link

qiita_db/test/test_util.py
```diff
@@ -421,6 +422,68 @@ def test_purge_filepaths(self):
...
+        self.assertTrue(exists(fp_exp_id))
+
+        # Unlink the filepath from the raw data
+        self.conn_handler.execute(
+            "DELETE FROM qiita.raw_filepath WHERE filepath_id=%s", (exp_id,))
+
+        # Only filepath 16 should be removed
+        sql = ("SELECT filepath_id FROM qiita.filepath ORDER BY "
+               "filepath_id")
+        exp_ids = [i[0] for i in self.conn_handler.execute_fetchall(sql)]
```
ElDeveloper: Minor but these list comprehensions could be replaced with skbio.util.flatten

qiita_db/test/test_util.py
```diff
@@ -421,6 +422,68 @@ def test_purge_filepaths(self):
...
+        exp_ids.pop()
+        purge_filepaths(self.conn_handler)
+        obs_ids = [i[0] for i in self.conn_handler.execute_fetchall(sql)]
+        self.assertEqual(obs_ids, exp_ids)
+
+        # Check that only the file for the removed filepath has been removed
```
ElDeveloper: Should these tests also check that all the other previously existing file paths *still* exist?

ElDeveloper:
A few comments, looks in good shape. The one thing that struck my attention was how complex the test method is, I would advise breaking it down into individual cases, but there's no reason why this should be a blocking comment.

qiita_db/test/test_util.py
josenavas: There is an issue using the SQL command `NOT IN` if the list contains null values. There is a link in line 428 with more information. It was for clarity, but I think I did not achieve that; so I'm going to remove the docstring and leave only the comment with the link

qiita_db/test/test_util.py
josenavas: Done

qiita_db/test/test_util.py
ElDeveloper: Ahhh, now I understand the phrasing of the docstring. Thanks!

qiita_db/test/test_util.py
josenavas: I've changed it for a couple of API calls (and fixed them in the previous test as it was a copy-paste...)

qiita_db/test/test_util.py
josenavas: Good call. Done!

qiita_db/test/test_util.py
josenavas: We are testing that only one of the other ones still exists.

Fun fact: the filepaths that are present in the database, they actually do not exist. Which means that if `purge_filepaths` try to remove them, it will fail, which will show up here. So, it is actually testing that is correctly working just by the fact that is not trying to remove those other files :-)

josenavas:
The method is quite complex, but is not testing multiple things. It is only testing one execution of the `purge_filepaths` function. Since it is a quite dangerous functionality, I decided to put quite few multiple checks and things that I think that functionality could potentially mess up

qiita_db/test/test_util.py
ElDeveloper: Hmmm if nobody else feels like this should be explicitly tested, then
I'm ok with the current state.

antgonza: The idea of the test framework is that before and after it always returns to the same state, right? So testing that this method erases those not used filepaths is not only safe but also a good exercise, right?

ElDeveloper: I too feel that way, or rather, I think it is only correct to verify
this and not to rely on a side-effect i.e. that the test would fail if
we tried to remove a file that doesn't exist.

josenavas:
This should be ready for re-re-re....-review! I've simplified tests and written a new function. It turns out that we already have unlinked files on the test database so I did not need to add a new unlinked file. Also, there was a bug on the fp removal, which is interesting since it was working before...

ElDeveloper:
This looks good, thanks for making these changes @josenavas :+1:

qiita_db/util.py
```diff
@@ -612,6 +612,28 @@ def get_mountpoint(mount_type, conn_handler=None, retrieve_all=False):
     return [(d, join(basedir, m, s)) for d, m, s in result]
 
 
+def get_mountpoint_path(mount_id, conn_handler=None):
```
antgonza: This can be confused with get_mountpoint, perhaps change to get_mountpoint_path_by_id?

ElDeveloper:
@josenavas, it looks like there are some flake8 problems, may be related to a newer version of flake8. Not your code, it's from the codebase in general.

josenavas:
Yes, I just saw them. I will try to fix them here, as the PR is ready to go...

josenavas:
@ElDeveloper @antgonza the issues due to the new version of pep8 should have been solved with my last commit.

josenavas:
@ElDeveloper @antgonza any remaining comment?

PR #839, opened 2015-02-10
Sample selection fix - Issue 783

squirrelo:
This now separates sample selection from datatype selection s the "select all/none/inverse" works properly.

### DIFF
--- a/qiita_pet/static/js/analysis.js
+++ b/qiita_pet/static/js/analysis.js
@@ -38,7 +38,7 @@ function select_deselect_samples_study(study) {
   var selected_datatypes = $('#study' + study + ' input:checkbox:checked').length;
   var sel = false;
   if (selected_datatypes > 0) { sel = true; }
-  select_deselect(study, sel)
+  select_deselect(study, sel);
 }
 
 function count_update(study) {
@@ -60,19 +60,20 @@ function count_update(study) {
 
 function select_deselect(study, select) {
   if(select == true) { 
-    $('.'+study).each(function() {this.checked = true;});
+    $('.sample.'+study).each(function() {this.checked = true;});
   }
   else { 
-    $('.'+study).each(function() {this.checked = false;});
+    $('.sample.'+study).each(function() {this.checked = false;});
   }
   count_update(study);
 }
 
 function select_inverse(study) {
-  $('.'+study).each(function() {
+  $('.sample.'+study).each(function() {
     if(this.checked == true) { this.checked = false; }
     else { this.checked = true; }
   });
+  count_update(study);
 }
 
 function pre_submit(action) {
--- a/qiita_pet/static/js/analysis.js
+++ b/qiita_pet/static/js/analysis.js
@@ -22,7 +22,7 @@ function displaySelected() {
 } 
 
 function select_category(category, study) {
-  if(study != '') { 
+  if(study !== '') { 
     $('.'+study+'.'+category).each(function() {this.checked = true;});
     count_update(study);
   }
@@ -47,19 +47,17 @@ function count_update(study) {
   document.getElementById('count' + study).innerHTML = selected;
   if(selected > 0) { 
     $('#study' + study).addClass('success');
-    studylink.disabled = false;
     studylink.style = "";
   }
   else {
     $('#study' + study).removeClass('success');
-    studylink.disabled = true;
     studylink.style = "text-decoration: none;";
     $('#study' + study + " input:checkbox").each(function() {this.checked = false;})
   }
 }
 
 function select_deselect(study, select) {
-  if(select == true) { 
+  if(select === true) { 
     $('.sample.'+study).each(function() {this.checked = true;});
   }
   else { 
@@ -70,7 +68,7 @@ function select_deselect(study, select) {
 
 function select_inverse(study) {
   $('.sample.'+study).each(function() {
-    if(this.checked == true) { this.checked = false; }
+    if(this.checked === true) { this.checked = false; }
     else { this.checked = true; }
   });
   count_update(study);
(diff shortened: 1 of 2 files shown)


antgonza:
:+1: Seems to work fine

josenavas:
The current behavior with this PR is really confusing...

First, the modal that shows the samples for a study sometimes is shown when I click the study name and others does not. Process to reproduce this behavior consistently: search for a study, open the modal, select one sample, close the modal, select the checkbox next to the datatype, unselect the checkbox, and try to open the modal again.

Second, It is confusing that I can open the modal, choose the sample that I want, and when I close the modal and click on "select samples" my entire search disappears and no samples are added (because I did not choose the data type). However, If I choose my samples and then I choose the datatype, all the samples are selected which I then need to go and make the selection again.

This is far from optimal. My suggestion here will be to fix the first issue (of course) and then do not allow the use to choose any sample from a study until he selects the data type. Of course, the selection of the datatype should not select all the samples, otherwise we get the second situation that I explained here.

squirrelo:
Last time we discussed this we agreed that selecting a datatype should by default select all samples. If this is no longer the case I can change it, but we need to settle on what we want this to do.

josenavas:
The other solution is then that when the user choses a sample, then the datatype is chosen. The problem is with the current set up in this PR, the user can spend a lot of time choosing samples and then all his work is lost if he clicks select samples without choosing the datatype.

I agree that we discuss things and they have to settle, but you should also notice that we have change the functionality of the interface, which could leverage to confusion, problems, or user time loss; as I explained in my previous comment.

What other @biocore/qiita-dev think? I don't agree with the current status of the PR, and I am open to different ways of solving the issues found here.

squirrelo:
I can put a check in to make sure a datatype is checked if samples are checked, but I don't think we can automatically select datatypes, since we will have multiple types eventually and we don't know what the user wants.

josenavas:
That is also a good solution. When the user clicks on "select samples", if he has samples selected in a study but not the datatype, it doesn't allow to continue and he has to either choose the datatype, or unselect those samples. At least he will not loose all the work done...

josenavas:
But in this case the choosing of the datatype should not select all the samples....

ElDeveloper:
I'm slightly confused by the behaviour, and I agree that this needs to
be revised. But I don't have specific suggestions on how to improve the
workflow.

squirrelo:
Ready for final review

antgonza:
What about having select all, none, inverse per data type? Perhaps near the data type header. This might make things easier.

squirrelo:
Sample selection and datatype selection are completely separate. As we discussed and agreed upon way back when this was first written, all samples you select will be selected for each datatype you want. There is no way to select samples for specific datatypes.

antgonza:
I think we should be able to select individual samples per data type.

josenavas:
Agree with @antgonza and it is the whole point of the PR actually, be able to select individual samples from a given data type.

@squirrelo as I explained before, I agree that think that where discussed and agreed should stay. But when the whole interface change it's behavior we need to be dynamic enough to see that some functionality needs to change in order to be able to be usable from the user point of view...

I also agree with the solution that @antgonza proposed. Having a couple of extra buttons either next to the datatype or next to the sample count will be more useful and clear.

squirrelo:
It's not a couple extra buttons. It's an extra checkbox by each sample for each datatype. It also means the counter needs to be changed to be by datatype. This is going to be much larger scope than originally planed, so I'll have something hopefully by end of day today.

josenavas:
What about using uimodules? You can re-use the code that you have and in the modal you simply add a tab screen, in which each tab is the data type.

squirrelo:
It's all doable, just a lot of UI and logic changes to cope with being able to select different samples per datatype.

josenavas:
Well another point that I would like to bring up is that there is the potential of one sample to be in one datatype, but not in the other one... so the samples that are shown in each tab are the ones for that data type. What do you think @antgonza

antgonza:
Agree and that's normally the case.

antgonza:
Pull from upstream master.

squirrelo:
After sleeping on it, I think this is adding too much to qiita, and if you want to do a specialized thing like that you should use straight up qiime, or do another analysis with the different samples. It opens up too many possibilities of doing cherry-picked analyses and/or just forgetting to select one sample of one type and wasting huge amounts of compute. 

I can add a "select all samples with [DATATYPE]" button in as a kind of compromise, since not all samples will have the different datatypes available, but that will significantly add to the compute costs as that requires opening each biom table and getting IDs available from it.

antgonza:
Remember that qiita will be the QIIME GUI so we want this.

squirrelo:
I still haven't heard a good argument for it, just "we want it".

antgonza:
The idea is that qiita will allow us to select any number of studies,
any number of samples within those samples, and any type of processing
on those samples to perform a meta-analysis so we can relate all of
them. To be able to do that, we need to have a GUI that allows us to
do this. Does this make sense?

rob-knight:
Agree 100%.


On Feb 12, 2015, at 7:46 AM, Antonio Gonzalez <notifications@github.com<mailto:notifications@github.com>> wrote:

Remember that qiita will be the QIIME GUI so we want this.

rob-knight:
That is a good argument.


On Feb 12, 2015, at 7:48 AM, Joshua Shorenstein <notifications@github.com<mailto:notifications@github.com>> wrote:


I still haven't heard a good argument for it, just "we want it".

squirrelo:
Ok, I just wanted to make absolutely sure before getting to work on this, since it's a bit compute heavy.

rob-knight:
I should re-phrase: requests from people who have done a lot of studies themselves successfully should be a sufficient argument for feature inclusion. Antonio's additional detail is a better argument though.


On Feb 12, 2015, at 8:21 AM, Joshua Shorenstein <notifications@github.com<mailto:notifications@github.com>> wrote:


Ok, I just wanted to make absolutely sure before getting to work on this, since it's a bit compute heavy.

squirrelo:
Since this has drifted way off the original intent of the pull request, which was to put in a quick fix for selection anomalies, I'm going to close this and create a new pull request for the selection by datatypes.

PR #840, opened 2015-02-10
BUG: Improve message when sample_name isn't found

### LINKED ISSUE #713: Improve error message
If a sample or a prep template does not include a column named `sample_name`, the following error will be displayed `"An error ocurred parsing the sample template: mapping.txt" 'sample name'`.

Solution is to change `load_template_to_dataframe` to catch this exception and raise one with a more meaningful message.

ElDeveloper:
Raise a more appropriate exception when the sample_name column is not found in
the template.

Fixes #713

### DIFF
--- a/qiita_db/metadata_template.py
+++ b/qiita_db/metadata_template.py
@@ -1895,6 +1895,8 @@ def load_template_to_dataframe(fn):
 
     Raises
     ------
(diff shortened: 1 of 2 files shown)


antgonza:
:+1:

PR #842, opened 2015-02-10
deletes raw data

antgonza:
Add the possibility of removing raw data. Raw data can only be removed if there are not prep templates added, this functionality will be in another PR. There are 2 delete levels: if there is more studies link to the raw data, only the study_raw_data connection is removed, and if the raw data is the last one, it will remove all connections.

This also adds the exists method at the BaseData level.

### DIFF
--- a/qiita_db/data.py
+++ b/qiita_db/data.py
@@ -85,7 +85,7 @@ from qiita_core.exceptions import IncompetentQiitaDeveloperError
 from .base import QiitaObject
 from .logger import LogEntry
 from .sql_connection import SQLConnectionHandler
-from .exceptions import QiitaDBError
+from .exceptions import QiitaDBError, QiitaDBUnknownIDError
 from .util import (exists_dynamic_table, insert_filepaths, convert_to_id,
                    convert_from_id, purge_filepaths, get_filepath_id,
                    get_mountpoint, move_filepaths_to_upload_folder)
@@ -255,7 +255,7 @@ class BaseData(QiitaObject):
         Returns
         -------
         bool
-            True if already exists. False otherwise.
+            True if exists, false otherwise.
         """
         conn_handler = SQLConnectionHandler()
 
@@ -342,8 +342,9 @@ class RawData(BaseData):
 
         Raises
         ------
-        ValueError
+        QiitaDBUnknownIDError
             If the raw data id doesn't exist
+        QiitaDBError
             If the raw data is not linked to that study_id
             If the raw data has prep templates associated
         """
@@ -351,26 +352,41 @@ class RawData(BaseData):
 
         # check if the raw data exist
         if not cls.exists(raw_data_id):
-            raise ValueError("Raw data: %s doesn't exist." % str(raw_data_id))
+            raise QiitaDBUnknownIDError(raw_data_id, "raw data")
 
         study_raw_data_exists = conn_handler.execute_fetchone(
             "SELECT EXISTS(SELECT * FROM qiita.study_raw_data WHERE "
             "study_id = {0} AND raw_data_id = {1})".format(study_id,
                                                            raw_data_id))[0]
         if not study_raw_data_exists:
-            raise ValueError(
-                "Raw data %s is not linked to study %s" % (raw_data_id,
-                                                           study_id))
+            raise QiitaDBError(
+                "Raw data %d is not linked to study %d or the study "
+                "doesn't exist" % (raw_data_id, study_id))
 
-        # check if there are any prep templates
+        # check if there are any prep templates for this study
         prep_template_count = conn_handler.execute_fetchone(
-            "SELECT COUNT(*) FROM qiita.prep_template WHERE "
-            "raw_data_id = {0}".format(raw_data_id))[0]
+            """SELECT COUNT(*) FROM qiita.prep_template AS pt
+            LEFT JOIN qiita.common_prep_info AS cpi ON
+                (pt.prep_template_id=cpi.prep_template_id)
+            LEFT JOIN qiita.required_sample_info AS rsi ON
+                (cpi.sample_id=rsi.sample_id)
+            WHERE raw_data_id = {0} and study_id = {1}""".format(raw_data_id,
+                                                                 study_id))[0]
 
         if (prep_template_count > 0):
-            raise ValueError("Raw data: %s, has %d prep templates associated "
-                             "so it can't be erased." % (str(raw_data_id),
-                                                         prep_template_count))
+            raise QiitaDBError(
+                "Raw data %d has prep template(s) associated so it can't be "
+                "erased." % raw_data_id)
+
+        # check if how many raw data are left, if last one, check that there
+        # are no linked files
+        raw_data_count = conn_handler.execute_fetchone(
+            "SELECT COUNT(*) FROM qiita.study_raw_data WHERE "
+            "raw_data_id = {0}".format(raw_data_id))[0]
+        if raw_data_count == 1 and RawData(raw_data_id).get_filepath_ids():
+            raise QiitaDBError(
+                "Raw data (%d) can't be remove because it has linked files. "
+                "To remove it, first unlink files." % raw_data_id)
 
         # delete
         conn_handler.execute("DELETE FROM qiita.study_raw_data WHERE "
--- a/qiita_db/data.py
+++ b/qiita_db/data.py
@@ -243,6 +243,27 @@ class BaseData(QiitaObject):
             "WHERE {1} = %s".format(self._table, self._data_filepath_column),
             (status, self._id))
 
+    @classmethod
+    def exists(cls, object_id):
+        r"""Checks if the given object_id exists
+
+        Parameters
+        ----------
+        id : str
+            The id of the object we are searching for
+
+        Returns
+        -------
+        bool
+            True if already exists. False otherwise.
+        """
+        conn_handler = SQLConnectionHandler()
+
+        return conn_handler.execute_fetchone(
+            "SELECT EXISTS(SELECT * FROM qiita.{0} WHERE "
+            "{1}=%s)".format(cls._table, cls._data_filepath_column),
+            (object_id, ))[0]
+
 
 class RawData(BaseData):
     r"""Object for dealing with raw data
@@ -307,6 +328,65 @@ class RawData(BaseData):
             rd.add_filepaths(filepaths, conn_handler)
 
         return rd
+
+    @classmethod
+    def delete(cls, raw_data_id, study_id):
+        """Removes the raw data with id raw_data_id
+
+        Parameters
+        ----------
+        raw_data_id : int
+            The raw data id
+        study_id : int
+            The study id
+
+        Raises
+        ------
+        ValueError
+            If the raw data id doesn't exist
+            If the raw data is not linked to that study_id
+            If the raw data has prep templates associated
+        """
+        conn_handler = SQLConnectionHandler()
+
+        # check if the raw data exist
+        if not cls.exists(raw_data_id):
+            raise ValueError("Raw data: %s doesn't exist." % str(raw_data_id))
+
+        study_raw_data_exists = conn_handler.execute_fetchone(
+            "SELECT EXISTS(SELECT * FROM qiita.study_raw_data WHERE "
+            "study_id = {0} AND raw_data_id = {1})".format(study_id,
+                                                           raw_data_id))[0]
+        if not study_raw_data_exists:
+            raise ValueError(
+                "Raw data %s is not linked to study %s" % (raw_data_id,
+                                                           study_id))
+
+        # check if there are any prep templates
+        prep_template_count = conn_handler.execute_fetchone(
+            "SELECT COUNT(*) FROM qiita.prep_template WHERE "
+            "raw_data_id = {0}".format(raw_data_id))[0]
+
+        if (prep_template_count > 0):
+            raise ValueError("Raw data: %s, has %d prep templates associated "
+                             "so it can't be erased." % (str(raw_data_id),
+                                                         prep_template_count))
+
+        # delete
+        conn_handler.execute("DELETE FROM qiita.study_raw_data WHERE "
+                             "raw_data_id = {0} AND "
+                             "study_id = {1}".format(raw_data_id, study_id))
+
+        # delete the connecting tables if there is no other study linked to
+        # the raw data
+        study_raw_data_count = conn_handler.execute_fetchone(
+            "SELECT COUNT(*) FROM qiita.study_raw_data WHERE "
+            "raw_data_id = {0}".format(raw_data_id))[0]
+        if study_raw_data_count == 0:
+            conn_handler.execute("DELETE FROM qiita.raw_filepath WHERE "
+                                 "raw_data_id = {0}".format(raw_data_id))
+            conn_handler.execute("DELETE FROM qiita.raw_data WHERE "
+                                 "raw_data_id = {0}".format(raw_data_id))
 
     @property
     def studies(self):
--- a/qiita_db/data.py
+++ b/qiita_db/data.py
@@ -365,20 +365,19 @@ class RawData(BaseData):
 
         # check if there are any prep templates for this study
         prep_template_count = conn_handler.execute_fetchone(
-            """SELECT COUNT(*) FROM qiita.prep_template AS pt
-            LEFT JOIN qiita.common_prep_info AS cpi ON
-                (pt.prep_template_id=cpi.prep_template_id)
-            LEFT JOIN qiita.required_sample_info AS rsi ON
-                (cpi.sample_id=rsi.sample_id)
-            WHERE raw_data_id = {0} and study_id = {1}""".format(raw_data_id,
-                                                                 study_id))[0]
-
-        if (prep_template_count > 0):
+            """
+            SELECT EXISTS(
+                SELECT * FROM qiita.prep_template AS pt
+                    JOIN qiita.study_raw_data AS srd on
+                        pt.raw_data_id=srd.raw_data_id
+                WHERE srd.raw_data_id={0} AND srd.study_id={1})
+            """.format(raw_data_id, study_id))[0]
+        if prep_template_count:
             raise QiitaDBError(
                 "Raw data %d has prep template(s) associated so it can't be "
                 "erased." % raw_data_id)
 
-        # check if how many raw data are left, if last one, check that there
+        # check how many raw data are left, if last one, check that there
         # are no linked files
         raw_data_count = conn_handler.execute_fetchone(
             "SELECT COUNT(*) FROM qiita.study_raw_data WHERE "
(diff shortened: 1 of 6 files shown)


qiita_db/data.py
```diff
@@ -243,6 +243,27 @@ def _set_link_filepaths_status(self, status):
...
+
+        Parameters
+        ----------
+        id : str
+            The id of the object we are searching for
+
+        Returns
+        -------
+        bool
+            True if already exists. False otherwise.
```
ElDeveloper: Can you rephrase the first sentence here?

qiita_db/data.py
```diff
@@ -308,6 +329,65 @@ def create(cls, filetype, studies, filepaths=None):
...
+        ValueError
+            If the raw data id doesn't exist
+            If the raw data is not linked to that study_id
+            If the raw data has prep templates associated
+        """
+        conn_handler = SQLConnectionHandler()
+
+        # check if the raw data exist
+        if not cls.exists(raw_data_id):
+            raise ValueError("Raw data: %s doesn't exist." % str(raw_data_id))
```
josenavas: `QiitaDBUnknownIDError`

qiita_db/data.py
```diff
@@ -308,6 +329,65 @@ def create(cls, filetype, studies, filepaths=None):
...
+
+        study_raw_data_exists = conn_handler.execute_fetchone(
+            "SELECT EXISTS(SELECT * FROM qiita.study_raw_data WHERE "
+            "study_id = {0} AND raw_data_id = {1})".format(study_id,
+                                                           raw_data_id))[0]
+        if not study_raw_data_exists:
+            raise ValueError(
```
josenavas: `QiitaDBError`

qiita_db/data.py
```diff
@@ -308,6 +329,65 @@ def create(cls, filetype, studies, filepaths=None):
...
+                "Raw data %s is not linked to study %s" % (raw_data_id,
+                                                           study_id))
+
+        # check if there are any prep templates
+        prep_template_count = conn_handler.execute_fetchone(
+            "SELECT COUNT(*) FROM qiita.prep_template WHERE "
+            "raw_data_id = {0}".format(raw_data_id))[0]
+
+        if (prep_template_count > 0):
+            raise ValueError("Raw data: %s, has %d prep templates associated "
```
josenavas: `QiitaDBError`

qiita_db/test/test_data.py
```diff
@@ -261,6 +261,28 @@ def test_remove_filepath_errors(self):
         with self.assertRaises(QiitaDBError):
             RawData(2).remove_filepath(fp)
 
+    def test_exists(self):
+        self.assertTrue(RawData.exists(1))
+        self.assertFalse(RawData.exists(1000))
+
+    def test_delete(self):
```
ElDeveloper: Just a comment, not necessary to change, but this test case could probably be divided into cases that raise an exception and cases that don't.

qiita_db/data.py
```diff
@@ -308,6 +329,65 @@ def create(cls, filetype, studies, filepaths=None):
...
+                             "raw_data_id = {0} AND "
+                             "study_id = {1}".format(raw_data_id, study_id))
+
+        # delete the connecting tables if there is no other study linked to
+        # the raw data
+        study_raw_data_count = conn_handler.execute_fetchone(
+            "SELECT COUNT(*) FROM qiita.study_raw_data WHERE "
+            "raw_data_id = {0}".format(raw_data_id))[0]
+        if study_raw_data_count == 0:
+            conn_handler.execute("DELETE FROM qiita.raw_filepath WHERE "
```
josenavas: What happens to the files that are linked with the raw data? I think this is removing the connection between the filepath and the raw data, which means that the filepaths are going to be in a "limbo" state in the filepath table and they are going to be deleted when somebody calls the `purge_filepath` function.

That would mean that there is no way in the interface to bring those files to the upload folder again.

qiita_pet/handlers/study_handlers/description_handlers.py
```diff
@@ -582,6 +582,33 @@ def display_template(self, study, user, msg, msg_level, top_tab=None,
...
+            The current user object
+        callback : function
+            The callback function to call with the results once the processing
+            is done
+        """
+        raw_data_id = self.get_argument('raw_data_id')
+
+        try:
+            RawData.delete(raw_data_id, study.id)
+            msg = ("Raw data %s has been deleted from study: "
```
ElDeveloper: you can change this %s to a %d and not cast raw_data_id ... just sayin'

qiita_db/test/test_data.py
```diff
@@ -261,6 +261,28 @@ def test_remove_filepath_errors(self):
...
+        # the raw data doesn't exist
+        with self.assertRaises(ValueError):
+            RawData.delete(1000, 1)
+
+        # the raw data and the study id are not linked
+        with self.assertRaises(ValueError):
```
josenavas: In this test is not only that the raw data and the study are not linked, is that the study does not exist. Can we check it with a study that actually exists?

qiita_pet/handlers/study_handlers/description_handlers.py
```diff
@@ -582,6 +582,33 @@ def display_template(self, study, user, msg, msg_level, top_tab=None,
...
+                   "<b><i>%s</i></b>" % (str(raw_data_id), study.title))
+            msg_level = "success"
+        except Exception as e:
+            msg = "Couldn't remove %s raw data: %s" % (str(raw_data_id),
```
ElDeveloper: Same as above.

josenavas:
A few comments @antgonza thanks!

During the review of this PR I was thinking on a example that I don't know if we fully support on the system and I would like to share it, to see what other @biocore/qiita-dev think:

We have to studies, A and B. Study A has an Illumina lane X. Study B has Illumina lanes X and Y. B shares one of its lanes with A, but not the other. In the current set up; study A is not going to be able to use the raw_data object from B, and vice-versa; which means that the Illumina lane X has to be uploaded twice.

I don't know how likely is this issue to happen, but I would say that we already have some example on the live system. What do we want to do in this case?

qiita_db/data.py
ElDeveloper: templates -> template(s)

qiita_pet/templates/study_description_templates/raw_data_editor_tab.html
```diff
@@ -60,7 +60,19 @@
...
         <br/>
-        <a class="btn btn-primary" onclick="add_prep_template({{raw_data_id}});">Add prep template</a>
+        <table border="0">
+          <tr>
+            <td>
+              <a class="btn btn-primary" onclick="add_prep_template({{raw_data_id}});">Add prep template</a>
+            </td>
+            <td>&nbsp; &nbsp;</td>
+            <td>
+              <a class="btn btn-danger" onclick="if ({{len(available_prep_templates)}} > 0) { alert('This raw data can\'t be erased. First remove the attached prep templates') } else { delete_raw_data({{raw_data_id}}) }">
```
ElDeveloper: The location of this button seems strange but I don't think there's a better place to put it? Could it maybe become a red cross (or the trash bin) by the name of the raw data? i.e. `FASTQ (ID: 7) X`. And only make it visible when the tab is on focus?

qiita_db/data.py
antgonza: That's correct. I didn't require the raw data to have no linked files because then you will not be able to erase a raw data from a second study without unlinking. One option, if you feel strongly about it, is to not allow erase it's the last raw data (remove all FK), if it has linked files.

qiita_db/test/test_data.py
antgonza: Will improving the error, like "or the study id doesn't exist" suffice?

qiita_db/data.py
josenavas: Yes, I think this is the safest path.

qiita_pet/templates/study_description_templates/raw_data_editor_tab.html
antgonza: Text, yes. Not sure what you mean with tab only on focus, isn't that the case already?

qiita_db/test/test_data.py
josenavas: Was just to make sure that it also works on that situation :smiley:

qiita_db/test/test_data.py
antgonza: What's the benefit, all those tests are testing the same method ...

qiita_pet/templates/study_description_templates/raw_data_editor_tab.html
ElDeveloper: What I meant is that the delete button should only be visible by the
name of the raw data when that object is selected on the interface i.e.
after you click on that tab.

qiita_db/data.py
```diff
@@ -308,6 +329,65 @@ def create(cls, filetype, studies, filepaths=None):
...
+            If the raw data is not linked to that study_id
+            If the raw data has prep templates associated
+        """
+        conn_handler = SQLConnectionHandler()
+
+        # check if the raw data exist
+        if not cls.exists(raw_data_id):
+            raise ValueError("Raw data: %s doesn't exist." % str(raw_data_id))
+
+        study_raw_data_exists = conn_handler.execute_fetchone(
```
squirrelo: Why not use the exists function just created in this pull request?

qiita_db/test/test_data.py
ElDeveloper: The organization of the tests would be the main benefit, plus ideally
when a test fails you want to be able to quickly identify what part is
malfunctioning, in this case this test could fail because the exception
is not being raised or because there's an internal logic error. Again,
just a comment. :neckbeard:

qiita_pet/handlers/study_handlers/description_handlers.py
```diff
@@ -582,6 +582,33 @@ def display_template(self, study, user, msg, msg_level, top_tab=None,
...
+        ----------
+        study : Study
+            The current study object
+        user : User
+            The current user object
+        callback : function
+            The callback function to call with the results once the processing
+            is done
+        """
+        raw_data_id = self.get_argument('raw_data_id')
```
squirrelo: `get_argument` always comes in as a string. It needs to be typecast to an int before use.

qiita_db/data.py
antgonza: done

qiita_db/data.py
antgonza: ok

qiita_db/data.py
antgonza: Not possible, the exists only checks that the raw data exists not that the combination of study_id and raw_data_id exist.

qiita_db/data.py
antgonza: ok

qiita_db/data.py
antgonza: ok

qiita_db/data.py
antgonza: OK, done.

qiita_db/test/test_data.py
antgonza: So improving the error or want me to use Study.exist or SQL to verify and raise an error due to that?

qiita_pet/handlers/study_handlers/description_handlers.py
antgonza: ok on both.

qiita_pet/handlers/study_handlers/description_handlers.py
antgonza: ok, casting now.

qiita_db/data.py
```diff
@@ -308,6 +329,81 @@ def create(cls, filetype, studies, filepaths=None):
...
+            "SELECT EXISTS(SELECT * FROM qiita.study_raw_data WHERE "
+            "study_id = {0} AND raw_data_id = {1})".format(study_id,
+                                                           raw_data_id))[0]
+        if not study_raw_data_exists:
+            raise QiitaDBError(
+                "Raw data %d is not linked to study %d or the study "
+                "doesn't exist" % (raw_data_id, study_id))
+
+        # check if there are any prep templates for this study
+        prep_template_count = conn_handler.execute_fetchone(
```
josenavas: I think this SQL query can be changed with this one:
```sql
SELECT COUNT(*) FROM qiita.prep_template AS pt
    JOIN qiita.study_raw_data AS srd on pt.raw_data_id=srd.raw_data_id
WHERE srd.study_id={0} AND srd.raw_data_id={1};
```
It has one less join and I think it is easier to understand.

josenavas: In addition, I don't think you are using the count, so probably with an `EXISTS` it will be enough.

qiita_db/data.py
```diff
@@ -308,6 +329,81 @@ def create(cls, filetype, studies, filepaths=None):
...
+                (cpi.sample_id=rsi.sample_id)
+            WHERE raw_data_id = {0} and study_id = {1}""".format(raw_data_id,
+                                                                 study_id))[0]
+
+        if (prep_template_count > 0):
+            raise QiitaDBError(
+                "Raw data %d has prep template(s) associated so it can't be "
+                "erased." % raw_data_id)
+
+        # check if how many raw data are left, if last one, check that there
```
josenavas: `# check if how many ...` -> `# check how many ...`

qiita_db/data.py
josenavas: This line can be removed now. The execution will arrive to this point if and only if there are no linked files, so no rows are going to be present on the `qiita.raw_filepath` table that links the raw_data with a file.

josenavas:
few more comments @antgonza 

Also I would like to hear about the example that I described in my previous review.

ElDeveloper:
Thanks for adding the `X` @antgonza, it looks really nice!

qiita_pet/templates/study_description.html
```diff
@@ -152,6 +152,25 @@
   }
 }
 
+
+function delete_raw_data(raw_data_id) {
+  if (confirm("Are you sure you want to delete raw data " + raw_data_id + "?")) {
```
ElDeveloper: Minor, but I think this could be `'Are you sure you want to delete FASTQ (ID: '+raw_data_id') ?'`.

ElDeveloper:
I have no other comments and think this should be good to be merged once @josenavas' comments are addressed.

qiita_pet/templates/study_description.html
wasade: Is it assured it'll be fastq here though?

wasade:
:+1:, curious re: @josenavas comment though

antgonza:
This is ready for merge.

antgonza:
Wait, just found an issue.

antgonza:
Again, ready!

ElDeveloper:
I think the only open end here is @josenavas' comment aside from that should be mergedy*! 

merge+ready=mergedy :laughing: :coffee: :laughing:

antgonza:
Which one? The SQL? If so, his suggestion doesn't work as it will
return any raw data with any study_id, when we are looking for
specific raw data linked to an specific study. His suggestion made me
realize that I wasn't testing this case correctly and fixed both on
the last PR.

ElDeveloper:
No, I was referring to this:

> We have to studies, A and B. Study A has an Illumina lane X. Study B has Illumina lanes X and Y. B shares one of its lanes with A, but not the other. In the current set up; study A is not going to be able to use the raw_data object from B, and vice-versa; which means that the Illumina lane X has to be uploaded twice.

> I don't know how likely is this issue to happen, but I would say that we already have some example on the live system. What do we want to do in this case?

ElDeveloper:
God dammit github+email, I was referring to this:

> We have to studies, A and B. Study A has an Illumina lane X. Study B has Illumina lanes X and Y. B shares one of its lanes with A, but not the other. In the current set up; study A is not going to be able to use the raw_data object from B, and vice-versa; which means that the Illumina lane X has to be uploaded twice.

> I don't know how likely is this issue to happen, but I would say that we already have some example on the live system. What do we want to do in this case?

antgonza:
I'm not sure that's actually possible and the solution will be to
upload X-Y to B and then have A use that raw data as vice versa will
not work.

ElDeveloper:
Seems good to me, if this is good for you @josenavas, can you merge this
PR?

josenavas:
Will review in a few minutes, I'm a bit behind some other issues

ElDeveloper:
Thanks!

josenavas:
@antgonza 
>Which one? The SQL? If so, his suggestion doesn't work as it will
return any raw data with any study_id, when we are looking for
specific raw data linked to an specific study. His suggestion made me
realize that I wasn't testing this case correctly and fixed both on
the last PR.

That is not true. Check the last line of my SQL (the WHERE clause limits the search for the study and the raw data that you're interested in):

```SQL
SELECT COUNT(*) FROM qiita.prep_template AS pt
                JOIN qiita.study_raw_data AS srd
                    ON pt.raw_data_id=srd.raw_data_id
                WHERE srd.study_id={0} AND srd.raw_data_id={1}
```

Actually, I just pull your PR, and I changed your SQL for this one and the tests pass...

> I'm not sure that's actually possible and the solution will be to
upload X-Y to B and then have A use that raw data as vice versa will
not work.

It is possible, there is a couple of examples already in the EMP datasets...
I guess it works on that specific situation. I'm unsure about other more complex situations though. What do you think @wasade ? You're also interested on this case.

josenavas:
Sorry, this is actually the SQL query that I tested:
```SQL
SELECT EXISTS (SELECT * FROM qiita.prep_template AS pt
                JOIN qiita.study_raw_data AS srd
                    ON pt.raw_data_id=srd.raw_data_id
                WHERE srd.study_id={0} AND srd.raw_data_id={1})
```

antgonza:
SQL: Strangely it didn't work on my installation, perhaps something to
do with pyc? Anyway, is that a blocker?

Studies A-B: is that concern really related to this PR or another
functionality issue (different PR)?

josenavas:
SQL: I suggest it should be a blocker since simpler SQL queries makes de code more maintainable. Also, one less JOIN should be faster, so we can potentially see performance benefits.

Studies A-B: forgot to mention that is unrelated to the PR but I wanted to bring it up I can create a different issue to discuss it or in our next call. My fault...

antgonza:
Can you test that's actually faster before suggesting it? Thanks.

josenavas:
I just run a quick benchmark with a small database and these are the results:
```
In [7]: %timeit conn_handler.execute_fetchone(antonio_sql)
1000 loops, best of 3: 760 µs per loop

In [8]: %timeit conn_handler.execute_fetchone(jose_sql)
1000 loops, best of 3: 375 µs per loop
```

josenavas:
SHAME ON ME!!!!! I've been destroyed :finnadie::finnadie::finnadie:

ElDeveloper:
But ... I thought your query was faster? Did I misunderstand that?

antgonza:
@josenavas query doesn't do what's expected and test fail. Faster != right

wasade:
@antgonza @josenavas, it's pretty common to combine studies sequencing runs, so I believe the situation @josenavas described happens a lot, no?

ElDeveloper:
Oh... I thought he said the test suite passed, but if it doesn't then I
was running under a false assumption. The "shame on him" makes sense
now (:+1: to that).

antgonza:
@josenavas, could you create an issue with these comments so this
discussion doesn't get lost?

josenavas:
Done: https://github.com/biocore/qiita/issues/851

antgonza:
Thank you

PR #843, opened 2015-02-10
Fixes #790

### LINKED ISSUE #790: Cannot download and re-upload a prep template

josenavas:
Checks that the samples have not been already prefixed with the study. If they are, it does not prefix them again. If not, it prefixes them with the study id.

### DIFF
--- a/qiita_db/metadata_template.py
+++ b/qiita_db/metadata_template.py
@@ -48,6 +48,7 @@ from functools import partial
 import pandas as pd
 import numpy as np
 import warnings
+warnings.simplefilter('always', UserWarning)
 
 from qiita_core.exceptions import IncompetentQiitaDeveloperError
 from .exceptions import (QiitaDBDuplicateError, QiitaDBColumnError,
@@ -145,13 +146,14 @@ def _prefix_sample_names_with_id(md_template, study):
         The study to which the metadata belongs to
     """
     # Get all the prefixes of the index, defined as any string before a '.'
-    prefixes = {idx.split('.')[0] for idx in md_template.index}
+    prefixes = {idx.split('.', 1)[0] for idx in md_template.index}
     # If the samples have been already prefixed with the study id, the prefixes
     # set will contain only one element and it will be the str representation
     # of the study id
     if len(prefixes) == 1 and prefixes.pop() == str(study.id):
         # The samples were already prefixed with the study id
-        warnings.warn("Sample names were already prefixed with the study id.")
+        warnings.warn("Sample names were already prefixed with the study id.",
+                      UserWarning)
     else:
         # Create a new pandas series in which all the values are the study_id
         # and it is indexed as the metadata template
--- a/qiita_db/metadata_template.py
+++ b/qiita_db/metadata_template.py
@@ -144,20 +144,30 @@ def _prefix_sample_names_with_id(md_template, study):
     study : Study
         The study to which the metadata belongs to
     """
-    # Create a new pandas series in which all the values are the study_id and
-    # it is indexed as the metadata template
-    study_ids = pd.Series([str(study.id)] * len(md_template.index),
-                          index=md_template.index)
-    # Create a new column on the metadata template that includes the metadata
-    # template indexes prefixed with the study id
-    md_template['sample_name_with_id'] = study_ids + '.' + md_template.index
-    # Assign the new previously created column as the new index
-    md_template.index = md_template.sample_name_with_id
-    # Delete the previously created column
-    del md_template['sample_name_with_id']
-    # The original metadata template had the index column unnamed - remove
-    # the name of the index
-    md_template.index.name = None
+    # Before we prefix the sample names with the study id, we should check if
+    # they have already prefixed with it
+    # Get all the prefixes of the index, defined as any string before a '.'
+    prefixes = set([idx.split('.')[0] for idx in md_template.index])
+    # If the samples have been already prefixed with the study id, the prefixes
+    # set will contain only one element and it will be the str representation
+    # of the study id
+    if not (len(prefixes) == 1 and prefixes.pop() == str(study.id)):
+        # The sample have not been prefixed, so prefix them with the study_id
+        # Create a new pandas series in which all the values are the study_id
+        # and it is indexed as the metadata template
+        study_ids = pd.Series([str(study.id)] * len(md_template.index),
+                              index=md_template.index)
+        # Create a new column on the metadata template that includes the
+        # metadata template indexes prefixed with the study id
+        md_template['sample_name_with_id'] = (study_ids + '.' +
+                                              md_template.index)
+        # Assign the new previously created column as the new index
+        md_template.index = md_template.sample_name_with_id
+        # Delete the previously created column
+        del md_template['sample_name_with_id']
+        # The original metadata template had the index column unnamed - remove
+        # the name of the index
+        md_template.index.name = None
 
 
 class BaseSample(QiitaObject):
--- a/qiita_db/metadata_template.py
+++ b/qiita_db/metadata_template.py
@@ -152,7 +152,9 @@ def _prefix_sample_names_with_id(md_template, study):
     # set will contain only one element and it will be the str representation
     # of the study id
     if not (len(prefixes) == 1 and prefixes.pop() == str(study.id)):
-        # The sample have not been prefixed, so prefix them with the study_id
+        warnings.warn("Sample names where not prefixed with the study id. "
+                      "Adding the study id as prefix")
+        # The samples have not been prefixed, so prefix them with the study_id
         # Create a new pandas series in which all the values are the study_id
         # and it is indexed as the metadata template
         study_ids = pd.Series([str(study.id)] * len(md_template.index),
--- a/qiita_db/metadata_template.py
+++ b/qiita_db/metadata_template.py
@@ -48,12 +48,12 @@ from functools import partial
 import pandas as pd
 import numpy as np
 import warnings
-warnings.simplefilter('always', UserWarning)
 
 from qiita_core.exceptions import IncompetentQiitaDeveloperError
 from .exceptions import (QiitaDBDuplicateError, QiitaDBColumnError,
                          QiitaDBUnknownIDError, QiitaDBNotImplementedError,
-                         QiitaDBDuplicateHeaderError, QiitaDBError)
+                         QiitaDBDuplicateHeaderError, QiitaDBError,
+                         QiitaDBWarning)
 from .base import QiitaObject
 from .sql_connection import SQLConnectionHandler
 from .ontology import Ontology
@@ -153,7 +153,7 @@ def _prefix_sample_names_with_id(md_template, study):
     if len(prefixes) == 1 and prefixes.pop() == str(study.id):
         # The samples were already prefixed with the study id
         warnings.warn("Sample names were already prefixed with the study id.",
-                      UserWarning)
+                      QiitaDBWarning)
     else:
         # Create a new pandas series in which all the values are the study_id
         # and it is indexed as the metadata template
@@ -1913,7 +1913,7 @@ def load_template_to_dataframe(fn):
     ------
     QiitaDBColumnError
         If the sample_name column is not present in the template.
-    UserWarning
+    QiitaDBWarning
         When columns are dropped because they have no content for any sample.
 
     Notes
@@ -1965,7 +1965,7 @@ def load_template_to_dataframe(fn):
     if dropped_cols:
         warnings.warn('The following column(s) were removed from the template '
                       'because all their values are empty: '
-                      '%s' % ', '.join(dropped_cols), UserWarning)
+                      '%s' % ', '.join(dropped_cols), QiitaDBWarning)
 
     return template
 
--- a/qiita_db/metadata_template.py
+++ b/qiita_db/metadata_template.py
@@ -144,17 +144,15 @@ def _prefix_sample_names_with_id(md_template, study):
     study : Study
         The study to which the metadata belongs to
     """
-    # Before we prefix the sample names with the study id, we should check if
-    # they have already prefixed with it
     # Get all the prefixes of the index, defined as any string before a '.'
-    prefixes = set([idx.split('.')[0] for idx in md_template.index])
+    prefixes = {idx.split('.')[0] for idx in md_template.index}
     # If the samples have been already prefixed with the study id, the prefixes
     # set will contain only one element and it will be the str representation
     # of the study id
     if len(prefixes) == 1 and prefixes.pop() == str(study.id):
+        # The samples were already prefixed with the study id
         warnings.warn("Sample names were already prefixed with the study id.")
     else:
-        # The samples have not been prefixed, so prefix them with the study_id
         # Create a new pandas series in which all the values are the study_id
         # and it is indexed as the metadata template
         study_ids = pd.Series([str(study.id)] * len(md_template.index),
@@ -163,12 +161,10 @@ def _prefix_sample_names_with_id(md_template, study):
         # metadata template indexes prefixed with the study id
         md_template['sample_name_with_id'] = (study_ids + '.' +
                                               md_template.index)
-        # Assign the new previously created column as the new index
         md_template.index = md_template.sample_name_with_id
-        # Delete the previously created column
         del md_template['sample_name_with_id']
         # The original metadata template had the index column unnamed - remove
-        # the name of the index
+        # the name of the index for consistency
         md_template.index.name = None
 
 
--- a/qiita_db/metadata_template.py
+++ b/qiita_db/metadata_template.py
@@ -151,9 +151,9 @@ def _prefix_sample_names_with_id(md_template, study):
     # If the samples have been already prefixed with the study id, the prefixes
     # set will contain only one element and it will be the str representation
     # of the study id
-    if not (len(prefixes) == 1 and prefixes.pop() == str(study.id)):
-        warnings.warn("Sample names where not prefixed with the study id. "
-                      "Adding the study id as prefix")
+    if len(prefixes) == 1 and prefixes.pop() == str(study.id):
+        warnings.warn("Sample names were already prefixed with the study id.")
+    else:
         # The samples have not been prefixed, so prefix them with the study_id
         # Create a new pandas series in which all the values are the study_id
         # and it is indexed as the metadata template
--- a/qiita_db/test/test_metadata_template.py
+++ b/qiita_db/test/test_metadata_template.py
@@ -26,7 +26,8 @@ from qiita_db.exceptions import (QiitaDBDuplicateError, QiitaDBUnknownIDError,
                                  QiitaDBNotImplementedError,
                                  QiitaDBDuplicateHeaderError,
                                  QiitaDBExecutionError,
-                                 QiitaDBColumnError, QiitaDBError)
+                                 QiitaDBColumnError, QiitaDBError,
+                                 QiitaDBWarning)
 from qiita_db.study import Study, StudyPerson
 from qiita_db.user import User
 from qiita_db.data import RawData
@@ -861,7 +862,7 @@ class TestSampleTemplate(TestCase):
 
     def test_create_already_prefixed_samples(self):
         """Creates a new SampleTemplate with the samples already prefixed"""
-        st = npt.assert_warns(UserWarning, SampleTemplate.create,
+        st = npt.assert_warns(QiitaDBWarning, SampleTemplate.create,
                               self.metadata_prefixed, self.new_study)
         # The returned object has the correct id
         self.assertEqual(st.id, 2)
@@ -1390,7 +1391,7 @@ class TestPrepTemplate(TestCase):
 
     def test_create_already_prefixed_samples(self):
         """Creates a new PrepTemplate"""
-        pt = npt.assert_warns(UserWarning, PrepTemplate.create,
+        pt = npt.assert_warns(QiitaDBWarning, PrepTemplate.create,
                               self.metadata_prefixed, self.new_raw_data,
                               self.test_study, self.data_type)
         # The returned object has the correct id
@@ -1891,7 +1892,7 @@ class TestUtilities(TestCase):
         assert_frame_equal(obs, exp)
 
     def test_load_template_to_dataframe_empty_columns(self):
-        obs = npt.assert_warns(UserWarning, load_template_to_dataframe,
+        obs = npt.assert_warns(QiitaDBWarning, load_template_to_dataframe,
                                StringIO(EXP_ST_SPACES_EMPTY_COLUMN))
         exp = pd.DataFrame.from_dict(SAMPLE_TEMPLATE_DICT_FORM)
         exp.index.name = 'sample_name'
@@ -1928,7 +1929,7 @@ class TestUtilities(TestCase):
         assert_frame_equal(obs, exp)
 
     def test_load_template_to_dataframe_empty_column(self):
-        obs = npt.assert_warns(UserWarning, load_template_to_dataframe,
+        obs = npt.assert_warns(QiitaDBWarning, load_template_to_dataframe,
                                StringIO(SAMPLE_TEMPLATE_EMPTY_COLUMN))
         exp = pd.DataFrame.from_dict(ST_EMPTY_COLUMN_DICT_FORM)
         exp.index.name = 'sample_name'
--- a/qiita_db/test/test_metadata_template.py
+++ b/qiita_db/test/test_metadata_template.py
@@ -538,6 +538,47 @@ class TestSampleTemplate(TestCase):
             }
         self.metadata = pd.DataFrame.from_dict(metadata_dict, orient='index')
 
+        metadata_prefixed_dict = {
+            '2.Sample1': {'physical_location': 'location1',
+                          'has_physical_specimen': True,
+                          'has_extracted_data': True,
+                          'sample_type': 'type1',
+                          'required_sample_info_status': 'received',
+                          'collection_timestamp':
+                          datetime(2014, 5, 29, 12, 24, 51),
+                          'host_subject_id': 'NotIdentified',
+                          'Description': 'Test Sample 1',
+                          'str_column': 'Value for sample 1',
+                          'latitude': 42.42,
+                          'longitude': 41.41},
+            '2.Sample2': {'physical_location': 'location1',
+                          'has_physical_specimen': True,
+                          'has_extracted_data': True,
+                          'sample_type': 'type1',
+                          'required_sample_info_status': 'received',
+                          'collection_timestamp':
+                          datetime(2014, 5, 29, 12, 24, 51),
+                          'host_subject_id': 'NotIdentified',
+                          'Description': 'Test Sample 2',
+                          'str_column': 'Value for sample 2',
+                          'latitude': 4.2,
+                          'longitude': 1.1},
+            '2.Sample3': {'physical_location': 'location1',
+                          'has_physical_specimen': True,
+                          'has_extracted_data': True,
+                          'sample_type': 'type1',
+                          'required_sample_info_status': 'received',
+                          'collection_timestamp':
+                          datetime(2014, 5, 29, 12, 24, 51),
+                          'host_subject_id': 'NotIdentified',
+                          'Description': 'Test Sample 3',
+                          'str_column': 'Value for sample 3',
+                          'latitude': 4.8,
+                          'longitude': 4.41},
+            }
+        self.metadata_prefixed = pd.DataFrame.from_dict(
+            metadata_prefixed_dict, orient='index')
+
         self.test_study = Study(1)
         info = {
             "timeseries_type_id": 1,
@@ -613,6 +654,48 @@ class TestSampleTemplate(TestCase):
     def test_create(self):
         """Creates a new SampleTemplate"""
         st = SampleTemplate.create(self.metadata, self.new_study)
+        # The returned object has the correct id
+        self.assertEqual(st.id, 2)
+
+        # The relevant rows to required_sample_info have been added.
+        obs = self.conn_handler.execute_fetchall(
+            "SELECT * FROM qiita.required_sample_info WHERE study_id=2")
+        # sample_id study_id physical_location has_physical_specimen
+        # has_extracted_data sample_type required_sample_info_status_id
+        # collection_timestamp host_subject_id description
+        exp = [["2.Sample1", 2, "location1", True, True, "type1", 1,
+                datetime(2014, 5, 29, 12, 24, 51), "NotIdentified",
+                "Test Sample 1", 42.42, 41.41],
+               ["2.Sample2", 2, "location1", True, True, "type1", 1,
+                datetime(2014, 5, 29, 12, 24, 51), "NotIdentified",
+                "Test Sample 2", 4.2, 1.1],
+               ["2.Sample3", 2, "location1", True, True, "type1", 1,
+                datetime(2014, 5, 29, 12, 24, 51), "NotIdentified",
+                "Test Sample 3", 4.8, 4.41]]
+        self.assertEqual(obs, exp)
+
+        # The relevant rows have been added to the study_sample_columns
+        obs = self.conn_handler.execute_fetchall(
+            "SELECT * FROM qiita.study_sample_columns WHERE study_id=2")
+        # study_id, column_name, column_type
+        exp = [[2, "str_column", "varchar"]]
+        self.assertEqual(obs, exp)
+
+        # The new table exists
+        self.assertTrue(exists_table("sample_2", self.conn_handler))
+
+        # The new table hosts the correct values
+        obs = self.conn_handler.execute_fetchall(
+            "SELECT * FROM qiita.sample_2")
+        # sample_id, str_column
+        exp = [['2.Sample1', "Value for sample 1"],
+               ['2.Sample2', "Value for sample 2"],
+               ['2.Sample3', "Value for sample 3"]]
+        self.assertEqual(obs, exp)
+
+    def test_create_already_prefixed_samples(self):
+        """Creates a new SampleTemplate with the samples already prefixed"""
+        st = SampleTemplate.create(self.metadata_prefixed, self.new_study)
         # The returned object has the correct id
         self.assertEqual(st.id, 2)
 
@@ -968,6 +1051,45 @@ class TestPrepTemplate(TestCase):
                             'experiment_design_description': 'BBBB'}
             }
         self.metadata = pd.DataFrame.from_dict(metadata_dict, orient='index')
+
+        metadata_prefixed_dict = {
+            '1.SKB8.640193': {'center_name': 'ANL',
+                              'center_project_name': 'Test Project',
+                              'ebi_submission_accession': None,
+                              'EMP_status': 'EMP',
+                              'str_column': 'Value for sample 1',
+                              'linkerprimersequence': 'GTGCCAGCMGCCGCGGTAA',
+                              'barcodesequence': 'GTCCGCAAGTTA',
+                              'run_prefix': "s_G1_L001_sequences",
+                              'platform': 'ILLUMINA',
+                              'library_construction_protocol': 'AAAA',
+                              'experiment_design_description': 'BBBB'},
+            '1.SKD8.640184': {'center_name': 'ANL',
+                              'center_project_name': 'Test Project',
+                              'ebi_submission_accession': None,
+                              'EMP_status': 'EMP',
+                              'str_column': 'Value for sample 2',
+                              'linkerprimersequence': 'GTGCCAGCMGCCGCGGTAA',
+                              'barcodesequence': 'CGTAGAGCTCTC',
+                              'run_prefix': "s_G1_L001_sequences",
+                              'platform': 'ILLUMINA',
+                              'library_construction_protocol': 'AAAA',
+                              'experiment_design_description': 'BBBB'},
+            '1.SKB7.640196': {'center_name': 'ANL',
+                              'center_project_name': 'Test Project',
+                              'ebi_submission_accession': None,
+                              'EMP_status': 'EMP',
+                              'str_column': 'Value for sample 3',
+                              'linkerprimersequence': 'GTGCCAGCMGCCGCGGTAA',
+                              'barcodesequence': 'CCTCTGAGAGCT',
+                              'run_prefix': "s_G1_L002_sequences",
+                              'platform': 'ILLUMINA',
+                              'library_construction_protocol': 'AAAA',
+                              'experiment_design_description': 'BBBB'}
+            }
+        self.metadata_prefixed = pd.DataFrame.from_dict(metadata_prefixed_dict,
+                                                        orient='index')
+
         self.test_raw_data = RawData(1)
         self.test_study = Study(1)
         self.data_type = "18S"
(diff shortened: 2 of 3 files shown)


qiita_db/metadata_template.py
```diff
@@ -144,20 +144,30 @@ def _prefix_sample_names_with_id(md_template, study):
...
-    md_template.index.name = None
+    # Before we prefix the sample names with the study id, we should check if
+    # they have already prefixed with it
+    # Get all the prefixes of the index, defined as any string before a '.'
+    prefixes = set([idx.split('.')[0] for idx in md_template.index])
+    # If the samples have been already prefixed with the study id, the prefixes
+    # set will contain only one element and it will be the str representation
+    # of the study id
+    if not (len(prefixes) == 1 and prefixes.pop() == str(study.id)):
+        # The sample have not been prefixed, so prefix them with the study_id
```
ElDeveloper: sample-> sample

josenavas: ?

qiita_db/test/test_metadata_template.py
```diff
@@ -652,6 +693,48 @@ def test_create(self):
                ['2.Sample3', "Value for sample 3"]]
         self.assertEqual(obs, exp)
 
+    def test_create_already_prefixed_samples(self):
```
ElDeveloper: Can you add a test case for samples that are prefixed with a number that doesn't match the study_id and one where the prefixes are different numbers/strings?

qiita_db/metadata_template.py
```diff
@@ -144,20 +144,30 @@ def _prefix_sample_names_with_id(md_template, study):
...
-    # the name of the index
-    md_template.index.name = None
+    # Before we prefix the sample names with the study id, we should check if
+    # they have already prefixed with it
+    # Get all the prefixes of the index, defined as any string before a '.'
+    prefixes = set([idx.split('.')[0] for idx in md_template.index])
+    # If the samples have been already prefixed with the study id, the prefixes
+    # set will contain only one element and it will be the str representation
+    # of the study id
+    if not (len(prefixes) == 1 and prefixes.pop() == str(study.id)):
```
ElDeveloper: Should this raise a warning?

qiita_db/metadata_template.py
ElDeveloper: Hahaha, sorry sample->samples

qiita_db/metadata_template.py
josenavas: I don't think it is needed. What others think?

qiita_db/metadata_template.py
josenavas: Done :wink:

qiita_db/metadata_template.py
antgonza: :+1: to warning so the user is aware.

josenavas: Done

qiita_db/test/test_metadata_template.py
josenavas: Done

qiita_db/metadata_template.py
```diff
@@ -144,20 +144,32 @@ def _prefix_sample_names_with_id(md_template, study):
...
+        warnings.warn("Sample names where not prefixed with the study id. "
```
ElDeveloper: Sorry for not communicating properly, but I think the case where the warning would be more appropriate is when the study identifiers are not being prepended i.e. when the user is loading a template that already includes the study identifier. 

Can you also test for the warning in the tests below?

josenavas: Oh, now makes sense. I was wondering why we should warn if we were not warning before... Changing right now!

josenavas:
@ElDeveloper I'm facing a weird situation here. If I only run that test, or the entire class where the failing test is, the test runs ok. However, if I run the entire test file, that test files. Is this some kind of catching that python is doing? i.e. if a warning have been raised it won't be raised again?

josenavas:
Ok, I just checked using an interactive session and it turns out that python catches the user warnings. This becomes a problem, as we cannot use warnings as a way to flag those issues. As far as I remember, @ElDeveloper added a warning on the function `load_template_to_dataframe` to flag that columns have been removed. If you open an interactive python session and execute that function twice with the same parameters, the first time it will raise the warning; while the second one will not, and it is actually removing the column to.

Given that we cannot use warnings for this, which will be the best option to flag these issues? @biocore/qiita-dev

qiita_db/metadata_template.py
ElDeveloper: Thanks!

qiita_db/metadata_template.py
```diff
@@ -144,20 +144,32 @@ def _prefix_sample_names_with_id(md_template, study):
...
-    md_template['sample_name_with_id'] = study_ids + '.' + md_template.index
-    # Assign the new previously created column as the new index
-    md_template.index = md_template.sample_name_with_id
-    # Delete the previously created column
-    del md_template['sample_name_with_id']
-    # The original metadata template had the index column unnamed - remove
-    # the name of the index
-    md_template.index.name = None
+    # Before we prefix the sample names with the study id, we should check if
+    # they have already prefixed with it
```
wasade: The wording of this comment is weird. Also, it does not appear to pertain to the logic in line 150, is that expected?

qiita_db/metadata_template.py
```diff
@@ -144,20 +144,32 @@ def _prefix_sample_names_with_id(md_template, study):
...
+    # Get all the prefixes of the index, defined as any string before a '.'
+    prefixes = set([idx.split('.')[0] for idx in md_template.index])
```
wasade: Should this split be done 1 or N times?

Minor, but will save on object construction, you can use a set comprehension. eg: `{idx.split('.', 1)[0] for idx in md_template.index}`

qiita_db/metadata_template.py
```diff
@@ -144,20 +144,32 @@ def _prefix_sample_names_with_id(md_template, study):
...
+    # If the samples have been already prefixed with the study id, the prefixes
+    # set will contain only one element and it will be the str representation
+    # of the study id
+    if len(prefixes) == 1 and prefixes.pop() == str(study.id):
+        warnings.warn("Sample names were already prefixed with the study id.")
+    else:
```
wasade: There are two other possibilities, the empty set and multiple prefixes. If either of those happen, does that impact this logic?

qiita_db/metadata_template.py
```diff
@@ -144,20 +144,32 @@ def _prefix_sample_names_with_id(md_template, study):
...
+        # The samples have not been prefixed, so prefix them with the study_id
+        # Create a new pandas series in which all the values are the study_id
+        # and it is indexed as the metadata template
+        study_ids = pd.Series([str(study.id)] * len(md_template.index),
+                              index=md_template.index)
+        # Create a new column on the metadata template that includes the
+        # metadata template indexes prefixed with the study id
+        md_template['sample_name_with_id'] = (study_ids + '.' +
+                                              md_template.index)
+        # Assign the new previously created column as the new index
```
wasade: Many of these comments are redundant given the code, and the larger issue with comments like this is the (easy) potential for them to get out of sync particularly during code maintenance. It can also be more difficult to read what's going on as the reviewer has to context switch frequently, or ignore the comments consciously. None of this is blocking, just a general comment on comments.

ElDeveloper:
Oh, I think you are right, I cannot think of a good, solution I'm gonna
try to find something on the intertubes. However I think the way we are
catching the exception in qiita_pet makes it so that we _always_ catch
the message.

qiita_db/test/test_metadata_template.py
```diff
@@ -652,6 +775,133 @@ def test_create(self):
...
+        # The returned object has the correct id
+        self.assertEqual(st.id, 2)
+
+        # The relevant rows to required_sample_info have been added.
+        obs = self.conn_handler.execute_fetchall(
+            "SELECT * FROM qiita.required_sample_info WHERE study_id=2")
+        # sample_id study_id physical_location has_physical_specimen
+        # has_extracted_data sample_type required_sample_info_status_id
+        # collection_timestamp host_subject_id description
+        exp = [["2.12.Sample1", 2, "location1", True, True, "type1", 1,
```
wasade: The delimiter we're using for the prefix/sample ID is a '.', but that is also allowable within the sample ID (`12.Sample1`)? One issue here is that, if given a random sample ID, such as `12.Sample1`, I do not know given the rules if the sample ID is `12.Sample1` or `Sample1`

qiita_db/metadata_template.py
josenavas: I don't get your first question...

Changed to use set comprehension

qiita_db/metadata_template.py
josenavas: If there is an empty set or the set has more than one element, the execution workflow goes to the else block of code; which is correct because that means that the samples have not been already prefixed with the sample id.

qiita_db/metadata_template.py
josenavas: I removed some comments and re-word others. Hope is clearer. (This entire block of code was already there, I just indented it :-)

qiita_db/test/test_metadata_template.py
josenavas: Yes. Following QIIME rules, the sample id can contain numbers, letters and the `.` character. In qiita we are adding `<study_id>.` as prefix to all sample ids. So everything after the first `.` is the "original" sample id. Note that in every other place in Qiita, the sample id is with the study id prefixed.

That construction of the sample ids was already added a while ago (august, september?) here I'm just adding the option to re-upload the templates downloaded from Qiita; but I'm not adding anything new to the sample ids.

qiita_db/metadata_template.py
wasade: The `split` is splitting on all `.`, shouldn't it only split on the first? Otherwise it's doing extra work for no benefit

qiita_db/metadata_template.py
wasade: Or that there are multiple prefixes?

qiita_db/metadata_template.py
wasade: Thanks!

qiita_db/test/test_metadata_template.py
wasade: The issue is the ambiguity, and that we are perpetuating the ambiguity. I don't think there is a great solution here, but we _may_ a more unique delimiter would reduce the potential for confusion

qiita_db/metadata_template.py
josenavas: Good point. Done.

2015-02-11 13:19 GMT-08:00 Daniel McDonald <notifications@github.com>:

> In qiita_db/metadata_template.py
> <https://github.com/biocore/qiita/pull/843#discussion_r24534293>:
>
> > -    study_ids = pd.Series([str(study.id)] * len(md_template.index),
> > -                          index=md_template.index)
> > -    # Create a new column on the metadata template that includes the metadata
> > -    # template indexes prefixed with the study id
> > -    md_template['sample_name_with_id'] = study_ids + '.' + md_template.index
> > -    # Assign the new previously created column as the new index
> > -    md_template.index = md_template.sample_name_with_id
> > -    # Delete the previously created column
> > -    del md_template['sample_name_with_id']
> > -    # The original metadata template had the index column unnamed - remove
> > -    # the name of the index
> > -    md_template.index.name = None
> > +    # Before we prefix the sample names with the study id, we should check if
> > +    # they have already prefixed with it
> > +    # Get all the prefixes of the index, defined as any string before a '.'
> > +    prefixes = set([idx.split('.')[0] for idx in md_template.index])
>
> The split is splitting on all ., shouldn't it only split on the first?
> Otherwise it's doing extra work for no benefit
>
> —
> <https://github.com/biocore/qiita/pull/843/files#r24534293>.
>



-- 
Jose Navas

qiita_db/metadata_template.py
josenavas: This code is adding a sample/prep template to a study. If there are more
than one prefix, means that is a user-defined prefix, it is not the study
id prefix added by Qiita (which will be unique for all the samples) meaning
that we need to add the Qiita prefix, so the final sample id is
"<study_id>.<whatever.the.user.specifiied.as.sample.id>"

2015-02-11 13:19 GMT-08:00 Daniel McDonald <notifications@github.com>:

> In qiita_db/metadata_template.py
> <https://github.com/biocore/qiita/pull/843#discussion_r24534324>:
>
> > -    md_template.index = md_template.sample_name_with_id
> > -    # Delete the previously created column
> > -    del md_template['sample_name_with_id']
> > -    # The original metadata template had the index column unnamed - remove
> > -    # the name of the index
> > -    md_template.index.name = None
> > +    # Before we prefix the sample names with the study id, we should check if
> > +    # they have already prefixed with it
> > +    # Get all the prefixes of the index, defined as any string before a '.'
> > +    prefixes = set([idx.split('.')[0] for idx in md_template.index])
> > +    # If the samples have been already prefixed with the study id, the prefixes
> > +    # set will contain only one element and it will be the str representation
> > +    # of the study id
> > +    if len(prefixes) == 1 and prefixes.pop() == str(study.id):
> > +        warnings.warn("Sample names were already prefixed with the study id.")
> > +    else:
>
> Or that there are multiple prefixes?
>
> —
> <https://github.com/biocore/qiita/pull/843/files#r24534324>.
>



-- 
Jose Navas

qiita_db/test/test_metadata_template.py
josenavas: I would say that is not that ambiguous. The first dot separates the
study_id from whatever the user used as a sample id. An the second part is
completely arbitrary...

2015-02-11 13:22 GMT-08:00 Daniel McDonald <notifications@github.com>:

> In qiita_db/test/test_metadata_template.py
> <https://github.com/biocore/qiita/pull/843#discussion_r24534521>:
>
> > @@ -652,6 +775,133 @@ def test_create(self):
> >                 ['2.Sample3', "Value for sample 3"]]
> >          self.assertEqual(obs, exp)
> >
> > +    def test_create_int_prefix(self):
> > +        """Creates a new SampleTemplate"""
> > +        st = SampleTemplate.create(self.metadata_int_pref, self.new_study)
> > +        # The returned object has the correct id
> > +        self.assertEqual(st.id, 2)
> > +
> > +        # The relevant rows to required_sample_info have been added.
> > +        obs = self.conn_handler.execute_fetchall(
> > +            "SELECT * FROM qiita.required_sample_info WHERE study_id=2")
> > +        # sample_id study_id physical_location has_physical_specimen
> > +        # has_extracted_data sample_type required_sample_info_status_id
> > +        # collection_timestamp host_subject_id description
> > +        exp = [["2.12.Sample1", 2, "location1", True, True, "type1", 1,
>
> The issue is the ambiguity, and that we are perpetuating the ambiguity. I
> don't think there is a great solution here, but we *may* a more unique
> delimiter would reduce the potential for confusion
>
> —
> <https://github.com/biocore/qiita/pull/843/files#r24534521>.
>



-- 
Jose Navas

wasade: What is the sample ID in this string: "123.foo"?

josenavas:
@ElDeveloper I don't think we are catching all. I think that specific part of the code is usually executed by the web server, which means that during the life time of the webserver this warning is only raised the first time that it happens. Even if that function is executed on the ipython workers, that warning will be raised only the first time the function is executed on that warning.

However, I just found the solution. By adding `warnings.simplefilter('always', UserWarning)` after warnings is imported, you are forcing that UserWarning are always raised, rather than just the first time.
We can even create our own warning if we want to me more explicit...

qiita_db/test/test_metadata_template.py
josenavas: It's the entire string. In Qiita, the entire string is the sample id.

2015-02-11 13:34 GMT-08:00 Daniel McDonald <notifications@github.com>:

> In qiita_db/test/test_metadata_template.py
> <https://github.com/biocore/qiita/pull/843#discussion_r24535664>:
>
> > @@ -652,6 +775,133 @@ def test_create(self):
> >                 ['2.Sample3', "Value for sample 3"]]
> >          self.assertEqual(obs, exp)
> >
> > +    def test_create_int_prefix(self):
> > +        """Creates a new SampleTemplate"""
> > +        st = SampleTemplate.create(self.metadata_int_pref, self.new_study)
> > +        # The returned object has the correct id
> > +        self.assertEqual(st.id, 2)
> > +
> > +        # The relevant rows to required_sample_info have been added.
> > +        obs = self.conn_handler.execute_fetchall(
> > +            "SELECT * FROM qiita.required_sample_info WHERE study_id=2")
> > +        # sample_id study_id physical_location has_physical_specimen
> > +        # has_extracted_data sample_type required_sample_info_status_id
> > +        # collection_timestamp host_subject_id description
> > +        exp = [["2.12.Sample1", 2, "location1", True, True, "type1", 1,
>
> What is the sample ID in this string: "123.foo"?
>
> —
> <https://github.com/biocore/qiita/pull/843/files#r24535664>.
>



-- 
Jose Navas

wasade: Okay, let me rephrase, does "123.foo" contain a study ID prefix or not?

josenavas: Hahaha I know that was your next question. And I agree that without
context, that is really confusing.
If that string is a sample id on Qiita, yes it does. If it is outside
Qiita, the study id is meaningless so you can arbitrarily decide if it does
or not...

2015-02-11 13:50 GMT-08:00 Daniel McDonald <notifications@github.com>:

> In qiita_db/test/test_metadata_template.py
> <https://github.com/biocore/qiita/pull/843#discussion_r24537015>:
>
> > @@ -652,6 +775,133 @@ def test_create(self):
> >                 ['2.Sample3', "Value for sample 3"]]
> >          self.assertEqual(obs, exp)
> >
> > +    def test_create_int_prefix(self):
> > +        """Creates a new SampleTemplate"""
> > +        st = SampleTemplate.create(self.metadata_int_pref, self.new_study)
> > +        # The returned object has the correct id
> > +        self.assertEqual(st.id, 2)
> > +
> > +        # The relevant rows to required_sample_info have been added.
> > +        obs = self.conn_handler.execute_fetchall(
> > +            "SELECT * FROM qiita.required_sample_info WHERE study_id=2")
> > +        # sample_id study_id physical_location has_physical_specimen
> > +        # has_extracted_data sample_type required_sample_info_status_id
> > +        # collection_timestamp host_subject_id description
> > +        exp = [["2.12.Sample1", 2, "location1", True, True, "type1", 1,
>
> Okay, let me rephrase, does "123.foo" contain a study ID prefix or not?
>
> —
> <https://github.com/biocore/qiita/pull/843/files#r24537015>.
>



-- 
Jose Navas

wasade: but even within qiita, we don't actually know unless we know exactly where
the value came from.

josenavas: That's not correct. If the sample has been added to the database (i.e. it
appears on required_sample_info table) then it is the study id, because all
the samples are always prefixed.

If we think that another character is need, I'm fine with modifying it, but
the only characters that are accepted in the sample ids in QIIME are
letters, numbers and '.' so I don't know what else we can use...

Also, having the study id in the sample name is a convenience, so we can
ensure unique sample ids across the entire database; but it is meaningless
in any other situation...

2015-02-11 13:56 GMT-08:00 Daniel McDonald <notifications@github.com>:

> In qiita_db/test/test_metadata_template.py
> <https://github.com/biocore/qiita/pull/843#discussion_r24537536>:
>
> > @@ -652,6 +775,133 @@ def test_create(self):
> >                 ['2.Sample3', "Value for sample 3"]]
> >          self.assertEqual(obs, exp)
> >
> > +    def test_create_int_prefix(self):
> > +        """Creates a new SampleTemplate"""
> > +        st = SampleTemplate.create(self.metadata_int_pref, self.new_study)
> > +        # The returned object has the correct id
> > +        self.assertEqual(st.id, 2)
> > +
> > +        # The relevant rows to required_sample_info have been added.
> > +        obs = self.conn_handler.execute_fetchall(
> > +            "SELECT * FROM qiita.required_sample_info WHERE study_id=2")
> > +        # sample_id study_id physical_location has_physical_specimen
> > +        # has_extracted_data sample_type required_sample_info_status_id
> > +        # collection_timestamp host_subject_id description
> > +        exp = [["2.12.Sample1", 2, "location1", True, True, "type1", 1,
>
> but even within qiita, we don't actually know unless we know exactly where
> the value came from.
> … <#14b7aa4b1aaafbe5_>
> On Wed, Feb 11, 2015 at 2:54 PM, josenavas <notifications@github.com>
> wrote: In qiita_db/test/test_metadata_template.py <
> https://github.com/biocore/qiita/pull/843#discussion_r24537379>: > @@
> -652,6 +775,133 @@ def test_create(self): > ['2.Sample3', "Value for sample
> 3"]] > self.assertEqual(obs, exp) > > + def test_create_int_prefix(self): >
> + """Creates a new SampleTemplate""" > + st =
> SampleTemplate.create(self.metadata_int_pref, self.new_study) > + # The
> returned object has the correct id > + self.assertEqual(st.id, 2) > + > +
> # The relevant rows to required_sample_info have been added. > + obs =
> self.conn_handler.execute_fetchall( > + "SELECT * FROM
> qiita.required_sample_info WHERE study_id=2") > + # sample_id study_id
> physical_location has_physical_specimen > + # has_extracted_data
> sample_type required_sample_info_status_id > + # collection_timestamp
> host_subject_id description > + exp = [["2.12.Sample1", 2, "location1",
> True, True, "type1", 1, Hahaha I know that was your next question. And I
> agree that without context, that is really confusing. If that string is a
> sample id on Qiita, yes it does. If it is outside Qiita, the study id is
> meaningless so you can arbitrarily decide if it does or not... 2015-02-11
> 13:50 GMT-08:00 Daniel McDonald <notifications@github.com>: …
> <#14b7aa3182127623_> In qiita_db/test/test_metadata_template.py <
> https://github.com/biocore/qiita/pull/843#discussion_r24537015>: > @@
> -652,6 +775,133 @@ def test_create(self): > ['2.Sample3', "Value for sample
> 3"]] > self.assertEqual(obs, exp) > > + def test_create_int_prefix(self): >
> + """Creates a new SampleTemplate""" > + st =
> SampleTemplate.create(self.metadata_int_pref, self.new_study) > + # The
> returned object has the correct id > + self.assertEqual(st.id, 2) > + > +
> # The relevant rows to required_sample_info have been added. > + obs =
> self.conn_handler.execute_fetchall( > + "SELECT * FROM
> qiita.required_sample_info WHERE study_id=2") > + # sample_id study_id
> physical_location has_physical_specimen > + # has_extracted_data
> sample_type required_sample_info_status_id > + # collection_timestamp
> host_subject_id description > + exp = [["2.12.Sample1", 2, "location1",
> True, True, "type1", 1, Okay, let me rephrase, does "123.foo" contain a
> study ID prefix or not? — Reply to this email directly or view it on GitHub
> <https://github.com/biocore/qiita/pull/843/files#r24537015>. -- Jose
> Navas — Reply to this email directly or view it on GitHub <
> https://github.com/biocore/qiita/pull/843/files#r24537379>.
>
> —
> <https://github.com/biocore/qiita/pull/843/files#r24537536>.
>



-- 
Jose Navas

wasade: Right, so you need external state maintained by a separate process in order to know how to interpret the variable -- resident memory, the variable, and its type are insufficient. That sucks as it opens the door for bugs and race conditions. 

If the downstream issue is producing valid qiime mapping files, then we can at least delay the construction of a valid qiime sample ID until the last moment, and represent these data internally in a manner that reflects their aggregate nature. 

If this ambiguity doesn't bother anyone else, then lets move on.

josenavas: Note that we cannot arbitrarily change the sample ids when we want, as the
demultiplexed files and the biom tables will include those sample ids...

Let's see what other Qiita devs think about this

2015-02-11 14:17 GMT-08:00 Daniel McDonald <notifications@github.com>:

> In qiita_db/test/test_metadata_template.py
> <https://github.com/biocore/qiita/pull/843#discussion_r24539391>:
>
> > @@ -652,6 +775,133 @@ def test_create(self):
> >                 ['2.Sample3', "Value for sample 3"]]
> >          self.assertEqual(obs, exp)
> >
> > +    def test_create_int_prefix(self):
> > +        """Creates a new SampleTemplate"""
> > +        st = SampleTemplate.create(self.metadata_int_pref, self.new_study)
> > +        # The returned object has the correct id
> > +        self.assertEqual(st.id, 2)
> > +
> > +        # The relevant rows to required_sample_info have been added.
> > +        obs = self.conn_handler.execute_fetchall(
> > +            "SELECT * FROM qiita.required_sample_info WHERE study_id=2")
> > +        # sample_id study_id physical_location has_physical_specimen
> > +        # has_extracted_data sample_type required_sample_info_status_id
> > +        # collection_timestamp host_subject_id description
> > +        exp = [["2.12.Sample1", 2, "location1", True, True, "type1", 1,
>
> Right, so you need external state maintained by a separate process in
> order to know how to interpret the variable -- resident memory, the
> variable, and its type are insufficient. That sucks as it opens the door
> for bugs and race conditions.
>
> If the downstream issue is producing valid qiime mapping files, then we
> can at least delay the construction of a valid qiime sample ID until the
> last moment, and represent these data internally in a manner that reflects
> their aggregate nature.
>
> If this ambiguity doesn't bother anyone else, then lets move on.
>
> —
> <https://github.com/biocore/qiita/pull/843/files#r24539391>.
>



-- 
Jose Navas

wasade: I have not proposed mutability.

josenavas:
The failure is stochastic. Looking at the output of travis, I can see that now there is a lot of warnings (coming from skbio). Given this situation, I would suggest creating our own Warning (subclassing the userwarning) and only allowing that one to be raised always. What do you think @biocore/qiita-dev ?

wasade:
Just to check, are valid paths in the code in which the method to add prefixes will be called to add prefixes to samples that are already prefixed? If so, instead of a warning (or an error?), can we just ensure this does not happen? Given the problem in #790, isn't this PR actually just trusting that the prefix provided by the user is correct? Based on the comments in #790, that is what it is doing, which seems very risky.

josenavas:
Is not trusting on the prefix provided by the user. Is actually checking that is correct (there is only one prefix correct - the study id). Check this line: https://github.com/biocore/qiita/pull/843/files#diff-0ffa6697592f287858f9880045642898R153

wasade:
Fine, but that is a warning and not an exception?

josenavas:
That's the whole point of the PR. The user will be able to upload a sample/prep template that has been already downloaded from qiita...

wasade:
So why not just allow the user to download unprefixed templates?

josenavas:
Cause it then will not match the biom table files or the split libraries files...

wasade:
It is trivial to add when needed. The other argument is that we do away completely with user specified sample IDs and instead always represent them (no matter where) with the prefixed study ID. Allowing both sample ID and the aggregate study_id + sample ID to be treated as one and the same, without a reliable means to deconvolute, is fragile for the reasons I stated already.

josenavas:
If I understand correctly, what you propose is to always represent the sample ids (no matter where) prefixed with the study ID, right? That is what we are trying to do in Qiita and in every part of the code this is treated like that. The only situation where this was broken is when a user tries to re-upload a sample/prep template that has been downloaded from Qiita. At this point the system was trying to prefix again the study id, giving as a result a failure in the insertion of the prep template, as the sample id's did not match.

This PR is only fixing that specific situation.

wasade:
Just to be clear, what you're saying is that the _only_ place in the qiita codebase, database and production filesystem where a sample ID might not be an aggregate is in this method?

josenavas:
It's not aggregated because it is already present.

wasade:
That didn't answer the question though. I ask because I have data pulled from qiita previously that does not have sample IDs aggregated with the study ID which suggests that there are multiple places in code where non-aggregated IDs are in use.

We're just going in loops. I'd be happy to continue on hangouts or skype, but I don't think it is productive to continue this discussion here.

josenavas:
That is actually unexpected. If you have data pulled from Qiita in which the study ids are not aggregated means that there is some bug in some other place in the codebase.

Which data is that one and when was it downloaded?

wasade:
AG 11, 13, 14. Last pull was November or so

josenavas:
Feature was added on November 10 (https://github.com/biocore/qiita/pull/583), before patching system in place; meaning that the data that you have was not generated with the live version of Qiita.

wasade:
I guarantee it was generated with the live version of qiita. I can fwd you
a tome of email if you'd like.

wasade:
Those studies were loaded prior to 11/10.

ElDeveloper:
If we created our own warning, we could centralize the import of
simplefilter to live inside a function, something like:

```
def raise_qiita_warning(msg):
    warnings.warn('Something kinda bad just happened ...')
```

That way the import to the module and the setup of the filter could live
in there and other parts of the code would only need to import
raise_qiita_warning.

ElDeveloper:
Yes, that makes sense :+1:, it also would be nice if we can get rid of
the skbio warnings in the travis build.

josenavas:
@wasade I remember now. The studies were loaded to the database; and the database was wiped-out and those studies were re-added keeping the study id, so that's why when you downloaded the data (previous to the official release) it did not include the study ids.

@ElDeveloper I don't think it is needed to create our own function. If we put that call in the qiita_db/__init__.py it should suffice to raise always our warnings. I will give it a try here!

ElDeveloper:
I think that makes sense, would this make it such that warnings from
other packages are always raised? I was thinking we could enclose this
in its own module, but maybe not ...

josenavas:
No if we are using our own warning class :)

ElDeveloper:
Makes sense! :+1:

antgonza:
Pull from upstream master.

wasade:
ah, @josenavas, that would explain it. However, the AG studies do not appear to have been reloaded. 

We still have an inconsistency issue though as those 3 AG rounds, and at least 2 other qiita studies that I can think of, were submitted to EBI prior to the change in representation meaning that the sample IDs in EBI for those studies are not prefixed.

josenavas:
@wasade I just checked with Gail and those studies are on the main Qiita site. They are still sandboxed and she is sharing them with you so can download the latest version. Can you double check that the current version that is on the system has the correct sample ids on the mapping file?

If they are correct, it would be worth to check if they are also correct in the post split libraries files and, in case they're wrong, we can reprocess them and update EBI.

wasade:
I can see AG 14, 15 and the BB 16S but not 11 or 13. 14 says there isn't
metadata associated with it

ackermag:
I have added you as a user on the other 2 studies

On Thu, Feb 12, 2015 at 9:40 AM, Daniel McDonald <notifications@github.com>
wrote:

> I can see AG 14, 15 and the BB 16S but not 11 or 13. 14 says there isn't
> metadata associated with it
>
> On Thu, Feb 12, 2015 at 10:37 AM, josenavas <notifications@github.com>
> wrote:
>
> > @wasade <https://github.com/wasade> I just checked with Gail and those
>
> > studies are on the main Qiita site. They are still sandboxed and she is
> > sharing them with you so can download the latest version. Can you double
> > check that the current version that is on the system has the correct
> sample
> > ids on the mapping file?
> >
> > If they are correct, it would be worth to check if they are also correct
> > in the post split libraries files and, in case they're wrong, we can
> > reprocess them and update EBI.
> >
> > —
> > Reply to this email directly or view it on GitHub
> > <https://github.com/biocore/qiita/pull/843#issuecomment-74115293>.
>
> >
>
> —
> <https://github.com/biocore/qiita/pull/843#issuecomment-74115927>.
>



-- 
Gail Ackermann
Knight Lab
UCSD

wasade:
Thanks. I see 11 and 13 now. It says "Metadata: False" and "Samples:
None/None" for 11, 13 and 14.

ackermag:
Does that mean I should reload and process?


On Thu, Feb 12, 2015 at 9:57 AM, Daniel McDonald <notifications@github.com>
wrote:

> Thanks. I see 11 and 13 now. It says "Metadata: False" and "Samples:
> None/None" for 11, 13 and 14.
>
>
> On Thu, Feb 12, 2015 at 10:54 AM, ackermag <notifications@github.com>
> wrote:
>
> > I have added you as a user on the other 2 studies
> >
> > On Thu, Feb 12, 2015 at 9:40 AM, Daniel McDonald <
> notifications@github.com
> > >
> > wrote:
> >
> > > I can see AG 14, 15 and the BB 16S but not 11 or 13. 14 says there
> isn't
> > > metadata associated with it
> > >
> > > On Thu, Feb 12, 2015 at 10:37 AM, josenavas <notifications@github.com>
> > > wrote:
> > >
> > > > @wasade <https://github.com/wasade> I just checked with Gail and
> those
> > >
> > > > studies are on the main Qiita site. They are still sandboxed and she
> is
> > > > sharing them with you so can download the latest version. Can you
> > double
> > > > check that the current version that is on the system has the correct
> > > sample
> > > > ids on the mapping file?
> > > >
> > > > If they are correct, it would be worth to check if they are also
> > correct
> > > > in the post split libraries files and, in case they're wrong, we can
> > > > reprocess them and update EBI.
> > > >
> > > > —
> > > > Reply to this email directly or view it on GitHub
> > > > <https://github.com/biocore/qiita/pull/843#issuecomment-74115293>.
> > >
> > > >
> > >
> > > —
> > > Reply to this email directly or view it on GitHub
> > > <https://github.com/biocore/qiita/pull/843#issuecomment-74115927>.
> > >
> >
> >
> >
> > --
> > Gail Ackermann
> > Knight Lab
> > UCSD
> > glackermann@ucsd.edu <ackermag@ucsd.edu>
> >
> > —
> > Reply to this email directly or view it on GitHub
> > <https://github.com/biocore/qiita/pull/843#issuecomment-74118351>.
>
> >
>
> —
> <https://github.com/biocore/qiita/pull/843#issuecomment-74118907>.
>



-- 
Gail Ackermann
Knight Lab
UCSD

antgonza:
If I remember correctly those studies were not actually processed in
qiita as they were urgent and qiita wasn't at that point then. We
processed by hand and then submitted using the qiita command line.

wasade:
demux was done by qiita, submission was done either via qiita's cmdline
interface or the web interface

@ackermag, I don't know

On Thu, Feb 12, 2015 at 11:04 AM, Antonio Gonzalez <notifications@github.com

josenavas:
I think we need to fix the metadata column on that page. That one is based in one column in the study table, which I would suggest that we no longer need. We can populate that value just by checking if a sample template exists.

@ackermag I don't think we need to reload, just reprocess.

josenavas:
This is ready for review!

ElDeveloper:
:+1: the warning addition looks good!

josenavas:
Thanks @ElDeveloper can I have another reviewer here?

qiita_db/test/test_metadata_template.py
antgonza: It seems like this is not resolved, @wasade do you have an actual solution? Agree that it's not ideal but is a solution. 

About your example: 123.foo, this has been prefixed if: 123 is the id of the study, and all the other study ids within that sample template have the same prefix (or is the only sample). This make this not as ambiguous, at least in my opinion.

wasade: It is still ambiguous as it is not assured to be true.

My concern is partially for this method, but more generally for qiita. The more we infer state that is vital for data integrity from an unstructured string (and in this PR it is untrusted as well), the more we open the door to very bad things happening in the future.

In this case, the most ideal solution is to never allow a user to upload something that has IDs of the form "study_id.sample_id". 

Within the rest of qiita, we should only ever manage sample IDs internally as a structured datatype that reflects its aggregate nature. We should only combine these datatypes at the last possible moment, and only when necessary.

I disagree that being lax with core datatypes that are vital to the integrity of our data model is a solution particularly as we're developing a system that will need to maintain its integrity for years.

josenavas: Are the study ids aggregated to the sample ids used anywhere in the code? AFAIK, there is no place where such study  id is used, and it is aggregated at the upload time. The only reason why they are added is because it is an easy way to define a DB global sample id...

wasade: As I indicated in this thread, one possible solution is to have global IDs which are the aggregate, as long as we do not have some places in code which expect "sample_id" and other places which expect "study_id.sample_id". The reason that solution is not ideal is that it denormalizes as well as encourages developers to think of the sample ID as more than just an ID. This method is an example of where we are violating the concept of a sample ID being just an ID. I do not know if there are other places in qiita impacted, but I would not be surprised if there are and I definitely would not be surprised if methods that do this get merged in in the future.

I've stated my concerns over these approaches. If no one else is concerned, then we should just drop it.

antgonza: I understand your concerns but not sure a global id is the actual way to go. For instance, if a user downloads 2 studies with the same sample names (blank.1, blank.2, ...) and tries to perform a meta-analysis, it will not work as expected as it will merge those samples. Having the idea of unique ids, will require the users to have an easy way to relate this unique id with their actual sample ids, which we do not have. The study_id.sample_id is not perfect but we haven't found a easier solution. Perhaps the way to go is to stop using text files and start using binary with extra metadata, like the hdf5 solutions we have talked about. Until those solutions are in place, I think this is our best bet. Perhaps, adding an issue so we do not forget and once we have solved the [next release](https://github.com/biocore/qiita/milestones/Alpha%200.1) issues we can address this and other ones in the same area.

josenavas: Ok. I think I see your point @wasade 
In the codebase, we are using the aggregated sample IDs as a unique sample id, without having the idea that they are actually aggregated data. So, this goes back to the real question of, how do we solve issue #790?
- Solution 1 (proposed on this PR): add a test on template load to check that the study ids were already present and we don't need to re-add them:
  * PROS: Completely transparent to the user
  * CONS: There is one place on the code base where we actually check that the sample id is `study_id.sample_id`
- Solution 2 (proposed by @wasade): add the ability to download un-prefixed templates
  * PROS: We don't need an implicit check
  * CONS: We still need to trim those study ids at download time. The user will need to know that for meta analyses he needs the prefixed template. The user will need to know that he has to download the un-prefixed version to be able to re-upload that template to QIITA.
- Solution 3: others?

Please, @biocore/qiita-dev , vote, add/remove and propose so we can reach a consensus on this

josenavas: My vote goes (of course) for 1, as it has the lowest impact on the user and we can deterministically perform that check with no case in which the user can mess the DB.

wasade: I have never advocated allowing a user to perform a meta analysis on non-unique IDs, but instead to delay the construction of those IDs until the last moment (and hopefully with sufficient explicit detail to disambiguate later).

Either I'm not expressing my concerns effectively, or no one else thinks this is a problem. As such, lets forget about it.

antgonza: :+1: to 1

wasade: We don't need to trim the study IDs if we do not store the IDs as aggregates. The user should not need to know what template is necessary for a meta analysis as you cannot do a meta analysis in qiita without having loaded your data in some capacity. The user likely wants to be able to download something isomorphic to what they uploaded?

1 is the easiest thing to do right now. 2 respects the nature of the data in the future but suggests changes to the database, meaning it will only be harder as time passes.

josenavas: @wasade Note that storing the study id and the sample id in two different columns will bring us back to the original design, where we actually found a lot of issues and we end up doing a lot changes in the DB to just aggregate the study id to the sample id (that was done on september, and I suffered it so I remember :tired_face: )

wasade: Yes, which makes sense as long as we only ever treat the ID as an ID.

josenavas:
ping!

adamrp:
The code looks okay on first inspection. Going to do some testing.

adamrp:
FWIW, I agree with @wasade's suggestions about not prefixing prematurely (as I have said in previous issues, but wanted to reiterate here for the record). But if the consensus is always to prefix, then I'll go with it.

adamrp:
:+1:

josenavas:
Can this be merged? I already have to :+1:

josenavas:
Thanks!

PR #844, opened 2015-02-10
Change way browser and version found

squirrelo:
There was a problem with firefox and opera where the version found was not the actual version of the browser. This new code fixes that.

### DIFF
--- a/qiita_pet/templates/sitebase.html
+++ b/qiita_pet/templates/sitebase.html
@@ -24,16 +24,16 @@
     <link rel="shortcut icon" href="/static/img/favicon.ico">
 
     <script>
-      $(document).ready(function() {
-        function overlay() {
-          overlay = document.getElementById("overlay");
-          overlay.style.visibility = "visible";
-          navbar = document.getElementById("navigation-bar");
-          navbar.style.visibility = "hidden";
-          jumbotron = document.getElementById("jumbotron");
-          jumbotron.style.visibility = "hidden";
-        }
+      function overlay() {
+        ol = document.getElementById("overlay");
+        ol.style.visibility = "hidden";
+        navbar = document.getElementById("navigation-bar");
+        navbar.style.visibility = "visible";
+        jumbotron = document.getElementById("jumbotron");
+        jumbotron.style.visibility = "visible";
+      }
 
+      $(document).ready(function() {
         // from http://stackoverflow.com/q/5916900/379593
         var ua=navigator.userAgent,tem,M=ua.match(/(opera|chrome|safari|firefox|msie|trident(?=\/))\/?\s*(\d+)/i) || []; 
         if(/trident/i.test(M[1])){
@@ -49,10 +49,10 @@
         objbrowserName = M[0];
         objBrMajorVersion = M[1];
 
-        if ((objbrowserName == "Chrome"  && objBrMajorVersion < {{allowed_min_browser_versions['chrome']}}) ||
-            (objbrowserName == "Firefox" && objBrMajorVersion < {{allowed_min_browser_versions['firefox']}}) ||
-            (objbrowserName == "Safari"  && objBrMajorVersion < {{allowed_min_browser_versions['safari']}}) ||
-            (objbrowserName == "IE" && objBrMajorVersion < {{allowed_min_browser_versions['IE']}})) {
+        if ((objbrowserName == "Chrome"  && objBrMajorVersion >= {{allowed_min_browser_versions['chrome']}}) ||
+            (objbrowserName == "Firefox" && objBrMajorVersion >= {{allowed_min_browser_versions['firefox']}}) ||
+            (objbrowserName == "Safari"  && objBrMajorVersion >= {{allowed_min_browser_versions['safari']}}) ||
+            (objbrowserName == "IE" && objBrMajorVersion >= {{allowed_min_browser_versions['IE']}})) {
           overlay();
         }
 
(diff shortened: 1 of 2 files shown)


qiita_pet/templates/sitebase.html
```diff
@@ -34,37 +34,20 @@
...
-        if ((ix=objfullVersion.indexOf(";"))!=-1)
-          objfullVersion=objfullVersion.substring(0,ix);
-        if ((ix=objfullVersion.indexOf(" "))!=-1)
-          objfullVersion=objfullVersion.substring(0,ix);
-        objBrMajorVersion = parseInt(''+objfullVersion,10);
-        if (isNaN(objBrMajorVersion)) {
-          objfullVersion = ''+parseFloat(navigator.appVersion);
-          objBrMajorVersion = parseInt(navigator.appVersion,10);
-        }
+        // from http://stackoverflow.com/questions/5916900/how-can-you-detect-the-version-of-a-browser
```
ElDeveloper: Can you use the permanent link instead? http://stackoverflow.com/q/5916900/379593

qiita_pet/templates/sitebase.html
```diff
@@ -34,37 +34,20 @@
...
+        var ua=navigator.userAgent,tem,M=ua.match(/(opera|chrome|safari|firefox|msie|trident(?=\/))\/?\s*(\d+)/i) || []; 
+    if(/trident/i.test(M[1])){
```
ElDeveloper: This whole block is under indented.

ElDeveloper:
This looks *good* except, there's not a good way to test this code, it has the "same" validity as the other chunk of code taken from a blog post found on the inter-tubes and that the code is not really maintainable. IMHO, should be ready to merge as soon as the comments I made are addressed but then again ... how can we make sure this new snippet is not introducing any problems that the previous block was accounting for? I would probably advocate for finding the bug in the previous block of code and correct it there.

antgonza:
The issue is that there is no perfect way and more when browsers change the way they report their version, which made the previous one obsolete. I think this is fine as long as we test in all valid browsers.

ElDeveloper:
Ok, sounds good. We should check Safari, Chrome, Opera, Internet
Explorer, Firefox and lynx? With the corresponding mobile versions where
possible?

antgonza:
Up to you. I normally check Safari, Chrome, and Firefox.

ElDeveloper:
Well the only browser that shows the error message is lynx, however the interface is unusable in internet explorer 8.

Just for fun:

screen shot 2015-02-10 at 3 51 53 pm

wasade:
@ElDeveloper, that is a fantastic screenshot

GabeAl:
Still no good. I'm being inappropriately flagged (Firefox 35.0.1)
baddetect
 and cannot access the site. I also do not use Chrome and do not run a Mac. Firefox is the only truly open, cross-platform, non-corporate browser that enjoys modern standards support. For Linux (and Windows) users who can't (or won't) use Chrome, it's probably better to botch the detection and include an "unsupported browser" blurb at the top than to blanket ban. 
Cheers

antgonza:
Are you testing on this branch or on the main site?

GabeAl:
Apologies -- actually, this is for the main site as it currently exists, not the proposed changes, thereby merely corroborating the finding in the opening message.

squirrelo:
So apparently flake8 updated and is throwing all sorts of fun errors. I'll fix that in aseperate emergency pull request.

ElDeveloper:
I'm on it, see #846

ElDeveloper:
almost done also ...

squirrelo:
Oops, pull request already in.

antgonza:
Pull from upstream master.

squirrelo:
ready for review again.

antgonza:
:+1: @ElDeveloper ?

ElDeveloper:
It still doesn't work in IE 8 i. e. no message about the website not working is shown. :confused:

squirrelo:
good to review again

ElDeveloper:
It now shows the appropriate message for IE8, however it shows the
message in IE10 and Opera (it wasn't before), from the displayed message
it seems that at least IE10 is supported.

squirrelo:
Bluh, The issue is that the older browsers flat out won't run the version check javascript, so I defaulted it to just throw the error by default. Not sure whats going on with IE 10, but we don't support Opera so it's getting the default error page.

ElDeveloper:
So then the only open issue is IE 10. And we may also need to rephrase
the error message to: "This site only works on these browsers".

squirrelo:
Yup. IE isn't firing off the document.ready so I'll dig into that.

squirrelo:
Went old school and used `body onload=''` to fire off the check, since that's supported by pretty much everything. This also means the site is visible by default now for non-checked browsers, so Opera will show the site again.

ElDeveloper:
If we don't support Opera should that be changed then?

squirrelo:
I think we can have in the documentation that we support X browsers above Y version, and while the site is accessible with other browsers , we make no claims to usability.

antgonza:
Agree, could you add that list to the help wiki?

antgonza:
@ElDeveloper could you check again that it works on the expected browsers? Seems like you have everything in your system to do this ... thanks!

squirrelo:
Added to the FAQ section.

antgonza:
Thanks, looks fine:
https://github.com/biocore/qiita/wiki/Frequently-Asked-Questions

qiita_pet/templates/sitebase.html
```diff
@@ -24,69 +24,41 @@
...
-        if ((objbrowserName == "Chrome"  && objBrMajorVersion < {{allowed_min_browser_versions['chrome']}}) ||
-            (objbrowserName == "Firefox" && objBrMajorVersion < {{allowed_min_browser_versions['firefox']}}) ||
-            (objbrowserName == "Safari"  && objBrMajorVersion < {{allowed_min_browser_versions['safari']}})) {
-          overlay();
-        }
-
-      });
-    </script>
+      }
+
```
ElDeveloper: Extra whitespace.

ElDeveloper:
Just tested this in all the browsers I have access to and it works just
fine! Lynx actually lets me log in (unlike IE8). Just one more minor
comment about whitespace and this should be merged right away!

squirrelo:
whitespace removed, merge ready.

PR #845, opened 2015-02-10
Issue 835

squirrelo:
This fixes the analysis filepath issues and adds a test to make sure it doesn't happen again.


ElDeveloper:
@squirrelo, there are legitimate flake8 errors in this build.

squirrelo:
Not any more.

ElDeveloper:
:boom: :boom:

qiita_db/test/test_analysis.py
```diff
@@ -427,6 +428,26 @@ def test_build_files_raises_value_error(self):
         with self.assertRaises(ValueError):
             self.analysis.build_files(-10)
 
+    def test_add_file(self):
+        fp = join(get_mountpoint('analysis')[0][1], 'testfile.txt')
+        try:
```
ElDeveloper: If I understand correctly: if you can write to the file path you add it to the analysis and regardless of the result of these two lines you check if it exists and if it exists you remove it? Should that fp be removed in the teardown method of the test class? So we can guarantee that it is removed, the finally block here is kinda confusing.

ElDeveloper:
This PR looks ok but I'm not super familiar with this codebase so it would be nice if someone else could review it in depth.

squirrelo:
Yes, the finally makes it so the test always removes the file created
regardless of if the test errors or not. I can move it to teardown I guess,
I just hate having test specific things in teardown instead of in the test.

ElDeveloper:
Ah got it. IMHO it's clearer if it's part of the teardown method (maybe
because that's what I'm used to see) than as an inline try/except block.
teardown methods are supposed to be dealing with these kind of things so
it's really up to you.

qiita_db/util.py
```diff
@@ -608,8 +608,8 @@ def get_mountpoint(mount_type, conn_handler=None, retrieve_all=False):
             "SELECT data_directory_id, mountpoint, subdirectory FROM "
             "qiita.data_directory WHERE data_type='%s' and active=true"
             % mount_type)]
-
-    return [(d, join(get_db_files_base_dir(), m, s)) for d, m, s in result]
+    basedir = get_db_files_base_dir()
```
antgonza: I do not understand this change, is it required or is for clarity?

squirrelo: The function does an SQL select every time it's called, so this makes it happen 1 time instead of N times.

antgonza: Excellent point! Thanks for catching that.

josenavas: Should this use `get_db_files_base_dir` or should we move to the `get_mountpoint` function that @antgonza added a while ago?

antgonza: This line is within the get_mountpoint function ....

qiita_db/analysis.py
```diff
@@ -753,7 +753,7 @@ def _add_file(self, filename, filetype, data_type=None, conn_handler=None):
                "checksum, checksum_algorithm_id, data_directory_id) VALUES "
                "(%s, %s, %s, %s, %s) RETURNING filepath_id")
         # magic number 1 is for crc32 checksum algorithm
-        fpid = conn_handler.execute_fetchone(sql, (fullpath, fptypeid,
+        fpid = conn_handler.execute_fetchone(sql, (filename, fptypeid,
```
josenavas: This line looks ok. However, I'm looking at the function in general `_add_file` and it looks pretty similar to `insert_filepaths` in the `util.py` file. Is it possible to remove this one and use the `insert_filepaths` function here, so no code duplication happens in any place? BTW, I would say that the code duplication was the source of this bug...

qiita_db/support_files/patches/14.sql
```diff
@@ -0,0 +1,2 @@
+--Placeholder for python to run
```
josenavas: Can you add the date to the patch?

qiita_db/support_files/patches/python_patches/14.py
```diff
@@ -0,0 +1,15 @@
+from qiita_db.sql_connection import SQLConnectionHandler
+from os.path import basename
+
+# This changes all analysis files to be relative path instead of absolute
```
josenavas: Add date and move all comments to the first line of the file (for consistency with other patches)

qiita_db/util.py
josenavas: Ooops, I just saw that this is actually the `get_mountpoint function`... :sweat_smile::innocent:

josenavas:
Few comments @squirrelo

qiita_db/support_files/patches/python_patches/14.py
```diff
@@ -0,0 +1,15 @@
...
+
+conn_handler = SQLConnectionHandler()
+
+filepaths = conn_handler.execute_fetchall(
+    'SELECT f.* from qiita.filepath f JOIN qiita.analysis_filepath afp ON '
+    'f.filepath_id = afp.filepath_id')
+for filepath in filepaths:
+    filename = basename(filepath['filepath'])
```
wasade: Just to verify: these files are still relative to a unique directory associated with a study or a meta-analysis, correct?

qiita_db/analysis.py
squirrelo: I can for the filepath insertion, but I still need it for the linking of analysis to the filepath.

josenavas: Ok, if you can use `insert_filepaths` to do part of the work here, use it, so we avoid code duplication. If you still need some code to finish other work it is fine to do it here.

qiita_db/support_files/patches/14.sql
squirrelo: done

antgonza:
Pull from upstream master.

squirrelo:
ready for review again.

antgonza:
The patch is not working in my machine because the mountpoint has an ending / and the file path doesn't
```
  File "qiita_db/support_files/patches/python_patches/14.py", line 23, in <module>
    mp_id = mountpoints[dirname(filepath['filepath'])]
KeyError: '/Users/antoniog/Desktop/CurrentWork/largeFiles/qiita/base_dir/analysis'
```
and I have these values in `filepath`, `mountpoints`
```
[308L, '/Users/antoniog/Desktop/CurrentWork/largeFiles/qiita/base_dir/analysis/1_analysis_mapping.txt', 9L, '833932644', 1L, 1L]
{'/Users/antoniog/Desktop/CurrentWork/largeFiles/qiita/base_dir/analysis/': 1L}
```

wasade:
Should the mount points have a trailing '/'? That technically refers to the _contents_ of the path, not that path

antgonza:
Got it but the current configuration allows you to have either so
validating both cases or not allowing one in the configuration should
be the way to go.

antgonza:
:+1:

antgonza:
@squirrelo conflicts, @ElDeveloper could you merge after conflicts are solved? Thanks.

qiita_db/support_files/patches/14.sql
ElDeveloper: Haha, either git or the conflict resolution got messed, this change should have only gone in 15.sql ... I'll let it slide :laughing:

PR #848, opened 2015-02-12
STY: Update flake8 requirements

### LINKED ISSUE #846: flake8 requirements have changed
Some of the builds are failing because flake8 has been updated and is now flagging things it wasn't before. To reproduce this problem, upgrade your version of flake8 and run it in your repo or have a look at these builds:

- [build 1](https://travis-ci.org/biocore/qiita/builds/50422108).
- [build 2](https://travis-ci.org/biocore/qiita/builds/50420752).

ElDeveloper:
Fixes #846

### DIFF
--- a/qiita_ware/wrapper.py
+++ b/qiita_ware/wrapper.py
@@ -1,13 +1,6 @@
 from __future__ import division
 
-__author__ = "Jose Antonio Navas Molina"
-__copyright__ = "Copyright 2011, The QIIME project"
-__credits__ = ["Jose Antonio Navas Molina"]
-__license__ = "BSD"
-__version__ = "1.8.0-dev"
-__maintainer__ = "Jose Navas Antonio Molina"
-# Taken from pull request #1616 in qiime, with permission from Jose
-# Permission was also given to release this code under the BSD licence
+from qiita_db.job import Job
 
 from shutil import rmtree
 from os import remove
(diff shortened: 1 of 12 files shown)


squirrelo:
Good sir you have beaten me. I concede. :+1:

ElDeveloper:
https://www.youtube.com/watch?v=TAryFIuRxmQ

qiita_ware/wrapper.py
```diff
@@ -17,7 +10,15 @@
 import networkx as nx
 from moi.job import system_call, submit, ctxs, ctx_default
 
-from qiita_db.job import Job
+
```
josenavas: I think these lines can be removed as they're leftover of a copy-paste.

josenavas:
Thanks @ElDeveloper Just one minor comment.

Just for clarity, the imports should not be organized now as:
```
std lib imports
<new line>
third party lib imports
<new line>
project imports
```
?

Which is now the rule for the imports? In order to save some developer time...

squirrelo:
The problem it's yelling about is that any imports after the `with .....` import are considered to be out of order. We can keep that same order we are using now and just move the `with` import to the bottom of that list.

ElDeveloper:
If you see the links in #824, flake8 was not very clear about this, but
it consistently complained about the module-imports not being at the top
of the file.  And for some reason it no longer likes *some* lambdas ...
weird.

ElDeveloper:
@squirrelo, you are right, I'll change that right away!

squirrelo:
:+1:

PR #850, opened 2015-02-12
Delete prep template

antgonza:
https://github.com/biocore/qiita/pull/842 should be merged first.

### DIFF
--- a/qiita_pet/handlers/study_handlers/description_handlers.py
+++ b/qiita_pet/handlers/study_handlers/description_handlers.py
@@ -627,16 +627,17 @@ class StudyDescriptionHandler(BaseHandler):
             is done
         """
         prep_template_id = int(self.get_argument('prep_template_id'))
+        tab_id = PrepTemplate(prep_template_id).raw_data
 
         try:
             PrepTemplate.delete(prep_template_id)
             msg = ("Prep template %d has been deleted" % prep_template_id)
             msg_level = "success"
-            tab_id = PrepTemplate(prep_template_id).raw_data
+            prep_id = None
         except Exception as e:
             msg = ("Couldn't remove prep template: %s" % str(e))
             msg_level = "danger"
-            tab_id = None
+            prep_id = prep_template_id
 
         callback((msg, msg_level, 'raw_data_tab', tab_id, None))
 
--- a/qiita_pet/handlers/study_handlers/description_handlers.py
+++ b/qiita_pet/handlers/study_handlers/description_handlers.py
@@ -639,7 +639,7 @@ class StudyDescriptionHandler(BaseHandler):
             msg_level = "danger"
             prep_id = prep_template_id
 
-        callback((msg, msg_level, 'raw_data_tab', tab_id, None))
+        callback((msg, msg_level, 'raw_data_tab', prep_id, None))
 
     @authenticated
     def get(self, study_id):
--- a/qiita_pet/handlers/study_handlers/description_handlers.py
+++ b/qiita_pet/handlers/study_handlers/description_handlers.py
@@ -627,7 +627,7 @@ class StudyDescriptionHandler(BaseHandler):
             is done
         """
         prep_template_id = int(self.get_argument('prep_template_id'))
-        tab_id = PrepTemplate(prep_template_id).raw_data
+        prep_id = PrepTemplate(prep_template_id).raw_data
 
         try:
             PrepTemplate.delete(prep_template_id)
@@ -637,7 +637,6 @@ class StudyDescriptionHandler(BaseHandler):
         except Exception as e:
             msg = ("Couldn't remove prep template: %s" % str(e))
             msg_level = "danger"
-            prep_id = tab_id
 
         callback((msg, msg_level, 'raw_data_tab', prep_id, None))
 
--- a/qiita_pet/handlers/study_handlers/description_handlers.py
+++ b/qiita_pet/handlers/study_handlers/description_handlers.py
@@ -613,6 +613,35 @@ class StudyDescriptionHandler(BaseHandler):
 
         callback((msg, msg_level, tab, tab_id, None))
 
+    def delete_prep_template(self, study, user, callback):
+        """Delete the selected prep template
+
+        Parameters
+        ----------
+        study : Study
+            The current study object
+        user : User
+            The current user object
+        callback : function
+            The callback function to call with the results once the processing
+            is done
+        """
+        prep_template_id = int(self.get_argument('prep_template_id'))
+
+        try:
+            PrepTemplate.delete(prep_template_id)
+            msg = ("Prep template %d has been deleted" % raw_data_id)
+            msg_level = "success"
+            tab = 'study_information_tab'
+            tab_id = None
+        except Exception as e:
+            msg = ("Couldn't remove prep template: %s" % str(e))
+            msg_level = "danger"
+            tab = 'raw_data_tab'
+            tab_id = PrepTemplate(prep_template_id).raw_data
+
+        callback((msg, msg_level, tab, tab_id, None))
+
     @authenticated
     def get(self, study_id):
         study, user = self._get_sudy_and_check_access(study_id)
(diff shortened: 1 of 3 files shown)


ElDeveloper:
I *think* you need to pull from upstream/master, but I could be wrong.

antgonza:
Done.

qiita_pet/handlers/study_handlers/description_handlers.py
```diff
@@ -613,6 +613,35 @@ def delete_raw_data(self, study, user, callback):
...
+            The callback function to call with the results once the processing
+            is done
+        """
+        prep_template_id = int(self.get_argument('prep_template_id'))
+
+        try:
+            PrepTemplate.delete(prep_template_id)
+            msg = ("Prep template %d has been deleted" % prep_template_id)
+            msg_level = "success"
+            tab = 'study_information_tab'
```
josenavas: Probably better to redirect in either case to the raw_data_tab? So from the user point of view the only change on the screen is that the prep template has been remove, rather than move to the RawData tab after the prep template is removed (he probably wants to add another prep template just after removing one).

antgonza: Done.

qiita_pet/handlers/study_handlers/description_handlers.py
```diff
@@ -613,6 +613,32 @@ def delete_raw_data(self, study, user, callback):
...
+
+        try:
+            PrepTemplate.delete(prep_template_id)
+            msg = ("Prep template %d has been deleted" % prep_template_id)
+            msg_level = "success"
+        except Exception as e:
+            msg = ("Couldn't remove prep template: %s" % str(e))
+            msg_level = "danger"
+
+        callback((msg, msg_level, 'raw_data_tab',
```
josenavas: This now is broken. If the removal worked successfully, the PrepTemplate with id `prep_template_id` no longer works. You just need to move the query to `PrepTemplate(prep_template_id).raw_data` before the try statement

antgonza: Duh!! Thanks for catching that.

qiita_pet/handlers/study_handlers/description_handlers.py
```diff
@@ -613,6 +613,33 @@ def delete_raw_data(self, study, user, callback):
 
         callback((msg, msg_level, tab, tab_id, None))
 
+    def delete_prep_template(self, study, user, callback):
```
josenavas: This is still broken... (sorry) the actual code should be this one:

```python
prep_template_id = int(self.get_argument('prep_template_id'))
tab_id = PrepTemplate(prep_template_id).raw_data

try:
	PrepTemplate.delete(prep_template_id)
	msg = ("Prep template %d has been deleted" % prep_template_id)
	msg_level = "success"
	prep_id = None
except Exception as e:
	msg = ("Couldn't remove prep template: %s" % str(e))
	msg_level = "danger"
	prep_id = prep_template_id

callback((msg, msg_level, 'raw_data_tab', tab_id, prep_template_id))

```

antgonza: Got it, I think you are right ...

josenavas:
I found another issue. Part of it is out of the scope of this PR but I think is easy enough to fix here too.

The removal of a raw data/prep template can be done independently of the study status. However, if a study is private/public raw data and prep templates can be removed. I don't think this should be allowed, at least if the study is public. If the study is private, however, I can see some discussion. But I would advocate that the user is adding metadata (prep template) so it should be checked...

antgonza:
I see you point but I do not think that's possible because to be able
to make it public you need processed object and if you have one, you
can not remove the preprocessed data, and if you ..., get it?

Either way, I can add that test, but depending where we do the
validation (qiita_db or qiita_pet), we will need to check the study
status. I will like to have it in qiita_db but that will require to
pass/create an study object within those objects. What do you think?

josenavas:
You can have a processed object from one raw data, but not from the other one. And you still will be able to delete that other one... Is just the idea that our "public" data should not be mutable in an easy way (or at least not by a normal user, just by an admin...).

I don't have strong feelings on where to put the check. Although it is a pretty light check on qiita_db that will not require interaction with the server, as the the study_status property is already used extensively in the study_description page. you can potentially propagate that value through the calls to the uimodules, so less DB interactions are needed.

(Also if you feel like it, adding a double check on qiita_db never hurts!)

qiita_pet/handlers/study_handlers/description_handlers.py
```diff
@@ -613,6 +613,34 @@ def delete_raw_data(self, study, user, callback):
...
+            PrepTemplate.delete(prep_template_id)
+            msg = ("Prep template %d has been deleted" % prep_template_id)
+            msg_level = "success"
+            prep_id = None
+        except Exception as e:
+            msg = ("Couldn't remove prep template: %s" % str(e))
+            msg_level = "danger"
+            prep_id = prep_template_id
+
+        callback((msg, msg_level, 'raw_data_tab', tab_id, None))
```
josenavas: You should pass `prep_id` instead of `None` here.

antgonza:
@josenavas and I discussed his comment offline and realized that we have a bigger issue and it's described here: https://github.com/biocore/qiita/issues/855

josenavas:
Given that we have created another issue to keep track of my concern, I would say that this PR is good to go. So if another developer can review this and agrees with the code, feel free to merge!

qiita_pet/handlers/study_handlers/description_handlers.py
```diff
@@ -613,6 +613,34 @@ def delete_raw_data(self, study, user, callback):
...
+        study : Study
+            The current study object
+        user : User
+            The current user object
+        callback : function
+            The callback function to call with the results once the processing
+            is done
+        """
+        prep_template_id = int(self.get_argument('prep_template_id'))
+        tab_id = PrepTemplate(prep_template_id).raw_data
```
ElDeveloper: It seems like `tab_id` could easily be replaced by `prep_id`. Plus I don't understand why the name of this variable is `tab..`.

josenavas: is not the `prep_id`. Is the `raw_data_id`. Also the name is tab because the callback function expect a message, the message level, the name of tab group to show, the id in that group (tab_id) and the id of the sub_tab in case that the previous ID points to another tab group.

ElDeveloper: Oh, ok. In any case using a single variable should work (I think).

qiita_pet/templates/study_description_templates/prep_template_panel.html
```diff
@@ -41,7 +41,10 @@ <h4 class="modal-title" id="myModalLabel">Choose preprocessing parameters</h4>
 <div class="panel panel-default">
   <div class="panel-heading">
     <h4 class="panel-title">
-      <a data-toggle="collapse" data-parent="#prep-accordion" href="#collapse{{prep_id}}">{{data_type}} (ID: {{prep_id}})</a>
+      <a data-toggle="collapse" data-parent="#prep-accordion" href="#collapse{{prep_id}}">
+        {{data_type}} (ID: {{prep_id}})
+        <button class="close" title="Remove this prep template" type="button" onclick="delete_prep_template('{{data_type}}', {{prep_id}})">&nbsp; ×</button>
```
ElDeveloper: Nice, I see this is consistent with the way raw datas are deleted. There is just one unintended side-effect. When I click this button and the alert comes up, if I choose Cancel, the prep template accordion is collapsed, not sure if we can bypass that so this doesn't happen?

ElDeveloper:
Just one comment, looks good otherwise.

antgonza:
OK, the latest push should address your two comments. Note that I completely deactivated the collapse on the prep templates cause (1) I spent a lot of time trying to figure out how to prevent it when you cancel the erase and (2) cause this is really confusing to users and the reason I deactivated by default in the first place.

ElDeveloper:
Thanks working in that, but now I can't click the delete button. On the
bright side, the accordion no longer collapses.

antgonza:
The tests passed but not sure why the PR wasn't updated.

ElDeveloper:
Tested it and it looks good!

PR #852, opened 2015-02-13
wip: raising coverage

antgonza:
Want to see if we can improve coverage but realized that perhaps we do not need most of the functions in qiita_core/environment_manager.py. @wasade, what do you think?


wasade:
I thought we did need methods that verify services?

antgonza:
Got it, there are used in `qiita_env test`. Any suggestion on how to test
the test? I started with something but not sure if it's the right path. I
will push again to fix the flake8 error and see what happens.

wasade:
Cut them out into a different module and have coverage omit them? If the methods are executed regularly by travis, then there should not be concern over their "correctness".

antgonza:
That will be fine with me. Basically all that is in environment_manager.py

qiita_core/test/__init__.py
```diff
@@ -0,0 +1,9 @@
+# -----------------------------------------------------------------------------
```
ElDeveloper: Test __init__.py should be empty, we just really want them so that our distributions ship with the test suite.

ElDeveloper:
This looks good and there was 0.47% coverage increase!

antgonza:
So you prefer this than the other option?

ElDeveloper:
Well this way we are explicitly testing and not ignoring so yes!!

antgonza:
OK so we are 1 and 1, we need more voters. Note that this specific file is
pretty hard to test and even the author (@wasade) thinks that it will be
better to follow the other approach.

josenavas:
My vote goes to the other PR not this one. Argument: who is gonna test the test of the test? (and continue forever and ever...)

ElDeveloper:
I would argue this is a service verification module and that we should
test the code that performs those verifications. Specially since we
depend on other tools we could catch compatibility problems with these
tests.

antgonza:
Closing this one as the internal voting preferred: https://github.com/biocore/qiita/pull/853

antgonza:
Closed the wrong one, reopening!

squirrelo:
Ping @antgonza  what is the status on this pull request?

antgonza:
Been busy, is this urgent?

squirrelo:
No, just status update.

josenavas:
@antgonza should be close this PR or are you still planning on working on this?

Issue #120, opened 2014-06-19
Test CLIs


ElDeveloper:
Likely using whatever click suggests.

RNAer:
I am working on this. The plan is to write the tests in `qiita/scripts/test/`, but the problem is that there seems no easy way to import the command functions from `qiita/scripts/qiita` and `qiita/scripts/qiita-env`, unless we set `qiita/scripts` as a submodule instead. @ElDeveloper @antgonza @squirrelo @josenavas , do you have any ideas?

josenavas:
I don't think there is a big problem of setting it as a submodule. @ElDeveloper @antgonza ?

antgonza:
I think submodules is not a big deal but could you explain why this is the best option? BTW what about setting another job to test this https://travis-ci.org/biocore/qiita ...

RNAer:
another job is certainly an option, but isn't it nicer to take advantage of
click's testing facility and to avoid writing boiler plate code ourselves?
To do that, I have to do imports. I am not sure alternative way for that
besides submodules...

antgonza:
Good point! However, I'm not sure what's the import issue (looking at http://click.pocoo.org/3/testing/), could you give an example? Thanks.

josenavas:
What if we put the tests in the same file than the functions are? We will not have the imports problem

RNAer:
The click doc examples defines interface inside test function. that's why
no imports is needed. My example is here:
https://github.com/RNAer/qiita/blob/clitest/scripts/test/test_qiita.py

antgonza:
Got it, thanks! What about following @josenavas suggestion? Have tests within scripts?

wasade:
In the long since forgotten rewrite of split_libraries.py, what we did was
had the command line logic part of an API module. This worked extremely
well and easily testable. The script would just import the relevant
methods. The test harness would import the API methods cleanly since
they're already importable. This has the added advantage that the scripts
code becomes very tiny.

wasade:
Relevant links below:

[qiime](https://github.com/wasade/qiime/blob/wf_qual_filter_fasta/scripts/qiime), this is the actual "script"
[cli](https://github.com/wasade/qiime/blob/wf_qual_filter_fasta/qiime/cli.py)
[click_commands](https://github.com/wasade/qiime/blob/wf_qual_filter_fasta/qiime/click_commands.py)
[...]

PR #853, opened 2015-02-13
ignoring qiita_core/environment_manager.py

antgonza:


josenavas:
It was a stochastic failure, I restarted the build

.coveragerc
```diff
@@ -4,7 +4,12 @@
 omit =
     */test*
     */__init__.py
-source = qiita_core,qiita_ware,qiita_db,qiita_pet 
+
+    # these files will be ommited from coveralls because they contain test
+    # of the environment and methods they are executed regularly by travis,
+    # then there should not be concern over their "correctness".
+    qiita_core/environment_manager.py
+source = qiita_core,qiita_ware,qiita_db,qiita_pet
```
ElDeveloper: Hmmm why is qiita_pet here if we have tests for that submodule?

antgonza: I do not think that's part of the omit but the source ...

squirrelo: Is there a way to test the click commands?

ElDeveloper: My bad ...

squirrelo: Also @ElDeveloper I believe source works as "Only look at these modules for test coverage counts"

ElDeveloper: There is, see also #120

antgonza: Baby steps, guys, baby steps. :baby:

squirrelo: Is more for whether we want to add `scripts` to source or not. If it's in there now with this pull request, it will lower the score and we'll remember to add those tests.

ElDeveloper: Yeah, that makes sense, or keeping #120 open and in this milestone. If
we can't add tests for the CLIs, it might also be worth to just run all
the scripts with the --help option and make sure none of them error
in the travis build, I think that would be a *baby step* :P

josenavas:
>Coverage remained the same at 78.5% when pulling 263f00e on antgonza:improve_cover_2 into 559556e on biocore:master.

Suspicious...
And checking on the travis build...
```
qiita_core.environment_manager                                   97     82     20     20    13%   14-18, 23-28, 33-36, 41-47, 52-57, 62-69, 74-96, 101-110, 122-162
```

It was not ignored!

.coveragerc
```diff
@@ -4,7 +4,12 @@
 omit =
     */test*
     */__init__.py
```
squirrelo: Also need to omit the qiita_pet/templates and qiita_pet/uimodules directories, since apparently they are also getting counted towards the test coverage even though they are HTML.

josenavas: the uimodules are not html, they're python code

squirrelo: Oh, right. But are they testable in any way?

josenavas: They should be testable through the handlers. Actually, the coverage of those files are >85%!

antgonza:
Closing this one as the prefered was https://github.com/biocore/qiita/pull/852

PR #856, opened 2015-02-15
fix #828

### LINKED ISSUE #828: password reset does not work
I forgot my QIITA password and so I hit the "Forgot password" link. I got a link in my e-mail to reset the password, but when I go to that link, there is no option to reset the password.

antgonza:


josenavas:
:+1:

qiita_pet/templates/change_lost_pass.html
```diff
@@ -8,7 +8,7 @@
 {%end%}
 
 {% block content %}
-  {% if not message %}
+  {% if display_body %}
```
squirrelo: This can be done much cleaner as 
```
if maintenance is None
```
and also would not require a new variable to be created.

squirrelo: This applies to all uses of `display_body`

antgonza: Will this work? Remember that the issue is that this templates used to rely on the presence/absence of a message so when a general message was present it didn't work. Not sure what has to do with just maintenance, could you explain?   Also, not sure on the benefit of None vs. True/False, could you elaborate?

squirrelo: If you look at sitebase, there are now 2 redis keys - maintenance and message. Maintenance mode disables login, while message jost holds the system message. Therefore you check for maintenance mode now instead of message.

squirrelo:
Can you also add qiita_pet tests to make sure something like this doesn't happen again?

antgonza:
Not sure how to do that, elaborate?

qiita_pet/templates/change_lost_pass.html
antgonza: I do not think we want to stop ant of these from working while in
maintenance, why should we?

squirrelo: Maintenance mode disable login to allow us to change things on teh site, so why would we want people changing passwords and doing other user db changes when the site is in maintenance mode?

squirrelo: Also the whole point of checking message here was that originally setting message was what put the site into maintenance mode.

antgonza: Not really, the idea of checking the message was to know if an error
occurred and do not display the rest of the page, please look at the
code. The current maintenance mode do not allow for code changes or
restarting the DB so we always need to stop and start the service,
thus maintenance mode is not actually doing what you think it does. If
you want to chat, just let me know.

What about the test?

squirrelo: Right, was getting sysmessage and message variables confused. Point still stands though: we do need to disable this in maintenance mode, as we don't want people trying to make database changes while in maintenance mode.

Could we also use `level == success` in this, since that implies it successfully did something? I'm trying to have as few new variables added to the templates as possible since it makes it more complicated to edit them down the line.

As for tests use the tornado testing base I made and set the redis key/send erroneous post data to make sure the pages render properly. Take a look at the other tests for how to do this.

antgonza: If you know which tests, will not be easier for you to point me to
them vs. telling me to look at all of them? Please be constructive.

Note that minimizing the number of variables makes sense, will fix,
but these templates have almost not variables so complexity is low.

squirrelo: I'm at dinner and don't know the specific tests off the top of my head. I just wanted to make sure that this was brought up before the request was merged by two other devs.

antgonza: OK will wait for your pointers.

squirrelo:
`test_post_select_samples` test function in `test_analysis_samples.py` is a good example of how to format the post dictionary, make the post call, and check response.body for specfic strings. response.body holds the entirety of the HTML generated for the page, and thus can't be sanely checked against a string without pages of text in the test, so searching through the text and making sure specific bits are there suffices.

antgonza:
Thanks for the explanation. How do you change the maintenance and
level values within the test framework so I can test the error?

squirrelo:
Import the redis object and use redis.set(KEY, VALUE).

squirrelo:
:+1:

PR #859, opened 2015-02-17
Adding easy preprocessing parameters

antgonza:
A collaborator under a deadline (thus, quick reviews will be greatly appreciated) asked me to allow for other preprocessing parameters (change the size of the barcode) but realized that doing this was gonna be a pain as we needed to change both the GUI and the database.

This PR allows to single add a new preprocessing in the DB and the GUI will recognize it. Currently only works for preprocess data but after chatting with @josenavas I realized that we can do the same for processing, which I will do after seeing what people think.

### DIFF
--- a/qiita_db/parameters.py
+++ b/qiita_db/parameters.py
@@ -70,18 +70,24 @@ class BaseParameters(QiitaObject):
 
     @classmethod
     def iter(cls):
-        """Iterates over all parameters
+        r"""Iterates over available parameters
 
         Returns
         -------
         generator
-            Yields a parameter instance
+            Yields a dict for each parameter set with each parameter as
+            column_name: value
         """
         conn_handler = SQLConnectionHandler()
-        sql = "SELECT {0} FROM qiita.{1}".format(cls._column_id, cls._table)
 
+        columns = flatten(conn_handler.execute_fetchall(
+            "SELECT column_name FROM information_schema.columns WHERE "
+            "table_name='{0}' and table_schema='qiita'".format(cls._table)))
+
+        columns_str = ', '.join(columns)
+        sql = "SELECT {0} FROM qiita.{1}".format(columns_str, cls._table)
         for result in conn_handler.execute_fetchall(sql):
-            yield cls(result[0])
+            yield dict(zip(columns, result))
 
     @property
     def name(self):
@@ -173,27 +179,6 @@ class BaseParameters(QiitaObject):
                 result.append("--%s %s" % (p_name, values[p_name]))
 
         return " ".join(result)
-
-    @classmethod
-    def iter(cls):
-        r"""Iterates over available parameters
-
-        Returns
-        -------
-        generator
-            Yields a dict for each parameter set with each parameter as
-            column_name: value
-        """
-        conn_handler = SQLConnectionHandler()
-
-        columns = flatten(conn_handler.execute_fetchall(
-            "SELECT column_name FROM information_schema.columns WHERE "
-            "table_name='{0}' and table_schema='qiita'".format(cls._table)))
-
-        columns_str = ', '.join(columns)
-        sql = "SELECT {0} FROM qiita.{1}".format(columns_str, cls._table)
-        for result in conn_handler.execute_fetchall(sql):
-            yield dict(zip(columns, result))
 
 
 class PreprocessedIlluminaParams(BaseParameters):
--- a/qiita_db/parameters.py
+++ b/qiita_db/parameters.py
@@ -7,6 +7,8 @@
 # -----------------------------------------------------------------------------
 from __future__ import division
 
+from skbio.util import flatten
+
 from .base import QiitaObject
 from .sql_connection import SQLConnectionHandler
 from .util import get_table_cols_w_type, get_table_cols
@@ -173,25 +175,25 @@ class BaseParameters(QiitaObject):
         return " ".join(result)
 
     @classmethod
-    def get_all_parameters(self):
-        r""" Generates a dictionary of all the available parameters
+    def iter(cls):
+        r"""Iterates over available parameters
 
         Returns
         -------
-        dict
-            A dictionary of column_name: value
+        generator
+            Yields a dict for each parameter set with each parameter as
+            column_name: value
         """
         conn_handler = SQLConnectionHandler()
 
-        columns = [r[0] for r in conn_handler.execute_fetchall(
+        columns = flatten(conn_handler.execute_fetchall(
             "SELECT column_name FROM information_schema.columns WHERE "
-            "table_name='{0}'".format(self._table))]
+            "table_name='{0}' and table_schema='qiita'".format(cls._table)))
 
         columns_str = ', '.join(columns)
-        result = [dict(zip(columns, v)) for v in conn_handler.execute_fetchall(
-            "SELECT {0} FROM qiita.{1}".format(columns_str, self._table))]
-
-        return result
+        sql = "SELECT {0} FROM qiita.{1}".format(columns_str, cls._table)
+        for result in conn_handler.execute_fetchall(sql):
+            yield dict(zip(columns, result))
 
 
 class PreprocessedIlluminaParams(BaseParameters):
--- a/qiita_db/parameters.py
+++ b/qiita_db/parameters.py
@@ -172,6 +172,27 @@ class BaseParameters(QiitaObject):
 
         return " ".join(result)
 
+    @classmethod
+    def get_all_parameters(self):
+        r""" Generates a dictionary of all the available parameters
+
+        Returns
+        -------
+        dict
+            A dictionary of column_name: value
+        """
+        conn_handler = SQLConnectionHandler()
+
+        columns = [r[0] for r in conn_handler.execute_fetchall(
+            "SELECT column_name FROM information_schema.columns WHERE "
+            "table_name='{0}'".format(self._table))]
+
+        columns_str = ', '.join(columns)
+        result = [dict(zip(columns, v)) for v in conn_handler.execute_fetchall(
+            "SELECT {0} FROM qiita.{1}".format(columns_str, self._table))]
+
+        return result
+
 
 class PreprocessedIlluminaParams(BaseParameters):
     r"""Gives access to the preprocessed parameters of illumina data"""
--- a/qiita_pet/handlers/preprocessing_handlers.py
+++ b/qiita_pet/handlers/preprocessing_handlers.py
@@ -7,8 +7,6 @@ from qiita_db.parameters import (PreprocessedIlluminaParams,
                                  Preprocessed454Params)
 from qiita_db.metadata_template import PrepTemplate
 from qiita_ware.context import submit
-from qiita_pet.uimodules.raw_data_tab import (PreprocessIlluminaParametersForm,
-                                              Preprocess454ParametersForm)
 
 
 class PreprocessHandler(BaseHandler):
@@ -17,46 +15,13 @@ class PreprocessHandler(BaseHandler):
         study_id = int(self.get_argument('study_id'))
         prep_template_id = int(self.get_argument('prep_template_id'))
         raw_data = RawData(PrepTemplate(prep_template_id).raw_data)
+        param_id = int(self.get_argument('preprocessing_parameters_id'))
 
         # Get the preprocessing parameters
         if raw_data.filetype == 'FASTQ':
-            form_data = PreprocessIlluminaParametersForm()
-            form_data.process(data=self.request.arguments)
-            rcomp_mapping_bcs = form_data.data['rev_comp_mapping_barcodes']
-            barcode_type = form_data.data['barcode_type'].strip("[']")
-            # currently only allow the user to change a single parameter of
-            # split libraries: --rev_comp_mapping_barcodes. The parameter ids 1
-            # and 2 contain the same set of values except for that flag. If
-            # param_id = 1, the flag is not activated, while in param_id = 2;
-            # it is.
-            if rcomp_mapping_bcs:
-                # Choose the parameter set with the --rev_comp_mapping_barcodes
-                # flag activated
-                if barcode_type == 'golay_12':
-                    param_id = 2
-                elif barcode_type == '8':
-                    param_id = 4
-                else:
-                    param_id = 6
-            else:
-                # Choose the parameter set with the --rev_comp_mapping_barcodes
-                # flag not activated
-                if barcode_type == 'golay_12':
-                    param_id = 1
-                elif barcode_type == '8':
-                    param_id = 3
-                else:
-                    param_id = 5
             param_constructor = PreprocessedIlluminaParams
         elif raw_data.filetype in ('FASTA', 'SFF'):
-            form_data = Preprocess454ParametersForm()
-            form_data.process(data=self.request.arguments)
             param_constructor = Preprocessed454Params
-            barcode_type = form_data.data['barcode_type'].strip("[']")
-            if barcode_type == 'golay_12':
-                param_id = 1
-            else:
-                param_id = 2
         else:
             raise ValueError('Unknown filetype')
 
--- a/qiita_pet/templates/study_description.html
+++ b/qiita_pet/templates/study_description.html
@@ -383,10 +383,11 @@ function unlink_all_files(raw_data_id) {
 }
 
 function submit_preprocess() {
-  if (confirm("Are you sure you want to submit a preprocessing job?")){
-    return true;
+  if ($("#preprocessing_parameters").children(":selected").attr("value") == 0) {
+    $("#display-extra-info").html('You need to select a valid preprocessing paramter to continue');
+    return false;
   }
-  return  false;
+  return true;
 };
 
 function submit_process() {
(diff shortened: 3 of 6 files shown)


qiita_db/parameters.py
```diff
@@ -172,6 +172,27 @@ def to_str(self):
...
+
+        Returns
+        -------
+        dict
+            A dictionary of column_name: value
+        """
+        conn_handler = SQLConnectionHandler()
+
+        columns = [r[0] for r in conn_handler.execute_fetchall(
+            "SELECT column_name FROM information_schema.columns WHERE "
```
wasade: This should include a restriction to `table_schema='qiita'`

qiita_db/parameters.py
```diff
@@ -172,6 +172,27 @@ def to_str(self):
...
+        r""" Generates a dictionary of all the available parameters
+
+        Returns
+        -------
+        dict
+            A dictionary of column_name: value
+        """
+        conn_handler = SQLConnectionHandler()
+
+        columns = [r[0] for r in conn_handler.execute_fetchall(
```
ElDeveloper: Should probably use `skbio.util.flatten`.

qiita_db/parameters.py
wasade: This query is not generally valid as qiime is inconsistent in its use of dashes and underscores in its parameters. Which sucks.

qiita_db/parameters.py
```diff
@@ -172,6 +172,27 @@ def to_str(self):
 
         return " ".join(result)
 
+    @classmethod
+    def get_all_parameters(self):
```
ElDeveloper: NABI: This name seems way too general, but I don't have suggestions to improve it.

------
NABI = not a blocking issue.

qiita_db/parameters.py
```diff
@@ -172,6 +172,27 @@ def to_str(self):
...
+        r""" Generates a dictionary of all the available parameters
```
ElDeveloper: Can you remove the extra space at the beginning of the docstring.

qiita_pet/uimodules/raw_data_tab.py
```diff
@@ -212,14 +175,26 @@ def render(self, prep, study_id, is_editable, ena_terms,
...
+            all_params = Preprocessed454Params.get_all_parameters()
         elif raw_data.filetype == 'FASTQ':
-            preprocess_form = PreprocessIlluminaParametersForm()
+            all_params = PreprocessIlluminaParametersForm.get_all_parameters()
         else:
             raise ValueError("Don't know what to do but this exception will "
                              "never actually get shown anywhere because why "
                              "would you want to see tracebacks?")
 
+        preprocess_options = []
```
wasade: Why replace WTForms?

qiita_db/parameters.py
```diff
@@ -172,6 +172,27 @@ def to_str(self):
...
+            A dictionary of column_name: value
+        """
+        conn_handler = SQLConnectionHandler()
+
+        columns = [r[0] for r in conn_handler.execute_fetchall(
+            "SELECT column_name FROM information_schema.columns WHERE "
+            "table_name='{0}'".format(self._table))]
+
+        columns_str = ', '.join(columns)
+        result = [dict(zip(columns, v)) for v in conn_handler.execute_fetchall(
```
ElDeveloper: Could this be turned into a dict comprehension statement?

qiita_pet/handlers/preprocessing_handlers.py
```diff
@@ -17,46 +15,13 @@ def post(self):
         study_id = int(self.get_argument('study_id'))
         prep_template_id = int(self.get_argument('prep_template_id'))
         raw_data = RawData(PrepTemplate(prep_template_id).raw_data)
+        param_id = int(self.get_argument('preprocessing_parameters_id'))
 
         # Get the preprocessing parameters
         if raw_data.filetype == 'FASTQ':
-            form_data = PreprocessIlluminaParametersForm()
```
wasade: I'm glad to see much of this hard coding removed, but am confused about the removal of WTForms?

wasade:
Few small comments but the big one is why remove WTForms here? Not opposed, just wondering what the motivation was

qiita_pet/templates/study_description.html
```diff
@@ -383,10 +383,11 @@
 }
 
 function submit_preprocess() {
-  if (confirm("Are you sure you want to submit a preprocessing job?")){
-    return true;
+  if ($("#preprocessing_parameters").children(":selected").attr("value") == 0) {
```
ElDeveloper: Use triple equals?

qiita_pet/templates/study_description.html
```diff
@@ -383,10 +383,11 @@
...
+    $("#display-extra-info").html('You need to select a valid preprocessing paramter to continue');
```
ElDeveloper: paramter -> parameter

qiita_pet/uimodules/raw_data_tab.py
```diff
@@ -212,14 +175,26 @@ def render(self, prep, study_id, is_editable, ena_terms,
...
+        for param in all_params:
+            text = []
+            for k, v in viewitems(param):
+                if k in ('preprocessed_params_id', 'param_set_name'):
```
ElDeveloper: Minor in this case since it is only two elements and the performance won't be terribly affected, but this tuple should be a set.

antgonza:
The reason we wanted to have WTForms was so we could validate the user inputs, which is not the case anymore. Additionally, to be able to make the GUI update a text field when you select paramA (now you get the full list of settings), would have taken a lot of work. Honestly, tried for an hour or 2 and realized that it was way too complex.

qiita_pet/uimodules/raw_data_tab.py
wasade: For that size, a `set` would more likely increase overhead

ElDeveloper:
Just did a quick review of the code, I'll deploy in my system when the build passes.

qiita_pet/uimodules/raw_data_tab.py
ElDeveloper: God dammit :L

qiita_db/parameters.py
adamrp: In other places (e.g., for the [user object](https://github.com/biocore/qiita/blob/master/qiita_db/user.py#L90-L104) or the [study person object](https://github.com/biocore/qiita/blob/master/qiita_db/study.py#L799-L815)) we have an `iter` method that accomplishes a similar goal, I think; perhaps follow that model?

qiita_db/parameters.py
adamrp: @wasade, I standardized on underscores for option longnames during the flagstaff sprint. See biocore/qiime#1792

qiita_db/parameters.py
adamrp: :+1:

antgonza:
Note that it failed due to flake8. Will address all comments (thanks!) and will add to processing soon.

adamrp:
Finished review -- code looks good, although I'm curious about @wasade's WTForms question, too.

qiita_db/parameters.py
wasade: is qiita using qiime 1.9 yet though?

antgonza: Yup, latest version repo.

qiita_db/parameters.py
antgonza: gonna change to iter()

qiita_db/parameters.py
wasade: A dev version, not a release? Weren't most of the OTUs picked using qiime 1.8?

antgonza: 1.8.0-dev so really 1.9.0.

wasade: It kinda depends on which commit... for instance, the dash/underscore issue
that @adamrp referred to was not in place when I put in the fastq demux
processing which was after the Flagstaff sprint. If qiime on
qiita.colorado.edu has been updated since, then it is possible there is an
inconsistency with the params table, that, or I may have omitted some of
the ambig options at the time

qiita_db/parameters.py
antgonza: Done

qiita_db/parameters.py
josenavas: We should update all our code to depend on the current release version of
QIIME as it should not be any further release on that 1.X branch.

2015-02-17 12:15 GMT-08:00 Daniel McDonald <notifications@github.com>:

> In qiita_db/parameters.py
> <https://github.com/biocore/qiita/pull/859#discussion_r24849687>:
>
> > @@ -172,6 +172,27 @@ def to_str(self):
> >
> >          return " ".join(result)
> >
> > +    @classmethod
> > +    def get_all_parameters(self):
> > +        r""" Generates a dictionary of all the available parameters
> > +
> > +        Returns
> > +        -------
> > +        dict
> > +            A dictionary of column_name: value
> > +        """
> > +
> > +        columns = [r[0] for r in conn_handler.execute_fetchall(
> > +            "SELECT column_name FROM information_schema.columns WHERE "
>
> It kinda depends on which commit... for instance, the dash/underscore
> issue that @adamrp <https://github.com/adamrp> referred to was not in
> place when I put in the fastq demux processing which was after the
> Flagstaff sprint. If qiime on qiita.colorado.edu has been updated since,
> then it is possible there is an inconsistency with the params table, that,
> or I may have omitted some of the ambig options at the time
> … <#14b992e6863b8a72_>
> On Tue, Feb 17, 2015 at 1:11 PM, Antonio Gonzalez <
> notifications@github.com> wrote: In qiita_db/parameters.py <
> https://github.com/biocore/qiita/pull/859#discussion_r24849395>: > @@
> -172,6 +172,27 @@ def to_str(self): > > return " ".join(result) > > +
> @classmethod <https://github.com/classmethod> > + def
> get_all_parameters(self): > + r""" Generates a dictionary of all the
> available parameters > + > + Returns > + ------- > + dict > + A dictionary
> of column_name: value > + """ > + conn_handler = SQLConnectionHandler() > +
> > + columns = [r[0] for r in conn_handler.execute_fetchall( > + "SELECT
> column_name FROM information_schema.columns WHERE " 1.8.0-dev so really
> 1.9.0. — Reply to this email directly or view it on GitHub <
> https://github.com/biocore/qiita/pull/859/files#r24849395>.
>
> —
> <https://github.com/biocore/qiita/pull/859/files#r24849687>.
>



-- 
Jose Navas

qiita_db/parameters.py
antgonza: Done

qiita_db/parameters.py
antgonza: Added restriction.

wasade: Agree. But also, updating QIIME needs to trigger a full reprocessing of the
database, or at the minimum, samples need to be tagged by what version of
qiime they were processed with. Is that in place?

qiita_db/parameters.py
antgonza: Changed it to:
```
        sql = "SELECT {0} FROM qiita.{1}".format(columns_str, cls._table)
        for result in conn_handler.execute_fetchall(sql):
            yield dict(zip(columns, result))
```

qiita_db/parameters.py
josenavas: Why is this? I don't think it depends right now...

2015-02-17 12:20 GMT-08:00 Daniel McDonald <notifications@github.com>:

> In qiita_db/parameters.py
> <https://github.com/biocore/qiita/pull/859#discussion_r24850101>:
>
> > @@ -172,6 +172,27 @@ def to_str(self):
> >
> >          return " ".join(result)
> >
> > +    @classmethod
> > +    def get_all_parameters(self):
> > +        r""" Generates a dictionary of all the available parameters
> > +
> > +        Returns
> > +        -------
> > +        dict
> > +            A dictionary of column_name: value
> > +        """
> > +
> > +        columns = [r[0] for r in conn_handler.execute_fetchall(
> > +            "SELECT column_name FROM information_schema.columns WHERE "
>
> Agree. But also, updating QIIME needs to trigger a full reprocessing of
> the database, or at the minimum, samples need to be tagged by what version
> of qiime they were processed with. Is that in place?
> … <#14b99332aea67944_>
> On Tue, Feb 17, 2015 at 1:18 PM, josenavas <notifications@github.com>
> wrote: In qiita_db/parameters.py <
> https://github.com/biocore/qiita/pull/859#discussion_r24849937>: > @@
> -172,6 +172,27 @@ def to_str(self): > > return " ".join(result) > > +
> @classmethod <https://github.com/classmethod> > + def
> get_all_parameters(self): > + r""" Generates a dictionary of all the
> available parameters > + > + Returns > + ------- > + dict > + A dictionary
> of column_name: value > + """ > + conn_handler = SQLConnectionHandler() > +
> > + columns = [r[0] for r in conn_handler.execute_fetchall( > + "SELECT
> column_name FROM information_schema.columns WHERE " We should update all
> our code to depend on the current release version of QIIME as it should not
> be any further release on that 1.X branch. 2015-02-17 12:15 GMT-08:00
> Daniel McDonald <notifications@github.com>: … <#14b9931785fc8a8a_> In
> qiita_db/parameters.py <
> https://github.com/biocore/qiita/pull/859#discussion_r24849687>: > @@
> -172,6 +172,27 @@ def to_str(self): > > return " ".join(result) > > +
> @classmethod <https://github.com/classmethod> <
> https://github.com/classmethod> > + def get_all_parameters(self): > +
> r""" Generates a dictionary of all the available parameters > + > + Returns
> > + ------- > + dict > + A dictionary of column_name: value > + """ > +
> conn_handler = SQLConnectionHandler() > + > + columns = [r[0] for r in
> conn_handler.execute_fetchall( > + "SELECT column_name FROM
> information_schema.columns WHERE " It kinda depends on which commit... for
> instance, the dash/underscore issue that @adamrp
> <https://github.com/adamrp> <https://github.com/adamrp> <
> https://github.com/adamrp> referred to was not in place when I put in the
> fastq demux processing which was after the Flagstaff sprint. If qiime on
> qiita.colorado.edu has been updated since, then it is possible there is
> an inconsistency with the params table, that, or I may have omitted some of
> the ambig options at the time … <#14b992e6863b8a72_> On Tue, Feb 17, 2015
> at 1:11 PM, Antonio Gonzalez < notifications@github.com> wrote: In
> qiita_db/parameters.py <
> https://github.com/biocore/qiita/pull/859#discussion_r24849395>: > @@
> -172,6 +172,27 @@ def to_str(self): > > return " ".join(result) > > +
> @classmethod <https://github.com/classmethod> <
> https://github.com/classmethod> < https://github.com/classmethod> > + def
> get_all_parameters(self): > + r""" Generates a dictionary of all the
> available parameters > + > + Returns > + ------- > + dict > + A dictionary
> of column_name: value > + """ > + conn_handler = SQLConnectionHandler() > +
> > + columns = [r[0] for r in conn_handler.execute_fetchall( > + "SELECT
> column_name FROM information_schema.columns WHERE " 1.8.0-dev so really
> 1.9.0. — Reply to this email directly or view it on GitHub <
> https://github.com/biocore/qiita/pull/859/files#r24849395>. — Reply to
> this email directly or view it on GitHub <
> https://github.com/biocore/qiita/pull/859/files#r24849687>. -- Jose Navas
> — Reply to this email directly or view it on GitHub <
> https://github.com/biocore/qiita/pull/859/files#r24849937>.
>
> —
> <https://github.com/biocore/qiita/pull/859/files#r24850101>.
>



-- 
Jose Navas

antgonza: No but AFAIK all of them were processed with the same versions ...

wasade: @rob-knight, it is critical that we do not unintentionally compare OTUs between versions of QIIME, correct?

antgonza: I think what is important is to keep track of the otu picking method and
reference ...

antgonza: and I really hope this discussion is for qiita in general and not for this PR.

qiita_pet/templates/study_description.html
antgonza: Done

qiita_pet/templates/study_description.html
antgonza: Done

qiita_db/parameters.py
wasade: qiita in general, though I'm concerned right now that we don't know what specific hash of what version of qiime was used when picking the current OTUs in the database

wasade: @antgonza, method and reference are insufficient as the method version might change, defaults might change, and there are quite a few pieces of logic that sit between all that including things like representative sequence, alignment method (if doing open-ref), taxonomy assignment, etc

antgonza: again, does it matter? I think no as long as we know the sortmerna and gg
versions ...

ElDeveloper:
Can someone else deploy and review this PR in their system, I cannot get it to work. :confused:

antgonza:
I think my last push messed up something ... taking a look now.

antgonza:
After taking a closer look at the iter main class method I realized that I could use that one to achieve the same as the methods I created here. Thus, will close this PR and create a new one. Thanks for the help so far and sorry for the confusion.

PR #862, opened 2015-02-18
ENH: natural sorting, fixes #831

### LINKED ISSUE #831: Samples column should not be sorted lexicographically
In the study results table (such as http://qiita.colorado.edu/study/public/), the Samples column reports the number of samples in the study. It currently sorts lexicographically, but this is not very useful. It should sort numerically.

wasade:
I only have 1 study in test, and my demo database is hosed, so I can't reallllly verify this one.

### DIFF
--- a/qiita_pet/templates/private_studies.html
+++ b/qiita_pet/templates/private_studies.html
@@ -3,6 +3,7 @@
 <link rel="stylesheet" href="/static/vendor/css/jquery.dataTables.css" type="text/css">
 
 <script src="/static/vendor/js/jquery.dataTables.min.js"></script>
+<script src="/static/vendor/js/jquery.dataTables.plugin.natural.js"></script>
 
 <script type="text/javascript">
 var current_study;
--- a/qiita_pet/templates/private_studies.html
+++ b/qiita_pet/templates/private_studies.html
@@ -8,8 +8,12 @@
 var current_study;
 
 $(document).ready(function() {
-    $('#user-studies-table').dataTable({"order": [[ 1, "asc" ]]});
-    $('#shared-studies-table').dataTable({"order": [[ 1, "asc" ]]});
+        $('#user-studies-table').dataTable({
+            order: [[ 1, "asc" ]],
+            columnDefs: 'natural'});
+        $('#shared-studies-table').dataTable({
+            order: [[ 1, "asc" ]],
+            columnDefs: 'natural'});
     $('#waiting').hide();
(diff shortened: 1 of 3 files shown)


qiita_pet/templates/private_studies.html
```diff
@@ -3,13 +3,18 @@
...
 
 $(document).ready(function() {
-    $('#user-studies-table').dataTable({"order": [[ 1, "asc" ]]});
-    $('#shared-studies-table').dataTable({"order": [[ 1, "asc" ]]});
+        $('#user-studies-table').dataTable({
+            order: [[ 1, "asc" ]],
+            columnDefs: 'natural'});
+        $('#shared-studies-table').dataTable({
+            order: [[ 1, "asc" ]],
+            columnDefs: 'natural'});
```
adamrp: haven't tested this yet, but the code here differs slightly from the example in natural.js

I'll test shortly.

wasade: you're right... not sure what i was thinking. lemme know if it works.

adamrp: Nope, I don't think it works see screenshot below of studies sorted by number of samples ascending):

image

Note that the problem only presents when the column is a mixed datatype (integers and strings in the column, as is the case when there are some studies that have `num_samples_collected` set to a number, and some studies that do not have it set (which displays as `None`)

adamrp: Even when I change the code to reflect the example, it doesn't work. There's also not great documentation on this plugin, which doesn't help...

wasade: No... and I can't test locally as my demo database is hosed

adamrp: Can you drop your environment (`qiita_env drop`) and create a new test environment (`qiita_env make --no-load-ontologies` with TEST_ENVIRONMENT=TRUE in your qiita_config)? If you do that, then create two studies on the test@foo.bar user, then change `number_samples_collected` on one of the studies using psql to, e.g., 210, you should be able to replicate the test I'm running here

adamrp: Got it -- didn't mean to duplicate effort on this, but your being unable to test led me to tinkering :smile:

I'll submit a PR against your fork

adamrp:
@wasade, when I was trying to do a PR against your fork, it kept pulling in a bunch of commits, so I opened a new PR; I think this one can be closed, since the other one contains the commits from this PR

PR #863, opened 2015-02-18
Easy (pre)processing params

antgonza:
This supersedes https://github.com/biocore/qiita/pull/859 and also makes the pre and the processing look the same + got rid of confirmation JS boxes. 

As in the previous PR fast reviews will be appreciated

### DIFF
--- a/qiita_db/support_files/patches/0.sql
+++ b/qiita_db/support_files/patches/0.sql
@@ -59,8 +59,3 @@ SELECT setval('qiita.study_study_id_seq', 10000, false);
 -- The first row has the default values on QIIME
 -- The second row has the default values on QIIME but rev_comp_mapping_barcodes is set to true
 INSERT INTO qiita.preprocessed_sequence_illumina_params (rev_comp_mapping_barcodes) VALUES (false), (true);
-
--- Initializing preprocessed_sequence_illumina_params to have 2 rows
--- First to have golay_12
--- Second to have hamming_8
-INSERT INTO qiita.preprocessed_sequence_454_params (barcode_type) VALUES ('golay_12'), ('hamming_8');
--- a/qiita_db/support_files/patches/0.sql
+++ b/qiita_db/support_files/patches/0.sql
@@ -59,3 +59,8 @@ SELECT setval('qiita.study_study_id_seq', 10000, false);
 -- The first row has the default values on QIIME
 -- The second row has the default values on QIIME but rev_comp_mapping_barcodes is set to true
 INSERT INTO qiita.preprocessed_sequence_illumina_params (rev_comp_mapping_barcodes) VALUES (false), (true);
+
+-- Initializing preprocessed_sequence_illumina_params to have 2 rows
+-- First to have golay_12
+-- Second to have hamming_8
+INSERT INTO qiita.preprocessed_sequence_454_params (barcode_type) VALUES ('golay_12'), ('hamming_8');
--- a/qiita_pet/templates/study_description_templates/prep_template_panel.html
+++ b/qiita_pet/templates/study_description_templates/prep_template_panel.html
@@ -9,23 +9,30 @@
         <button type="button" class="close" data-dismiss="modal" aria-hidden="true">&times;</button>
         <h4 class="modal-title" id="myModalLabel">Choose preprocessing parameters</h4>
       </div>
-      <form action="/study/preprocess" method="post" onsubmit='return submit_preprocess();'>
+      <form action="/study/preprocess" method="post" onsubmit='return submit_preprocess({{prep_id}});'>
         <div class="modal-body">
           <div>
             <div class="form-group">
               <table>
-                {% for idx, form_item in enumerate(preprocess_form, start=1) %}
-                  <tr>
-                    <td width="50%">{% raw form_item.label %}</td>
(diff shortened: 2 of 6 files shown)


antgonza:
Forgot that the first commit of #859  was added to my master and thus here too, removing now.

antgonza:
Changing history is scary! Anyway, this is ready for final review.

josenavas:
Thanks @antgonza 

Just a comment. In the preprocessing modal, no parameters are selected by default, while in the processing one the "Default" parameter set is selected. I would say that it will simplify code and improve consistency if in both cases the "Default" parameter set is chosen (the function `submit_preprocess ` will be removed, and both modals will work exactly equal).

antgonza:
I'm not sure which one will be the default for preprocessing as there
are sooooo many options. Thus, the idea of having the user select one
and have no default.

ElDeveloper:
It is scary indeed, but someone has to do it!

josenavas:
I would say the default will be the one with golay_12 and not 'rev_comp_mapping_barcodes' (which is the one that has default as name). However I don't think this is blocking, so if other devs are happy as it is right now is :+1: for me to merge.

qiita_pet/templates/study_description_templates/prep_template_panel.html
```diff
@@ -9,24 +9,31 @@
...
-                  <tr>
-                    <td width="50%">{% raw form_item.label %}</td>
-                    <td width="50%">{% raw form_item(tabindex=idx, class_='form-control') %}</td>
-                  </tr>
-                {% end %}
                 <tr>
                   <td>
+                    <select id="preprocessing_parameters_{{prep_id}}" name="preprocessing_parameters_id" class="form-control" onchange="$('#display-extra-info-{{prep_id}}').html($(this).children(':selected').attr('extra-info'))">
+                      <option value="0" extra-info="">Select an option ...</option>
+                      {% for idx, name, text in preprocess_options %}
```
ElDeveloper: We are guaranteed to always have preprocessing options right? so this would never be a dropdown menu that's essentially empty?

ElDeveloper:
Looks good, only comment on the GUI is that when you get to the parameter selection screen, it is very unclear what are you actually selecting. It reads at the top `choose preprocessing parameters`, as a naive user this means nothing to me, even as an experienced QIIME user this means very little. Maybe it would be worth to add a short description (at the bottom of the modal) of what steps will take place and possibly link to the docs for `split_libraries_fastq.py` and `split_libraries.py`?

Aside from this, this PR is ready.

qiita_pet/templates/study_description_templates/prep_template_panel.html
antgonza: At least for illumina:
https://github.com/biocore/qiita/blob/master/qiita_db/support_files/patches/0.sql#L61

ElDeveloper: So if it's a different type, then we are not going to be able to
preprocess the data until we manually add parameters? Not part of this
PR but that's a real issue.

adamrp:
Code looks good to me, about to test it on my system.

adamrp:
:+1: when tests pass

qiita_pet/templates/study_description_templates/prep_template_panel.html
adamrp: @ElDeveloper IIRC, for both preprocessing and processing, the decision was made to supply users with a selection of predetermined parameter sets, which is what @antgonza has implemented here.

wasade: That's what I remember too

qiita_pet/uimodules/raw_data_tab.py
```diff
@@ -212,14 +174,22 @@ def render(self, prep, study_id, is_editable, ena_terms,
...
-            preprocess_form = PreprocessIlluminaParametersForm()
+            param_iter = PreprocessedIlluminaParams.iter()
         else:
             raise ValueError("Don't know what to do but this exception will "
                              "never actually get shown anywhere because why "
                              "would you want to see tracebacks?")
 
+        preprocess_options = []
+        for param in param_iter:
+            text = ["<b>%s:</b> %s" % (k, v)
```
wasade: Very minor, but this can be a generator expression. The benefit is that the code on 187-188 will do a traversal of `param.values`, and then the join on 191 will do a traversal of the resulting list. A generator expression here will reduce this to a single traversal of the data. 

```python
text = ("<b>%s:</b> %s" % (k, v) for k, v in viewitems(param.values))
```

qiita_db/support_files/patches/0.sql
```diff
@@ -59,3 +59,8 @@ SELECT setval('qiita.study_study_id_seq', 10000, false);
 -- The first row has the default values on QIIME
 -- The second row has the default values on QIIME but rev_comp_mapping_barcodes is set to true
 INSERT INTO qiita.preprocessed_sequence_illumina_params (rev_comp_mapping_barcodes) VALUES (false), (true);
+
```
wasade: I don't think we should be modifying existing patches, correct?

antgonza: Removed.

antgonza:
Ready again.

ElDeveloper:
:+1:

antgonza:
Can this be merged?

Issue #692, opened 2014-11-25
sample templates in search criteria


LucasMS:
It would be cool to have the search function also looking inside the sample templates. For instance, I found studies searching for the keyword "ocean" but none using the keywords "seawater" or "sea water".

Issue #283, opened 2014-09-12
Ensure unique IDs for metaanalysis


wasade:
...in the case of the same samples run on multiple sequencing platforms, the resulting IDs need to remain unique.

antgonza:
I think this is not resolved for meta-analysis. Let's imagine that we have 1 sample that was ran with 454 and Illumina named: 'sample1', you preprocess and process, then you will have 2 otu tables with study-id.sample1, then you want to do a meta-analysis of both. This will result in a merge of the 2 samples ...

josenavas:
Can we then change the title of the issue? It no longer applies to the sample/prep template upload (that is checked) it applies to when the OTU table for meta-analysis are created.

antgonza:
Done.

antgonza:
Closing this in favor of: https://github.com/biocore/qiita/issues/1236

PR #865, opened 2015-02-18
Select by datatype and processed data id - NO MERGE

squirrelo:
This pull request makes search and selection of samples happen by datatype and processed ID. Previously you would select samples by study, and they would be selected for all processed data IDs in the study.

To do this, I have added two tab lists to each modal, so you can select the datatype and process ID to list samples for. This list will only include samples available in that processed data, so we no longer have the issue of being able to select non-existant samples for certain data types.

*READY FOR REVIEW*


squirrelo:
Ready for review

qiita_db/data.py
```diff
@@ -1167,6 +1167,21 @@ def data_type(self, ret_id=False):
...
+        Returns
+        -------
+        set of str
+            Sample_ids available according to the prep template
+        """
+        # Get the samples available for the prep template
+        pid = PreprocessedData(self.preprocessed_data).prep_template
+        conn_handler = SQLConnectionHandler()
+        sql = "SELECT sample_id FROM qiita.prep_{0}".format(pid)
+        return {x[0] for x in conn_handler.execute_fetchall(sql)}
```
antgonza: I think this can be a generator, no?

qiita_pet/webserver.py
```diff
@@ -39,7 +39,8 @@
 STATIC_PATH = join(DIRNAME, "static")
 TEMPLATE_PATH = join(DIRNAME, "templates")  # base folder for webpages
 _, RES_PATH = get_mountpoint('job')[0]
-COOKIE_SECRET = b64encode(uuid4().bytes + uuid4().bytes)
+COOKIE_SECRET = 'TEST' if qiita_config.test_environment else \
```
antgonza: Why is this needed? Perhaps a comment about it will be useful

squirrelo: It makes it so if you're in a testing environment you don't have to constantly re-login if you change something. I can add a comment.

qiita_db/data.py
squirrelo: Could you explain the benefit of doing so instead of just returning a set? Note `execute_fetchall` is a list of lists so it's not going to save memory by being a generator.

antgonza: Yes, the main benefit is not having to transverse the list several
times. I was thinking on something like this:
https://github.com/biocore/qiita/blob/master/qiita_db/parameters.py#L70

qiita_db/data.py
```diff
@@ -1167,6 +1167,22 @@ def data_type(self, ret_id=False):
         return data_type[0]
 
     @property
+    def samples(self):
+        """Returns samples available in this processed data
+
+        Returns
+        -------
+        set of str
```
adamrp: This needs to be updated -- no longer a set. Also use the "generator" keyword (or say that it yields sample IDs or something to tip off the reader that this is not constructing the set and returning it)

qiita_db/search.py
```diff
@@ -176,6 +177,11 @@ def __call__(self, searchstr, user):
             Search string to use
         user : User object
             User making the search. Needed for permissions checks.
+        remove_selected : bool, optional
+            If search in on an analysis that already has selected samples,
+            remove the selected samples from the results. Default False.
```
adamrp: in on an -> is on an.  I don't really understand the utility of this option.

qiita_db/search.py
```diff
@@ -176,6 +177,11 @@ def __call__(self, searchstr, user):
...
+        analysis : int, optional
+            Analysis ID search is run for. Required of remove_selected is True
```
adamrp: of -> if

qiita_db/search.py
squirrelo: If you have an analysis with samples previously selected, this will remove those samples from the results returned. it keeps you from doing two searches and selecting the same sample twice because it showed up in both searches. This was originally filtered at the python level, but filtering it at the SQL level is faster.

adamrp: Ahh, I see now, thanks.

qiita_db/search.py
```diff
@@ -232,6 +240,11 @@ def _parse_study_search_string(self, searchstr,
             SQL query for each study to get the sample ids that mach the query
         meta_headers : list
             metadata categories in the query string in alphabetical order
+        remove_selected : bool, optional
+            If search in on an analysis that already has selected samples,
```
adamrp: in on an -> is on an

qiita_db/search.py
```diff
@@ -232,6 +240,11 @@ def _parse_study_search_string(self, searchstr,
...
+            remove the selected samples from the results. Default False.
+        analysis : int, optional
+            Analysis ID search is run for. Required of remove_selected is True
```
adamrp: of -> if

qiita_pet/templates/search_sample_list.html
```diff
@@ -0,0 +1,66 @@
...
+    </tr>
+    {% if meta_headers %}
+      {% for sample in proc_data_samples[pid]%}
+    <tr><td>
+      <input type='checkbox' name='samples{{sel}}'
+      value='{{pid}}#{{sample[0]}}' class='study{{sid}} proc{{pid}} {{ " ".join(clean_str(s) for s in sample) }} {{sel}}' onchange='count_update({{sid}})'>
+      <td><a href="#" onclick="select_category('{{clean_str(sample[0])}}', '')">All Datatypes</a></td>
+      </td><td>{{sample[0]}}</td>
+        {% for pos in range(len(meta_headers)) %}
+      <td>{{sample[pos+1]}}</td>
```
adamrp: why is the 0th index handled explicitly? The format looks to be the same as what you're looping over, right?

squirrelo: The 0th index is the sample ID, so it's ignored when printing the metadata.

qiita_pet/templates/search_sample_list.html
```diff
@@ -0,0 +1,66 @@
...
+      <th>Sample ID</th>  
+    {% for meta in meta_headers %}
+      <th>{{meta}}</th>
+    {% end %}
+    </tr>
+    {% if meta_headers %}
+      {% for sample in proc_data_samples[pid]%}
+    <tr><td>
+      <input type='checkbox' name='samples{{sel}}'
+      value='{{pid}}#{{sample[0]}}' class='study{{sid}} proc{{pid}} {{ " ".join(clean_str(s) for s in sample) }} {{sel}}' onchange='count_update({{sid}})'>
```
adamrp: This checkbox has to have the class of every sample? Isn't this potentially a lot of classes? There is apparently no hard limit (googling suggests that the limit is somewhere over 2000), but it does seem a little odd to have so many

qiita_pet/templates/search_sample_list.html
squirrelo: Oh, I see what you mean, yes. Originally there was data between the sample ID and the metadata, but now there isn't, so yeah that can all be one loop now.

qiita_pet/templates/search_studies.html
```diff
@@ -1,41 +1,93 @@
 {%extends sitebase.html%}
 {%block head%}
+{% from future.utils import viewitems, viewkeys %}
+{% from qiita_db.study import Study %}
+{% from qiita_db.data import ProcessedData %}
+{% from qiita_pet.util import clean_str %}
+<script type="text/javascript">
+  STUDIES = {{[str(i) for i in results]}};
```
adamrp: Doesn't this require square brackets and single quotes?

```javascript
STUDIES = ['{{"', '".join([str(i) for i in results])}}']
```

qiita_pet/util.py
```diff
@@ -52,7 +52,7 @@ def clean_str(item):
     str
         cleaned string
     """
-    return str(item).replace(" ", "_").replace(":", "")
+    return str(item).replace(" ", "_").replace(":", "").replace(".", "")
```
adamrp: Are we sure that modifying this function will not impact assumptions made in other places in the codebase? Also, can you update the docstring for this function to reflect the code that's currently implemented (including your addition).

qiita_ware/search.py
```diff
@@ -0,0 +1,103 @@
...
+        counts for each found metadata value over all studies, in the format
+        {meta_col1: {value1: count, value2: count, ...}, ...}
+    studycount : dict of dict of dicts
+        counts for each found metadata value for each study, in the format
+        {study_id: {meta_col1: {value1: count, value2: count, ...}, ...}, ...}
+    """
+    fullcount = {}
+    # zip all samples so that each metadata column found is its own list
+    meta_vals = zip(*[samples[sample] for samples in viewvalues(results)
+                    for sample in range(len(samples))])
```
adamrp: This is somewhat hard to understand, and in the past we have shied away from using nested for loops in list comprehensions. I would break this up, but it would be good to get others' opinions

ElDeveloper: :+1: to breaking this up.

josenavas: :+1: on breaking this up. It can potentially use generators so no overhead for object creation is incurred.

qiita_ware/search.py
```diff
@@ -0,0 +1,103 @@
...
+            # skip processed data if it doesn't fit the given datatypes
+            if datatypes is not None and datatype not in datatypes:
+                continue
+            proc_data_samples[proc_data_id] = []
+            samps_available = set(proc_data.samples)
+            for sample in results[study_id]:
+                # filter to samples available in this proc data
+                if sample[0] in samps_available:
+                    proc_data_samples[proc_data_id].append(sample)
+            if proc_data_samples[proc_data_id] is []:
```
adamrp: This does not work:

```python
>>> t = []
>>> l = []
>>> t is l
False
>>> t is []
False
>>> l is []
False
>>> [] is []
False
```

adamrp:
Finished going over the code. There is at least one bug (testing equivalence using `is`) that will need to be fixed before I test on my system.  Thanks!

qiita_ware/search.py
squirrelo: good catch, should be `==`

qiita_ware/search.py
squirrelo: OK, and breaking it up is actually slightly faster apparently.
```
def thing(results, meta_counts):
            cats = [[] for x in range(meta_counts+1)]
            for samples in viewvalues(results):
                for sample in range(meta_counts):
                    for pos, val in enumerate(samples[sample]):
                        cats[pos].append(val)
            return cats

timeit zip(*[samples[sample] for samples in viewvalues(results) for sample in range(len(samples))])
100 loops, best of 3: 3.89 ms per loop

In [51]: timeit thing(results, 5)
1000 loops, best of 3: 1.36 ms per loop
```

qiita_db/data.py
squirrelo: done

qiita_ware/search.py
adamrp: Awesome! Please call the function `thing` in the production code.  :laughing:

Also, change `cats` to `kittens`

qiita_pet/templates/search_sample_list.html
squirrelo: This has the cleaned sample ID and all metadata categories for a single sample. This means that there is one class entry for each metadata value for a sample plus the study, procid, and a few other overhead classes, so probably like 10 class entries per checkbox. It facilitates the select by metadata process.

qiita_pet/templates/search_studies.html
squirrelo: yup, my bad.

qiita_ware/search.py
josenavas: I think the pythonic way of doing this is:
```python
if not proc_data_samples[proc_data_id]:
    # do stuff
```

qiita_pet/util.py
squirrelo: Yup. I'm sure it doesn't change anything, as this function is for cleaning out string items that would screw up jquery use, so having a period in the text would definitely screw up a jquery check by class. I can update the docstring to reflect this.

qiita_ware/search.py
squirrelo: :+1: to kittens

squirrelo: Wait, there was a bug in that code. This is the real times with the bug fixed and run through a test to make sure:
```
timeit zip(*[samples[sample] for samples in viewvalues(results) for sample in range(len(samples))])
100 loops, best of 3: 3.51 ms per loop

timeit thing(results, [1,2,3,4,5])
10 loops, best of 3: 22.5 ms per loop
```

squirrelo:
Ready to go for more testing.

qiita_db/data.py
```diff
@@ -1167,6 +1167,22 @@ def data_type(self, ret_id=False):
...
+        Returns
+        -------
+        generator of str
+            Sample_ids available according to the prep template
+        """
+        # Get the samples available for the prep template
+        pid = PreprocessedData(self.preprocessed_data).prep_template
+        conn_handler = SQLConnectionHandler()
+        sql = "SELECT sample_id FROM qiita.prep_{0}".format(pid)
+        for x in conn_handler.execute_fetchall(sql):
```
wasade: Could also do as:

```python
return (x[0] for x in conn_handler.execute_fetchall(sql))
```

qiita_db/search.py
```diff
@@ -194,7 +200,8 @@ def __call__(self, searchstr, user):
         Metadata column names and string searches are case-sensitive
         """
         study_sql, sample_sql, meta_headers = \
-            self._parse_study_search_string(searchstr, True)
```
wasade: This method is already incredibly complex, and running on 150 lines... I think it should likely be decomposed prior to adding in more functionality.

The method name is `_parse_study_search_string`, but isn't the removal of samples from an analysis different than parsing a study search string?

qiita_pet/handlers/analysis_handlers.py
```diff
@@ -54,72 +54,30 @@ def check_analysis_access(user, analysis):
...
+            study_proc_data[proc_data.study][proc_data.data_type()].append(pid)
+        return study_proc_data, proc_data_samples
+
+    def _parse_deselect_checkbox_data(self):
+        """splits the list values into proc_data_id and sample_id, then builds
+           dict of results as {proc_data_id: [ samp1, samp2, ...], ...} """
+        data = self.get_arguments('samples-sel')
+        res = defaultdict(list)
+        for val in data:
+            hold = val.split('#')
```
wasade: Can this be a `split('#', 1)` please?

qiita_db/search.py
squirrelo: Yes, but this is the only way to build a working SQL that will auto-filter the samples. I can go back to doing it pythonically but it is slower.

qiita_db/data.py
squirrelo: That doesn't return a tuple? Cool.

qiita_pet/static/js/analysis.js
```diff
@@ -34,42 +34,28 @@ function select_category(category, study) {
...
+  else { $('#study' + study).removeClass('success'); }
 }
 
-function select_deselect(study, select) {
-  if(select == true) { 
-    $('.'+study).each(function() {this.checked = true;});
+function select_deselect(study, pdid, type, select) {
+  boxes = '.study'+study+'.proc'+pdid
+  if (type = '-sel') { boxes += '.-sel' }
+  if(select == true) {
```
wasade: `===`?

qiita_pet/static/js/analysis.js
```diff
@@ -34,42 +34,28 @@ function select_category(category, study) {
...
+    $(boxes).each(function() {this.checked = false;});
   }
   count_update(study);
 }
 
-function select_inverse(study) {
-  $('.'+study).each(function() {
+function select_inverse(study, pdid, type) {
+  boxes = '.study'+study+'.proc'+pdid
+  if (type = '-sel') { boxes += '.-sel' }
```
wasade: `=`?

qiita_pet/templates/search_sample_list.html
```diff
@@ -0,0 +1,65 @@
...
+      <th>{{meta}}</th>
+    {% end %}
+    </tr>
+    {% if meta_headers %}
+      {% for sample in proc_data_samples[pid]%}
+    <tr><td>
+      <input type='checkbox' name='samples{{sel}}'
+      value='{{pid}}#{{sample[0]}}' class='study{{sid}} proc{{pid}} {{ " ".join(clean_str(s) for s in sample) }} {{sel}}' onchange='count_update({{sid}})'></td>
+      <td><a href="#" onclick="select_category('{{clean_str(sample[0])}}', '')">All Datatypes</a></td>
+        {% for val in sample %}
```
wasade: Is this a quad nested for loop?

squirrelo: Yes! Need to show every study, then every proc id for the study, then each sample under that proc id, then the metadata values for that study.

qiita_pet/test/test_analysis_handlers.py
```diff
@@ -75,16 +75,22 @@ def test_post_deselect_samples(self):
...
-            'dt1': '1',
-            'sel1': 'SKB8.640193'}
+            'samples-sel': '1#1.SKB8.640193'}
 
+        # first make sure sample is there to begin with
+        response = self.get('/analysis/2?aid=1')
+        # Make sure page response loaded sucessfully
+        self.assertEqual(response.code, 200)
+        # make sure sample removed
+        self.assertFalse("1.SKB8.640193" not in str(response.body))
```
wasade: `self.assertNotIn`

qiita_pet/test/test_analysis_handlers.py
```diff
@@ -75,16 +75,22 @@ def test_post_deselect_samples(self):
...
+
+        # now test deselect posting
         response = self.post('/analysis/2', post_args)
 
         # Make sure page response loaded sucessfully
         self.assertEqual(response.code, 200)
         # make sure sample removed
-        self.assertTrue("SKB8.640193" not in str(response.body))
+        self.assertTrue("1.SKB8.640193" not in str(response.body))
```
wasade: Same

qiita_pet/uimodules/search_sample_list.py
```diff
@@ -0,0 +1,16 @@
+from .base_uimodule import BaseUIModule
+
+
+class SearchSampleList(BaseUIModule):
+    def render(self, study_id, meta_headers,
+               proc_data_dict, proc_data_samples):
+        if not meta_headers:
```
wasade: `selected = not meta_headers`

qiita_pet/util.py
```diff
@@ -51,8 +51,13 @@ def clean_str(item):
...
     str
         cleaned string
+
+    Notes
+    -----
+    This function removes colons and periods from strings, and replaces spaces
+    with underscores.
     """
-    return str(item).replace(" ", "_").replace(":", "")
+    return str(item).replace(" ", "_").replace(":", "").replace(".", "")
```
wasade: If `clean_str` is used regularly, then a more efficient replace should be done. Either a regex or translation table

qiita_ware/search.py
```diff
@@ -0,0 +1,107 @@
...
+    Returns
+    -------
+    fullcount : dict of dicts
+        counts for each found metadata value over all studies, in the format
+        {meta_col1: {value1: count, value2: count, ...}, ...}
+    studycount : dict of dict of dicts
+        counts for each found metadata value for each study, in the format
+        {study_id: {meta_col1: {value1: count, value2: count, ...}, ...}, ...}
+    """
+    def double_comprehension(results):
```
wasade: This can be replaced with `itertools.product`

wasade: wait, scratch that.

qiita_ware/search.py
```diff
@@ -0,0 +1,107 @@
...
+        study_proc_ids[study_id] = defaultdict(list)
+        for proc_data_id in study.processed_data():
+            proc_data = ProcessedData(proc_data_id)
+            datatype = proc_data.data_type()
+            # skip processed data if it doesn't fit the given datatypes
+            if datatypes is not None and datatype not in datatypes:
+                continue
+            proc_data_samples[proc_data_id] = []
+            samps_available = set(proc_data.samples)
+            for sample in results[study_id]:
```
wasade: `proc_data_samples[proc_data_id] = [s for s in results[study_id] if s[0] in samps_available]`

wasade: this would let you remove the if/else that follows, as well as the explicit creation of the list on line 88

qiita_db/search.py
wasade: Decomposition doesn't necessarily mean doing this in python, and python is not necessarily slower. What I'm recommending is to break the function into smaller blocks as it will be much easier to verify that the individual pieces function as expected. It is very difficult to verify that a large function with complex branching logic is correct.

qiita_db/data.py
wasade: Generator expression, awesome, eh?

qiita_pet/templates/search_sample_list.html
```diff
@@ -0,0 +1,65 @@
...
+      <th>Sample ID</th>  
+    {% for meta in meta_headers %}
+      <th>{{meta}}</th>
+    {% end %}
+    </tr>
+    {% if meta_headers %}
+      {% for sample in proc_data_samples[pid]%}
+    <tr><td>
+      <input type='checkbox' name='samples{{sel}}'
+      value='{{pid}}#{{sample[0]}}' class='study{{sid}} proc{{pid}} {{ " ".join(clean_str(s) for s in sample) }} {{sel}}' onchange='count_update({{sid}})'></td>
```
wasade: ah, `clean_str` is used _a lot_. This probably should be done more efficiently than multiple replaces.

qiita_pet/templates/search_sample_list.html
wasade: What type of load does this put onto the server, and does this scale? This quickly gets to be very scary in big-O

squirrelo: It's actually not that expensive since it's just putting values in a template, and those are compiled before use. I'm not entirely sure about scale though, as i don't have a large enough database to test it on.

wasade: Compiled server side each render

qiita_pet/static/js/analysis.js
squirrelo: yeah, typo. That explains some odd behavior.

qiita_pet/templates/search_sample_list.html
squirrelo: Well, it's an extra layer on what we already have (triple loop) so with some of the optimizations done in the code for the search, it may actually be as expensive or less so than the current setup.

qiita_pet/templates/search_sample_list.html
squirrelo: now one replace for spaces and a compiled regex for removing all others.

qiita_pet/handlers/analysis_handlers.py
squirrelo: done

qiita_pet/static/js/analysis.js
squirrelo: fixed. also linted so more JS fixed.

qiita_pet/test/test_analysis_handlers.py
squirrelo: That's a new one for me, cool. Fixed.

qiita_pet/util.py
squirrelo: replaced with a compiled regex.

qiita_ware/search.py
squirrelo: I believe that end if/else is still needed since this will still end up as an empty list if nothing matches samps_available, correct?

qiita_ware/search.py
squirrelo: Scratched. Also, why is there no record emoji? Sad.

qiita_pet/uimodules/search_sample_list.py
squirrelo: done

qiita_pet/templates/search_sample_list.html
wasade: Let's run through with a small example, say, 10 studies, each with 2 processed IDs, each with 100 samples, and each with 20 metadata values. This would result in 40,000 executions of the inner loop, which is building a large HTML table (without knowing the internals of the templating engine, this may be very expensive to do). In addition to the HTML table, the 2nd nesting will end up doing 44,000 regex calls (lines 45, 57 and 58). In other words, this quickly gets very expensive especially if we have concurrent users. 

Let's further imagine that the compilation of this page takes 100ms. This is a reasonable estimate given that doing a (naive) regex 44,000 times takes ~50ms:

```python
> %%timeit
for i in range(44000):
  re.search('foo', 'asdasdasdasdasdasda')
   ....:
10 loops, best of 3: 53.5 ms per loop
```

What that means is that, with 10 concurrent users doing similar small searches, the server is blocked for a full second. This will greatly impact the user experience. There is also the potential for a single user to form a search that grinds the site to a full halt for an extended period. An expense here limits our ability to scale to concurrent users among other variables. 

For something like search, we very likely want to cache and index (external to postgres) aggressively. Given that we're approaching alpha, I think we probably want to hold off on a caching exploration, but it should be something we discuss soon.

qiita_ware/search.py
wasade: hah, :clap:

qiita_db/search.py
squirrelo: Ok, broken into three blocks: parsing query into the where clause and metadata headers searched over, building the actual query, and a helper to prepend SQL table labels.

wasade: Awesome :)

qiita_pet/templates/search_sample_list.html
squirrelo: Agreed, we have redis for a reason and we should probably start integrating the SQLConnectionManager with redis soon. Something similar to [this](https://clasense4.wordpress.com/2012/07/29/python-redis-how-to-cache-python-mysql-result-using-redis/)

squirrelo: also, if I introduce another object into the fray, I can cut down on the regex calls. This would mean there is a dict of lists with proc_data_ids keyed to their studies, a dict of lists with each sample id keyed to the proc_data_id it is in, and new third that is a dict of list of tuple, each tuple containing the samples with metadata, and each list of samples keyed to the metadata. This means we can run through the metadata cleaning once for all samples instead of each time it is printed, which could be multiple times per study.

squirrelo:
I managed to remove one layer of for loops using a string join, and it is at the lowest sample layer so it should therefore remove the most repetitions. @antgonza, if you could do a heavy search like `ph > 0` on the demo database before and after pulling this to get a sense of before and after page creation times, that would help a lot. I currently have a broken demo database and can't test this myself.

qiita_pet/util.py
```diff
@@ -51,8 +56,13 @@ def clean_str(item):
...
     str
         cleaned string
+
+    Notes
+    -----
+    This function removes colons and periods from strings, and replaces spaces
+    with underscores.
     """
-    return str(item).replace(" ", "_").replace(":", "")
+    return re.sub(clean_str_regex, "", str(item).replace(" ", "_"))
```
wasade: should be possible to do the replacement of the " " as well. @adamrp, resident regexpert, insight?

qiita_pet/util.py
```diff
@@ -38,9 +40,12 @@ def linkify(link_template, item):
     """
     return link_template.format(*item)
 
+# global compiled regex so compiled once and super fast after that
+clean_str_regex = re.compile("(\.|:)")
```
adamrp: These are capturing parentheses. Although this won't make the `clean_str` function fail, it would be better to use a character class:

```python
clean_str_regex = re.compile("[.:]")
```

qiita_pet/util.py
adamrp: Unfortunately no, since the `.` and `:` are being replaced with something different than the space (empty string versus underscore, respectively). It could be implemented like this:

```python
clean_str_regex = re.compile("[.:]")
def repl_fn(matchobj):
    if matchobj.group(0) == ' ':
        return '_'
    else:
        return ''

def clean_str(item):
    return re.sub(clean_str_regex, repl_fn, str(item))
```

but I'm not sure if that would be faster (it's calling a function for each occurrence of any of the three characters specified).

wasade: oh, that sucks :(

what about a translation table?

adamrp: Yes, I think a translation table is probably the way to go here, although if this function is really used a ton, then it's worth benching against the implementation suggested by @squirrelo

squirrelo: For what this is being used for, stripping spaces out completely might work fine as well. Testing now.

squirrelo: Someone has apparently already benched this type of thing and translation tables are the winner: http://stackoverflow.com/questions/265960/best-way-to-strip-punctuation-from-a-string-in-python

wasade: Likely just implements an array mapped against the ascii set or something,
so should be extremely quick.

adamrp: Excellent! Thanks for checking it out.

squirrelo:
Ready to go for another round of review.

antgonza:
Conflicts

squirrelo:
Conflicts resolved, ready for review again.

squirrelo:
Anyone want to do a final pass review?

squirrelo:
Please note this makes changes to the search object, so I can't do any more work on that until this gets merged.

qiita_db/search.py
```diff
@@ -223,24 +230,34 @@ def _parse_study_search_string(self, searchstr,
...
+        all_headers : list of str
+            all information columns searched over in the query
+        all_types : list of str
+            list of SQL types for each entry in all_headers, in the same order
+            as all_headers
 
         Notes
         -----
-        All searches are case-sensitive
+        All searches are NOT case-sensitive
```
wasade: case-insensitive

qiita_db/search.py
```diff
@@ -288,6 +304,41 @@ def _parse_study_search_string(self, searchstr,
...
+        Parameters
+        ----------
+        searchstr : str
+            The string to parse
+        remove_selected : bool, optional
+            If search is on an analysis that already has selected samples,
+            remove the selected samples from the results. Default False.
+        analysis : int, optional
+            Analysis ID search is run for. Required if remove_selected is True
+
```
wasade: extra space

qiita_db/search.py
```diff
@@ -288,6 +304,41 @@ def _parse_study_search_string(self, searchstr,
...
+        study_sql : str
+            SQL query for selecting studies with the required metadata columns
+        sample_sql : str
+            SQL query for each study to get the sample ids that mach the query
+        meta_headers : list
+            metadata categories in the query string in alphabetical order
+
+        Notes
+        -----
+        All searches are NOT case-sensitive
```
wasade: case-insensitive

qiita_db/search.py
```diff
@@ -303,9 +354,9 @@ def _parse_study_search_string(self, searchstr,
                     meta_header_type_lookup[header] = 'varchar'
 
         # create the study finding SQL
-        # remove metadata headers that are in required_sample_info table
-        meta_headers = meta_headers.difference(self.required_cols).difference(
-            self.study_cols)
+        # remove metadata headers that are in non-dynamic tables
+        meta_headers = meta_headers.difference(
```
wasade: `meta_headers = meta_headers - self.required_cols - self.study_cols`

wasade: ...not sure if that helps reduce the size of the line or not

squirrelo: Can only use operators if both are sets, and all these operations are on lists.

wasade: should `self.required_samp_cols` be a `list`? It likely should be at least immutable. Less sure about `self.study_cols` but I'd assume the same

wasade: `frozenset`?

squirrelo: I can tuple them both for safety.

squirrelo:
Ready to go.

antgonza:
This looks so much nicer! A couple of comments:
- The quick select is gone, this is a really nice feature, could you put back? Perhaps a series of checkboxes with datatypes will do it.
- These links, see image, don't do anything. 

screen shot 2015-03-05 at 5 09 02 am

squirrelo:
I'm not sure where those are coming from. In safari, firefox, and chrome I do not see that second set of select all/none/inverse you are pointing out. The ones under the word Samples though are the quick select by datatype.

antgonza:
1. Those appear when you click on already selected studies
2. Quick checkboxes are desired, see:
https://github.com/biocore/qiita/issues/692

squirrelo:
Removed the errant code for the extra select box. 

The "select all/select none" vs a checkbox is entirely stylistic, and removes the ability to have select inverse, so leaving as is.

squirrelo:
Ready to go.

antgonza:
I like the new distribution in the GUI but some of the things that were removed are really useful and should be returned. These are:
- There is no easy feedback of what has been selected and what hasn't. In the past we had two ways to see this: the study that was selected turned green and there was a counter of how many samples were available in the study and how many have been selected
- Easy select from main display either by a link or a checkbox. Having to click on your query tab and then having to select from there is not as easy as having a checkbox or a link.

rob-knight:
+1 especially on the second item


On Mar 6, 2015, at 8:49 AM, Antonio Gonzalez <notifications@github.com<mailto:notifications@github.com>> wrote:


I like the new distribution in the GUI but some of the things that were removed are really useful and should be returned. These are:

 *   There is no easy feedback of what has been selected and what hasn't. In the past we had two ways to see this: the study that was selected turned green and there was a counter of how many samples were available in the study and how many have been selected
 *   Easy select from main display either by a link or a checkbox. Having to click on your query tab and then having to select from there is not as easy as having a checkbox or a link.

squirrelo:
Ok, so before I implement it, should it then check every sample in every processed data for that datatype, regardless of the possibility of duplicates or not?

squirrelo:
Also, there is not easy way to do a counter now, since there is the possibility of multiple copies of the same sample within different processed data in the same or different datatypes. The study row still turns green when at least one is selected though.

antgonza:
I think there should be 4 levels of select/check all:
- Select all samples from specific study from a specific data type
(ie: select all samples from processed data X for study Y)
- Select all samples from all studies in a specific data type (ie:
select all samples for all studies returned from the search that were
processed in X way = 16S or Y parameters)
- Select all samples from all data types in a single study (ie: select
all samples from all processed tables for study X)
- Select everything from a query.

Additionally, it will be nice to have a deselect all or select
inverse, these can be a general ones.

Not sure what you mean about duplicates, is it:
https://github.com/biocore/qiita/issues/283 ? If so then, yes we want
to select all.

squirrelo:
Yeah that's what I meant about duplicates. I'll add this, but I am wary of the confusion for the user that the overlap of all these different selection methods will cause.

antgonza:
Agree and that's why we need to solve that issue too.

squirrelo:
Added selects. Ready for another review.

antgonza:
Some more issues:
- There is no way to deselect an already selected study, "Remove samples" button
- Select inverse doesn't work within a study sample selection (frame)
- This might be out of the scope of this PR (OK to open another issue
about it) but currently you can select 16S, WGS or whatever but what
we really want is to be able to select by processing method. This
should not only allow us to select by data type but also differentiate
between GG_13_8, Silva, etc.

wasade:
It should really be by processing parameters as we care about the methods,
versions, and parameters used

antgonza:
That's what I was trying to say.

squirrelo:
We had a weeks-long discussion about selecting samples vs selecting processed data when this page was first created, and the can of worms opened up by allowing users to select by sample vs processing. I'm implementing the simplest form of selection we agreed upon then, but if we want to rehash that discussion we can during the code sprint. I'll open an issue for it.

antgonza:
OK, let's discuss but note that the first 2 points are still missing.

squirrelo:
Ready to go once tests pass.

antgonza:
:+1:

qiita_db/search.py
```diff
@@ -288,6 +304,40 @@ def _parse_study_search_string(self, searchstr,
...
+
+        Parameters
+        ----------
+        searchstr : str
+            The string to parse
+        remove_selected : bool, optional
+            If search is on an analysis that already has selected samples,
+            remove the selected samples from the results. Default False.
+        analysis : int, optional
+            Analysis ID search is run for. Required if remove_selected is True
```
adamrp: `only_with_processed_data` is not documented

qiita_pet/handlers/analysis_handlers.py
```diff
@@ -59,72 +59,30 @@ def check_analysis_access(user, analysis):
...
+            study_proc_data[proc_data.study][proc_data.data_type()].append(pid)
+        return study_proc_data, proc_data_samples
+
+    def _parse_deselect_checkbox_data(self):
+        """splits the list values into proc_data_id and sample_id, then builds
+           dict of results as {proc_data_id: [ samp1, samp2, ...], ...} """
+        data = self.get_arguments('samples-sel')
+        res = defaultdict(list)
+        for val in data:
+            hold = val.split('#', 1)
```
adamrp: IMHO, I think this deserves a comment -- what is the expected format here? I can guess at it, but it would be good to have explicitly stated, probably.

qiita_pet/static/js/analysis.js
```diff
@@ -22,57 +22,49 @@ function displaySelected() {
...
-  select_deselect(study, sel)
+function count_update(study, sel) {
+  format = '#modal' + study;
+  if(sel == "-sel") { format = format + "-sel" }
+    format = format + ' input:checkbox:checked';
+  var selected = $(format).length;
+  if(selected > 0 && sel != "-sel") { $('#study' + study).addClass('success'); }
+  else if(sel != "-sel") { $('#study' + study).removeClass('success'); }
+  else if(selected > 0 && sel == "-sel") { $('#study' + study+"-sel").addClass('danger'); }
+  else if(sel == "-sel") { $('#study' + study + "-sel").removeClass('danger'); }
```
adamrp: Use of `===` and `!==` inconsistent with code above

qiita_pet/static/js/analysis.js
```diff
@@ -22,57 +22,49 @@ function displaySelected() {
...
-  count_update(study);
+  count_update(study, sel);
 }
 
-function select_inverse(study) {
-  $('.'+study).each(function() {
-    if(this.checked == true) { this.checked = false; }
+function select_inverse(study, filter, sel, type) {
+  filter = filter + ".study" + study;
+  if(sel == "-sel") { filter += ".-sel"; }
```
adamrp: `===`

qiita_pet/static/js/analysis.js
```diff
@@ -22,57 +22,49 @@ function displaySelected() {
...
-    $('#study' + study + " input:checkbox").each(function() {this.checked = false;})
-  }
-}
-
-function select_deselect(study, select) {
-  if(select == true) { 
-    $('.'+study).each(function() {this.checked = true;});
+function select_deselect(study, filter, sel, select) {
+  filter = '.study' + study + filter;
+  if(sel == '-sel') { filter = filter + '.-sel'; }
```
adamrp: `===`

qiita_pet/templates/search_sample_list.html
```diff
@@ -0,0 +1,64 @@
+<!--Study datatypes tabs-->
+{% from future.utils import viewitems %}
+{% from qiita_pet.util import clean_str %}
+{% set sel = "" %}
+{% if selected is True %}
+    {% set sel = "-sel" %}
+  {% end %}
```
adamrp: indented slightly too far

qiita_pet/uimodules/search_sample_list.py
```diff
@@ -0,0 +1,12 @@
...
+
+class SearchSampleList(BaseUIModule):
+    def render(self, study_id, meta_headers,
+               proc_data_dict, proc_data_samples):
+        selected = not meta_headers
+
+        return self.render_string(
+            "search_sample_list.html", meta_headers=meta_headers, sid=study_id,
+            proc_data_dict=proc_data_dict, proc_data_samples=proc_data_samples,
+            selected=selected)
```
adamrp: is this also done in the template? It doesn't seem necessary to keep track of it separately in the handler since it is always just the boolean NOT of meta_headers. If this *is* necessary, please add a comment, since I was confused by it and others might also be.

qiita_ware/search.py
```diff
@@ -0,0 +1,108 @@
...
+            proc_data = ProcessedData(proc_data_id)
+            datatype = proc_data.data_type()
+            # skip processed data if it doesn't fit the given datatypes
+            if datatypes is not None and datatype not in datatypes:
+                continue
+            samps_available = set(proc_data.samples)
+            proc_data_samples[proc_data_id] = [s for s in results[study_id]
+                                               if s[0] in samps_available]
+            # Add number of samples left to the total for that datatype
+            dtcount[datatype] += len(proc_data_samples[proc_data_id])
```
adamrp: calculate `len(proc_data_samples[proc_data_id])` once, and then in the `if` statement on 95 check if length is 0

adamrp: not blocking

squirrelo:
MERGE ME!

ElDeveloper:
:loudspeaker: do not merge :warning:

squirrelo:
Referencing #978 here to keep a link, as this starts to fix the issue.

PR #866, opened 2015-02-18
Queues for adding templates

josenavas:
Modifies the insertion of the templates to use the queue system. This will help to avoid the problem in which "floating" samples are added and cannot be automatically removed.

### DIFF
--- a/qiita_db/metadata_template.py
+++ b/qiita_db/metadata_template.py
@@ -1479,6 +1479,8 @@ class PrepTemplate(MetadataTemplate):
 
         # Get a connection handler
         conn_handler = SQLConnectionHandler()
+        queue_name = "CREATE_PREP_TEMPLATE_%d" % raw_data.id
+        conn_handler.create_queue(queue_name)
 
         # Check if the data_type is the id or the string
         if isinstance(data_type, (int, long)):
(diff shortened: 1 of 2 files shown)


ElDeveloper:
Looks great, only thing is: Do you think we could add a test to ensure the problem we were having doesn't come up again? Seems like a tricky test to design but would be nice to have it.

josenavas:
I will try to add a test, although as you said, is going to be tricky...

adamrp:
:+1: when tests pass.

josenavas:
Does somebody knows a specific case where this issue was showing up? I'm having problems on defining that situation.

adamrp:
I do not, sorry -- I think the queues should eliminate the possibility. Is deletion of md templates also taken care of via queue? If creation and deletion are both atomic operations (via queue), then the problematic situation should not be possible.

ElDeveloper:
Yeah, I think you may be right. Do you agree @josenavas or did you find
a way? :O

josenavas:
@antgonza helped me to find a way. Tests are added now

josenavas:
BTW, I agree that it should not happen, but better to be sure with tests!

josenavas:
This should be ready to go!

antgonza:
:+1:

josenavas:
@ElDeveloper can you take a quick look and merge if you think it is correct?

ElDeveloper:
Looks good to me!

ElDeveloper:
Thanks for figuring out a way to test this!

PR #868, opened 2015-02-18
Typo and docstring

adamrp:
minor, noticed while reviewing other PRs

### DIFF
--- a/qiita_db/metadata_template.py
+++ b/qiita_db/metadata_template.py
@@ -1181,9 +1181,8 @@ class SampleTemplate(MetadataTemplate):
         ----------
         md_template : DataFrame
             The metadata template file contents indexed by samples Ids
-        study : Study or RawData
-            The study to which the metadata template belongs to. Study in case
-            of SampleTemplate and RawData in case of PrepTemplate
+        study : Study
+            The study to which the sample template belongs to.
         """
         cls._check_subclass()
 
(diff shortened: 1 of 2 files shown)


qiita_db/metadata_template.py
```diff
@@ -1181,9 +1181,8 @@ def create(cls, md_template, study):
         ----------
         md_template : DataFrame
             The metadata template file contents indexed by samples Ids
-        study : Study or RawData
-            The study to which the metadata template belongs to. Study in case
-            of SampleTemplate and RawData in case of PrepTemplate
+        study : Study
+            The study to which the sample template belongs to.
```
adamrp: I think this is an accurate change, but please correct me if I'm wrong, and I can revert

josenavas: Yes it is accurate. Good catch!

qiita_pet/handlers/study_handlers/description_handlers.py
```diff
@@ -77,7 +77,7 @@ def _to_int(value):
 
 class StudyDescriptionHandler(BaseHandler):
 
-    def _get_sudy_and_check_access(self, study_id):
+    def get_study_and_check_access(self, study_id):
```
josenavas: Why is no longer private?

adamrp: Woops, find/replace mistake. correcting now...

ElDeveloper:
:+1:

PR #870, opened 2015-02-18
Cleaner error when prep template contains non-sample template IDs

squirrelo:
This pull request adds a check and filter to the prep template addition. It makes sure the samples in the prep template exist in the sample template, and whichever don't are removed and raised as a warning. 

This allows us to support a prep template with samples from multiple studies on it, which is frequently the case.

### DIFF
--- a/qiita_db/metadata_template.py
+++ b/qiita_db/metadata_template.py
@@ -53,7 +53,7 @@ from qiita_core.exceptions import IncompetentQiitaDeveloperError
 from .exceptions import (QiitaDBDuplicateError, QiitaDBColumnError,
                          QiitaDBUnknownIDError, QiitaDBNotImplementedError,
                          QiitaDBDuplicateHeaderError, QiitaDBError,
-                         QiitaDBWarning)
+                         QiitaDBWarning, QiitaDBExecutionError)
 from .base import QiitaObject
 from .sql_connection import SQLConnectionHandler
 from .ontology import Ontology
@@ -1491,23 +1491,6 @@ class PrepTemplate(MetadataTemplate):
         queue_name = "CREATE_PREP_TEMPLATE_%d" % raw_data.id
         conn_handler.create_queue(queue_name)
 
-        # Check if there are sample IDs present here but not in sample template
-        sql = ("SELECT sample_id from qiita.required_sample_info WHERE "
-               "study_id = %s")
-        # Get list of study sample IDs, prep template study IDs,
-        # and their intersection
-        study_samples = set(s[0] for s in conn_handler.execute_fetchall(
-            sql, [study.id]))
-        prep_samples = set(md_template.index.values)
-        both_samples = study_samples.intersection(prep_samples)
-        # filter prep template to just samples available in the study
-        md_template = md_template.loc[both_samples]
-        if len(both_samples) != len(prep_samples):
-            prep_samples.difference_update(both_samples)
-            raise QiitaDBWarning(
-                'Samples found in prep template but not sample template: %s'
-                % ', '.join(prep_samples))
-
         # Check if the data_type is the id or the string
         if isinstance(data_type, (int, long)):
             data_type_id = data_type
@@ -1607,6 +1590,25 @@ class PrepTemplate(MetadataTemplate):
             conn_handler.execute(
                 "DELETE FROM qiita.prep_template where "
                 "{0} = %s".format(cls._id_column), (prep_id,))
+
+            # Check if sample IDs present here but not in sample template
+            sql = ("SELECT sample_id from qiita.required_sample_info WHERE "
+                   "study_id = %s")
+            # Get list of study sample IDs, prep template study IDs,
+            # and their intersection
+            study_samples = set(s[0] for s in conn_handler.execute_fetchall(
+                sql, [study.id]))
+            prep_samples = set(md_template.index.values)
+            both_samples = study_samples.intersection(prep_samples)
+            # filter prep template to just samples available in the study
+            md_template = md_template.loc[both_samples]
+            if len(both_samples) != len(prep_samples):
+                prep_samples.difference_update(both_samples)
+                raise QiitaDBExecutionError(
+                    'Samples found in prep template but not sample template: '
+                    '%s' % ', '.join(prep_samples))
+
+            # some other error we haven't seen before so raise it
             raise
 
         # figuring out the filepath of the backup
--- a/qiita_db/metadata_template.py
+++ b/qiita_db/metadata_template.py
@@ -1596,17 +1596,13 @@ class PrepTemplate(MetadataTemplate):
                    "study_id = %s")
             # Get list of study sample IDs, prep template study IDs,
             # and their intersection
-            study_samples = set(s[0] for s in conn_handler.execute_fetchall(
-                sql, [study.id]))
             prep_samples = set(md_template.index.values)
-            both_samples = study_samples.intersection(prep_samples)
-            # filter prep template to just samples available in the study
-            md_template = md_template.loc[both_samples]
-            if len(both_samples) != len(prep_samples):
-                prep_samples.difference_update(both_samples)
+            unknown_samples = prep_samples.difference(
+                s[0] for s in conn_handler.execute_fetchall(sql, [study.id]))
+            if unknown_samples:
                 raise QiitaDBExecutionError(
                     'Samples found in prep template but not sample template: '
-                    '%s' % ', '.join(prep_samples))
+                    '%s' % ', '.join(unknown_samples))
 
             # some other error we haven't seen before so raise it
             raise
--- a/qiita_db/metadata_template.py
+++ b/qiita_db/metadata_template.py
@@ -1407,7 +1407,7 @@ class SampleTemplate(MetadataTemplate):
 
 
 class PrepTemplate(MetadataTemplate):
-    r"""Represent the PrepTemplate of a raw dat. Provides access to the
+    r"""Represent the PrepTemplate of a raw data. Provides access to the
     tables in the DB that holds the sample preparation information.
 
     See Also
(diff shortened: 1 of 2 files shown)


squirrelo:
Ready for review

ElDeveloper:
Code looks ok, there's just a few things though:

- Can you add a test case with a prep template that has fewer samples than the sample template? I think you are only checking the case where an exception is raised.

- Would it be feasible to get a patch such that all studies have meet this constraint?

- Since this is going to be ignoring a few rows in the templates, it would be useful for the user to get a warning about this.

josenavas:
Sorry for a late comment but, I though the prep templates where always done in a per-study basis?
You state that is frequently the case, but I still have not face it... @ackermag is that true that prep templates contain samples from multiple studies?

squirrelo:
From my understanding you can get a prep template for a single run that has multiple studies on it. This pull request allows you to not have to filter that prep template by study and instead just upload the same one to each study, then have the system auto-filter to the samples relevant to the study.

squirrelo:
@ElDeveloper 
* I can add that test, yeah. Good catch. 
* I don't think there is anything to patch with this, since it's working on prep templates as they are added to the system. We already have the fk set up so you can't add samples that don't exist in the study, this just makes it cleaner and not have the long awful SQL error.
* Currently it raises a warning with a list of samples removed. From my understanding of the code, this warning will then be printed to the screen upon addition of the prep template finishing.

josenavas:
Where did you get that information? AFAIK the prep templates are prepared in a per-study basis, and I've still not seen any prep templates with multiple studies on it. However I could be wrong.

My concern is that if then the user has the sample id's with typos, then they will be ignored, instead of failing because they don't match the study ones...

ElDeveloper:
*Thanks!
*Ok
*I see, what I was looking for was a warnings.warn call instead of that
raise block. Can you change that? See load_template_to_dataframe for an
example of what I mean.

antgonza:
I agree with @josenavas on the fact that they shouldn't be ignored and that an error with the list of "filtered" out samples should be added. If the user doesn't want that sample to be present, it can really easily remove it.

squirrelo:
Pinging @ackermag to comment on the prevalence of prep templates with multiple studies in them. Also I can always add an extend function to the prep template, similar to #876, if we want to be able to add back misspelled samples.

squirrelo:
After discussion with @ackermag and @josenavas, implementing the error raising instead of add and warn.

josenavas:
@squirrelo can you close this PR an open a new one? I think is going to be misleading if we continue working in here.

Thanks!

ackermag:
To clarify:
The prep template is on a per study basis only.  No samples are  included
from other studies that may have been processed on the same run.  I think
the information @squirrelo <https://github.com/squirrelo> got from the
wetlab pertained to the minimal mapping file produced by the wetlab
to inform all the barcodes in a single run of data (which may include
several studies).
I think the only check that needs to be done is that the list of samples in
the sample template includes all the samples in the prep template.


On Mon, Feb 23, 2015 at 9:44 AM, Joshua Shorenstein <
notifications@github.com> wrote:

> After discussion with @ackermag <https://github.com/ackermag> and
> @josenavas <https://github.com/josenavas>, implementing the error raising
> instead of add and warn.
>
> —
> <https://github.com/biocore/qiita/pull/870#issuecomment-75592587>.
>



-- 
Gail Ackermann
Knight Lab
UCSD

squirrelo:
@josenavas it's just moving a block of code down so not really an issue. I can also rename the pull request.

adamrp:
@ackermag, thanks for the clarification, that's my understanding as well.

squirrelo:
Comments addressed, once tests pass is good for another review.

qiita_db/test/test_metadata_template.py
```diff
@@ -1363,6 +1364,35 @@ def test_create_bad_sample_names(self):
             PrepTemplate.create(self.metadata, self.new_raw_data,
                                 self.test_study, self.data_type)
 
+    def test_create_unknown_sample_names(self):
+        # set two real and one fake sample name
+        self.metadata_dict['NOTREAL'] = self.metadata_dict['SKB7.640196']
+        del self.metadata_dict['SKB7.640196']
+        self.metadata = pd.DataFrame.from_dict(self.metadata_dict,
+                                               orient='index')
+        # Test warning raised and correct warning given
```
josenavas: Needs updating comment

josenavas:
just a small comment @squirrelo

squirrelo:
Fixed things, ready for review.

qiita_db/metadata_template.py
```diff
@@ -1590,6 +1590,25 @@ def create(cls, md_template, raw_data, study, data_type,
...
+                sql, [study.id]))
+            prep_samples = set(md_template.index.values)
+            both_samples = study_samples.intersection(prep_samples)
+            # filter prep template to just samples available in the study
+            md_template = md_template.loc[both_samples]
+            if len(both_samples) != len(prep_samples):
+                prep_samples.difference_update(both_samples)
+                raise QiitaDBExecutionError(
+                    'Samples found in prep template but not sample template: '
+                    '%s' % ', '.join(prep_samples))
```
adamrp: The code confuses me slightly, and I don't think it can be simplified:

```python
unknown_sample_ids = prep_samples - study_samples
if unknown_sample_ids:
    # Raise the QiitaDBExecutionError as before
```

qiita_db/metadata_template.py
```diff
@@ -1590,6 +1590,25 @@ def create(cls, md_template, raw_data, study, data_type,
...
+            study_samples = set(s[0] for s in conn_handler.execute_fetchall(
+                sql, [study.id]))
+            prep_samples = set(md_template.index.values)
+            both_samples = study_samples.intersection(prep_samples)
+            # filter prep template to just samples available in the study
+            md_template = md_template.loc[both_samples]
+            if len(both_samples) != len(prep_samples):
+                prep_samples.difference_update(both_samples)
+                raise QiitaDBExecutionError(
+                    'Samples found in prep template but not sample template: '
```
ElDeveloper: This error message could be improved.

qiita_db/metadata_template.py
```diff
@@ -1590,6 +1590,25 @@ def create(cls, md_template, raw_data, study, data_type,
             conn_handler.execute(
                 "DELETE FROM qiita.prep_template where "
                 "{0} = %s".format(cls._id_column), (prep_id,))
+
+            # Check if sample IDs present here but not in sample template
+            sql = ("SELECT sample_id from qiita.required_sample_info WHERE "
+                   "study_id = %s")
+            # Get list of study sample IDs, prep template study IDs,
+            # and their intersection
+            study_samples = set(s[0] for s in conn_handler.execute_fetchall(
```
ElDeveloper: This cast into a set could be removed if you use DISTINCT in the sql query.

qiita_db/metadata_template.py
```diff
@@ -1590,6 +1590,25 @@ def create(cls, md_template, raw_data, study, data_type,
             conn_handler.execute(
                 "DELETE FROM qiita.prep_template where "
                 "{0} = %s".format(cls._id_column), (prep_id,))
+
```
ElDeveloper: This whole new patch lives inside a `try/except Exception`, is there any way to avoid this?

qiita_db/metadata_template.py
squirrelo: The cast is so you can do the set operations (difference_update, etc) on it, not to filter to distinct samples.

qiita_db/metadata_template.py
squirrelo: Fair, was testing both ways around before but not needed any more.

qiita_db/metadata_template.py
squirrelo: How so? It seems clear to me.

qiita_db/metadata_template.py
squirrelo: Not that I can think of if we want to do it the way that was discussed, which is that we only do the checks if insertion fails.

adamrp:
done with review, code looks good except for the comments that @ElDeveloper and I added after @josenavas was done. Thanks!

qiita_db/metadata_template.py
squirrelo: I've cleaned up the logic here and leveraged generators to save memory.

qiita_db/test/test_metadata_template.py
```diff
@@ -1363,6 +1364,38 @@ def test_create_bad_sample_names(self):
...
 
+    def test_create_unknown_sample_names(self):
+        # set two real and one fake sample name
+        self.metadata_dict['NOTREAL'] = self.metadata_dict['SKB7.640196']
+        del self.metadata_dict['SKB7.640196']
+        self.metadata = pd.DataFrame.from_dict(self.metadata_dict,
+                                               orient='index')
+        # Test error raised and correct error given
+        # Try/Except used so can test the text of the error raised
+        try:
```
ElDeveloper: The problem with this approach is that if no exception is raised this test could still potentially pass, you need to assertRaises (with a context manager) and then check for the appropriate values.

squirrelo: The problem with assertRaises is that I can't check the string of the error message. I think I can combine both in the test though to make absolutely sure.

qiita_db/test/test_metadata_template.py
```diff
@@ -1363,6 +1364,42 @@ def test_create_bad_sample_names(self):
...
                                 self.test_study, self.data_type)
 
+    def test_create_unknown_sample_names(self):
+        # set two real and one fake sample name
+        self.metadata_dict['NOTREAL'] = self.metadata_dict['SKB7.640196']
+        del self.metadata_dict['SKB7.640196']
+        self.metadata = pd.DataFrame.from_dict(self.metadata_dict,
+                                               orient='index')
+        # Test error raised and correct error given
+        with self.assertRaises(QiitaDBExecutionError):
```
ElDeveloper: Try this instead:

```python
from unittest import main, TestCase

def foo():
    raise ValueError('lolsies')

class LOL(TestCase):
    def setUp(self):
        pass
    def test_lol(self):
        pass
    def test_rofl(self):
        with self.assertRaises(ValueError) as e:
            foo()
            self.assertEqual(str(e.exception), 'lolsies')

if __name__ == '__main__':
    main()


```

squirrelo: I had it that way originally, but that only works with warnings. Anything inside the with after the error is raised is ignored.

ElDeveloper: You can also move the assertEqual outside the context manager.

squirrelo: Oh python and it's silly scope issues. That worked.

squirrelo:
This passed tests, not sure why it's not updating to show that here. Regardless, ready for another review.

ElDeveloper:
:+1:

josenavas:
:+1:

josenavas:
@squirrelo I'm going to restart de build here to see if it updates now.

Issue #684, opened 2014-11-24
make the download validator non blocking


antgonza:
The download is not blocking but the validator (permissions) is.

antgonza:
This one is covered by the multiple workers and by solving/speeding-up https://github.com/biocore/qiita/issues/1010 so closing.

PR #873, opened 2015-02-19
Results permissions issue 869

adamrp:
This makes sure that a user has access to requested analysis results files.

### DIFF
--- a/qiita_db/analysis.py
+++ b/qiita_db/analysis.py
@@ -282,7 +282,7 @@ class Analysis(QiitaStatusObject):
 
     @property
     def all_associated_filepath_ids(self):
-        """Get all associated filepath_ids EXCEPT job results filepaths
+        """Get all associated filepath_ids
 
         Returns
         -------
@@ -295,6 +295,19 @@ class Analysis(QiitaStatusObject):
               WHERE af.analysis_id = %s"""
         filepaths = {row[0]
                      for row in conn_handler.execute_fetchall(sql, [self._id])}
+
+        sql = """SELECT fp.filepath_id
+              FROM qiita.analysis_job aj
+                JOIN qiita.job j ON aj.job_id = j.job_id
+                JOIN qiita.job_results_filepath jrfp ON aj.job_id = jrfp.job_id
+                JOIN qiita.filepath fp ON jrfp.filepath_id = fp.filepath_id
+              WHERE aj.analysis_id = %s"""
+
+        job_filepaths = {row[0]
+                         for row in conn_handler.execute_fetchall(sql,
+                                                                  [self._id])}
+
+        filepaths = filepaths.union(job_filepaths)
 
         return filepaths
 
--- a/qiita_pet/handlers/analysis_handlers.py
+++ b/qiita_pet/handlers/analysis_handlers.py
@@ -333,8 +333,9 @@ class ResultsHandler(StaticFileHandler, BaseHandler):
         """
         u = self.current_user
 
+        db_files_base_dir = get_db_files_base_dir()
         accessible_filepaths = get_accessible_filepath_ids(u)
-        accessible_filepaths = {join(get_db_files_base_dir(),
+        accessible_filepaths = {join(db_files_base_dir,
                                      filepath_id_to_rel_path(fpid))
                                 for fpid in accessible_filepaths}
 
--- a/qiita_pet/handlers/analysis_handlers.py
+++ b/qiita_pet/handlers/analysis_handlers.py
@@ -12,7 +12,7 @@ Qitta analysis handlers for the Tornado webserver.
 from __future__ import division
 from future.utils import viewitems
 from collections import defaultdict, Counter
-from os.path import join, isdir, sep
+from os.path import join, sep, commonprefix
 
 from tornado.web import authenticated, HTTPError, StaticFileHandler
 from pyparsing import ParseException
@@ -29,10 +29,10 @@ from qiita_db.metadata_template import SampleTemplate
 from qiita_db.job import Job, Command
 from qiita_db.util import (get_db_files_base_dir, get_table_cols,
                            filepath_id_to_rel_path)
-from qiita_db.meta_util import get_accessible_filepath_ids
 from qiita_db.search import QiitaStudySearch
 from qiita_db.exceptions import (
     QiitaDBIncompatibleDatatypeError, QiitaDBUnknownIDError)
+from qiita_db.sql_connection import SQLConnectionHandler
 
 SELECT_SAMPLES = 2
 SELECT_COMMANDS = 3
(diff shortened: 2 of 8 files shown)


qiita_pet/handlers/analysis_handlers.py
```diff
@@ -321,3 +325,31 @@ def get(self):
...
+
+
+class ResultsHandler(StaticFileHandler, BaseHandler):
+    def validate_absolute_path(self, root, absolute_path):
+        """Overrides StaticFileHandler's method to include authentication
+        """
+        u = self.current_user
+
+        accessible_filepaths = get_accessible_filepath_ids(u)
+        accessible_filepaths = {join(get_db_files_base_dir(),
```
wasade: The call to `get_db_files_base_dir` only needs to happen once, possible to pull that out of the loop?

qiita_pet/handlers/analysis_handlers.py
```diff
@@ -321,3 +325,31 @@ def get(self):
...
+                                for fpid in accessible_filepaths}
+
+        for accessible_filepath in accessible_filepaths:
+            access = False
+
+            if (accessible_filepath == absolute_path):
+                access = True
+
+            elif isdir(accessible_filepath) and \
+                    absolute_path.startswith(accessible_filepath + sep):
```
wasade: is `accessible_filepath` assured to not already have a '/'?

qiita_pet/handlers/analysis_handlers.py
```diff
@@ -321,3 +325,31 @@ def get(self):
...
+                access = True
+
+            if access:
+                return super(ResultsHandler, self).validate_absolute_path(
+                    root, absolute_path)
+
+        raise QiitaPetAuthorizationError(u.id, absolute_path)
```
wasade: From a probing perspective, this might help reveal details about valid paths (despite access) and could expose more details than necessary. Totally minor... but we could also just throw either 404 or just 403 (regardless of if the path exists or not)

qiita_pet/handlers/analysis_handlers.py
adamrp: I think it depends on the job's parameters (if the job specifies an output directory with a trailing slash, then maybe that's what gets recorded in the filepath table). In any case, this check is necessary because job results can be directories, and the static handler has to be able to serve the files from within those directories if they have access, but you wouldn't want to allow access to `5_summarize_taxa_through_plots.py_output_dir_buahahahaha` if the user only has access to `5_summarize_taxa_through_plots.py_output_dir`. I couldn't think of a better way to do this, but I'm open to suggestions!

qiita_pet/handlers/analysis_handlers.py
antgonza: A couple of issues with functionality:
- A regular user can not open its own studies. However, a user can not
open an admin study, which is fine.
- On a production site (with at least the EMP studies) these changes
make more evident this issue:
https://github.com/biocore/qiita/issues/684. Do you think is possible
to check for a specific file path id and user vs. getting all valid
file paths ids and then checking if the specific one is within that
group? This should make this faster. Note that this is specially bad
for admins as it returns basically all file path ids!

adamrp: When this exception is raised, the client receives a 500 error (which is raised whenever an exception is raised in a Qiita handler). This 500 response will be sent whether or not `absolute_path` exists or not, if the logged in user does not have access to the `absolute_path` (and it *should* be the case that a probe will not be logged in anyways, right?)

qiita_pet/handlers/analysis_handlers.py
adamrp: Good call

qiita_pet/handlers/analysis_handlers.py
adamrp: @antgonza can you clarify your first point? I don't understand what the issue is. By studies, do you mean analysis results? In this implementation, a regular user is capable of opening their own analysis results.

> Do you think is possible to check for a specific file path id...

I do not believe it is possible (assuming we still want to use `StaticFileHandler` or a subclass of it to serve results), for a couple reasons.

1) Not all results files are in the filepath table (e.g., the output of summarize taxa is a directory, which is in the filepath table, but the files contained within that directory are not in the filepath table, and these are the paths that are being served by the `StaticFileHandler`)
2) The static filehandler is only aware of the actual path being requested, and not the filepath ID.

antgonza:
1. The user couldn't open it's own results. The issue was that I had a broken study due to some tests and was getting and error.
2. Have you tested on a machine that has the EMP studies and have some meta-analysis performed with a non admin user? It takes a lot (haven't been able to open one, note that it opens fine for an admin) to open for all the checks it has to do, I believe due to `get_accessible_filepath_ids` because it uses the object vs. doing a simple query to retrieve what's needed. If we actually deploy this in qiita.server we will bring the server down. Is there a way to make it less expensive?

qiita_pet/handlers/analysis_handlers.py
squirrelo: maybe commonprefix?

```python
if isdir(accessible_filepath) and commonprefix([absolute_path, accessible_filepath]) == accessible_filepath.strip('/\\'):
    # do stuff
```

adamrp:
1. So if i understand correctly this is no longer an issue.
2. I tested on only my own machine which has only a few studies, and a
couple meta analyses i used for testing. If we want to use
staticfilehandler, i do not know a cheaper way to do this, since the only
information exposed to the handler is the file path and the user (not even
the analysis id). There are other cheaper ways to do this using regular
handlers, i think. Were you able to test on a system where the expense of
the function caused usability issues?

I'm unable to replicate the issue you are seeing where the user cannot open
an analysis that they have access to. Note that since this is a static file
handler, you have to do a hard refresh to avoid loading from cache; perhaps
that is the issue?

antgonza:
1. It could be if there are Studies that failed but I think the queue solution that Jose implemented should prevent this.
2. My "real" case environment is basically the same than qiita: has all the emp studies, with 2 users (mine, regular, and demo, as admin), and when I test a new feature I add new studies.

I tested that and the issue is performance, cause once I try to open a study as a user you can hear my laptop going nuts and you can see an increment of 20% on CPU, 70% on RAM, and heat, without doing anything else + is even worst if you decide to have 2 tabs.

adamrp:
You say study again; do you mean analysis results? Specifically analysis
results from the /results/... URI?

if this is too expensive, then i don't think we can use a static file
handler (or at least i don't know how to implement it).

antgonza:
Yup, study on the last paragraph really means meta-analysis, sorry for
confusion.

I do not _think_ the issue is how the static handlers perform but how
get_accessible_filepath_ids retrieves the results: get's all the
studies using the Study object, then looks on each of their raw data,
preprocessed data, and process data to retrieve all file paths; then
does the same for analysis. All these is extremely expensive for all
the objects and loops created. I _believe_ that changing this for a
"complex" SQL command will make things way better. Let me know if you
want to chat.

adamrp:
The issue, as I've said, is that you only have a file path and a user.
Let's chat this morning, I'll be in touch soon.

wasade:
As Antonio suggested, might this be phrased as a set membership question
against postgres? It likely will still be expensive though, so it might be
worth some thought on either different structures or a more appropriately
structured cache

adamrp:
@antgonza pointed out that the path requested by the user is always either going to be in the `BASE_DATA_DIR/job/` directory (or whatever the current mount is), or in a subdirectory of it. Therefore, we can look at the first path element under `job/` and just check whether the user has access to that file or directory.

@antgonza can you test this on your database (which is more populated than mine) when you get a chance?

antgonza:
Thanks @adamrp. This works sooooo much better!

A few comments:
- Admins do not have access to all meta-analysis
- Raising errors and loging them is fine but it doesn't tell the user
that the meta-analysis doesn't exist or they have not the right
permissions, which might be more useful
- I know is not part of this PR but we have discussed a lot on is so
will it be possible to:
 + Improve speed of opening the main meta-analysis page, this is still
slow compared to opening a job. My guess a lot of object creation.
 + Improve either how get_accessible_filepath_ids retrieves all ids or
like in this case change it that so it doesn't retrieve all fp ids but
just checking if fileA can be accessed by UserB

squirrelo:
For @antgonza comment part 1, you can easily add an `if user.admin or user.dev: return True` in tehre and get around having to even check admin and dev access.

adamrp:
>- Admins do not have access to all meta-analysis

Added

>- Raising errors and loging them is fine but it doesn't tell the user
that the meta-analysis doesn't exist or they have not the right
permissions, which might be more useful

People should never be accessing the `/results/` URI directly, so getting a 500 is okay I think.

>- I know is not part of this PR but we have discussed a lot on is so
will it be possible to:
 + Improve speed of opening the main meta-analysis page, this is still
slow compared to opening a job. My guess a lot of object creation.
 + Improve either how get_accessible_filepath_ids retrieves all ids or
like in this case change it that so it doesn't retrieve all fp ids but
just checking if fileA can be accessed by UserB

I think these comments should be addressed in a different PR, as you suggest.

I also addressed @squirrelo's comments, so if/when tests pass, this should be ready for another review. Thanks, guys!

adamrp:
Ready for another review

qiita_db/analysis.py
```diff
@@ -282,7 +282,7 @@ def shared_with(self):
 
     @property
     def all_associated_filepath_ids(self):
-        """Get all associated filepath_ids EXCEPT job results filepaths
```
squirrelo: I understand that there are less object instantiations this way, but do we want the analysis object interacting with jobs like this? Not sure how much blending of the two objects we want to allow.

squirrelo:
One philosophical question, but not blocking :+1:

qiita_db/analysis.py
josenavas: @squirrelo I don't think this is an issue... If you want to retrieve all the files associated with an analysis, I would say that this makes sense. The 2 objects are really related so improving SQL queries rather than rely on python interactions can be really good.

josenavas:
Code looks good to me. @antgonza you've been reviewing this one. Can you check it again and merge if you agree with the code?

PR #876, opened 2015-02-20
Add extend function to sample template

squirrelo:
This is the first of two pull requests for the ability to add samples to previously created studies. This adds the `extend` function to the SampleTemplate object, which takes in a metadata dataframe and essentially appends it to the existing sample template. It also checks to make sure there are no duplicated samples during this process.

### DIFF
--- a/qiita_db/metadata_template.py
+++ b/qiita_db/metadata_template.py
@@ -60,7 +60,7 @@ from .ontology import Ontology
 from .util import (exists_table, get_table_cols, get_emp_status,
                    get_required_sample_info_status, convert_to_id,
                    convert_from_id, find_repeated, get_mountpoint,
-                   insert_filepaths)
+                   insert_filepaths, scrub_data)
 from .logger import LogEntry
 
 if PY3:
@@ -135,14 +135,14 @@ def _as_python_types(metadata_map, headers):
     return values
 
 
-def _prefix_sample_names_with_id(md_template, study):
+def _prefix_sample_names_with_id(md_template, study_id):
     r"""prefix the sample_names in md_template with the study id
 
     Parameters
     ----------
     md_template : DataFrame
         The metadata template to modify
-    study : Study
+    study_id : int
         The study to which the metadata belongs to
     """
     # Get all the prefixes of the index, defined as any string before a '.'
@@ -150,14 +150,14 @@ def _prefix_sample_names_with_id(md_template, study):
     # If the samples have been already prefixed with the study id, the prefixes
     # set will contain only one element and it will be the str representation
     # of the study id
-    if len(prefixes) == 1 and prefixes.pop() == str(study.id):
+    if len(prefixes) == 1 and prefixes.pop() == str(study_id):
         # The samples were already prefixed with the study id
         warnings.warn("Sample names were already prefixed with the study id.",
                       QiitaDBWarning)
     else:
         # Create a new pandas series in which all the values are the study_id
         # and it is indexed as the metadata template
-        study_ids = pd.Series([str(study.id)] * len(md_template.index),
+        study_ids = pd.Series([str(study_id)] * len(md_template.index),
                               index=md_template.index)
         # Create a new column on the metadata template that includes the
         # metadata template indexes prefixed with the study id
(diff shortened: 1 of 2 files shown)


squirrelo:
Ready for review

qiita_db/metadata_template.py
```diff
@@ -135,14 +135,14 @@ def _as_python_types(metadata_map, headers):
     return values
 
 
-def _prefix_sample_names_with_id(md_template, study):
+def _prefix_sample_names_with_id(md_template, study_id):
```
antgonza: Here and other instance below: why the change? what's the benefit?

qiita_db/metadata_template.py
```diff
@@ -1324,6 +1355,95 @@ def study_id(self):
...
+        """
+        conn_handler = SQLConnectionHandler()
+        queue_name = "EXTEND_SAMPLE_TEMPLATE_%d" % self.id
+        conn_handler.create_queue(queue_name)
+
+        md_template = self._clean_validate_template(md_template, self.study_id,
+                                                    conn_handler)
+
+        # Get some useful information from the metadata template
+        sample_ids = md_template.index.tolist()
```
antgonza: Should this be sets? Better performance for big cases

antgonza:
What's the plan for updating existing samples and extending the number of metadata columns? Wondering cause the method extend can be used for the latter and this is not possible.

qiita_db/metadata_template.py
squirrelo: This is a copy paste of what was originally there for validation, so it's probably fine. If you look a few lines down it's used for the intersection with curr_samples, so it being a set or list shouldn't make a difference on performance.

qiita_db/metadata_template.py
squirrelo: I don't have the study object when extending, just the study ID. Therefore I can't do study.id and instead just send the ID directly.

qiita_db/metadata_template.py
squirrelo: Also it needs to retain it's order for insertion, since the datatypes list is assuming the sample in position X has the datatype in position X.

squirrelo:
Right now the code adds the new fields to the `sample_X` table and they are null for all existing samples. Conversely, if a new sample extends without a non-required column, it will have that field null. There are DB level functions to edit metadata on a per-column basis, so I was going to leverage that in the UI to update those on user input.

qiita_db/metadata_template.py
josenavas: I disagree with this change. Leave is as it was and provide the study. You state that you don't have the study when extending just the id. Make the caller to pass the study...

squirrelo: What happens if they pass the wrong study for the existing prep template? This way you have automatic data integrity since the study id comes from the object itself.

squirrelo: Talked with @josenavas and we agreed that it should stay with the ID, not the study object itself.

squirrelo:
Can his get final review/merge?

qiita_db/test/test_metadata_template.py
```diff
@@ -1214,6 +1214,77 @@ def test_get_filepath(self):
...
+               ['1.SKB5.640181'], ['1.SKB6.640176'], ['1.SKM2.640199'],
+               ['1.SKM5.640177'], ['1.SKB1.640202'], ['1.SKD8.640184'],
+               ['1.SKD4.640185'], ['1.SKB3.640195'], ['1.SKM1.640183'],
+               ['1.SKB7.640196'], ['1.SKD3.640198'], ['1.SKD7.640191'],
+               ['1.SKD6.640190'], ['1.SKB2.640194'], ['1.SKM9.640192'],
+               ['1.SKM6.640187'], ['1.SKD5.640186'], ['1.SKD1.640179'],
+               ['1.Sample1'], ['1.Sample2'], ['1.Sample3']]
+        self.assertEqual(obs, exp)
+
+        # test new columns were added to *_cols table and dynamic table
```
josenavas: It doesn't look like you have added new columns when you added samples.

qiita_db/test/test_metadata_template.py
```diff
@@ -1214,6 +1214,77 @@ def test_get_filepath(self):
...
+               [1, 'DEPTH', 'float8'], [1, 'TAXON_ID', 'varchar'],
+               [1, 'TEXTURE', 'varchar'],
+               [1, 'ASSIGNED_FROM_GEO', 'varchar'],
+               [1, 'SEASON_ENVIRONMENT', 'varchar'],
+               [1, 'sample_id', 'varchar']]
+        self.assertEqual(obs, exp)
+
+    def test_extend_duplicated_samples(self):
+        with self.assertRaises(QiitaDBDuplicateError):
+            self.tester.extend(self.metadata)
```
josenavas: Why is the extend 2 times here? Is it needeD?

josenavas:
@squirrelo 2 comments

qiita_db/test/test_metadata_template.py
squirrelo: `newcol` and `str_column` at the end.

qiita_db/test/test_metadata_template.py
squirrelo: Once to add the samples, then again to add the same samples and trigger the error.

qiita_db/test/test_metadata_template.py
josenavas: oh thanks, since you didn't modify `self.metadata` to add new columns I was confused.

qiita_db/test/test_metadata_template.py
josenavas: Can you then put the first one outside of the assert raises? This way we are sure that the one that is raising the error is the second one. Thanks!

squirrelo: Done.

qiita_db/test/test_metadata_template.py
squirrelo: https://github.com/biocore/qiita/pull/876/files#diff-71873eee34c39768872ef71a74452029R1218

squirrelo:
Ready for merge.

antgonza:
:+1: @josenavas could you merge if your 2 comments have been addressed?

PR #878, opened 2015-02-21
BUG: Prevent pandas from inferring data types

### LINKED ISSUE #872: Don't assume data types for required columns
`load_template_to_dataframe` should not assume the data types of the required columns.

ElDeveloper:
This was causing some required columns were being casted into unusable types.
For example if the values in HOST_SUBJECT_ID were all int-looking
(00011, 0011.2, etc), these were being casted into integers which later caused
troubles with the pipeline.

We now restrict pandas from casting certain columns into undesirable types and
enforce other types where possible.

Fixes #872

### DIFF
--- a/qiita_db/metadata_template.py
+++ b/qiita_db/metadata_template.py
@@ -119,7 +119,7 @@ def _as_python_types(metadata_map, headers):
     Returns
     -------
     list of lists
-        The values of the columns in metadata_map pointed by headers casted to
+        The values of the columns in metadata_map pointed by headers cast to
         python types.
     """
     values = []
@@ -1936,7 +1936,7 @@ def load_template_to_dataframe(fn):
     ------
     QiitaDBColumnError
         If the sample_name column is not present in the template.
-        If there's a value in one of the reserved columns that cannot be casted
+        If there's a value in one of the reserved columns that cannot be cast
         to the needed type.
     QiitaDBWarning
         When columns are dropped because they have no content for any sample.
@@ -1944,7 +1944,7 @@ def load_template_to_dataframe(fn):
     Notes
     -----
     The index attribute of the DataFrame will be forced to be 'sample_name'
-    and will be casted to a string. Additionally rows that start with a '\t'
+    and will be cast to a string. Additionally rows that start with a '\t'
     character will be ignored and columns that are empty will be removed. Empty
     sample names will be removed from the DataFrame.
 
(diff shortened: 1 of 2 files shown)


antgonza:
:+1: 

Could you add a section about metadata and have the fields that are required, their description, some examples and what's the preferred default for these in [here](https://docs.google.com/document/d/1u7kwLP31NM513-8xwpwvLbSQxYu0ehI6Jau1APR13e0/edit#)

qiita_db/metadata_template.py
```diff
@@ -1965,7 +1993,29 @@ def load_template_to_dataframe(fn):
                            keep_default_na=False, na_values=[''],
                            parse_dates=True, index_col=False, comment='\t',
                            mangle_dupe_cols=False, converters={
-                               'sample_name': lambda x: str(x).strip()})
```
josenavas: Instead of hardcoding these types, can we get them from the database? See [here](http://stackoverflow.com/questions/2146705/select-datatype-of-the-field-in-postgres). This way if something on the db changes, we will not need to update this code.

qiita_db/metadata_template.py
```diff
@@ -1936,6 +1936,8 @@ def load_template_to_dataframe(fn):
     ------
     QiitaDBColumnError
         If the sample_name column is not present in the template.
+        If there's a value in one of the reserved columns that cannot be casted
```
adamrp: casted -> cast

adamrp:
Code looks good to me, thanks @ElDeveloper

qiita_db/metadata_template.py
ElDeveloper: I wanted to do something like that but I ended up opting for this solution. The reason being that pandas will try to infer data types (which is useful) unless you want to keep the string as is. If you see the types for latitude/longitude are not enforced in this dictionary, instead I let pandas infer those types and if those are not inferred as floats, we raise an error.

Since this is a two-step process I don't think there's a good way to automate this.

josenavas: Thanks for the explanation!

josenavas:
:+1:

ElDeveloper:
Good idea, I've added that section to the Google doc you linked to.

qiita_db/metadata_template.py
ElDeveloper: No problem!

ElDeveloper:
This PR is ready to be merged.

PR #881, opened 2015-02-23
Issue 551

squirrelo:
Dead simple pull request to use scikit bio flatten instead of homebrew now that the required version is high enough.


ElDeveloper:
... actually we should change our codebase to rely on our home-brewed version as this function will be deprecated in the near future. See this issue: https://github.com/biocore/scikit-bio/pull/858

josenavas:
shall we keep using flatten or as @gregcaporaso suggest is a best practice to use the pythonic one liner across the code?

ElDeveloper:
I would actually advocate for the connection handler to be able to
return "flattened" results. This way we could shortcut the pattern where
we retrieve the results from a select and we then extract the first
element.

josenavas:
Ok, it will need a bit of magic on the connection handler but is doable. Can you add an issue on glowing dangerzone? I'm working on other more important issues in Qiita right now, but once those are done I will try to get qiita using glowing-dangerzone

ElDeveloper:
Sounds good: https://github.com/biocore/glowing-dangerzone/issues/4

In the meantime, maybe this PR should be closed.

adamrp:
Yes I think this PR should be closed. Having GD be able to flatten is a
good idea. Are there other places (besdies SQL handler results) where we
would need to flatten? If so, I think we should have a flatten method,
since it would actually be easier/require fewer changes across the code
base (we would just have to change imports).

ElDeveloper:
If there are other places where we are using flatten or an equivalent, I
would support the addition of that convenience function!

gregcaporaso:
Note that skbio's flatten isn't going away until 0.4.0, and qiita depends on skbio < 0.3.0, so dealing with this change likely isn't urgent.

squirrelo:
This is the only place flatten is used. Closing now.

Issue #855, opened 2015-02-13
shared status shouldn't be by study but by processed data


antgonza:
This issue was initially discussed in https://github.com/biocore/qiita/pull/850 but we realized that this will require bigger modifications to actually work well.

Description:
Let's imagine that a study has 2 raw datas (A/B) and 2 prep templates (1/2) in each one (for clarity, we will have A1, A2, B1, and B2). Then the users makes public the study because A1 was published, then all the other raw datas + prep templates will also be public, which is not OK. What we will need to have is that the user makes public a process data, then all the files that generated that "biom" become public. We will display public studies based on their public process data.

josenavas:
The proposed changes will be to add the "status" property in a per-object basis, and the study one can automatically populated (or removed) by looking at the status of its objects.

Pros:
- Once the Metadata has been accepted, the user can make public any analysis, (pre)process data that has been already published while he is working in other unpublished work within the same study.
- We are going to avoid study duplication: in case that a user has some published data but he is working a follow-up unpublished work, he can still work in the same study.

Cons:
- HORRIBLE database/codebase changes.

Even the amount of work behind this issue is considerable. I would suggest that this issue is really important and should be solved ASAP, before the code base continues growing...

ElDeveloper:
> The proposed changes will be to add the "status" property in a per-object basis, and the study one can automatically populated (or removed) by looking at the status of its objects.

:fire: :scream: :running: :skull:

How much more complex is the system going to become if we add this
feature? Are the benefits much larger than the amount of
work/modifications we need to do to accommodate this?

I'm not worried about the work or modifications, I'm worried about this
complicating the way users interact with their datasets.

antgonza:
I agree with @ElDeveloper, I think the users should only be able to
make public 1 thing: the processed data and the permissions should
carry down. Per object seems tooooo complicated, without any benefits,
at least that I can see clearly.

josenavas:
Ok. I agree with the solution that we want to do it per processed data (not by per prep template - as the title says).

The only reason why I suggested in a per object basis is in the case that the user used the system to upload some sequences and then submit those sequences to EBI. Then those sequences are publicly available on EBI, and I think it would be also useful to be publicly available on QIITA.

What do you think about this case?

antgonza:
Excellent point but we also need to check the map file so we can submit to
EBI and we do that by PCoAs ...

josenavas:
[...]

PR #882, opened 2015-02-23
Issue 855 patch

josenavas:
Adds the status attribute to the processed data object. Is the first step to start solving #855

### DIFF
--- a/qiita_db/support_files/patches/16.sql
+++ b/qiita_db/support_files/patches/16.sql
@@ -1,0 +1,12 @@
+-- Feb 23, 2015
+-- This patch adds the status to the processed data object
+-- We take advantage of the study status table as the values should be the same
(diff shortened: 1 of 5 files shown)


ElDeveloper:
@josenavas, there's a legitimate flake8 failure.

josenavas:
Thanks for the heads up @ElDeveloper !

I was testing the code directly in SQL and I did not execute the patch... Now it should be working and ready for review/merge!

josenavas:
After discussing with @antgonza looks like the best approach is to remove the study status from the db, and the property value will be inferred from the processed data contained in such study.

Given that the changes required due to the removal, it will be easier to close this PR and perform all the changes at once, rather than first introduce the patch and then modify the code.

ElDeveloper:
Thanks!

Issue #857, opened 2015-02-17
Implement an analysis collection object


squirrelo:
Similar to the idea of an investigation, which contains multiple studies, we need a collection object of some sort that contains many analysis objects. This will allow the many sub-analyses that go into one total analysis/paper to be easily referenced and found during final use for figure creation and paper writing. 

It would also be useful for this to have an "important" button for the analysis results in a collection, allowing you to highlight specific important results across the analyses in a collection object. For example, if you have 7 analyses in the object, and the beta diversity from analysis 3 is particularly important, you can highlight it like in gmail.

josenavas:
That's confusing to me. I though that we will able to add more jobs to a given analyses once it is done. I'm missing something?

squirrelo:
My understanding is that once a job is done running it's set in stone, but we can branch it using the analysis chain table. These two would then go into the collections object. There's also the case of subsetting your data and running another analysis, which would also end up as two separate analyses in our current setup, but still be in one overarching project. This just makes it easier for users to bookkeep their projects as they run more exploratory analyses.

josenavas:
Thanks. Agree re: subsetting. Disagree on setting it in stone. If the data is the same, we should be able to add a job. Eg: if we have a meta-analysis with taxa summaries and we want to run beta diversity, we should be able to add that job to the meta-analysis rather than branching.

squirrelo:
That makes sense in terms of not needing branching. So both setups need to be applied to the UI in this case: Adding more jobs to an analysis and having a collections object.

josenavas:
Yes. In terms of expanding the discussion. Which will be the workflow? The user creates the analysis group and then the analysis or you create the analysis as it is currently and then you can later add them to an analysis group? I like the later...

adamrp:
I like the latter as well.

squirrelo:
Both would be best. You can either create some an analyses, realize you need a collection object, then create one and add them to it. Alternately you have already done several analyses and created a collection for them, so when you create another analysis there should be an easy way to automatically add that new analysis to the existing collection.

josenavas:
Ok, but the creation of the analysis group should not be required. I.e. when you are creating an analysis, you should be able to choose an analysis group, but not required. This will also require changes on the analysis list page, so we have a list of available analysis groups and analysis themselves.

squirrelo:
Yup, agree with all of that.

antgonza:
Collection was completely deprecated in the last analysis refactor, thus closing.

PR #885, opened 2015-02-25
DB CHANGES: Analysis collection object

squirrelo:
This adds 5 new tables for the collections object. One tracks the main info, one tracks status, one allows sharing, one tracks analyses added to the collection, and one tracks jobs highlighted as important to the collection.

### DIFF
--- a/qiita_db/analysis.py
+++ b/qiita_db/analysis.py
@@ -752,3 +752,116 @@ class Analysis(QiitaStatusObject):
         sql = ("INSERT INTO qiita.analysis_filepath (analysis_id, filepath_id"
                "{0}) VALUES (%s, %s{1})".format(col, dtid))
         conn_handler.execute(sql, (self._id, fpid))
+
+
+class Collection(QiitaStatusObject):
+    """
+    Analysis overview object to track a multi-analysis collection.
+
+    Attributes
+    ----------
+    name: str
+        Name of the Collection
+    description: str
+        Description of what the collection is investigating
+    owner: User object
+        Owner of the Collection
+    analyses: list of Analysis Objects
+        all analyses that are part of the collection
+    highlights : list of Job objects
+        Important job results related to the collection
+
+    Methods
+    -------
+    add_analysis
+    remove_analysis
+    highlight_job
+    remove_highlight
+    """
+    _table = "collection"
+
+    @classmethod
+    def create(cls, owner, name, description=None):
+        """Creates a new collection on the database
+
+        Parameters
+        ----------
+        owner : User object
+            Owner of the collection
+        name : str
+            Name of the collection
+        description : str, optional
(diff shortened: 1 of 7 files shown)


squirrelo:
DB changes for #857

squirrelo:
Travs is apparently having a bad day and not updating github with final status, so I'll manually say that all tests passed and this is good to review.

josenavas:
I just take a quick look. Why do you need the collection-job table? You should have `collection<->analysis<->job` so the `collection<->job` is redundant...

squirrelo:
collecton<->job is so you can highlight specific jobs as important to the analysis, e.g. the alpha diversity in analysis 1 is a figure I want in a paper, and the beta diversity from analysis 2. This can then all be dropped on a single page for overview.

josenavas:
uhmmm... ok I'm still wrapping this on my mind... Probably better to change the name of the table (e.g. important jobs) and put collection_id, analysis_id and job_id so we can ensure some integrity with the data (you can potentially try to match a job that does not belong to that colleciton)

squirrelo:
OK, can do.

squirrelo:
After trying to find a way to do this integrity check with foreign keys, it looks like we're going to have to do that integrity check with python. Altering the table to have analysis_id would just be redundant with the collection_analysis table data in that case.

josenavas:
We have not been using triggers so far on the project... what other people think about creating integrity checkers using triggers? http://www.postgresql.org/docs/9.1/static/sql-createtrigger.html

squirrelo:
So reading this, it would be
```SQL
CREATE TRIGGER insert_collection_job
    BEFORE INSERT ON collection_job
    FOR EACH ROW
    EXECUTE PROCEDURE check_collection_access();
```
With `check_collection_access()` being a check to make sure the analysis the job is in is part of the collection.

squirrelo:
Do we have a way of putting procedures in the database using patches currently?

wasade:
Patches execute arbitrary sql, so yes. However, this did not seem to work properly for AG web. Not sure if its due to schemas or not.

josenavas:
@squirrelo which is the status of this? I need to modify the database too, and I don't want to throw away all your code. However, I need an ETA, as the changes that I need to do they have some kind of time pressure

squirrelo:
Triggers in postgres are tricky, but  I should have it in by end of day.

josenavas:
Thanks!!

squirrelo:
This is good to go.

josenavas:
@squirrelo can you add a test to check that the actual trigger works?

squirrelo:
They're already in there. `test_sql.py`

josenavas:
Ooops ok! For some reason it was not showing (or I did not see it...) :+1:

PR #888, opened 2015-03-02
ipython 2.4.1

antgonza:


antgonza:
Ready for review.

squirrelo:
:+1:

wasade:
2.4.1 <= x < 2.5

Would be better, unless we think we will go to 3.0 on very short order

squirrelo:
My understanding is we're going to be on 3.0 as fast as we can, and this is a stopgap.

antgonza:
@squirrelo I already made @wasade suggestion and the test pass. I think is fine to validate that range and simply change to 3.0 once our code supports it.

squirrelo:
Works for me.

PR #891, opened 2015-03-03
Move generic navbar links to far left

adamrp:
Also some minor modifications to the navbar links.

### DIFF
--- a/qiita_pet/templates/sitebase.html
+++ b/qiita_pet/templates/sitebase.html
@@ -138,7 +138,7 @@
             </li>
               {% elif user_level == 'dev' %}
             <li class="dropdown">
-              Dev<b class="caret"></b>
+              <a href="#" data-toggle="dropdown" class="dropdown-toggle">Dev<b class="caret"></b></a>
               <ul class="dropdown-menu">
                 <li><a href="/admin/error/">View Errors</a></li>
               </ul>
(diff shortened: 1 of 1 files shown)


squirrelo:
:+1:

ElDeveloper:
... the PR looks fine and the code works as expected, however I don't like that the first option you get when you are logged in is Help, Stats, GitHub and finally the actually useful things like Meta Analysis, Study and Admin. :-1: to this change.

antgonza:
I think  the best will be to have 3 sections: user menu, login, and other.
- menu: create account, reset forgotten passwd, when not log in;
metaanalysis, study, etc once you are log in
- login: login text and button when not log in; email with account
update when log in
- other: stats, help, etc.

what do you guys think?

ElDeveloper:
I like the organization as is, but perhaps the only group that we should
create is "Other", here's where Stats, Help, and GitHub could live.

squirrelo:
I'd rename "Other" to "Info" or "Help" but :+1: to @ElDeveloper idea of leaving as is but condensing the three others.

antgonza:
Note that the sections do not need to be labeled, is just grouping so
the last (other) stays static to the right.

adamrp:
There is no separator to create sections, and bootstrap will resize things based on screen real estate, so I don't think it's feasible to create 3 different sections that will dependably display consistently.

@ElDeveloper by a "separate group" for Stats, Help, and Github, do you mean another dropdown?

ElDeveloper:
Yes, that's what I mean. Sorry for not being clear.

adamrp:
Is this better?

ElDeveloper:
:+1: I like this a lot more!

PR #892, opened 2015-03-03
Add extend functionality to UI

squirrelo:
Adds ability to upload a second sample template to the UI and add it to the existing study sample template.

### DIFF
--- a/qiita_db/metadata_template.py
+++ b/qiita_db/metadata_template.py
@@ -1376,13 +1376,17 @@ class SampleTemplate(MetadataTemplate):
         num_samples = len(sample_ids)
         headers = list(md_template.keys())
 
-        # Make sure all samples being added are not already existing
+        # Raise warning and filter out existing samples
         sql = ("SELECT sample_id FROM qiita.required_sample_info WHERE "
                "study_id = %d" % self.id)
         curr_samples = set(s[0] for s in conn_handler.execute_fetchall(sql))
-        if len(curr_samples.intersection(sample_ids)) > 0:
-            raise QiitaDBDuplicateError(
-                "sample_id", ", ".join(curr_samples.intersection(sample_ids)))
+        existing_samples = curr_samples.intersection(sample_ids)
+        if len(existing_samples) > 0:
+            raise QiitaDBWarning(
+                "The following samples already exist and will be ignored: "
+                "%s" % ", ".join(curr_samples.intersection(
+                                 sorted(existing_samples))))
+            md_template.drop(existing_samples)
 
         # Get the required columns from the DB
         db_cols = get_table_cols(self._table, conn_handler)
--- a/qiita_db/metadata_template.py
+++ b/qiita_db/metadata_template.py
@@ -1371,22 +1371,23 @@ class SampleTemplate(MetadataTemplate):
         md_template = self._clean_validate_template(md_template, self.study_id,
                                                     conn_handler)
 
-        # Get some useful information from the metadata template
-        sample_ids = md_template.index.tolist()
-        num_samples = len(sample_ids)
-        headers = list(md_template.keys())
-
         # Raise warning and filter out existing samples
+        sample_ids = md_template.index.tolist()
         sql = ("SELECT sample_id FROM qiita.required_sample_info WHERE "
                "study_id = %d" % self.id)
         curr_samples = set(s[0] for s in conn_handler.execute_fetchall(sql))
         existing_samples = curr_samples.intersection(sample_ids)
-        if len(existing_samples) > 0:
-            raise QiitaDBWarning(
+        if existing_samples:
+            warnings.warn(
                 "The following samples already exist and will be ignored: "
                 "%s" % ", ".join(curr_samples.intersection(
-                                 sorted(existing_samples))))
-            md_template.drop(existing_samples)
+                                 sorted(existing_samples))), QiitaDBWarning)
+            md_template.drop(existing_samples, inplace=True)
+
+        # Get some useful information from the metadata template
+        sample_ids = md_template.index.tolist()
+        num_samples = len(sample_ids)
+        headers = list(md_template.keys())
 
         # Get the required columns from the DB
         db_cols = get_table_cols(self._table, conn_handler)
--- a/qiita_pet/handlers/study_handlers/description_handlers.py
+++ b/qiita_pet/handlers/study_handlers/description_handlers.py
@@ -168,14 +168,11 @@ class StudyDescriptionHandler(BaseHandler):
         fp_rsp = join(base_fp, str(study.id), sample_template)
 
         if not exists(fp_rsp):
-            # The file does not exists, fail nicely
-            raise HTTPError(400, "This file doesn't exist: %s" % fp_rsp)
+            # The file does not exist, fail nicely
+            raise HTTPError(404, "This file doesn't exist: %s" % fp_rsp)
 
         try:
             with warnings.catch_warnings(record=True) as warns:
-                # force all warnings to always be triggered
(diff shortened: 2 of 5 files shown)


squirrelo:
Final thing needed for adding samples (Part of #822)

antgonza:
Can you 
- [ ] add some text with the rules to extend the sample template, like: the new samples should have the same metadata than the old one, only new samples are allowed.
- [ ] add a nicer message when columns do not overlap with the existing sample template. The message should contain all columns that do not match. Currently the error is just the SQL error.

squirrelo:
First one is made moot by the raising of a warning instead of an error for duplicated samples and auto-filtering, which we agreed on in the meeting today.

Can you explain the second one? The extend code that was merged auto-adds any column it finds that wasn't in the first sample template.

squirrelo:
Now raises warning instead of error. Ready for review.

qiita_db/metadata_template.py
```diff
@@ -1376,13 +1376,17 @@ def extend(self, md_template):
...
+        # Raise warning and filter out existing samples
         sql = ("SELECT sample_id FROM qiita.required_sample_info WHERE "
                "study_id = %d" % self.id)
         curr_samples = set(s[0] for s in conn_handler.execute_fetchall(sql))
-        if len(curr_samples.intersection(sample_ids)) > 0:
-            raise QiitaDBDuplicateError(
-                "sample_id", ", ".join(curr_samples.intersection(sample_ids)))
+        existing_samples = curr_samples.intersection(sample_ids)
+        if len(existing_samples) > 0:
+            raise QiitaDBWarning(
```
ElDeveloper: Can you use warnings.warn instead of raise?

qiita_pet/handlers/study_handlers/description_handlers.py
```diff
@@ -197,6 +197,66 @@ def process_sample_template(self, study, user, callback):
...
+        fp_rsp = join(base_fp, str(study.id), sample_template)
+
+        if not exists(fp_rsp):
+            # The file does not exists, fail nicely
+            raise HTTPError(400, "This file doesn't exist: %s" % fp_rsp)
+
+        try:
+            with warnings.catch_warnings(record=True) as warns:
+                # force all warnings to always be triggered
+                warnings.simplefilter("always")
```
ElDeveloper: If I recall correctly, with the changes that @josenavas made, we shouldn't need this anymore.

squirrelo: Which changes? Want to be sure before removing.

ElDeveloper: These: https://github.com/biocore/qiita/blob/master/qiita_db/exceptions.py#L78-82

qiita_db/metadata_template.py
squirrelo: That's exactly what QiitaDBWarning does, just with a qiita-specific warning wrapper. What do we gain using warning directly?

ElDeveloper: Not really, one of them raises an exception and the other one warns:

```python

In [1]: from qiita_db.exceptions import QiitaDBWarning

In [2]: def raising():
   ...:     raise QiitaDBWarning
   ...: 

In [3]: import warnings

In [4]: def warning():
   ...:     warnings.warn('lol', QiitaDBWarning)
   ...:     

In [8]: warning()
/Users/yoshikivazquezbaeza/.virtualenvs/qiita/bin/ipython:2: QiitaDBWarning: lol
  

In [9]: raising()
QiitaDBWarning                            Traceback (most recent call last)
<ipython-input-9-929624e4202b> in <module>()
----> 1 raising()

<ipython-input-2-027989dcbf59> in raising()
      1 def raising():
----> 2     raise QiitaDBWarning
      3 

QiitaDBWarning: 
```

qiita_pet/handlers/study_handlers/description_handlers.py
```diff
@@ -197,6 +197,66 @@ def process_sample_template(self, study, user, callback):
...
+               "added" % sample_template)
+        msg_level = "success"
+        # Get the uploads folder
+        _, base_fp = get_mountpoint("uploads")[0]
+        # Get the path of the sample template in the uploads folder
+        fp_rsp = join(base_fp, str(study.id), sample_template)
+
+        if not exists(fp_rsp):
+            # The file does not exists, fail nicely
+            raise HTTPError(400, "This file doesn't exist: %s" % fp_rsp)
```
ElDeveloper: Testing the GUI, I got to this error, and while it is accurate, as a user who is not an admin, this information is relevant but not presented anywhere. Is there any easy way to fix this?

ElDeveloper: Thinking more about it, this shouldn't be an HTTPError, it should be a different class.

qiita_pet/templates/study_description_templates/sample_template_tab.html
```diff
@@ -9,7 +9,7 @@
       {% end %}
     </select>
     <br/><br/>
-    <a class="btn btn-primary" onclick="process_sample_template();">Process sample template</a>
+    <a class="btn btn-primary" onclick="process_sample_template();">Process sample template</a> <a class="btn btn-primary" onclick="extend_sample_template();">Add samples to study</a>
```
ElDeveloper: I just realized that if the dropdown menu has no files and you click any of these two buttons you'll get an error, while it is not problem introduced by this PR we certainly have to fix this for the buttons to be disabled unless there's a file selected.

qiita_db/metadata_template.py
```diff
@@ -1376,13 +1376,17 @@ def extend(self, md_template):
...
-        # Make sure all samples being added are not already existing
+        # Raise warning and filter out existing samples
         sql = ("SELECT sample_id FROM qiita.required_sample_info WHERE "
                "study_id = %d" % self.id)
         curr_samples = set(s[0] for s in conn_handler.execute_fetchall(sql))
-        if len(curr_samples.intersection(sample_ids)) > 0:
-            raise QiitaDBDuplicateError(
-                "sample_id", ", ".join(curr_samples.intersection(sample_ids)))
+        existing_samples = curr_samples.intersection(sample_ids)
+        if len(existing_samples) > 0:
```
ElDeveloper: The greater than zero comparison is not needed.

qiita_db/metadata_template.py
squirrelo: That's....incredibly odd. Ok. will fix.

ElDeveloper: Thanks!

qiita_pet/handlers/study_handlers/description_handlers.py
squirrelo: Yup, 404. Fixed.

qiita_pet/handlers/study_handlers/description_handlers.py
squirrelo: Cool. removed.

qiita_pet/templates/study_description_templates/sample_template_tab.html
squirrelo: I can do that for all buttons in a separate pull request. Can you open an issue?

ElDeveloper: You got it: https://github.com/biocore/qiita/issues/903

squirrelo:
This should be good to go.

ElDeveloper:
:+1:

ElDeveloper:
If someone else reviews this PR it should be ready to be merged.

PR #899, opened 2015-03-04
Regenerate html for DB Schema. Fix #897

### LINKED ISSUE #897: latest version of qiita-db.html is wrong
http://htmlpreview.github.io/?https://github.com/biocore/qiita/blob/master/qiita_db/support_files/qiita-db.html

squirrelo:
Does what it says on the tin.

Preview it [here](http://htmlpreview.github.io/?https://github.com/squirrelo/qiita/blob/regen-html-for-db/qiita_db/support_files/qiita-db.html)


antgonza:
:+1:

Issue #292, opened 2014-09-16
Convert status tables to have same convention as other constraint tables


squirrelo:
Currently the status tables do not follow the same convention as the filepath_type and other constraint tables. This means we can't use the convert_to_id and convert_from_id functions on the status tables. Needs to be fixed so we can use these functions.

squirrelo:
We agree this needs to happen, as soon as the 3 main issues currently on the table are resolved.

antgonza:
Closing in favor of: https://github.com/biocore/qiita/issues/410

PR #900, opened 2015-03-04
Issue 855

### LINKED ISSUE #855: shared status shouldn't be by study but by processed data

josenavas:
:feelsgood::feelsgood::feelsgood::feelsgood:

Fixes #855 

I've moved the status column from the study object to the processed data and now the status of the study is inferred by the status of its processed data.

I've all the tests passing (at least on my machine) but I don't think the functionality is currently consistent!

The issue is that the metadata pipeline will show all the processed data for a given study, even if a given processed data is not public... Since this PR is making a deep change in the db structure and has impacts in a lot of parts of the codebase; I think the most secure way of moving this forward is making a PR so people can download the PR and check different contexts, in case that I missed something.

Missing things (that I can think of - please add as needed):
- [x] Modify interface so the `make public / revert to sandbox / etc` buttons are in per processed data basis.
- [ ] The meta-analysis pipeline should only show the processed data that is public or that the user is the owner (see https://github.com/biocore/qiita/issues/1005)
- [x] If accessing a public study, the study description page should only show the public raw and (pre)processed data from the public processed data.
- [x] The studies awaiting approval page for admins should show which processed data is awaiting approval.


wasade:
@josenavas, conflicts

qiita_db/study.py
```diff
@@ -170,10 +194,23 @@ def get_by_status(cls, status):
...
-               "s.study_status_id = ss.study_status_id WHERE "
-               "ss.status = %s".format(cls._table))
-        return [x[0] for x in conn_handler.execute_fetchall(sql, (status, ))]
+        sql = """SELECT study_id FROM qiita.study_processed_data spd
+                JOIN qiita.processed_data pd
+                    ON spd.processed_data_id=pd.processed_data_id
+                JOIN qiita.processed_data_status pds
+                    ON pds.processed_data_status_id=pd.processed_data_status_id
+                WHERE pds.processed_data_status=%s"""
+        studies = [x[0] for x in
```
wasade: can this be a `set`? See #961

josenavas:
Note that there is still one item in the list not implemented, but that is something that cannot be implemented right now. The system still does not allow to select individual processed data for meta-analysis, so I cannot make that change. Also, that change requires multiple changes in the search engine, which will be refactored soon, once all DB changes that we are planning to do are in place. So, from my point of view, it is not worth to put more duct tape and developer effort in a code that is going to be removed. That being said, this is ready to go.

josenavas:
Added a new issue to keep track of the missing point: https://github.com/biocore/qiita/issues/1005

josenavas:
Finally I got tests passing again! Ready for review. Note that this PR is doing database changes, so it is blocking any further development on that direction

qiita_db/data.py
```diff
@@ -615,6 +615,68 @@ def remove_filepath(self, fp):
...
+        The status of the raw data is inferred by the status of the processed
+        data generated from this raw data. Since the raw data can be shared by
+        multiple studies; the raw data can have multiple status, so the status
+        is defined by the view from a study. If not processed data was
+        generated on the given study from this raw data; then the status is
+        'sandbox'.
+        """
+        if self._id not in study.raw_data():
+            raise QiitaDBStatusError(
+                "The study %s do not have access to the raw data %s"
```
squirrelo: does not have access

qiita_db/data.py
```diff
@@ -615,6 +615,68 @@ def remove_filepath(self, fp):
...
+        conn_handler = SQLConnectionHandler()
+        sql = """SELECT processed_data_status
+                FROM qiita.processed_data_status pds
+                  JOIN qiita.processed_data pd
+                    ON pds.processed_data_status_id=pd.processed_data_status_id
+                  JOIN qiita.preprocessed_processed_data ppd_pd
+                    ON ppd_pd.processed_data_id=pd.processed_data_id
+                  JOIN qiita.prep_template_preprocessed_data pt_ppd
+                    ON pt_ppd.preprocessed_data_id=ppd_pd.preprocessed_data_id
+                  JOIN qiita.prep_template pt
```
squirrelo: Why the prep template joins in here? If we just need to get up to the processed data, using raw_data, study_raw_data, and study_preprocessed_data would be easier and faster than prep template.

qiita_db/data.py
```diff
@@ -615,6 +615,68 @@ def remove_filepath(self, fp):
...
+                    ON srd.raw_data_id=pt.raw_data_id
+                WHERE pt.raw_data_id=%s AND srd.study_id=%s"""
+        pd_statuses = conn_handler.execute_fetchall(sql, (self._id, study.id))
+
+        if not pd_statuses:
+            # If there are no processed data, then the status is sandbox
+            status = 'sandbox'
+        else:
+            pd_statuses = set(s[0] for s in pd_statuses)
+            if 'public' in pd_statuses:
```
squirrelo: This seems like a bad idea, since you are going by the most open version of permissions. I'd go by least open for now, since in this case you could have a private and public pointing to this data, the public wins, and both datasets go out with the raw data, even though the private shouldn't have.

qiita_db/data.py
```diff
@@ -995,6 +1057,48 @@ def processing_status(self, state):
...
+                    ON ppd_pd.processed_data_id=pd.processed_data_id
+                WHERE ppd_pd.preprocessed_data_id=%s"""
+        pd_statuses = conn_handler.execute_fetchall(sql, (self._id,))
+
+        if not pd_statuses:
+            # If there are no processed data, then the status is sandbox
+            status = 'sandbox'
+        else:
+            pd_statuses = set(s[0] for s in pd_statuses)
+            if 'public' in pd_statuses:
```
squirrelo: Same comment as above.

qiita_db/data.py
```diff
@@ -1021,6 +1125,61 @@ class ProcessedData(BaseData):
     _preprocessed_processed_table = "preprocessed_processed_data"
 
     @classmethod
+    def get_by_status(cls, status):
```
squirrelo: Since we have this for processed data, we should have it for preprocessed and raw data as well to keep things consistent. Anything that has a status really should have this classmethod.

qiita_db/data.py
```diff
@@ -1021,6 +1125,61 @@ class ProcessedData(BaseData):
...
+        conn_handler = SQLConnectionHandler()
+        sql = """SELECT spd.study_id,
+            array_agg(pd.processed_data_id ORDER BY pd.processed_data_id)
+            FROM qiita.processed_data pd
+                JOIN qiita.processed_data_status pds
+                    ON pd.processed_data_status_id=pds.processed_data_status_id
+                JOIN qiita.study_processed_data spd
+                    ON spd.processed_data_id=pd.processed_data_id
+            WHERE pds.processed_data_status = %s
+            GROUP BY study_id;"""
```
squirrelo: `spd.study_id` just to make it explicit.

qiita_db/data.py
```diff
@@ -1173,3 +1332,39 @@ def processed_date(self):
         return conn_handler.execute_fetchone(
             "SELECT processed_date FROM qiita.{0} WHERE "
             "processed_data_id=%s".format(self._table), (self.id,))[0]
+
+    @property
+    def status(self):
+        conn_handler = SQLConnectionHandler()
+        sql = """SELECT processed_data_status
```
squirrelo: `pds.procesed_data_status` to keep it explicit.

qiita_db/data.py
```diff
@@ -1173,3 +1332,39 @@ def processed_date(self):
...
+        ------
+        QiitaDBStatusError
+            If the processed data status is public
+        """
+        if self.status == 'public':
+            raise QiitaDBStatusError(
+                "Illegal operation on public processed data")
+
+        conn_handler = SQLConnectionHandler()
+        sql = """UPDATE qiita.{0} SET processed_data_status_id = (
```
squirrelo: using `convert_to_id` would probably be better here. Also, why is this override needed? This function is built into the base class.

qiita_db/meta_util.py
```diff
@@ -81,9 +81,8 @@ def get_accessible_filepath_ids(user):
         return set(f[0] for f in fpids)
 
     # First, the studies
-    # There are public, private, and shared studies
-    study_ids = Study.get_by_status('public') | user.user_studies | \
-        user.shared_studies
+    # There are private and shared studies
+    study_ids = user.user_studies | user.shared_studies
```
squirrelo: Why remove the public from here?

squirrelo: nevermind, see it below.

qiita_db/meta_util.py
```diff
@@ -118,6 +116,40 @@ def get_accessible_filepath_ids(user):
                              in SampleTemplate(study_id).get_filepaths()]
             filepath_ids.update(sample_fp_ids)
 
+    # Next, the public processed data
+    processed_data_ids = ProcessedData.get_by_status('public')
+    for pd_id in processed_data_ids:
```
squirrelo: This seems amazingly expensive, especially if this is going to be used to check file access and needs to happen every download. Can you create a PrepTemplate(x).associated_files or some other way of getting these out with far less database calls.

qiita_db/metadata_template.py
```diff
@@ -2206,6 +2206,50 @@ def create_qiime_mapping_file(self, prep_template_fp):
...
+                    ON pt_ppd.preprocessed_data_id=ppd_pd.preprocessed_data_id
+                WHERE pt_ppd.prep_template_id=%s"""
+        pd_statuses = conn_handler.execute_fetchall(sql, (self._id,))
+
+        if not pd_statuses:
+            # If there are no processed data, then the status is sandbox
+            status = 'sandbox'
+        else:
+            pd_statuses = set(s[0] for s in pd_statuses)
+            if 'public' in pd_statuses:
```
squirrelo: Same as before about multiple status condensing.

qiita_db/study.py
```diff
@@ -142,18 +142,42 @@ class Study(QiitaStatusObject):
...
+                    ON spd.processed_data_id=pd.processed_data_id
+                WHERE spd.study_id = %s"""
+        pd_statuses = conn_handler.execute_fetchall(sql, (self._id,))
+
+        if not pd_statuses:
+            # If there are no processed data, then the status is sandbox
+            study_status = 'sandbox'
+        else:
+            pd_statuses = set(s[0] for s in pd_statuses)
+            if 'public' in pd_statuses:
```
squirrelo: If we are hiding the non-public procesed data this makes sense to have it in this order here.

qiita_db/support_files/patches/18.sql
```diff
@@ -1,3 +1,3 @@
 -- March 18, 2015
 -- Add column to analysis table to mark shopping cart
-ALTER TABLE qiita.analysis ADD dflt bool NOT NULL DEFAULT false;
\ No newline at end of file
+ALTER TABLE qiita.analysis ADD dflt bool NOT NULL DEFAULT false;
```
squirrelo: ???

qiita_db/support_files/patches/19.sql
```diff
@@ -0,0 +1,34 @@
...
+    FROM study_values sv
+    WHERE pd.processed_data_id IN (
+        SELECT processed_data_id FROM qiita.study_processed_data
+        WHERE study_id = sv.study_id);
+
+
+-- We no longer need the status column on the study table
+ALTER TABLE qiita.study DROP CONSTRAINT fk_study_study_status;
+DROP INDEX qiita.idx_study_0;
+ALTER TABLE qiita.study DROP COLUMN study_status_id;
```
squirrelo: If you're fixing missing newlines at EOF, this needs one too.

qiita_pet/handlers/study_handlers/description_handlers.py
```diff
@@ -107,7 +107,10 @@ def _get_study_and_check_access(self, study_id):
         else:
             check_access(user, study, raise_error=True)
 
-        return study, user
+        full_access = (study.id in user.user_studies | user.shared_studies or
+                       user.level == 'admin')
```
squirrelo: check for admin first and return, that way you don't need to do the user database calls if not needed.

squirrelo:
Really fast first pass done. Some comments.

qiita_db/data.py

qiita_db/data.py
josenavas: Note that I have to check the specific processed data that has been generated from this raw data, not any processed data. That path is only available through the prep template.

qiita_db/data.py
josenavas: That is not a bug, is a feature hahaha explanation:
If a processed data becomes available, anything that belongs to it, becomes available. Imagine that a single processed data exists and its public. If in a year from now I start another analysis changing some parameters, that processed data is not public, but I cannot revert the status to private. Thus, this has been thought carefully and this the correct path.

qiita_db/data.py
josenavas: Same answer as above :smile: I just noticed that I can move this function to util.py so there is less code duplication

qiita_db/data.py
josenavas: I've been thinking about this... I can do that, although that functionality has not been required for any part of the code right now; so I figured that it will be easier to open an issue for that and do not block more DB development with this PR: #1009

qiita_db/data.py

qiita_db/data.py

qiita_db/data.py
josenavas: Note that these objects do not extend the QiitaObjectStatus class, they extend the QiitaObject class

qiita_db/meta_util.py
josenavas: :smile:

qiita_db/meta_util.py
josenavas: I'm just following the pattern on this function. I already noticed that this function is quite bad and needs some refactoring, but I think it needs to be done outside of this PR. See https://github.com/biocore/qiita/issues/1010

qiita_db/metadata_template.py
josenavas: Same as before...

qiita_db/study.py
josenavas: All the status property follows the same structure

qiita_db/support_files/patches/18.sql
josenavas: No idea, probably while fixing merge conflicts. Will revert

qiita_db/support_files/patches/19.sql
josenavas: Nope, I was not! I don't think any sql has the new line so will leave as it is.

qiita_pet/handlers/study_handlers/description_handlers.py
josenavas: excellent suggestion. Done

josenavas:
Thanks @squirrelo comments addressed/answered

ElDeveloper:
@josenavas, can you pull from upstream/master?

josenavas:
Done

josenavas:
stochastic failure, re-starting the build. It would be good if I can receive the reviews here so we can merge this, as this is blocking development in any other database changes related work. @biocore/qiita-dev

qiita_db/data.py
```diff
@@ -615,6 +615,68 @@ def remove_filepath(self, fp):
...
+        generated on the given study from this raw data; then the status is
+        'sandbox'.
+        """
+        if self._id not in study.raw_data():
+            raise QiitaDBStatusError(
+                "The study %s does not have access to the raw data %s"
+                % (study.id, self.id))
+
+        conn_handler = SQLConnectionHandler()
+        sql = """SELECT processed_data_status
```
squirrelo: Not blocking, but these would be cleaner/easier to read if you used `USING` instead of `ON`

qiita_db/data.py
```diff
@@ -1173,3 +1332,39 @@ def processed_date(self):
...
+        conn_handler = SQLConnectionHandler()
+        sql = """SELECT pds.processed_data_status
+                FROM qiita.processed_data_status pds
+                  JOIN qiita.processed_data pd
+                    ON pd.processed_data_status_id=pds.processed_data_status_id
+                WHERE pd.processed_data_id=%s"""
+        return conn_handler.execute_fetchone(sql, (self._id,))[0]
+
+    @status.setter
+    def status(self, status):
```
squirrelo: There is no sanity check for the status, so when I tested it with status 'chicken' it errors out. Better to use `convert_to_id` and let that error. This goes for all the other status setters.

squirrelo: And by errors out I mean it says it would violate the NOT NULL constraint of the column.

qiita_db/support_files/patches/19.sql
```diff
@@ -0,0 +1,34 @@
+-- March 19, 2015
+-- Removes the status from the study and adds it to the processed data table.
+
+-- Modify the study_status_table so it becomes the processed_data_status table
+ALTER TABLE qiita.study_status RENAME TO processed_data_status;
+ALTER TABLE qiita.processed_data_status RENAME COLUMN study_status_id TO processed_data_status_id;
+ALTER TABLE qiita.processed_data_status RENAME COLUMN status TO processed_data_status;
```
squirrelo: Why do this? We have a setup where the text of the status is in column named `status` so we can easily use convert_to_id and convert_from_id.

qiita_db/test/test_data.py
```diff
@@ -868,6 +932,59 @@ def test_processed_date(self):
...
+        pd = ProcessedData.create(self.params_table, self.params_id,
+                                  self.filepaths,
+                                  preprocessed_data=self.preprocessed_data)
+        self.assertEqual(pd.status, 'sandbox')
+
+    def test_status_setter(self):
+        pd = ProcessedData(1)
+        self.assertEqual(pd.status, 'private')
+
+        pd.status = 'sandbox'
```
squirrelo: also need to test setting it to something not in the status table.

qiita_db/test/test_meta_util.py
```diff
@@ -17,7 +17,12 @@
 @qiita_test_checker()
 class MetaUtilTests(TestCase):
     def _set_studies_private(self):
```
squirrelo: _set_processed_data_private

qiita_db/test/test_meta_util.py
```diff
@@ -17,7 +17,12 @@
...
-        self.conn_handler.execute("UPDATE qiita.study SET study_status_id=3")
+        self.conn_handler.execute(
+            "UPDATE qiita.processed_data SET processed_data_status_id=3")
+
+    def _set_studies_public(self):
```
squirrelo: _set_processed_data_public

qiita_db/test/test_study.py
```diff
@@ -169,20 +170,21 @@ def setUp(self):
...
             'number_samples_collected': 27}
 
-    def _make_private(self):
-        # make studies private
-        self.conn_handler.execute("UPDATE qiita.study SET study_status_id = 4")
-
-    def _make_sandbox(self):
-        # make studies private
-        self.conn_handler.execute("UPDATE qiita.study SET study_status_id = 1")
+    def _change_studies_status(self, new_status):
```
squirrelo: _change_processed_data_status

qiita_db/test/test_study.py
```diff
@@ -169,20 +170,21 @@ def setUp(self):
...
+        # Change the status of the studies by changing the status of their
+        # processed data
+        id_status = convert_to_id(new_status, 'processed_data_status',
+                                  self.conn_handler)
+        self.conn_handler.execute(
```
squirrelo: why not use the processed data setter here?

squirrelo:
Few more comments

qiita_db/data.py
josenavas: Thanks! I changed all instances that I've added in this PR.

qiita_db/data.py
josenavas: Good catch. Fixed, note that this is the only status setter implemented on this PR (as all the other status are inferred from the processed data status)

qiita_db/support_files/patches/19.sql
josenavas: Sorry, we had mixed set ups and I thought that this one was the one that we wanted to standardize... Removing that line.

squirrelo: Problem is convert_to_id takes one format and convert_from_id takes the other. Do we want to use this pull request to standardize all this, since it already is about status objects? Issue #292 for reference, so it's been there a while.

qiita_db/data.py
adamrp: Nice, that's some nice shorthand to know! Thanks @squirrelo

qiita_db/support_files/patches/19.sql
josenavas: Since this is an open discussion and we have to agree in what are we going to standardize, I'm not going to change this and if it has to be fixed, it will be done as part of fixing the issue #292

qiita_db/test/test_data.py
josenavas: Done

qiita_db/test/test_meta_util.py
josenavas: Done

qiita_db/test/test_meta_util.py
josenavas: Done

qiita_db/test/test_study.py
josenavas: Done

qiita_db/test/test_study.py
josenavas: Cause I'm changing all processed_data statuses not a given one.

josenavas:
Ready for another review.

qiita_db/data.py
```diff
@@ -615,6 +616,54 @@ def remove_filepath(self, fp):
...
+        ----------
+        study : Study
+            The study that is looking to the raw data status
+
+        Returns
+        -------
+        str
+            The status of the raw data
+
+        Raises
```
ElDeveloper: You might want to remove this section as there are not specific exceptions raised in this method.

qiita_db/data.py
```diff
@@ -615,6 +616,54 @@ def remove_filepath(self, fp):
...
+        ------
+
+        Notes
+        -----
+        The status of the raw data is inferred by the status of the processed
```
ElDeveloper: It's a little hard to follow this paragraph, would you mind rephrasing it?

qiita_db/data.py
```diff
@@ -1021,6 +1098,61 @@ class ProcessedData(BaseData):
...
+
+        Parameters
+        ----------
+        status : str
+            Status to search for
+
+        Returns
+        -------
+        list of int
+            All the processed data id that match the given status
```
ElDeveloper: id -> ids

qiita_db/data.py
```diff
@@ -1021,6 +1098,61 @@ class ProcessedData(BaseData):
...
+        """
+        conn_handler = SQLConnectionHandler()
+        sql = """SELECT processed_data_id FROM qiita.processed_data pd
+                JOIN qiita.processed_data_status pds
+                    USING (processed_data_status_id)
+                WHERE pds.processed_data_status=%s"""
+        result = conn_handler.execute_fetchall(sql, (status,))
+        if result:
+            pds = [x[0] for x in result]
```
ElDeveloper: Minor, but should this be a set? I doubt we could be getting repeated identifiers from the query above, but maybe.

qiita_pet/util.py
```diff
@@ -26,6 +26,13 @@
 from qiita_db.reference import Reference
 
 
+status_styler = {
```
ElDeveloper: Minor, but I *think* this should be in all caps.

antgonza:
The code looks fine but couldn't test due to:
```
  File "/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/site-packages/tornado/web.py", line 1144, in _when_complete
  File "/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/site-packages/tornado/concurrent.py", line 129, in result
  File "/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/site-packages/tornado/gen.py", line 221, in wrapper
    runner.run()
  File "/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/site-packages/tornado/gen.py", line 505, in run
  File "/Users/antoniog/svn_programs/qiita/qiita_pet/handlers/study_handlers/listing_handlers.py", line 92, in get
    user_studies = yield Task(self._get_private, user)
  File "/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/site-packages/tornado/gen.py", line 533, in run
    self.yield_point.start(self)
  File "/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/site-packages/tornado/gen.py", line 371, in start
  File "/Users/antoniog/svn_programs/qiita/qiita_pet/handlers/study_handlers/listing_handlers.py", line 101, in _get_private
    callback(_build_study_info("private", user))
  File "/Users/antoniog/svn_programs/qiita/qiita_pet/handlers/study_handlers/listing_handlers.py", line 59, in _build_study_info
    status = study.status
  File "/Users/antoniog/svn_programs/qiita/qiita_db/study.py", line 164, in status
    pd_statuses = conn_handler.execute_fetchall(sql, (self._id,))
  File "/Users/antoniog/svn_programs/qiita/qiita_db/sql_connection.py", line 454, in execute_fetchall
    with self._sql_executor(sql, sql_args) as pgcursor:
  File "/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/contextlib.py", line 17, in __enter__
    return self.gen.next()
  File "/Users/antoniog/svn_programs/qiita/qiita_db/sql_connection.py", line 303, in _sql_executor
    (sql, str(sql_args), e)))
QiitaDBExecutionError: 
Error running SQL query: SELECT processed_data_status
                FROM qiita.processed_data_status pds
                  JOIN qiita.processed_data pd
                    USING (processed_data_status_id)
                  JOIN qiita.study_processed_data spd
                    USING (processed_data_id)
                WHERE spd.study_id = %s
ARGS: (10002L,)
Error: relation "qiita.processed_data_status" does not exist
LINE 2:                 FROM qiita.processed_data_status pds
                             ^
```

antgonza:
The error was due to not having the latest patch, erasing these comments and continuing review.

ElDeveloper:
I think this is looking good, and it works great!

The only other thing that I noticed (aside from the minor comments above), is that the icon in the "show prep template summary" button should probably removed, as the eye or the crossed-out eye imply information about the privacy status of the data objects.

screen shot 2015-03-25 at 3 31 56 pm

ElDeveloper:
Forgot to add, :+1:.

qiita_db/data.py
josenavas: I actually forgot to add the Error raised :astonished:

qiita_db/data.py
josenavas: Sure, Done!

qiita_db/data.py
josenavas: Done!

qiita_db/data.py
josenavas: Better be safe than sorry! Changing.

qiita_pet/util.py
josenavas: You might be right as this is a global. Done!

josenavas:
Good catch @ElDeveloper I've also removed the eye icon from the show sample template summary so the interface is consistent. I tried to find another icon, but given that we already have the test and I cannot find another "good enough" icon, I just left the text. Thanks for the reviews!

ElDeveloper:
Thanks!! That seems reasonable (re: button). :+1:

Yoshiki Vázquez-Baeza

josenavas:
That's a new one! I received a "Killed" message... I re-started the build...

ElDeveloper:
Yeah, I was getting that last week. I've started taking note of the
worker node identifiers and the output of `redis-server --version`
(maybe there's something funky there. BTW, I restarted the build
(again).

qiita_pet/templates/study_description_templates/processed_data_info_tab.html
```diff
@@ -14,4 +14,14 @@
       </tr>
     {% end %}
   </table>
+  {% if btn_to_show == 'request_approval' %}
+    <a class="btn btn-default glyphicon glyphicon-eye-open" onClick="request_approval({{pd_id}});" style="word-spacing: -10px;"> Request Approval</a>
```
antgonza: Can you be more specific, users will not know what approval means? Perhaps, "Request making this data public". Same to the other buttons.

antgonza:
IMOO it's confusing that awaiting for awaiting for private approval has an orange eye and when public a green one. The confusing part is that private has a gray eye. What about: waiting for approval: orange open eye + 'glyphicon glyphicon-lock', private yellow open eye + 'glyphicon glyphicon-lock', public green open eye + 'glyphicon glyphicon-globe'?

antgonza:
Two comments. After those :+1:

ElDeveloper:
Build restarted :alarm_clock:

antgonza:
... and again.

josenavas:
This is getting annoying... we need to find a solution for that...

josenavas:
I've increased the timeout time to 10 sec instead of 5, to see if now the tests pass....

josenavas:
Now a redis connection problem :cry: re-starting...

josenavas:
Ok, I got the tests passing now. I'm going to re-start them once to see if they are passing again. If so, this should be ready to final review/merge

ElDeveloper:
Thanks @josenavas!

josenavas:
:boom:

PR #901, opened 2015-03-04
Change columns for natsort on shared studies table

### LINKED ISSUE #894: js error loading the studies page
Getting this error when trying to open the study list: error

My guess is something to do with one of the studies not having certain value. We can't update main site until this is solved.

adamrp:
Fix #894

### DIFF
--- a/qiita_pet/templates/private_studies.html
+++ b/qiita_pet/templates/private_studies.html
@@ -15,7 +15,7 @@ $(document).ready(function() {
         });
         $('#shared-studies-table').dataTable({
             order: [[ 0, "asc" ]],
-            columnDefs: [{type:'natural', targets:[1,2,3,7]}]
+            columnDefs: [{type:'natural', targets:[2,3,4]}]
         });
     $('#waiting').hide();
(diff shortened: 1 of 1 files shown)


qiita_pet/templates/private_studies.html
```diff
@@ -15,7 +15,7 @@
         });
         $('#shared-studies-table').dataTable({
             order: [[ 0, "asc" ]],
-            columnDefs: [{type:'natural', targets:[1,2,3,7]}]
+            columnDefs: [{type:'natural', targets:[2,3,4]}]
```
antgonza: Could you add a comment about which column each number represents?

adamrp: It's just all the columns that don't have HTML in them. The HTML throws off the natsort plugin.

antgonza:
:+1:

ElDeveloper:
This works fine on my system, though I don't have a way to reproduce the problem.

ElDeveloper:
:+1:

PR #902, opened 2015-03-05
DOC: Improve CONTRIBUTING.md and INSTALL.md

ElDeveloper:
There's still a link missing, I don't know of any documentation about           
the Qiita configuration file, but someone pointed at it during the call         
earlier today. If I can get that, I'll add it right away.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -14,7 +14,7 @@ Adding source code to Qiita, can take place in three different modules:
 
 ###Configuration file
 
-The Qiita configuration file determines how the package interacts with your system’s resources (redis, postgres and the IPython cluster). Thus you should review the documentation detailed here (WE ARE STILL MISSING THIS LINK), but especially bear in mind the following points:
+The Qiita configuration file determines how the package interacts with your system’s resources (redis, postgres and the IPython cluster). Thus you should review the documentation detailed [here](https://docs.google.com/document/d/1u7kwLP31NM513-8xwpwvLbSQxYu0ehI6Jau1APR13e0/edit#), but especially bear in mind the following points:
 
 * An example version of this file can be found here `qiita_core/support_files/qiita_config.txt` and if you don’t set a `QIITA_CONFIG_FP` environment variable, that’s the file that Qiita will use.
 * The `[main]` section sets a `TEST_ENVIRONMENT` variable, which determines whether your system will be running unit tests or if it a demo/production system. You will want to set the value to TRUE if you are running the unit tests.
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,13 +1,54 @@
-#Making Database Changes
+#Contributing to Qiita
+
+Qiita is an open source software package, and we welcome community contributions. You can find the source code and test code for Qiita under public revision control in the Qiita git repository on [GitHub](http://github.com/biocore/qiita). We very much welcome contributions.
+
+This document covers what you should do to get started with contributing to Qiita. You should read this whole document before considering submitting code to Qiita. This will save time for both you and the Qiita developers.
+
+#General Notes on Development
+
+Adding source code to Qiita, can take place in three different modules:
+
+* `qiita_pet`: Contains the graphical user interface layer of the system.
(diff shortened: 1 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -1,13 +1,54 @@
...
+
+Qiita is an open source software package, and we welcome community contributions. You can find the source code and test code for Qiita under public revision control in the Qiita git repository on [GitHub](http://github.com/biocore/qiita). We very much welcome contributions.
+
+This document covers what you should do to get started with contributing to Qiita. You should read this whole document before considering submitting code to Qiita. This will save time for both you and the Qiita developers.
+
+#General Notes on Development
+
+Adding source code to Qiita, can take place in three different modules:
+
+* `qiita_pet`: Contains the graphical user interface layer of the system.
```
antgonza: Do you think is worth adding the languages each of this rely on?

antgonza:
Also, what about adding a link to the files and script to load the EMP tables? This should create a nice test environment? We could host the files somewhere in microbio.me

ElDeveloper:
In general I would agree with that, but realistically those files and
commands always become outdated and there's nobody who has committed to
update them accordingly, thus the result is a frequently broken shell
script. If someone were to volunteer though, I would :+1: this
suggestion. Perhaps the real way to solve this on the long run is by
creating a DB dump as pointed out in the other issue.

antgonza:
I think there hasn't been a change to that script in over 2 months, which is a lot in our current release cycle.

josenavas:
@antgonza did you check if such script still works?

josenavas:
@ElDeveloper an alternative solution will be to include those files in the repo and try to execute that command at the end of the build.

antgonza:
Yup, we used it on Thu (or Wed?) last week on a fresh install.

josenavas:
Thanks @antgonza 

If we want to use that script as a "standard" demo testing, we should include that script on the repo and execute it on travis. We can either include the files too, or making them available through the ftp so travis can download the tgz...

CONTRIBUTING.md
ElDeveloper: Sure, done!

ElDeveloper:
Hmmm how much time would that add to the Travis builds, they are already
rather long? Maybe using a subset instead?

antgonza:
I do not like the idea of adding all but we could add the script and 1
single study so we now that it works all the time.

ElDeveloper:
That would make sense, loading two demo studies.

josenavas:
Sounds reasonable, new issue for that or will be fixed here?

ElDeveloper:
New issue, this is a documentation improvement.

CONTRIBUTING.md
```diff
@@ -1,13 +1,54 @@
...
+
+Adding source code to Qiita, can take place in three different modules:
+
+* `qiita_pet`: Contains the graphical user interface layer of the system, mainly written in Python, JavaScript and HTML (see [Tornado templates](http://tornado.readthedocs.org/en/latest/template.html)).
+* `qiita_db`: Contains the bridge layer between the Python objects and the SQL database. In general this subpackage is mainly written in Python with a fair amount of inline PostgreSQL statements (see the section below on how to make database changes).
+* `qiita_ware`: Contains the logic of the system and functions that can generally be called from a Python script (see the scripts directory), and it is mostly written in Python. Several workflows that can be achieved using the GUI, can also be reproduced through the command line using this subpackage.
+
+###Configuration file
+
+The Qiita configuration file determines how the package interacts with your system’s resources (redis, postgres and the IPython cluster). Thus you should review the documentation detailed here (WE ARE STILL MISSING THIS LINK), but especially bear in mind the following points:
```
ElDeveloper: Does anyone know where this link is?

antgonza: Currently is a Google doc, not sure if you want to point to that. If so,
look at the hackpad.

ElDeveloper: Got it, I thought this was referring to a document about the qiita
config file. This PR should be ready for merge.

antgonza:
:+1:

ElDeveloper:
Can someone review/merge this PR?

PR #904, opened 2015-03-05
Update sample template

antgonza:
This PR add being able to update a sample template

### DIFF
--- a/qiita_db/metadata_template.py
+++ b/qiita_db/metadata_template.py
@@ -44,6 +44,7 @@ from os import close
 from time import strftime
 from functools import partial
 from tempfile import mkstemp
+from os.path import basename
 
 import pandas as pd
 import numpy as np
@@ -655,12 +656,12 @@ class Sample(BaseSample):
                 value_type = type(value).__name__
 
                 if column_type != value_type:
-                    raise ValueError("""The new value being added to column:
-                        "{0}" is "{1}" (type: "{2}"). However, this
-                        column in the DB is of type "{3}". Please change the
-                        value in your updated template or reprocess your sample
-                        template.""".format(column, value, value_type,
-                                            column_type))
+                    raise ValueError(
+                        'The new value being added to column: "{0}" is "{1}" '
+                        '(type: "{2}"). However, this column in the DB is of '
+                        'type "{3}". Please change the value in your updated '
+                        'template or reprocess your sample template.'.format(
+                            column, value, value_type, column_type))
                 else:
                     raise e
         elif exists_required:
@@ -1562,7 +1563,7 @@ class SampleTemplate(MetadataTemplate):
                 for _, fp in pt.get_filepaths():
                     # the difference between a prep and a qiime template is the
                     # word qiime within the name of the file
-                    if '_qiime_' not in fp:
+                    if '_qiime_' not in basename(fp):
                         pt.create_qiime_mapping_file(fp)
 
     def remove_category(self, category):
--- a/qiita_db/metadata_template.py
+++ b/qiita_db/metadata_template.py
@@ -1535,11 +1535,12 @@ class SampleTemplate(MetadataTemplate):
         # generating all new QIIME mapping files
         for rd_id in Study(self.id).raw_data():
             for pt_id in RawData(rd_id).prep_templates:
-                # the difference between a prep and a qiime template is the
-                # word qiime within the name of the file
                 pt = PrepTemplate(pt_id)
-                fps = [pt.create_qiime_mapping_file(fp)
-                       for _, fp in pt.get_filepaths() if 'qiime' not in fp]
+                for _, fp in pt.get_filepaths():
+                    # the difference between a prep and a qiime template is the
+                    # word qiime within the name of the file
+                    if 'qiime' not in fp:
+                        pt.create_qiime_mapping_file(fp)
 
     def remove_category(self, category):
         """Remove a category from the sample template
--- a/qiita_db/metadata_template.py
+++ b/qiita_db/metadata_template.py
@@ -346,14 +346,9 @@ class BaseSample(QiitaObject):
             "sample_id=%s".format(self._dynamic_table),
             (self._id, )))
         d.update(dynamic_d)
-        del d['sample_id']
         del d[self._id_column]
-        try:
-            # The study_id could potentially be already removed with _id_column
-            # so wrapping in a try except
-            del d['study_id']
-        except KeyError:
-            pass
+        d.pop('study_id', None)
+
         # Modify all the *_id columns to include the string instead of the id
         for k, v in viewitems(self._md_template.translate_cols_dict):
             d[v] = self._md_template.str_cols_handlers[k][d[k]]
@@ -1032,10 +1027,7 @@ class MetadataTemplate(QiitaObject):
                 metadata_map[k][value] = self.str_cols_handlers[key][id_]
                 del metadata_map[k][key]
             metadata_map[k].update(dyn_vals[k])
-            try:
-                del metadata_map[k]['study_id']
-            except KeyError:
-                pass
+            metadata_map[k].pop('study_id', None)
 
         # Remove samples that are not in the samples list, if it was supplied
         if samples is not None:
@@ -1491,10 +1483,7 @@ class SampleTemplate(MetadataTemplate):
 
         for k in current_map:
             current_map[k].update(dyn_vals[k])
-            try:
-                del current_map[k]['study_id']
-            except KeyError:
-                pass
+            current_map[k].pop('study_id', None)
 
         # converting sql results to dataframe
         current_map = pd.DataFrame.from_dict(current_map, orient='index')
--- a/qiita_db/metadata_template.py
+++ b/qiita_db/metadata_template.py
@@ -622,31 +622,34 @@ class Sample(BaseSample):
 
         # try dynamic tables
         exists_dynamic = conn_handler.execute_fetchone("""
+        SELECT EXISTS (
             SELECT column_name
             FROM information_schema.columns
             WHERE table_name='{0}'
                 AND table_schema='qiita'
-                AND column_name='{1}'""".format(self._dynamic_table, column))
+                AND column_name='{1}')""".format(self._dynamic_table,
+                                                 column))[0]
         # try required_sample_info
         exists_required = conn_handler.execute_fetchone("""
+        SELECT EXISTS (
             SELECT column_name
             FROM information_schema.columns
             WHERE table_name='required_sample_info'
                 AND table_schema='qiita'
-                AND column_name='{0}'""".format(column))
+                AND column_name='{0}')""".format(column))[0]
 
-        if exists_dynamic is not None:
+        if exists_dynamic:
             conn_handler.execute("""
                 UPDATE qiita.{0}
-                SET {1}={2}
-                WHERE sample_id='{3}'""".format(self._dynamic_table, column,
-                                                value, self._id))
-        elif exists_required is not None:
+                SET {1}=%s
+                WHERE sample_id=%s""".format(self._dynamic_table,
+                                             column), (value, self._id))
+        elif exists_required:
             conn_handler.execute("""
                 UPDATE qiita.required_sample_info
-                SET {0}='{1}'
-                WHERE sample_id='{2}'
-                """.format(column, value, self._id))
+                SET {0}=%s
+                WHERE sample_id=%s
+                """.format(column), (value, self._id))
         else:
             raise QiitaDBColumnError("Column %s does not exist in %s" %
                                      (column, self._dynamic_table))
--- a/qiita_db/metadata_template.py
+++ b/qiita_db/metadata_template.py
@@ -346,6 +346,7 @@ class BaseSample(QiitaObject):
             "sample_id=%s".format(self._dynamic_table),
             (self._id, )))
         d.update(dynamic_d)
+        del d['sample_id']
         del d[self._id_column]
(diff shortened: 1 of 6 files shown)


qiita_db/metadata_template.py
```diff
@@ -620,23 +620,37 @@ def __setitem__(self, column, value):
...
+        exists_dynamic = conn_handler.execute_fetchone("""
             SELECT column_name
             FROM information_schema.columns
             WHERE table_name='{0}'
                 AND table_schema='qiita'
                 AND column_name='{1}'""".format(self._dynamic_table, column))
-
-        if exists is None:
+        # try required_sample_info
+        exists_required = conn_handler.execute_fetchone("""
```
squirrelo: We have the utility function `get_table_cols` that does exactly this.

antgonza: I can use that one but note that `get_table_cols` doesn't use `AND table_schema='qiita'`, any reason for this?

squirrelo: Oh, I see what's going on here. Please either use `exists` in the SQL to turn it into a bool, or preferably the get_table_cols and `column in table_cols`. Cleaner logic.

antgonza: Good point, will do, I was following what was already there ...

qiita_db/metadata_template.py
```diff
@@ -620,23 +620,37 @@ def __setitem__(self, column, value):
...
+            FROM information_schema.columns
+            WHERE table_name='required_sample_info'
+                AND table_schema='qiita'
+                AND column_name='{0}'""".format(column))
+
+        if exists_dynamic is not None:
+            conn_handler.execute("""
+                UPDATE qiita.{0}
+                SET {1}={2}
+                WHERE sample_id='{3}'""".format(self._dynamic_table, column,
```
squirrelo: Using format here is extremely dangerous as it opens us up to SQL attacks. You can use the string replacement in the execute function to add the value and id.

qiita_db/metadata_template.py
```diff
@@ -620,23 +620,37 @@ def __setitem__(self, column, value):
...
+                                                value, self._id))
+        elif exists_required is not None:
+            conn_handler.execute("""
+                UPDATE qiita.required_sample_info
+                SET {0}='{1}'
+                WHERE sample_id='{2}'
+                """.format(column, value, self._id))
```
squirrelo: Same here

antgonza:
Tests are not actually failing. The new interface marks the build as failure cause the Coverage decreased

qiita_db/metadata_template.py
```diff
@@ -1250,7 +1267,7 @@ def _clean_validate_template(cls, md_template, study_id,
 
     @classmethod
     def create(cls, md_template, study):
-        r"""Creates the metadata template in the database
+        r"""Creates the sample metadata template in the database
```
ElDeveloper: I think the word metadata can be removed from this line.

qiita_db/metadata_template.py
```diff
@@ -1445,6 +1462,89 @@ def extend(self, md_template):
         # adding the fp to the object
         self.add_filepath(fp)
 
+    def update(self, md_template):
+        r"""Update values in the sample metadata template
```
ElDeveloper: Same here and anywhere else, nobody calls them sample metadata templates, I think everyone calls them sample templates. However this is not a blocking issue, it's more of a nit-pick.

qiita_db/metadata_template.py
```diff
@@ -1445,6 +1462,89 @@ def extend(self, md_template):
...
+        current_map = self._transform_to_dict(conn_handler.execute_fetchall(
+            "SELECT * FROM qiita.{0} WHERE {1}=%s".format(self._table,
+                                                          self._id_column),
+            (self.id,)))
+        dyn_vals = self._transform_to_dict(conn_handler.execute_fetchall(
+            "SELECT * FROM qiita.{0}".format(self._table_name(self.id))))
+
+        for k in current_map:
+            current_map[k].update(dyn_vals[k])
+            try:
```
ElDeveloper: I think it's better to check `if 'study_id' in current_map[k]:` instead of a try/except block.

antgonza: Wouldn't that make it slower? This way we are not searching within a list every single time

qiita_db/metadata_template.py
```diff
@@ -1445,6 +1462,89 @@ def extend(self, md_template):
...
+                               'stored in database by these samples names: %s'
+                               % ', '.join(samples_diff))
+        columns_diff = set(new_map.columns) - set(current_map.columns)
+        if columns_diff:
+            raise QiitaDBError('The new sample template differs from what is '
+                               'stored in database by these columns names: %s'
+                               % ', '.join(columns_diff))
+
+        # here we are comparing two dataframes following:
+        # http://stackoverflow.com/questions/24495695
```
ElDeveloper: This is pointing to a question, would you mind pointing to the answer that you are referring to?

qiita_db/metadata_template.py
```diff
@@ -1445,6 +1462,89 @@ def extend(self, md_template):
...
+        self.add_filepath(fp)
+
+        # generating all new QIIME mapping files
+        for rd_id in Study(self.id).raw_data():
+            for pt_id in RawData(rd_id).prep_templates:
+                pt = PrepTemplate(pt_id)
+                for _, fp in pt.get_filepaths():
+                    # the difference between a prep and a qiime template is the
+                    # word qiime within the name of the file
+                    if 'qiime' not in fp:
```
ElDeveloper: Can you check that `_qiime_` is in the `basename` of the `fp`? We could have a pathological case where someone's filesystem's name is `/qiime/` and then all paths would be picked-up by this if statement.

qiita_pet/handlers/study_handlers/description_handlers.py
```diff
@@ -197,6 +197,66 @@ def process_sample_template(self, study, user, callback):
...
+        fp_rsp = join(base_fp, str(study.id), sample_template)
+
+        if not exists(fp_rsp):
+            # The file does not exists, fail nicely
+            raise HTTPError(400, "This file doesn't exist: %s" % fp_rsp)
+
+        try:
+            with warnings.catch_warnings(record=True) as warns:
+                # force all warnings to always be triggered
+                warnings.simplefilter("always")
```
ElDeveloper: We don't need this line.

qiita_pet/handlers/study_handlers/description_handlers.py
```diff
@@ -197,6 +197,66 @@ def process_sample_template(self, study, user, callback):
...
+                if warns:
+                    msg = '; '.join([str(w.message) for w in warns])
+                    msg_level = 'warning'
+
+        except (TypeError, QiitaDBColumnError, QiitaDBExecutionError,
+                QiitaDBDuplicateError, IOError, ValueError, KeyError,
+                CParserError, QiitaDBDuplicateHeaderError, QiitaDBError) as e:
+            # Some error occurred while processing the sample template
+            # Show the error to the user so he can fix the template
+            msg = html_error_message % ('parsing the sample template:',
```
ElDeveloper: parsing -> updating?

qiita_db/metadata_template.py
ElDeveloper: Ah ... I see what you are saying. From searching on the intertubes, it
seems like this is a perfect use-case for dict.pop, for example:

```
In [1]: x = {'a': 1, 'b': 2, 'c': 3}

In [2]: x.pop('a', None)
Out[2]: 1

In [3]: x.pop('LOLERCOASTER', None)
```

ElDeveloper:
I added a sample template to a study that already had a preprocessed data. And since this is not allowed, I was sent to the error page. Which is not useful for a regular user, this error should be described and presented to the user, along with a suggestion on how to deal with this:

```python
  File "/Users/yoshikivazquezbaeza/.virtualenvs/qiita/lib/python2.7/site-packages/tornado/web.py", line 1144, in _when_complete
  File "/Users/yoshikivazquezbaeza/.virtualenvs/qiita/lib/python2.7/site-packages/tornado/concurrent.py", line 129, in result
  File "/Users/yoshikivazquezbaeza/.virtualenvs/qiita/lib/python2.7/site-packages/tornado/gen.py", line 221, in wrapper
    runner.run()
  File "/Users/yoshikivazquezbaeza/.virtualenvs/qiita/lib/python2.7/site-packages/tornado/gen.py", line 505, in run
  File "/Users/yoshikivazquezbaeza/git_sw/qiita/qiita_pet/handlers/study_handlers/description_handlers.py", line 738, in post
    study, user)
  File "/Users/yoshikivazquezbaeza/.virtualenvs/qiita/lib/python2.7/site-packages/tornado/gen.py", line 533, in run
    self.yield_point.start(self)
  File "/Users/yoshikivazquezbaeza/.virtualenvs/qiita/lib/python2.7/site-packages/tornado/gen.py", line 371, in start
  File "/Users/yoshikivazquezbaeza/git_sw/qiita/qiita_pet/handlers/study_handlers/description_handlers.py", line 181, in process_sample_template
    fp_rsp)
  File "/Users/yoshikivazquezbaeza/git_sw/qiita/qiita_pet/handlers/study_handlers/description_handlers.py", line 548, in remove_add_study_template
    PrepTemplate.delete(pt)
  File "/Users/yoshikivazquezbaeza/git_sw/qiita/qiita_db/metadata_template.py", line 1921, in delete
    "using it." % id_)
 QiitaDBError: Cannot remove prep template 101 because a preprocessed data has been already generated using it.
```

From trying to update another sample template in a different study I got this on the user interface (note that updating worked in other cases, but this is the first time I see this error):

```python
Error running SQL query: UPDATE qiita.sample_10006 SET depth=%s WHERE sample_id=%s ARGS: ('no_data', '10006.NEWSAMPLEB1') Error: invalid input syntax for type double precision: "no_data" LINE 3: SET depth='no_data' ^
```

-------

Aside from those two things the interface looks fine.

antgonza:
Ready for final review.

qiita_db/metadata_template.py
```diff
@@ -623,23 +619,62 @@ def __setitem__(self, column, value):
...
+            except Exception as e:
+                column_type = conn_handler.execute_fetchone("""
+                    SELECT data_type
+                    FROM information_schema.columns
+                    WHERE column_name=%s AND table_schema='qiita'
+                    """, (column,))[0]
+                value_type = type(value).__name__
+
+                if column_type != value_type:
+                    raise ValueError("""The new value being added to column:
```
ElDeveloper: This error message is going to come out oddly because of the whitespace. Can you change it such that you are not using the triple quotation marks?

qiita_pet/handlers/study_handlers/description_handlers.py
```diff
@@ -194,6 +194,61 @@ def process_sample_template(self, study, user, callback):
...
+                # will be ignored if an exception is raised
+                if warns:
+                    msg = '; '.join([str(w.message) for w in warns])
+                    msg_level = 'warning'
+
+        except (TypeError, QiitaDBColumnError, QiitaDBExecutionError,
+                QiitaDBDuplicateError, IOError, ValueError, KeyError,
+                CParserError, QiitaDBDuplicateHeaderError, QiitaDBError) as e:
+            # Some error occurred while processing the sample template
+            # Show the error to the user so he can fix the template
```
ElDeveloper: he -> they

qiita_pet/handlers/study_handlers/description_handlers.py
```diff
@@ -194,6 +194,61 @@ def process_sample_template(self, study, user, callback):
...
+        # Define here the message and message level in case of success
+        msg = "The sample template '%s' has been updated" % sample_template
+        msg_level = "success"
+        # Get the uploads folder
+        _, base_fp = get_mountpoint("uploads")[0]
+        # Get the path of the sample template in the uploads folder
+        fp_rsp = join(base_fp, str(study.id), sample_template)
+
+        if not exists(fp_rsp):
+            # The file does not exists, fail nicely
```
ElDeveloper: exists->exist

qiita_db/metadata_template.py
```diff
@@ -1453,6 +1485,86 @@ def extend(self, md_template):
...
+        self.add_filepath(fp)
+
+        # generating all new QIIME mapping files
+        for rd_id in Study(self.id).raw_data():
+            for pt_id in RawData(rd_id).prep_templates:
+                pt = PrepTemplate(pt_id)
+                for _, fp in pt.get_filepaths():
+                    # the difference between a prep and a qiime template is the
+                    # word qiime within the name of the file
+                    if '_qiime_' not in fp:
```
ElDeveloper: Just to be extra cautious, can you check for:

```
from os.path import basename

if '_qiime_' not in basename(fp):
    # do the dinosaur
```

ElDeveloper:
Looks good, a few minor comments. I tried to find an issue for the first problem that I noted in my previous comment, but couldn't find it, can you open a separate issue if we indeed don't have an issue about it, else I'll open it.

qiita_db/metadata_template.py
antgonza: Changing but not really as in html multiple spaces are changed to one single space.

antgonza:
Added: https://github.com/biocore/qiita/issues/909

qiita_db/metadata_template.py
ElDeveloper: Thanks, it would have been a problem through a non HTML interface
though.

ElDeveloper:
Thanks! :+1:, this is ready to be merged.

PR #906, opened 2015-03-05
Add option, strip whitespace from template values

### LINKED ISSUE #883: issue with ebi submission
We have a blind failure in qiita where submissions to EBI will not be successful cause we only accept https://github.com/biocore/qiita/blob/master/qiita_ware/ebi.py#L374 so if the template has Ilumina or IlluMina it will not work. What will be the best solution? Note that to have the studies that we want got to EBI without a problem we could update the DB (so it is ILLUMINA) or the code (so it accepts any casing).

adamrp:
By default it is true, so whitespaces will be stripped from template values
prior to being loaded into a dataframe.

Fix #883

### DIFF
--- a/qiita_db/metadata_template.py
+++ b/qiita_db/metadata_template.py
@@ -2058,14 +2058,14 @@ class PrepTemplate(MetadataTemplate):
         return filepath
 
 
-def load_template_to_dataframe(fn, strip_values=True):
+def load_template_to_dataframe(fn, strip_whitespace=True):
     """Load a sample or a prep template into a data frame
 
     Parameters
     ----------
     fn : str
         filename of the template to load
-    strip_values : bool, optional
+    strip_whitespace : bool, optional
         Defaults to True. Whether or not to strip whitespace from values in the
         input file
 
@@ -2117,7 +2117,7 @@ def load_template_to_dataframe(fn, strip_values=True):
     """
 
     # First, strip all values from the cells in the input file, if requested
-    if strip_values:
+    if strip_whitespace:
         fd, fp = mkstemp()
         close(fd)
 
--- a/qiita_db/metadata_template.py
+++ b/qiita_db/metadata_template.py
@@ -42,13 +42,16 @@ from future.builtins import zip
 from future.utils import viewitems, PY3
 from copy import deepcopy
 from os.path import join
+from os import close
 from time import strftime
 from functools import partial
+from tempfile import mkstemp
 
 import pandas as pd
 import numpy as np
 import warnings
 from skbio.util import find_duplicates
+from skbio.io.util import open_file
 
 from qiita_core.exceptions import IncompetentQiitaDeveloperError
 from .exceptions import (QiitaDBDuplicateError, QiitaDBColumnError,
@@ -2055,13 +2058,16 @@ class PrepTemplate(MetadataTemplate):
         return filepath
 
 
-def load_template_to_dataframe(fn):
+def load_template_to_dataframe(fn, strip_values=True):
     """Load a sample or a prep template into a data frame
 
     Parameters
     ----------
     fn : str
         filename of the template to load
+    strip_values : bool, optional
+        Defaults to True. Whether or not to strip whitespace from values in the
+        input file
 
     Returns
     -------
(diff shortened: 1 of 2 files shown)


ElDeveloper:
I'm on the fence about this, opening the file and removing the whitespace, writing the file and then reopening it to load it into the data frame seems like there's overhead we can cut. Maybe once the data frame is loaded use the [`applymap`](http://pandas.pydata.org/pandas-docs/dev/generated/pandas.DataFrame.applymap.html) function to strip all the whitespace?

ElDeveloper:
Also, we would need a DB patch to fix all the values stored with leading and trailing white spaces.

adamrp:
I don't think a database patch is necessary, since this only applies to new metadata templates being loaded.

I don't think creating a temporary file is much overhead. The benefit to doing it this way, prior to loading into a dataframe, is that I don't have to worry about types. In fact, it might even eliminate incorrect dataframe parsing in the case that the column contains whitespace-padded numerical values.

antgonza:
I like the idea of a patch to fix the database entries. I do not have
an opinion between file/pandas ...

adamrp:
If I patch all the sample and prep template values, this potentially affects all preprocessed data, processed data, and analyses generated from studies whose templates are modified. I don't think we really wan to do that, do we?

antgonza:
Nop, just database and perhaps recreate sample/prep/qiime templates.

adamrp:
Why would (pre)processed data and analyses not be affected? Changing underlying metadata for existing files is not ideal.

ElDeveloper:
I see what you are saying, what about updating the QIIME mapping files?
I feel like having those trailing/leading are a very undetectable source
of frustration, that once you realize is there, you are screwed :(

As for pandas being able to infer datatypes better, you are probably
right, so I think this on its own is good enough that going over the
file isn't too bad.

qiita_db/metadata_template.py
```diff
@@ -2110,6 +2116,19 @@ def load_template_to_dataframe(fn):
...
     """
 
+    # First, strip all values from the cells in the input file, if requested
+    if strip_values:
+        fd, fp = mkstemp()
+        close(fd)
+
+        with open_file(fn, 'U') as input_f, open(fp, 'w') as new_f:
+            for line in input_f:
+                line_elements = [x.strip()
```
ElDeveloper: I'm kinda surprised you didn't use a regex here.

qiita_db/metadata_template.py
```diff
@@ -2055,13 +2058,16 @@ def create_qiime_mapping_file(self, prep_template_fp):
...
 
-def load_template_to_dataframe(fn):
+def load_template_to_dataframe(fn, strip_values=True):
     """Load a sample or a prep template into a data frame
 
     Parameters
     ----------
     fn : str
         filename of the template to load
+    strip_values : bool, optional
```
ElDeveloper: I would probably change this to `strip_whitespace`, but that's just a personal opinion and non-blocking to this PR.

antgonza:
I never said they weren't ;-) but fixing the entries in the DB will
solve future issues like the one that generated this PR ...

qiita_db/metadata_template.py
adamrp: Haha regex have their time and place, but here isn't one of them :smile:

qiita_db/metadata_template.py
adamrp: I can make that change

adamrp:
Okay, so what's the consensus? I honestly don't think that this PR should modify old DB values since it gets us to a place of inconsistency between the database and the existing files that have been generated from the database.

ElDeveloper:
I vote only QIIME mapping file.

ElDeveloper:
Chatting with @adamrp offline, I think I now understand the problem. We cannot change the sample and prep templates, otherwise we will have inconsistencies across the DB. Since QIIME mapping files are generated from the sample+prep templates, updating them would imply that we need to update the sample and prep templates.

:+1: for merge.

adamrp:
@ElDeveloper exactly, updating those values in the database would have far-reaching impacts and create inconsistencies everywhere :smile:

ElDeveloper:
There's nothing to :smile: about, this is a serious issue :man:.

ElDeveloper:
Can someone review+merge this issue?

PR #917, opened 2015-03-09
ENH: change default webserver port

### LINKED ISSUE #914: Check port return
Qiita runs by default at localhost:8888.
IPython also runs at localhost:8888.

Would be nice if Qiita could ping the local host instances which are running and start a new instance, rather than manual override.

jairideout:
Changed default port from 8888 to 21174, which is (cursive Q)iita in 1337sp34k. This port is unlikely to conflict with other services (e.g., IPython). If it does, a descriptive error message is raised.

Fixes #914.

Thanks to @ebolyen and @ElDeveloper for the input on this!


antgonza:
:+1: when test pass.

PR #918, opened 2015-03-09
Displays sample counts in prep/sample template.

### LINKED ISSUE #608: show number of samples
The study/prep/sample summaries should show the number of samples.

JWDebelius:
Fixes #608

### DIFF
--- a/qiita_pet/templates/metadata_summary.html
+++ b/qiita_pet/templates/metadata_summary.html
@@ -4,6 +4,10 @@
 
 <h1>Metadata Summary for Study {{study_title}}</h1>
 
+<div>
+    <p>There are <strong>{{num_samples}}</strong> samples in this study.</p>
+</div>
+
 <div class="panel panel-default">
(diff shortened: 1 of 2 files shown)


qiita_pet/templates/metadata_summary.html
```diff
@@ -4,6 +4,10 @@
 
 <h1>Metadata Summary for Study {{study_title}}</h1>
 
+<div>
+    <p>There are <strong>{{num_samples}}</strong> samples in this study.</p>
```
wasade: There is an extra space, only need once space on the side of `num_samples`

wasade:
One comment, :+1: otherwise

qiita_pet/templates/metadata_summary.html
wasade: ya, im an idiot

JWDebelius: I think it needs both spaces to render sanely.

PR #923, opened 2015-03-10
adds a back link to the top of the metadata summary pages

gregcaporaso:


antgonza:
:+1:

wasade:
:+1:

qiita_pet/templates/metadata_summary.html
```diff
@@ -5,6 +5,10 @@
 <h1>Metadata Summary for Study {{study_title}}</h1>
 
 <div>
+    <a class="glyphicon glyphicon-arrow-left" href="{{back_button_path}}">Back</a>
```
ElDeveloper: Can you add the button class to this tag? and word-spacing so there's some space between the icon and the word? 


Like this:

```
 <a class="btn btn-default glyphicon glyphicon-....." href="...." style="word-spacing: -10px;">....</a>
```

See this [line](https://github.com/biocore/qiita/blob/master/qiita_pet/templates/study_description_templates/prep_template_panel.html#L68) for an exampl.

gregcaporaso:
@ElDeveloper, I made this a button class. ``word-spacing``doesn't affect the spacing between the glyphicon and the text, just the spacing between words in the text (so no effect in the case where there is only one word). I'm haven't figured out what the right style would be for this, but it seems like we should do something that would globally be in place for glyphicon, e.g. via CSS.

gregcaporaso:
The [bootstrap page for glyphicons](http://getbootstrap.com/components/) has examples of how these should be used for creating buttons. I updated these two cases to match those examples, and this fixes the spacing.

Issue #16, opened 2014-02-13
Ability to download beta-diversity text file


ElDeveloper:
Ported from Qiita-pet.

```
squirrelo commented a month ago
Current Qiime database does not give the beta-diversity distance matrix. Could this be made available in the new release?
 Yoshiki Vázquez Baeza
 
ElDeveloper commented a month ago
I'm pretty sure the distance matrix is part of the zipped file that comes as a result of a meta-analysis if you decided to compute any beta-diversity metric to plot it in a three dimensional space. But I agree that there should be a direct link to this file instead of just to the 3d plot.
```

tracykteal:
+1 on being able to download the distance matrix or the FASTQ files for all the samples selected for analysis. Right now I don't see a zipped file that can be downloaded, just an index.html page link after running the meta-analysis.

antgonza:
This should be really easy at this point. The tables exist and are available but we just need to put a link to  #_beta_div.../unweighted_unifrac_dm.txt or weighted_unifrac_dm.txt.

tracykteal:
Right, I could download them once you told me the URL, but there wasn't an indication that this was possible on the site. I think this would be a really important feature! Then people could do their own statistics and data analysis which is really key.

antgonza:
Agree. Note that next week we will have a code sprint to solve this and other issues ... in case you want to join us virtually. Here are some notes: https://hackpad.com/Biocore-code-sprint-v15.3-X4ypG5V0mSz

JWDebelius:
I'd like to be assigned to this.

tracykteal:
Great to see this being implemented. Let me know if you need a beta (alpha?) tester.

PR #929, opened 2015-03-10
Updates jobs to add distance matrix files

JWDebelius:
Fixes issue #16 ?

### DIFF
--- a/qiita_db/job.py
+++ b/qiita_db/job.py
@@ -347,6 +347,11 @@ class Job(QiitaStatusObject):
                     glob(join(basedir, check_dir, "*.html")):
                 result_fps.append(relpath(res, basedir))
 
+        def add_beta_diversity(basedir, check_dir, result_fps):
+            for res in glob(join(basedir, check_dir, "*_dm.txt")) + \
+                    glob(join(basedir, check_dir, "*_dm.txt")):
+                result_fps.append(relpath(res, basedir))
+
         # create new list, with relative paths from db base
         result_fps = []
         for fp in results:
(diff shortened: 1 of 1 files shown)


JWDebelius:
The support files for testing need to be added to the populate_test_db.sql. @squirrelo, @josenavas: can you help, please?

ElDeveloper:
:+1:, can we add a python patch to retrieve the currently existing distance matrices?

ElDeveloper:
Thanks for the explanation @antgonza and @JWDebelius.

ElDeveloper:
:+1:

ElDeveloper:
@JWDebelius, this looks fine, except for the fact that when you click on the name of the distance matrix, you get the data displayed in the iframe, instead of it being downloaded to your computer (which is what I was expecting and what I think we need this to do).

ElDeveloper:
http://stackoverflow.com/a/18678698/379593

wasade:
Distance matrices can get very large, will these be going through nginx for
the nom-tornado download? Might be useful to compress too

ElDeveloper:
@wasade, to answer your question as per @adamrp's description: yes it will go through nginx. As for compression, this would be something that we need to fix in the processing pipeline, not necessarily in the display.

wasade:
Great, thanks!

JWDebelius:
@squirrelo: Not sure how to do nginx or tornado, can you take over.

ElDeveloper:
@JWDebelius, note that you don't have to make any nginx/tornado changes,
just make it such that when clicking the link a file is downloaded
instead of displayed in the user interface.

josenavas:
@JWDebelius ETA on this PR? re: download check how the links are created for downloading the sample template and you'll have an example.

JWDebelius:
This is a problem of being able to download a file, or a display a file,
depending on the file. Each behavior is trivial to replicate, given the
codebase. However, the logic behind piping the figures is non-trivial.

ElDeveloper:
@JWDebelius Let me know if you need help figuring this out.

PR #931, opened 2015-03-10
Merged sample template with study information tab

### LINKED ISSUE #896: hide sample template tab
From @AmandaBirmingham: Given that the "Add raw data" tab and various other tabs don't show up until you have uploaded some files to use them on, are you able to do the same thing with the "Sample template" tab? It would actually be less confusing for the user if it WASN'T there until after she did her uploads ...

I will add that it will be even nicer if tabs were appearing as they have data: if we do not have raw data, we shouldn't have the preprocessed or processed tabs ...

wasade:
Fixes #896


wasade:
Found some dangling requests, thanks @antgonza

qiita_pet/templates/study_description_templates/study_information_tab.html
```diff
@@ -5,4 +7,48 @@
   <b>Principal investigator:</b> {% raw principal_investigator %} <br/>
   <b>Samples:</b> {{number_samples_promised}}/{{number_samples_collected}}<br/>
   <b>Metadata:</b> {{metadata_complete}} <br/>
+
```
antgonza: Could you add a couple of `<br/>` or a `<hr>`, just to have a visual division between the top and the bottom?

antgonza:
One minor comment. However, the issue also includes that the preprocess tab doesn't show until you have added a raw data, and the process tab until you have a loaded prep template ....

wasade:
@antgonza, I read that as a "nice to have", not a necessary part of closing out that issue. Can explore it if you'd like though

qiita_pet/templates/study_description_templates/study_information_tab.html
wasade: done

antgonza:
Yes, it's a nice to have, doesn't decrement functionality.

squirrelo:
:+1:

PR #932, opened 2015-03-10
removed some unnecessary pop-ups

### LINKED ISSUE #754: Remove unnecessary popup dialogs
See discusion here https://github.com/biocore/qiita/pull/732#discussion_r22235787

gregcaporaso:
Fixes #754. 

Specifically I removed the pop-ups that confirm creation of new raw data entries and prep templates. These were the two remaining pop-ups that were discussed in the issue that should be deleted.


wasade:
:+!:

PR #936, opened 2015-03-10
Unindent over-indented functions

adamrp:
For some reason (probably erroneous copy and paste) many functions in
qiita_db/util.py were indented too much. No changes were made here other than
to correct the indentation level of these functions.

### DIFF
--- a/qiita_db/util.py
+++ b/qiita_db/util.py
@@ -262,7 +262,7 @@ def create_rand_string(length, punct=True):
     length: int
         Length of string to return
     punct: bool, optional
-        Include punctiation as well as letters and numbers. Default True.
+        Include punctuation as well as letters and numbers. Default True.
     """
     chars = ''.join((ascii_letters, digits))
     if punct:
@@ -271,7 +271,7 @@ def create_rand_string(length, punct=True):
 
 
 def hash_password(password, hashedpw=None):
-    """ Hashes password
+    """Hashes password
 
     Parameters
     ----------
@@ -869,7 +869,7 @@ def filepath_ids_to_rel_paths(filepath_ids):
 
 
 def convert_to_id(value, table, conn_handler=None):
-    """Converts a string value to it's corresponding table identifier
+    """Converts a string value to its corresponding table identifier
 
     Parameters
     ----------
@@ -901,7 +901,7 @@ def convert_to_id(value, table, conn_handler=None):
 
 
 def convert_from_id(value, table, conn_handler=None):
-    """Converts an id value to it's corresponding string value
+    """Converts an id value to its corresponding string value
 
     Parameters
     ----------
--- a/qiita_db/util.py
+++ b/qiita_db/util.py
@@ -255,53 +255,53 @@ def get_emp_status(key='emp_status'):
 
 
 def create_rand_string(length, punct=True):
-        """Returns a string of random ascii characters
+    """Returns a string of random ascii characters
 
-        Parameters
-        ----------
-        length: int
-            Length of string to return
-        punct: bool, optional
-            Include punctiation as well as letters and numbers. Default True.
-        """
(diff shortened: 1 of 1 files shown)


qiita_db/util.py
```diff
@@ -255,53 +255,53 @@ def get_emp_status(key='emp_status'):
...
-        chars = ''.join((ascii_letters, digits))
-        if punct:
-            chars = ''.join((chars, punctuation))
-        return ''.join(choice(chars) for i in range(length))
+    Parameters
+    ----------
+    length: int
+        Length of string to return
+    punct: bool, optional
+        Include punctiation as well as letters and numbers. Default True.
```
wasade: punctuation

qiita_db/util.py
```diff
@@ -255,53 +255,53 @@ def get_emp_status(key='emp_status'):
...
-        if hashedpw is None:
-            hashedpw = gensalt()
-        else:
-            hashedpw = hashedpw.encode('utf-8')
-        password = password.encode('utf-8')
-        output = hashpw(password, hashedpw)
-        if isinstance(output, bytes):
-            output = output.decode("utf-8")
-        return output
+    """ Hashes password
```
wasade: Extra space

qiita_db/util.py
```diff
@@ -255,53 +255,53 @@ def get_emp_status(key='emp_status'):
...
+    -----
+    Relies on bcrypt library to hash passwords, which stores the salt as
+    part of the hashed password. Don't need to actually store the salt
+    because of this.
+    """
+    # all the encode/decode as a python 3 workaround for bcrypt
+    if hashedpw is None:
+        hashedpw = gensalt()
+    else:
+        hashedpw = hashedpw.encode('utf-8')
```
wasade: why is this necessary? the encrypting here is much more complex than what we're doing on american gut

qiita_db/util.py
```diff
@@ -869,66 +869,66 @@ def filepath_ids_to_rel_paths(filepath_ids):
...
-
-        Parameters
-        ----------
-        value : str
-            The string value to convert
-        table : str
-            The table that has the conversion
-        conn_handler : SQLConnectionHandler, optional
-            The sql connection object
+    """Converts a string value to it's corresponding table identifier
```
wasade: its

qiita_db/util.py
```diff
@@ -869,66 +869,66 @@ def filepath_ids_to_rel_paths(filepath_ids):
...
-
-        Parameters
-        ----------
-        value : int
-            The id value to convert
-        table : str
-            The table that has the conversion
-        conn_handler : SQLConnectionHandler, optional
-            The sql connection object
+    """Converts an id value to it's corresponding string value
```
wasade: its

wasade:
few comments

qiita_db/util.py
squirrelo: We are using the bcrypt library directly, which only works with utf-8 data. Americangut is using a wrapper around bcrypt library.

squirrelo:
Reviewed and no comments added on top of @wasade 's

qiita_db/util.py
wasade: question was in regards to dealing with the salt

squirrelo: Ah, this is also because we're using the bcrypt library directly. the `hashpw` needs the string and a salt. This salt can either be a new salt if it's not comparing anything, or the hashed password and it pulls the hash from that password for comparison later.

adamrp: @wasade or @squirrelo I'm going to leave this as is right now, but does one of you want to create an issue about this?

squirrelo: Good as is, just no wrapper from other package. No need for issue unless we want to change to the same package as American Gut.

PR #942, opened 2015-03-10
Only show relevant extensions for sample template

### LINKED ISSUE #940: Sample template selection could only allow relevant file extensions
If the file extensions that are allowed for upload are restricted, we can internally make some choices about which files are suggested for processing sample templates (etc). Currently ``.fna`` files are suggested for this:

screenshot 2015-03-10 10 55 27

adamrp:
Fix #940 

Shows txt and tsv files. Any other extensions? I considered csv, but I don't think our parsers actually handle that format (and no one likes CSV files anyways, right?)


ElDeveloper:
Can you add a text describing that only txt and tsv extensions are shown under that dropdown menu?

ElDeveloper:
Also, this should also be added for the prep template.

squirrelo:
:+1:  once timeout failure is done.

qiita_pet/templates/study_description_templates/raw_data_editor_tab.html
```diff
@@ -12,10 +12,12 @@
         <ol>
           <!-- Step 1: choose the prep template file -->
           <li>
-            select your prep template file
+            select your prep template file (only files with the "txt" and "tsv" file extensions will be displayed here):
             <select id="add_prep_template_{{raw_data_id}}">
               {% for f in files %}
+                {% if f.endswith(('txt', 'tsv')) %}
```
wasade: did not know that was allowable...

squirrelo:
Can you table-ize this? It's a bit ragge and messy-looking as is:
screen shot 2015-03-11 at 12 30 11 pm

qiita_pet/templates/study_description_templates/raw_data_editor_tab.html
```diff
@@ -12,10 +12,12 @@
         <ol>
           <!-- Step 1: choose the prep template file -->
           <li>
-            select your prep template file
+            select your prep template file (only files with the "txt" and "tsv" file extensions will be displayed here):
```
squirrelo: Can you do a `<br />` before the added text to break it up a bit? It's really long right now. Same for below.

qiita_pet/templates/study_description_templates/raw_data_editor_tab.html
adamrp: I *think* I actually learned this from reviewing some of your code at some point last year, although I may be misremembering :laughing:

PR #943, opened 2015-03-10
Add tsv to default acceptable upload extensiosn

### LINKED ISSUE #939: .tsv should be an allowed upload extension
This is the default for outputting from Google Spreadsheets as tab-separated text, so it'd be convenient to support this extension.

adamrp:
This is the extension that is added to tab-separated value files when exported
from google docs.

Fix #939


ElDeveloper:
:+1:

PR #945, opened 2015-03-10
Fix #935 - TST: add test studies

### LINKED ISSUE #935: add script for emp study addition
We have a non committed script that adds emp studies to your database that will be good to add. We agreed that having 2-3 small EMP studies in the repo for testing will be OK (remember that we only need the study config file, sample/prep templates, and the biom table) but have all of them in ftp.microbio.me

Jorge-C:
Add three studies that can be cleanly imported into qiita by running the `commands.sh` file by Antonio. Should this be integrated more closely into qiita?

Re. adding every study, there's a lot of work to clean them up: some issues are easy to deal with (invalid characters, nitpicky word casing, duplicated data...) but e.g. what to do when columns with the same name that hold different data (!?)


ElDeveloper:
Thanks, we also need to add this to the .travis.yml file so that we check this works with every build.

squirrelo:
:+1:

.travis.yml
```diff
@@ -1,6 +1,10 @@
 language: python
 env:
-  - PYTHON_VERSION=2.7
+  global:
+    - PYTHON_VERSION=2.7
+  matrix:
```
ElDeveloper: This looks fine, but can you add  a check for `if test_add_studies` in the `after_success` section so we are not running coveralls twice?

test_data_studies/commands.sh
```diff
@@ -0,0 +1,74 @@
...
+    mv ${otu_table}_backup $otu_table
+    echo "Ok"
+
+    # # Making study public
+    echo "\tmaking study public... "
+    echo -e "from qiita_db.study import Study\nStudy(${i}).status = 'public'\n\n" | python
+    echo "Ok"
+done
+
+# Study 1748 is only one line
```
wasade: ?

ElDeveloper:
Please do not merge this, @Jorge-C is working on an unreported failure.

test_data_studies/commands.sh
```diff
@@ -0,0 +1,74 @@
...
+    # Get all needed filepaths for the study
+    conf_fp=test_data_studies/studies/$i/study_config.txt
+    sample_file=test_data_studies/studies/$i/sample_template_$i.txt
+    prep_file=test_data_studies/studies/$i/prep_template_$i.txt
+    otu_table=test_data_studies/studies/$i/otu_table.biom
+
+    echo "Study $i:"
+    # Generate the study config file
+    echo "\tgenerating config file... "
+    echo -e "[required]\ntimeseries_type_id = 1\nmetadata_complete = True\nmixs_compliant = True\nportal_type_id = 3\nprincipal_investigator = Earth Microbiome Project, emp@earthmicrobiome.org, CU Boulder\nreprocess = False\nstudy_alias = $title\nstudy_description = $description\nstudy_abstract = $description\nefo_ids = 1\n[optional]\nstudy_id = $i" > $conf_fp
```
ElDeveloper: CU Boulder -> UCSD? :+1:

test_data_studies/commands.sh
```diff
@@ -0,0 +1,74 @@
...
+    echo "Ok"
+
+    # Loading prep template
+    echo "\tloading prep template... "
+    qiita db load_prep_template $prep_file --raw_data $raw_id --study $study_id --data_type "16S"
+    echo "Ok"
+
+    # Loading preprocessed data
+    echo "\tloading preprocessed data... "
+    mkdir -p temp
```
ElDeveloper: I think this can be moved outside the for loop.

test_data_studies/studies/895/study_config.txt
```diff
@@ -0,0 +1,13 @@
+[required]
```
ElDeveloper: You don't need to add this file or the other config file, these are created for each study.

test_data_studies/commands.sh
```diff
@@ -0,0 +1,74 @@
...
+export PGDATESTYLE="ISO, MDY"
+
+studies=`ls test_data_studies/studies`
+
+echo "DROPPING ENV... "
+qiita_env drop --no-ask-for-confirmation
+echo "Ok"
+
+echo "MAKING ENV... "
+qiita_env make --add-demo-user --no-load-ontologies # --download-reference
```
ElDeveloper: Can you remove the `--download-reference`.

test_data_studies/commands.sh
Jorge-C: :ok: done

test_data_studies/commands.sh
Jorge-C: and remove it at the end with `rmdir`, thanks Yoshiki!

ElDeveloper:
This is ready to be merged, @antgonza can you have a look since you are a residen-bash-script-wizard.

.travis.yml
```diff
@@ -23,13 +27,14 @@ script:
   - ipython profile create qiita_general --parallel
   - qiita_env start_cluster qiita_general
   - qiita_env make --no-load-ontologies
-  - qiita_test_install
-  - nosetests --with-doctest --with-coverage
-  - flake8 qiita_* setup.py scripts/qiita scripts/qiita_env scripts/qiita_test_install
+  - if [ ${TEST_ADD_STUDIES} == "True" ]; then test_data_studies/commands.sh ; fi
+  - if [ ${TEST_ADD_STUDIES} == "False" ]; then qiita_test_install ; fi
```
antgonza: Why multiple ifs? Just curious ...

test_data_studies/commands.sh
```diff
@@ -0,0 +1,75 @@
...
+
+echo "MAKING ENV... "
+qiita_env make --add-demo-user --no-load-ontologies
+echo "Ok"
+
+# Inserting all the information for each study
+echo "INSERTING STUDIES"
+mkdir -p temp
+base_id=0
+for i in ${studies[@]}; do
```
antgonza: This will not be possible to run more than once on the same system without re-downloading some files, right? Will it be possible to fix that?

antgonza:
One comment and one question.

.travis.yml
Jorge-C: If everything was in the same `if` all but the last return codes would be hidden (this already caused a false green build when one command was `cd ; failed command; cd ..`).

Agree it's not optimal, and if travis.yml gets more complex we could add a makefile like in skbio.

test_data_studies/commands.sh
Jorge-C: Do you mean being able to re-run without having to re-download some files? Not sure where this is downloading anything.

.travis.yml
antgonza: OK, thanks.

test_data_studies/commands.sh
antgonza: Something like that. I was thinking in making copies of all the files that get moved and then put them back in place. Similar to the otu table.

josenavas:
Which is the status of this PR?

Jorge-C:
Thanks for pinging Jose! I think I've addressed Antonio's comment (so that now running the script does not add untracked files to the repo).

Jorge-C:
The build failed because a recent change in the metadata loading functionality stopped accepting y/n as values for `has_physical_specimen` and now requires booleans. I'll update the metadata.

PR #949, opened 2015-03-10
Use chosen and validator plugins to validate study creation. Fix #664

### LINKED ISSUE #664: Focus on the PI validation on study creation/edition does not work
Looks like an issue in bootstrap. We need to take a deeper look

squirrelo:


wasade:
I think that's a stochastic failure, restarting tests

qiita_pet/templates/edit_study.html
```diff
@@ -20,147 +20,37 @@
...
-            return {
-              label: text,
-              value: text,
-              option: this
-            };
-        }) );
+$(document).ready(function() {
+    $("#new_person_div").hide();
+     $( "#{{creation_form.principal_investigator.id}}" ).chosen({
+        allow_single_deselect: true,
```
wasade: comma at the end?

wasade:
:+1: on static analysis, pulling down

qiita_pet/templates/edit_study.html
squirrelo: left since it was the first in a list before. Doesn't hurt anything though.

squirrelo:
This can be merged

qiita_pet/templates/edit_study.html
```diff
@@ -202,6 +95,8 @@
 
     $("#{{creation_form.principal_investigator.id}}").append(new_element);
     $("#{{creation_form.lab_person.id}}" ).append(new_element);
+    $("#{{creation_form.principal_investigator.id}}").trigger("chosen:updated");
+      $("#{{creation_form.lab_person.id}}" ).trigger("chosen:updated");
```
ElDeveloper: This indentation is funny.

qiita_pet/templates/edit_study.html
```diff
@@ -172,14 +62,17 @@
...
                 res = false;
             }
           },
       async: false
     })
-    return res;
+    if(res) {
+      form.submit();
+    } else {
+      alert("Study '" + $("#study_title")[0].value + "' already exists! Please choose a different name.");
```
ElDeveloper: Can you use `bootstrapAlert`?

ElDeveloper:
I think this may be another issue, but would be nice if we could fix, see how the submit button gets grayed out:

pr

adamrp:
@ElDeveloper @squirrelo I would say that's blocking, since users cannot proceed with study creation until the issue is fixed

adamrp:
I have finished my review, and I have no further comments after those offered by @ElDeveloper are resolved, but the submit button graying out is a blocker.

qiita_pet/templates/edit_study.html
squirrelo: Not my code but sure.

ElDeveloper: Thanks!

squirrelo:
ready again

ElDeveloper:
@squirrelo looks good, there's just one thing though, the "Principal Investigator" error at the top doesn't clear when the chosen widget changes, see gif:

pr

squirrelo:
There is not a clean way to change that, since it's not actually changing the select but the weird span that  chosen makes. It will still update when you hit the create button. Is this blocking?

josenavas:
My 2 cents. I think it is misleading... It has to be a way to do it, so please explore...

squirrelo:
After a lot of googling and testing it looks like the fact that chosen or any other selection plugin runs on divs and spans makes it nigh impossible to make that message go away after updates.

josenavas:
What is confusing to me is that if you're able to put that message on, why you're not able to put that message away? Note that in the for loop when you're adding the form items, you can put an if on them and add kwargs, where you can add some callback function for the "onchange" or some of those actions. In the previous implementation that I tried (which finally didn't make it) I was able to detect the change and make the highlight go away. The reason why it didn't make it is because the way PI was shown to the user that was required was different from the other fields...

adamrp:
IMHO fixing this is blocking since showing a misleading and inaccurate
error message instead of not showing one is only marginally (arguably)
better than not showing one when it should be shown, which was the behavior
before.

squirrelo:
The issue is that the chosen span changing does not seem to update the underlying select box. Not sure what we want to do with this pull request in light of all this.

squirrelo:
Forced it to work with javascript hacking. I now force a validation when the chosen box is changed.

qiita_pet/templates/edit_study.html
```diff
@@ -20,168 +20,71 @@
...
-            this._trigger( "select", event, {
-              item: ui.item.option
-            });
-          },
-
-          autocompletechange: "_removeIfInvalid"
+$(document).ready(function() {
+    $("#new_person_div").hide();
+     $( "#{{creation_form.principal_investigator.id}}" ).chosen({
+        allow_single_deselect: true,
```
ElDeveloper: Can you remove the trailing comma here? Or is it needed?

ElDeveloper:
One tiny comment, aside from that this now works flawlessly!

qiita_pet/templates/edit_study.html
squirrelo: Not strictly needed but also doesn't harm anything, so I'd call it ready to go.

ElDeveloper: So you don't want to remove this?

squirrelo: I can, but the time it takes to do this, and any type of changes like this in general for future pull requests, would be better spent elsewhere. It's a purely stylistic thing that doesn't do anything to function.

ElDeveloper: Just FYI, the motivation to remove this is to keep (as much as possible) a homogeneous code-base and I pointed this out in case you missed it or added it by accident since this is not done in a consistent way throughout this PR or anywhere else (that I remember).

I wouldn't mind merging as is (granted that someone else decides this should be overlooked), but then I'm unsure what should I highlight as part of the code review and what I shouldn't. Your point is indeed true that there's many other things where time could be better rather than discussing this, and in that light I would like to not discuss this any further.

antgonza: I think we should try avoiding last elements with `,s` as they behave differently in different browsers. For example, the length of this could be 1 or 2 but without the comma it will always be 1. Note that if this change was done when it was first reported, this will had been already merged!!

squirrelo:
ready for merge

PR #952, opened 2015-03-10
ENH: Change exception being raised on prep.remove

### LINKED ISSUE #909: Reuploading sample template
In the past we didn't allow to erase sample/prep templates via GUI so the easiest way to update them was to upload a sample template and automatically delete prep templates. However, if a sample/prep template was already preprocessed it will raise an error only seen by the devs, which is not useful for the user. The current solution will be to not allow reupload as now we can delete prep templates, and have update and extend methods for the sample 

Original comment from @ElDeveloper: 
I added a sample template to a study that already had a preprocessed data. And since this is not allowed, I was sent to the error page. Which is not useful for a regular user, this error should be described and presented to the user, along with a suggestion on how to deal with this:

```
  File "/Users/yoshikivazquezbaeza/.virtualenvs/qiita/lib/python2.7/site-packages/tornado/web.py", line 1144, in _when_complete
  File "/Users/yoshikivazquezbaeza/.virtualenvs/qiita/lib/python2.7/site-packages/tornado/concurrent.py", line 129, in result
  File "/Users/yoshikivazquezbaeza/.virtualenvs/qiita/lib/python2.7/site-packages/tornado/gen.py", line 221, in wrapper
    runner.run()
  File "/Users/yoshikivazquezbaeza/.virtualenvs/qiita/lib/python2.7/site-packages/tornado/gen.py", line 505, in run
  File "/Users/yoshikivazquezbaeza/git_sw/qiita/qiita_pet/handlers/study_handlers/description_handlers.py", line 738, in post
    study, user)
  File "/Users/yoshikivazquezbaeza/.virtualenvs/qiita/lib/python2.7/site-packages/tornado/gen.py", line 533, in run
    self.yield_point.start(self)
  File "/Users/yoshikivazquezbaeza/.virtualenvs/qiita/lib/python2.7/site-packages/tornado/gen.py", line 371, in start
  File "/Users/yoshikivazquezbaeza/git_sw/qiita/qiita_pet/handlers/study_handlers/description_handlers.py", line 181, in process_sample_template
    fp_rsp)
  File "/Users/yoshikivazquezbaeza/git_sw/qiita/qiita_pet/handlers/study_handlers/description_handlers.py", line 548, in remove_add_study_template
    PrepTemplate.delete(pt)
  File "/Users/yoshikivazquezbaeza/git_sw/qiita/qiita_db/metadata_template.py", line 1921, in delete
    "using it." % id_)
 QiitaDBError: Cannot remove prep template 101 because a preprocessed data has been already generated using it. 
```

ElDeveloper:
When removing a prep template that already had processed data, the
remove method would raise an exception that was not caught by the
handler. The solution was to change the QiitaDBError (which is a base
class and I think is to generic) to a QiitaDBExecutionError.

Fixes #909

### DIFF
--- a/qiita_db/metadata_template.py
+++ b/qiita_db/metadata_template.py
@@ -1931,7 +1931,7 @@ class PrepTemplate(MetadataTemplate):
 
         Raises
         ------
-        QiitaDBError
+        QiitaDBExecutionError
             If the prep template already has a preprocessed data
         QiitaDBUnknownIDError
             If no prep template with id = id_ exists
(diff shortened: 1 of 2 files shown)


wasade:
:+1:

PR #956, opened 2015-03-10
Fixes #884

### LINKED ISSUE #884: EBI submission incorrectly reporting metadata headers
Seems to be a count of all the cells in the mapping file, not just number of columns

JWDebelius:
Adds a study-specific metadata header object.

Fixes #884


qiita_db/metadata_template.py
```diff
@@ -1486,6 +1486,15 @@ def extend(self, md_template):
         # adding the fp to the object
         self.add_filepath(fp)
 
+    @property
+    def headers(self):
+        """Identifies the metadata columns present in a template"""
+
+        cols = get_table_cols(self._table_name(self._id))
+        cols.extend(get_table_cols(self._table)[1:])
+
```
squirrelo: Can you run this through the `self.translate_cols_dict` so the column names are consistent with the templates instead of the database?

squirrelo:
One comment.

qiita_pet/handlers/study_handlers/ebi_handlers.py
```diff
@@ -39,7 +39,7 @@ def display_template(self, preprocessed_data_id, msg, msg_level):
         study = Study(preprocessed_data.study)
         stats = [('Number of samples', len(prep_template)),
                  ('Number of metadata headers',
-                  len(sample_template.metadata_headers()))]
+                  len(sample_template.header))]
```
ElDeveloper: missing :zap: s.

qiita_pet/handlers/study_handlers/vamps_handlers.py
```diff
@@ -39,7 +39,7 @@ def display_template(self, preprocessed_data_id, msg, msg_level):
         study = Study(preprocessed_data.study)
         stats = [('Number of samples', len(prep_template)),
                  ('Number of metadata headers',
-                  len(sample_template.metadata_headers()))]
+                  len(sample_template.header))]
```
ElDeveloper: Same here.

qiita_db/metadata_template.py
```diff
@@ -1486,6 +1486,19 @@ def extend(self, md_template):
...
+        for idx, c in enumerate(cols):
+            if c in self.translate_cols_dict:
```
wasade: `cols[idx] = self.translate_cols_dict.get(c, c)`, only a single lookup necessary

wasade:
Isn't this essentially the same as `MetadataTemplate.categories`?

qiita_db/metadata_template.py
```diff
@@ -1146,26 +1146,22 @@ def get_filepaths(self, conn_handler=None):
...
+        cols.extend(get_table_cols(self._table)[1:])
 
-        categories = {c[0] for c in raw}
-        categories.remove('sample_id')
+        for idx, c in enumerate(cols):
+            if c in self.translate_cols_dict:
+                cols[idx] = self.translate_cols_dict[c]
 
-        return categories
+        return set(cols)
```
wasade: the work done in 1160 is not doing anything due to the cast to `set`

the docstring indicates the return is a `list`, however the return is a `set`. It would be great if the return type isn't modified.

ElDeveloper:
Searching the codebase, I found a place where we are still using metadata_headers, I think it has to be changed to categories.

See:
https://github.com/biocore/qiita/blob/b65e614aadfebd77f98bbb3fefedd01e98ce9bb4/qiita_pet/handlers/analysis_handlers.py#L142
https://github.com/biocore/qiita/blob/b65e614aadfebd77f98bbb3fefedd01e98ce9bb4/qiita_pet/handlers/analysis_handlers.py#L217

See also:
https://github.com/biocore/qiita/search?utf8=✓&q=metadata_headers

squirrelo:
Those are supposed to be there: They are for populating the search samples list of available metadta columns

ElDeveloper:
Good, so then this is ready to be merged, thanks!

Issue #915, opened 2015-03-09
`qiita webserver start` error message improvements


jairideout:
From a fresh Qiita install, I get the following error:

```
$ qiita webserver start
Traceback (most recent call last):
  File "/Users/jairideout/.virtualenvs/qiita/bin/qiita", line 10, in <module>
    execfile(__file__)
  File "/Users/jairideout/dev/qiita/scripts/qiita", line 20, in <module>
    from qiita_db.commands import (load_sample_template_from_cmd,
  File "/Users/jairideout/dev/qiita/qiita_db/commands.py", line 23, in <module>
    from .metadata_template import (SampleTemplate, PrepTemplate,
  File "/Users/jairideout/dev/qiita/qiita_db/metadata_template.py", line 1138, in <module>
    class SampleTemplate(MetadataTemplate):
  File "/Users/jairideout/dev/qiita/qiita_db/metadata_template.py", line 1154, in SampleTemplate
    'required_sample_info_status_id': get_required_sample_info_status()}
  File "/Users/jairideout/dev/qiita/qiita_db/util.py", line 217, in get_required_sample_info_status
    con = SQLConnectionHandler()
  File "/Users/jairideout/dev/qiita/qiita_db/sql_connection.py", line 120, in __init__
    self._open_connection()
  File "/Users/jairideout/dev/qiita/qiita_db/sql_connection.py", line 155, in _open_connection
    raise RuntimeError("Cannot connect to database: %s" % str(e))
RuntimeError: Cannot connect to database: FATAL:  role "postgres" does not exist
```

It would be helpful to point users to their Qiita config file and specify what setting(s) need to be added/modified.

wdwvt1:
addressed in #957

wdwvt1:
solved in #957.

PR #957, opened 2015-03-10
issue #915

wdwvt1:


qiita_db/sql_connection.py
```diff
@@ -150,9 +152,19 @@ def _open_connection(self):
 
         try:
             self._connection = connect(**args)
-        except Exception as e:
+        except Exception:
             # catch any exception and raise as runtime error
-            raise RuntimeError("Cannot connect to database: %s" % str(e))
+            # attempt to point the user to their config file
+            conf_fp = environ['QIITA_CONFIG_FP']
```
wasade: this isn't ensured to work. possible for the error to indicate that the qiita config cannot be found either if it doesn't exist?

wasade: or, alternatively, these details could be sourced from `qiita_config`

adamrp:
:+1:

PR #959, opened 2015-03-11
ENH: Use bootstrapAlert function to replace popups

ElDeveloper:
bootstrapAlert is a convenience function to replace the annoying calls
to alert that would show a pop-up.


squirrelo:
:+1:

Issue #919, opened 2015-03-09
Button missing when modifying "Shared With These Users"


Jorge-C:
On the list of "your studies", clicking modify in the column "Shared With These Users" displays a pop up, and after modifying it hitting enter saves the changes, but there is no UI button to accept/cancel the changes.

Jorge-C:
Maybe a quick-fix?

adamrp:
The current implementation immediately applies your changes as you add and remove users from the chosen multiselect using AJAX, so there is no confirm/cancel button. If this is not desirable, then it could be changed. Thoughts?

ElDeveloper:
I think adding some text to describe this behaviour, would be
sufficient.

squirrelo:
Agree, just some text saying it automatically updates would be fine.

ElDeveloper:
Fixed as part of #919

PR #963, opened 2015-03-11
Solves issue #919

wdwvt1:
Adds text that alterts the user to the fact that sharing is updated dynamically and they don't need to click some hidden accept button.


ElDeveloper:
:+1: Looks good to me.

ElDeveloper:
Note this is a spurious failure.

PR #967, opened 2015-03-11
Issue 965

wdwvt1:


qiita_db/sql_connection.py
```diff
@@ -150,16 +150,32 @@ def _open_connection(self):
...
-            # catch any exception and raise as runtime error
-            # attempt to point the user to their config file
-            error_text = ("Can't connect to database; user `%s` is not known. "
-                          "Your QIITA configuration file, located at `%s` must"
-                          " specify the correct user, via the `USER` parameter"
-                          " under the `[postgres]` header. Please refer to "
-                          "`INSTALL.md` in the QIITA base directory.")
-            error_text = error_text % (qiita_config.user, qiita_config.conf_fp)
-            raise RuntimeError(error_text)
+        except Exception as e:
```
ElDeveloper: I just read through the psycopg2 documentation and we should be able to look for specific exceptions here, see the exceptions here:

http://initd.org/psycopg/docs/module.html#psycopg2.InterfaceError 

I'm up to discuss this.

qiita_db/sql_connection.py
```diff
@@ -150,16 +150,32 @@ def _open_connection(self):
...
+            # catch threee known common exceptions and raise runtime errors
+            ebase = ('An error with the following text occurred\n\n\t%s\n%s '
+                     'Please review `INSTALL.md`.')
+            # parse string of expected errors - if too weird, we can't suggest
+            # a solution
+            try:
+                etype = str(e).split(':')[1].split()[0]
```
ElDeveloper: If the raised error didn't include a `:` this would except, so I don't think doing this is a good idea.

qiita_db/sql_connection.py
ElDeveloper: If we are able to find the errors by the exceptions then we should totally use the messages you have below!

wdwvt1:
@ElDeveloper 

I took a look this morning at the `pyscopg2.Error` class and it looked promising. There is an attributes of the error `e.pgcode` that would have been very useful. Unfortunately, the errors that we are dealing with here don't return `e.pcode` because they occur prior to actual initialization of the database. The `pgcode`'s look like they are primarily for problems with accessing a running database with an active `cursor`. As such, I am not sure how we should change whats there now.

ElDeveloper:
Thanks @wdwvt1, got it, thanks for checking that!  It's unfortunate that `pgcode` is not available, what about catching/looking at the specific exception types? Are different (read specific) exceptions raised for the errors that you are trying to catch?

wdwvt1:
I have updated to use `OperationalError` as that is the only type of error that we can catch here prior to the creation of the database. 

As far as the comment about the `split(':')` raising an error - even if the string doesn't contain `:` this will not error - it just returns a list of one element.

ElDeveloper:
Thanks, I will review right now, this is what I meant with the split call:

```python
In [8]: e = 'abcdf'

In [9]: str(e).split(':')[1]
IndexError                                Traceback (most recent call last)
<ipython-input-9-a5be9a456e29> in <module>()
----> 1 str(e).split(':')[1]

```

ElDeveloper:
:+1: if someone else can review this, it should be ready to be merged.

qiita_db/sql_connection.py
```diff
@@ -150,16 +151,33 @@ def _open_connection(self):
...
+                         (qiita_config.user, qiita_config.conf_fp))
+            elif etype == 'Connection':
+                etext = ('This is likely because postgres isn\'t '
+                         'running. Check that postgres is correctly '
+                         'installed and is running.')
+            else:
+                # we recieved a really unanticipated error with a colon
+                etext = ''
+            ebase = ('An OperationalError with the following message occured'
+                     '\n\n\t%s\n%s For more information, review `INSTALL.md`'
```
adamrp: Minor, but could you add another `%s` between `e.message` and `etext`? I think having a blank line there will make the error message easier to read. This is not blocking if we want to get this PR merged now.

adamrp:
Minor nonblocking comment, testing now.

adamrp:
Not sure if we *want* to (probably a UX consideration?), but it might be nice if possible to suppress the traceback output. I think either @jairideout or @ebolyen has done this in the past

adamrp:
The functionality is all there, looks good!

Issue #651, opened 2014-11-20
study abstracts in search criteria 


antgonza:
From @clozupone:
In the information pages for the different studies, it would nice to have the study abstracts in there when available and to make that searchable too.

PR #970, opened 2015-03-11
Ability to view and search abstracts on list pages- Issue 651

squirrelo:
This allows you to search over the abstracts when searching the table of public, private, or shared studies. It also adds links for modals with the study abstracts in them.


rob-knight:
As well as links, can you put the abstract text in mouseover? Often critical for figuring out if the study is worth further investigation from a long list.

On Mar 11, 2015, at 2:23 PM, Joshua Shorenstein <notifications@github.com<mailto:notifications@github.com>> wrote:


This allows you to search over the abstracts when searching the table of public, private, or shared studies. It also adds links for modals with the study abstracts in them.

________________________________
You can view, comment on, or merge this pull request online at:

  https://github.com/biocore/qiita/pull/970

Commit Summary

 *   Add abstract searching/display to public studies list
 *   Add abstract searching/display to private studies list. Fix #651

File Changes

 *   M qiita_pet/handlers/study_handlers/listing_handlers.py<https://github.com/biocore/qiita/pull/970/files#diff-0> (5)
 *   M qiita_pet/templates/private_studies.html<https://github.com/biocore/qiita/pull/970/files#diff-1> (49)
 *   M qiita_pet/templates/public_studies.html<https://github.com/biocore/qiita/pull/970/files#diff-2> (33)

Patch Links:

 *   https://github.com/biocore/qiita/pull/970.patch
 *   https://github.com/biocore/qiita/pull/970.diff

squirrelo:
Abstract text is in a modal that you click on a button to view. Very similar to a mouseover popup but able to copy/paste from it and move your mouse around.

qiita_pet/templates/private_studies.html
```diff
@@ -134,6 +137,22 @@ <h4 class="modal-title" id="myModalLabel">Modify Sharing Settings</h4>
             {% end %}
         </tbody>
     </table>
+
+<!--User study Modals-->
+{% for s in user_studies %}
```
ElDeveloper: Instead of adding all these hidden objects to the DOM, can we instead do this through JavaScript? Also, can you add a close button to the modal window?

qiita_pet/templates/private_studies.html
```diff
@@ -179,6 +200,22 @@ <h4 class="modal-title" id="myModalLabel">Modify Sharing Settings</h4>
             {% end %}
         </tbody>
     </table>
+
+<!--Shared study Modals-->
```
ElDeveloper: Same as above.

ElDeveloper:
:+1: looks good and works great, though there's a major improvement we can make.

qiita_pet/templates/private_studies.html
squirrelo: Nope, it needs to be this way to allow dataframes to search over the abstracts. I can add the close button though.

qiita_pet/templates/private_studies.html
squirrelo: Also same as above

qiita_pet/templates/private_studies.html
```diff
@@ -134,6 +137,25 @@ <h4 class="modal-title" id="myModalLabel">Modify Sharing Settings</h4>
...
+  <div class="modal-dialog modal-med">
+    <div class="modal-content">
+    <div class="modal-header">
+       <h3>{{ s.title }}</h3>
+    </div>
+    <div class="modal-body">
+        {% raw s.abstract %}
+      </div>
+    <div class="modal-footer">
+      <button type="button" class="btn btn-default" data-dismiss="modal">Close</button>
```
ElDeveloper: I was referring to an `x` on the top right.

qiita_pet/templates/private_studies.html
ElDeveloper: I thought the abstract was added to the table itself so it could be searched for (which is done above). These divs are representing the modal windows (which are all going to be written into the page).

squirrelo: Yes. How do you propose doing this in javascript? I'm not sure I can even think of a way to do so. This is the easiest way to do it, sonce it makes the modals and shows them as needed.

ElDeveloper: A way to do this would be to:
1.- add a single div with the modal window.
2.- dynamically populate this modal window using the information that
already exists in the data table.
3.- When the button "show abstract" button is clicked, populate the
single div in window 1 (this should be easily done with jQuery).

If this is still not clear, we can chat off-line.

squirrelo:
This is ready to go.

qiita_pet/templates/private_studies.html
```diff
@@ -44,13 +44,16 @@
             $("#users_list").trigger("chosen:updated");
 
         });
-};
+}
 
+function fillabstract(table, row) {
```
ElDeveloper: This function is being used in two different templates, would you mind centralizing it somewhere? Perhaps qiita.js? And change the name from fillabstract->fillAbstract.

ElDeveloper:
If someone else reviews this, it should be ready to go.

ElDeveloper:
Friendly ping on this needing just one more reviewer.

Issue #953, opened 2015-03-10
Allow to process FASTA studies without quality information


ElDeveloper:
Currently it is only possible to process FASTA studies that include quality files, otherwise you will get a pop-up saying that you need QUAL files.

gregcaporaso:
This seems to be a bigger problem that we initially thought. It turns out that Qiita tries to support this by not passing ``-q`` and ``-d`` if there are no qual files provided, but ``--min_qual_score 25`` also needs to be excluded from the command in this case. 

This pull request adds a more useful error message if a user runs *Preprocess* without both a fasta and qual file.

wasade:
Any luck using a different/new preprocessed parameters set?

gregcaporaso:
I think that's one solution, but I think that's a bit deeper than I'm able to dive into Qiita at the moment (it sounds like that would have to be a change to the database schema, and that those are challenging to put in place). #972 improves the error message in the meantime.

wasade:
Does the act of setting the min qual score value (even to None) force
split_libraries to need qual?

gregcaporaso:
Passing any value for ``--min_qual_score`` makes ``-q`` required. I don't
think passing None would parse as the option parser expects an int.

wasade:
That's annoying... so ya, even a new set of processed params wouldn't work.
It actually would be a small fix to make split_libraries handle this but
that would necessitate 1.9.1 and that would not be fun

antgonza:
As a group we decided to not allow just FASTA to be added in to the system as we can't actually use. Renaming the issue.

antgonza:
We need to change:
- [ ] code and configuration of plugins
- [ ] code in Qiita
- [ ] documentation in Qiita
- [ ] perhaps the upload configuration

PR #972, opened 2015-03-11
improved error handling when attempting to preprocess a fasta without qual

gregcaporaso:
this partially addresses #953


ElDeveloper:
I would like to bring this to the attention of the rest of the @biocore/qiita-dev group. The problem is that the minimum quality value gets added to the `split_libraries.py` command even though you would be processing a study with no quality FASTA files. So the solution is to add a new entry to the database with parameters for sequences with no quality.

If I'm underestimating the problem and this is a bigger issue, this PR does a great job at ameliorating the current situation for the time being, so :+1: .

antgonza:
IMOO this is fine as it makes it clearer in the DB that is another
kind of preprocessing: no quality checking.

wasade:
This sounds good

ElDeveloper:
:+1:

josenavas:
I discussed this with @gregcaporaso and I agree with @antgonza: it is either a new kind of preprocessing (no quality checking) or we need to do quite a few modifications to the database/code base to support that.

PR #974, opened 2015-03-11
Automagically lowercase all required columns. Fix #908

### LINKED ISSUE #908: accept sample_name (lower case) and SAMPLE_NAME (upper case in sample and prep templates

squirrelo:
Does what it says on the tin. Lowercase required columns so we can make sure they exist.

### DIFF
--- a/qiita_db/metadata_template.py
+++ b/qiita_db/metadata_template.py
@@ -45,6 +45,7 @@ from time import strftime
 from functools import partial
 from tempfile import mkstemp
 from os.path import basename
+from StringIO import StringIO
 
 import pandas as pd
 import numpy as np
@@ -2180,8 +2181,8 @@ def load_template_to_dataframe(fn, strip_whitespace=True):
 
     Parameters
     ----------
-    fn : str
-        filename of the template to load
+    fn : str or open file
+        filename of the template to load, or an already open template file
     strip_whitespace : bool, optional
         Defaults to True. Whether or not to strip whitespace from values in the
         input file
@@ -2233,19 +2234,30 @@ def load_template_to_dataframe(fn, strip_whitespace=True):
     |             longitude |        float |
     +-----------------------+--------------+
     """
+    # Load in file lines
+    holdfile = None
+    if isinstance(fn, str):
+        # filepath passed
+        with open(fn) as f:
+            holdfile = f.readlines()
+    else:
+        # open file passed
+        holdfile = fn.readlines()
 
-    # First, strip all values from the cells in the input file, if requested
+    # get and clean the required columns
+    reqcols = set(get_table_cols("required_sample_info"))
+    reqcols.add('sample_name')
+    reqcols.add('required_sample_info_status')
+    reqcols.discard('required_sample_info_status_id')
+    # clean all the column names
+    cols = holdfile[0].split('\t')
+    holdfile[0] = '\t'.join(c.lower() if c.lower() in reqcols else c
+                            for c in cols)
+
+    # Strip all values in the cells in the input file, if requested
     if strip_whitespace:
-        fd, fp = mkstemp()
-        close(fd)
-
-        with open_file(fn, 'U') as input_f, open(fp, 'w') as new_f:
-            for line in input_f:
-                line_elements = [x.strip()
-                                 for x in line.rstrip('\n').split('\t')]
-                new_f.write('\t'.join(line_elements) + '\n')
-
-        fn = fp
+        for pos, line in enumerate(holdfile):
+            holdfile[pos] = '\t'.join(d.strip(' ') for d in line.split('\t'))
 
     # index_col:
     #   is set as False, otherwise it is cast as a float and we want a string
@@ -2262,7 +2274,8 @@ def load_template_to_dataframe(fn, strip_whitespace=True):
     # comment:
     #   using the tab character as "comment" we remove rows that are
     #   constituted only by delimiters i. e. empty rows.
-    template = pd.read_csv(fn, sep='\t', infer_datetime_format=True,
+    template = pd.read_csv(StringIO(''.join(holdfile)), sep='\t',
+                           infer_datetime_format=True,
                            keep_default_na=False, na_values=[''],
                            parse_dates=True, index_col=False, comment='\t',
                            mangle_dupe_cols=False, converters={
@@ -2289,15 +2302,6 @@ def load_template_to_dataframe(fn, strip_whitespace=True):
                 raise QiitaDBColumnError("The '%s' column includes values that"
                                          " cannot be cast into a %s "
                                          "type." % (n, c_dtype))
-
-    # get and clean the required columns
-    reqcols = set(get_table_cols("required_sample_info"))
-    reqcols.add('sample_name')
-    reqcols.add('required_sample_info_status')
-    reqcols.discard('required_sample_info_status_id')
-    # lowercase the required columns for easy matching
-    template.rename(columns=lambda x: x.lower() if x.lower() in reqcols else x,
-                    inplace=True)
 
     initial_columns = set(template.columns)
 
--- a/qiita_db/metadata_template.py
+++ b/qiita_db/metadata_template.py
@@ -49,6 +49,7 @@ import pandas as pd
 import numpy as np
 import warnings
 from skbio.util import find_duplicates
+from skbio.io.util import open_file
 
 from qiita_core.exceptions import IncompetentQiitaDeveloperError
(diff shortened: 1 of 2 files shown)


qiita_db/metadata_template.py
```diff
@@ -2289,6 +2290,15 @@ def load_template_to_dataframe(fn, strip_whitespace=True):
                                          " cannot be cast into a %s "
                                          "type." % (n, c_dtype))
 
+    # get and clean the required columns
```
ElDeveloper: Although this works, this does not enforce the requirements above i.e. latitude/longitude being a number, sample_name always being a string, etc. The normalization to lowercase has to be done before we parse through pandas, otherwise we cannot guarantee these requirements are met.

ElDeveloper: I think a way to do this is by not writing to a temporary file (see above) and instead write to a StringIO object where you can lowercase all the column names of the first line, look find the location of the columns and then replace only those (leaving the non-required columns untouched). I'm happy to discuss this, cc @adamrp

ElDeveloper:
:-1: with potential of becoming :+1:

squirrelo:
Added the changes so everything (including strip whitespace) is done in memory.

qiita_db/metadata_template.py
```diff
@@ -2180,8 +2181,8 @@ def load_template_to_dataframe(fn, strip_whitespace=True):
 
     Parameters
     ----------
-    fn : str
-        filename of the template to load
+    fn : str or open file
```
ElDeveloper: `open file` -> `file-like`.

qiita_db/metadata_template.py
```diff
@@ -2232,19 +2234,30 @@ def load_template_to_dataframe(fn, strip_whitespace=True):
...
-
-        with open_file(fn, 'U') as input_f, open(fp, 'w') as new_f:
-            for line in input_f:
-                line_elements = [x.strip()
-                                 for x in line.rstrip('\n').split('\t')]
-                new_f.write('\t'.join(line_elements) + '\n')
-
-        fn = fp
+        for pos, line in enumerate(holdfile):
+            holdfile[pos] = '\t'.join(d.strip(' ') for d in line.split('\t'))
```
ElDeveloper: Not that you are not stripping the whitespace in the same way as before.

ElDeveloper: Ignore this comment.

qiita_db/metadata_template.py
```diff
@@ -2232,19 +2234,30 @@ def load_template_to_dataframe(fn, strip_whitespace=True):
...
+        holdfile = fn.readlines()
+
+    # get and clean the required columns
+    reqcols = set(get_table_cols("required_sample_info"))
+    reqcols.add('sample_name')
+    reqcols.add('required_sample_info_status')
+    reqcols.discard('required_sample_info_status_id')
+    # clean all the column names
+    cols = holdfile[0].split('\t')
+    holdfile[0] = '\t'.join(c.lower() if c.lower() in reqcols else c
```
ElDeveloper: You have to go through the strip_whitespace block before checking for this otherwise a string like `'lolercoaster '` would not be found, if `'lolercoaster'` was a required column.

qiita_db/test/test_metadata_template.py
```diff
@@ -2451,6 +2458,17 @@ def test_invalid_lat_long(self):
     "2.Sample3\t2014-05-29 12:24:51\tTest Sample 3\tTrue\tTrue\tNotIdentified"
     "\t3\t4.8\t4.41\tlocation1\treceived\ttype1\tValue for sample 3\n")
 
+EXP_SAMPLE_TEMPLATE_MULTICASE = (
+    "Sample_naMe\tcoLLection_timestamp\tdesCription\thas_exTracted_data\t"
```
ElDeveloper: It would be great to have another tests that guaranteed that non-required columns remain with the casing unaltered.

qiita_db/metadata_template.py
```diff
@@ -45,6 +45,7 @@
 from functools import partial
 from tempfile import mkstemp
 from os.path import basename
+from StringIO import StringIO
```
ElDeveloper: Can you import this from the python-future package?

qiita_db/test/test_metadata_template.py
squirrelo: Test updated

ElDeveloper:
:+1:, @adamrp can you review this one?

qiita_db/metadata_template.py
```diff
@@ -2228,20 +2229,32 @@ def load_template_to_dataframe(fn, strip_whitespace=True):
...
     """
+    # Load in file lines
+    holdfile = None
+    if isinstance(fn, (str, unicode)):
+        # filepath passed
+        with open(fn) as f:
+            holdfile = f.readlines()
+    else:
+        # open file passed
+        holdfile = fn.readlines()
```
adamrp: This block (lines 2232-2240) is the exact functionality that skbio.util.io.open_file provides, is there a reason not to use it?

qiita_db/metadata_template.py
```diff
@@ -2228,20 +2229,32 @@ def load_template_to_dataframe(fn, strip_whitespace=True):
...
+    if len(holdfile) == 0:
```
adamrp: Can just be `if not holdfile` since it will either be a list of lines (non-empty file was read), an empty list (empty file was passed), or an exception would have been raised somewhere between lines 2234 and 2240.

qiita_db/metadata_template.py
```diff
@@ -2228,20 +2229,32 @@ def load_template_to_dataframe(fn, strip_whitespace=True):
...
-        with open_file(fn, 'U') as input_f, open(fp, 'w') as new_f:
-            for line in input_f:
-                line_elements = [x.strip()
-                                 for x in line.rstrip('\n').split('\t')]
-                new_f.write('\t'.join(line_elements) + '\n')
-
-        fn = fp
-
+        for pos, line in enumerate(holdfile):
+            holdfile[pos] = '\t'.join(d.strip(' ') for d in line.split('\t'))
```
adamrp: This is only stripping spaces, and not other whitespace characters. Although tab and newline have special meaning in the TSV file and will probably not be found, there are other whitespace characters that might be present in the file that should also be stripped (such as space, tab, linefeed, return, formfeed, or vertical tab). I recommend leaving the argument to `strip` blank.

qiita_db/metadata_template.py
```diff
@@ -2257,7 +2270,8 @@ def load_template_to_dataframe(fn, strip_whitespace=True):
     # comment:
     #   using the tab character as "comment" we remove rows that are
     #   constituted only by delimiters i. e. empty rows.
-    template = pd.read_csv(fn, sep='\t', infer_datetime_format=True,
+    template = pd.read_csv(StringIO(''.join(holdfile)), sep='\t',
```
adamrp: Pretty sure you need to join the elements of the list with `'\n'` instead of `''`

adamrp: actually, you only need to join with `'\n'` if you implement suggestion above to leave argument to `strip` empty

adamrp:
Some comments. Thanks @squirrelo

qiita_db/metadata_template.py
squirrelo: No need since not stripping newlines during whitespace cleaning.

squirrelo:
stochastic failure. Good to go.

adamrp:
Restarted the build, it would be nice if we could luck out and get the stochastic failure passing before merging

qiita_db/metadata_template.py
```diff
@@ -2228,20 +2229,33 @@ def load_template_to_dataframe(fn, strip_whitespace=True):
...
-        with open_file(fn, 'U') as input_f, open(fp, 'w') as new_f:
-            for line in input_f:
-                line_elements = [x.strip()
-                                 for x in line.rstrip('\n').split('\t')]
-                new_f.write('\t'.join(line_elements) + '\n')
-
-        fn = fp
-
+        for pos, line in enumerate(holdfile):
+            holdfile[pos] = '\t'.join(d.strip(" \r\x0b\x0c")
```
ElDeveloper: I'm not sure this was here before.

squirrelo: it was, teh line just got too long for pep8 and I had to cut it in two.

ElDeveloper:
If @adamrp approves, this should be good to go.

qiita_db/metadata_template.py
squirrelo: didnt know it existed. Using it now

PR #980, opened 2015-03-12
Change typechecking to number from float

### LINKED ISSUE #938: processing sample template requires that latitude and longitude has a decimal
For example, ``latitude`` of 40 does not work, but 40.0 does work. Is this intentional, or that that requirement be relaxed?

adamrp:
Fix #938

### DIFF
--- a/qiita_db/metadata_template.py
+++ b/qiita_db/metadata_template.py
@@ -2278,16 +2278,17 @@ def load_template_to_dataframe(fn, strip_whitespace=True):
 
     # let pandas infer the dtypes of these columns, if the inference is
     # not correct, then we have to raise an error
-    columns_to_dtype = [(['latitude', 'longitude'], np.number),
+    columns_to_dtype = [(['latitude', 'longitude'], (np.int, np.float),
+                         'integer or decimal'),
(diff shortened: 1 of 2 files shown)


ElDeveloper:
@adamrp, just one minor thing, can you add a regression test for this problem?

ElDeveloper:
:+1: otherwise.

adamrp:
Thanks @ElDeveloper added regression test, and made some more modifications to the way the types are being checked

ElDeveloper:
@adamrp this faliure is showing up in the Travis build, would you mind having a look?

```
FAIL: test_categories (qiita_db.test.test_metadata_template.DecoratedClass)
  File "/home/travis/build/biocore/qiita/qiita_db/test/test_metadata_template.py", line 1275, in test_categories
    self.assertEqual(obs, exp)
AssertionError: Lists differ: ['env_feature', 'description_d... != ['sample_id', 'season_environm...
First differing element 0:
env_feature
sample_id
Diff is 1213 characters long. Set self.maxDiff to None to see it.
```

adamrp:
Okay, I think it should be good now!  Thanks for the reviews.

wdwvt1:
:+1:

PR #984, opened 2015-03-13
Stop failures maybe?

squirrelo:


ElDeveloper:
Let's repeat this build a few times.

adamrp:
@ElDeveloper :+1:

squirrelo:
This has passed 4 restarts. If it passes this one, I've got the solution and this should be merged. Note this permanently turns off debug mode though, so any changes will require a server restart to see.

antgonza:
NOOOOOOOOOOO!! Any way to have both?

squirrelo:
OK, so here's the gist (hah) of what's going on: In debug mode a poll is added to the ioloop to check files every half second for updates. Since we have so many tests, the checks pile up and end up choking the system and it times out. I've looked into forcing debug to false on the tests, but it seems to be read-only once created. I'll look a little further, but if I can't find a solution tonight this is more critical to merge as is, in my opinion.

squirrelo:
Sometimes the docs are wrong apparently. I can force it because it's just a dict. Ready to go once tests pass.

squirrelo:
Going to close this and open a new one since this was more an exploratory branch.

Issue #950, opened 2015-03-10
study description page timing out in tests


squirrelo:
The study description page is taking a long time to render during tests, which is throwing a 5 second timeout error pretty regularly now.

```python
======================================================================
FAIL: test_post (qiita_pet.test.test_study_handlers.TestStudyDescriptionHandler)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/travis/build/biocore/qiita/qiita_pet/test/test_study_handlers.py", line 114, in test_post
    response = self.post('/study/description/1', post_args)
  File "/home/travis/build/biocore/qiita/qiita_pet/test/tornado_test_base.py", line 57, in post
    return self._fetch(url, 'POST', data, headers)
  File "/home/travis/build/biocore/qiita/qiita_pet/test/tornado_test_base.py", line 62, in _fetch
    return self.wait()
  File "/home/travis/miniconda3/envs/env_name/lib/python2.7/site-packages/tornado/testing.py", line 269, in wait
    self.__rethrow()
  File "/home/travis/miniconda3/envs/env_name/lib/python2.7/site-packages/tornado/testing.py", line 205, in __rethrow
    raise_exc_info(failure)
  File "/home/travis/miniconda3/envs/env_name/lib/python2.7/site-packages/tornado/testing.py", line 253, in timeout_func
    timeout)
AssertionError: Async operation timed out after 5 seconds
-------------------- >> begin captured logging << --------------------
tornado.access: INFO: 200 POST /study/description/1 (127.0.0.1) 5702.12ms
--------------------- >> end captured logging << ---------------------
```

squirrelo:
This is failing for any test in TestStudyDescriptionHandler, get or post, randomly.

antgonza:
This is still failing randomly.

squirrelo:
Where? I've seen travis fail for other things, but not this issue any more.

antgonza:
Got the "Broken" email from master several times last night and had to
restart those builds several times. Those errors have disappeared as
they are now passing.

antgonza:
This hasn't happen in a long time, closing now.

PR #986, opened 2015-03-14
Force debug=False for testing. - URGENT MERGE NEEDED

squirrelo:
This turns debug off for tests. Essentially what was happening is, because of the debug setting, a filecheck was added to ioloop to check files every 500ms. Since each test opened an ioloop, this file check started to get io bound and cause the stochastic failures.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,7 @@ language: python
 env:
   - PYTHON_VERSION=2.7
 before_install:
+  - redis-server --version
   - wget http://repo.continuum.io/miniconda/Miniconda3-3.7.3-Linux-x86_64.sh -O miniconda.sh
   - chmod +x miniconda.sh
   - ./miniconda.sh -b
--- a/qiita_pet/test/tornado_test_base.py
+++ b/qiita_pet/test/tornado_test_base.py
@@ -20,6 +20,7 @@ class TestHandlerBase(AsyncHTTPTestCase):
 
     def get_app(self):
         BaseHandler.get_current_user = Mock(return_value=User("test@foo.bar"))
+        self.app.settings['debug'] = False
         return self.app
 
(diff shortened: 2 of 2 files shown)


squirrelo:
Fixes #950, Found and already tested by 5 restarts of travis in #984

squirrelo:
Unsure why this suddenly kills the connection to the database on travis. Tests pass locally. Looking into it.

squirrelo:
apparently it was just that travis box. All good now.

squirrelo:
UREGENT TO MERGE since this gets rid of the stochastic failures

ElDeveloper:
:+1: @squirrelo, sorry I can't merge it, needs another reviewer :worried:

.travis.yml
```diff
@@ -2,6 +2,7 @@ language: python
 env:
   - PYTHON_VERSION=2.7
 before_install:
+  - redis-server --version
```
josenavas: Can you remove this line? I don't think we need it here. After that, it should be good to go.

squirrelo: It's actually useful since I did need to know what version of the redis server was running. It doesn't harm anything and is good for debugging, so probably best to keep it.

josenavas: Forgot to mention that this is not blocking, meaning that if other @biocore/qiita-dev thinks that this is ok to leave; then he can merge...

josenavas: @squirrelo you mention that it was only in that box. Is that because that box has an older version of redis? If so, this means that we can hit that box again and run into the same issue that in this PR. Is it possible to specify the redis version that we need to travis do not send the build to a box with an outdated version of redis?

squirrelo: It wasn't an outdated version, just not configured properly. I think I have another pull request doing the same thing.

PR #988, opened 2015-03-16
Start improving search

squirrelo:
This puts base changes in place to make the search object more easily extendable. It also breaks the computational part of searches into qiita_ware, and centralizes the search stuff into qiita_ware as a passthrough and computational work.

Additionally, it allows searching over prep template data and data-type of processed data.


squirrelo:
@antgonza @adamrp @ElDeveloper  This is ready for review, and needs to be merged before some work on #900 can continue.

squirrelo:
This isn't actually failing, but for some reason the following failure happense infrequently (see how the other test session passed)
```python
FAIL: test_categories (qiita_db.test.test_metadata_template.DecoratedClass)
  File "/home/travis/build/biocore/qiita/qiita_db/test/test_metadata_template.py", line 1311, in test_categories
    self.assertEqual(obs, exp)
AssertionError: Items in the second set but not the first:
'longitude'
```
I've opened issue for it

qiita_db/data.py
```diff
@@ -1144,6 +1144,21 @@ def study(self):
             "processed_data_id=%s".format(self._study_processed_table),
             [self._id])[0]
 
+    @property
+    def samples(self):
```
josenavas: Change this function to make use of a single SQL query. Right now is executing at least 4.

squirrelo: Ah, this is a vestige of the old way of finding processed data samples. Do we want to keep this function still?

qiita_ware/search.py
```diff
@@ -0,0 +1,59 @@
...
+    fullcount = {}
+    # rearrange all samples so that each metadata column found is its own list
+    meta_vals = zip(*double_comprehension(results))
+    for pos, cat in enumerate(meta_cols):
+        # use Counter object to count all metadata values for a column
+        # pos+1 so we skip the sample names list
+        fullcount[cat] = Counter(meta_vals[pos + 1])
+
+    # Now get metadata counts for each study, removing sample ids as before
+    studycount = {}
```
josenavas: `studycount = defaultdict(dict)` and remove line 43

qiita_ware/search.py
```diff
@@ -0,0 +1,59 @@
+from collections import Counter
+
+from future.utils import viewvalues
+
+from qiita_db.search import QiitaStudySearch
+
+
+def count_metadata(results, meta_cols):
```
josenavas: this is not counting anything, so the name and the description are misleading. Can you change the name to something like `collapse_search_results` or something similar? Also I'm unsure if this function should be exposed to qiita_pet or should be private and be included in the `search` function below, which will make it useful.

qiita_ware/search.py
```diff
@@ -0,0 +1,59 @@
...
+        meta_vals = zip(*(sample for sample in results[study_id]))
+        for pos, cat in enumerate(meta_cols):
+            # use Counter object to count all metadata values for a column
+            # pos+1 so we skip the sample names list
+            studycount[study_id][cat] = Counter(meta_vals[pos + 1])
+
+    return fullcount, studycount
+
+
+def search(searchstr, user, study=None):
```
josenavas: Include the `metadata_counts` function in here, otherwise this function is not worth. Note that there is no problem on using the search object directly on qiita_pet, as we are using the other "data" objects in general.

josenavas:
@squirrelo few comments

qiita_ware/search.py
squirrelo: It is counting the number of occurrences of each metadata value for each metadata colun. It's there specifically for qiita_pet.

qiita_db/data.py
josenavas: If you don't need this property, why are you adding it here? Remove then.

qiita_ware/search.py
squirrelo: It's not needed at all times, so I've broken them up specifically so they can search and get summary as two separate actions. It's not super expensive to get the counts, but it takes long enough to do that I'd rather have them separate than always do it.

qiita_ware/search.py
josenavas: Oh, sorry, I got confused with the code.

qiita_ware/search.py
squirrelo: Would rather leave it as is, since with a defaultdict it won't keyerror if you try and get a study that doesn't exist in the results.

qiita_ware/search.py
josenavas: Ok, in that case, remove this function. There is no need of it (note that on qiita_pey we are using the qiita_db objects directly). Passthrough functions only increase the code complexity, so unless they are doing something else on the results, it is not worth it.

E.g. a probably better solution might be to include a flag `include_counts`. If true, it also returns the counts. If you think that this is not worth, then remove function.

squirrelo: Also to address this function specifically: From the original design specs we wanted no communication between qiita_pet and qiita_db, so this is me trying to do that as much as I can in this case. We are slowing down and stepping back since we aren't in demo heavy mode any more, so I want to start getting the original design ideas going again.

josenavas: We are not going to do Passthrough functions fort every single API call that we have on the qiita_db objects. Please, remove this.

qiita_ware/search.py
josenavas: oh, ok, thanks!

josenavas:
thanks @squirrelo pulled down and it looks like works correctly.
@antgonza can you review/merge? This PR is blocking some development in my PR as I have to change the search object too and, have it merged will be very useful rather than trying to resolve the merge conflicts...

squirrelo:
Also @antgonza note the demo DB you have is not consistent and will not work with this pull request because of that. There are missing prep tables referenced, etc.

antgonza:
Thanks, I fixed those when @josenavas added queues for submission.
However, while trying to search for studies:
```
ERROR:tornado.application:Uncaught exception POST /analysis/2 (::1)
HTTPRequest(protocol='http', host='localhost:21174', method='POST',
uri='/analysis/2', version='HTTP/1.1', remote_ip='::1',
headers={'Origin': 'http://localhost:21174', 'Content-Length': '66',
'Accept-Language': 'en-US,en;q=0.8,es;q=0.6', 'Accept-Encoding':
'gzip, deflate', 'Host': 'localhost:21174', 'Accept':
'text/html,application/xhtml+xml,application/xml;q=0.9,image/webp,*/*;q=0.8',
'User-Agent': 'Mozilla/5.0 (Macintosh; Intel Mac OS X 10_10_2)
AppleWebKit/537.36 (KHTML, like Gecko) Chrome/41.0.2272.89
Safari/537.36', 'Connection': 'keep-alive', 'Referer':
'http://localhost:21174/analysis/2', 'Cache-Control': 'max-age=0',
'Cookie': '_ga=GA1.1.641551533.1412272534;
user="ImFudGdvbnphQGdtYWlsLmNvbSI=|1426701791|debed46f478f366d37f16a99184ef92a6dc26654"',
'Content-Type': 'application/x-www-form-urlencoded'})
  File "/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/site-packages/tornado/web.py",
line 1141, in _when_complete
    callback()
  File "/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/site-packages/tornado/web.py",
line 1162, in _execute_method
  File "/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/site-packages/tornado/web.py",
line 2297, in wrapper
  File "/Users/antoniog/svn_programs/qiita/qiita_pet/handlers/analysis_handlers.py",
line 147, in post
    results, meta_headers = search(query, user)
  File "/Users/antoniog/svn_programs/qiita/qiita_db/search.py", line
275, in __call__
    results = self._process_to_dict(conn_handler.execute_queue('search'),
  File "/Users/antoniog/svn_programs/qiita/qiita_db/sql_connection.py",
line 349, in execute_queue
    self._rollback_raise_error(queue, sql, sql_args, e)
  File "/Users/antoniog/svn_programs/qiita/qiita_db/sql_connection.py",
line 312, in _rollback_raise_error
    str(sql_args), e)))
QiitaDBExecutionError:
Error running SQL query in queue search: SELECT
sr.study_id,sr.processed_data_id,sr.sample_id,sr.study_title FROM
(SELECT * FROM qiita.study s RIGHT JOIN qiita.study_processed_data
USING (study_id) JOIN qiita.required_sample_info USING (study_id) JOIN
qiita.common_prep_info USING (sample_id) JOIN qiita.processed_data
USING (processed_data_id) JOIN qiita.data_type USING (data_type_id)
JOIN qiita.sample_969 USING (sample_id) JOIN qiita.prep_142 USING
(sample_id) ) AS sr WHERE LOWER(sr.study_title) LIKE '%test%'
ARGS: None
Error: column reference "study_id" is ambiguous
LINE 1: SELECT sr.study_id,sr.processed_data_id,sr.sample_id,sr.stud...
               ^

ERROR:tornado.access:500 POST /analysis/2 (::1) 174.84ms
```

josenavas:
Thanks @antgonza just noticed that this PR depends on https://github.com/biocore/qiita/pull/992 so it cannot be merged until that one is merged.

josenavas:
I just noticed that @squirrelo closed the other PR. Anyways, the changes done in here limit the metadata columns, so this PR cannot be merged until the columns are limited in the metadata template and a data patch is in place to fix the existing templates

qiita_db/support_files/patches/python_patches/18.py
```diff
@@ -0,0 +1,27 @@
...
+
+# Get all sample and prep templates with offending columns
+sql = """SELECT DISTINCT table_name from information_schema.columns where
+    table_schema = 'qiita' AND
+    (table_name SIMILAR TO '%%sample[_][0-9,]%%' OR
+    table_name SIMILAR TO '%%prep[_][0-9,]%%') AND
+    (column_name = 'study_id' OR column_name = 'processed_data_id')"""
+
+tables = [c[0] for c in conn_handler.execute_fetchall(sql)]
+for t in tables:
```
josenavas: It is not ok to drop the columns, as you don't know which data are those columns holding. Please re-name the column names.

josenavas:
@squirrelo one more comment

josenavas:
Thanks @squirrelo 
@antgonza can you re-try on your database to see if this fixes the issue?

antgonza:
Nop, :'(

josenavas:
Nop that you can't try or nop it didn't fix the issue?

antgonza:
didn't fix the issue.

squirrelo:
Same error or new error?

josenavas:
@antgonza did you apply the patch?

josenavas:
@squirrelo just noticed that the patch is not complete. You need to re-generate the filepaths for those sample/prep templates that changed and the QIIME mapping file. Check patch num 14 to see how it's done.

squirrelo:
Given the increasing complexity of this pull request and the offline discussion about the schema changes that will have a drastic effect on this pull request, I'm closing this for now.

PR #1000, opened 2015-03-18
Add def column to analysis - Prelim for cart system

squirrelo:
This change adds the 'def' column to the analysis table. This column will be used to mark the default analysis to be used to store selected samples while the selection is still in progress and has not been made an official analysis. Essentially, every user (or, by patch, existing users) will get an analysis that holds these samples, and the def column set to true. All other actually created analyses will be false.

This is the most minimal way to make changes on the DB to support the new selection system, with the code to actually select things coming in a subsequent pull request. In keeping with the mentality of having the DB change pull requests as minimal as possible, I'm adding just the DB patch and nothing else here.

### DIFF
--- a/qiita_db/support_files/patches/18.sql
+++ b/qiita_db/support_files/patches/18.sql
@@ -1,3 +1,3 @@
 -- March 18, 2015
 -- Add column to analysis table to mark shopping cart
-ALTER TABLE qiita.analysis ADD def bool NOT NULL DEFAULT false;
+ALTER TABLE qiita.analysis ADD dflt bool NOT NULL DEFAULT false;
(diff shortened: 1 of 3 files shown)


qiita_db/support_files/patches/18.sql
```diff
@@ -0,0 +1,3 @@
+-- March 18, 2015
+-- Add column to analysis table to mark shopping cart
+ALTER TABLE qiita.analysis ADD def bool NOT NULL DEFAULT false;
```
josenavas: I know you can't use "default" as a column name, but def is confusing. Can you use `dflt` as that is one of the standard default acronyms in CS?

squirrelo: Sure, fixed.

josenavas:
:+1:

squirrelo:
longitude failed again, so good to merge since that is stochastic.

squirrelo:
Anyone? Needed to continue the cart work.

ElDeveloper:
Issue :hash: :one: :zero: :zero: :zero:  :chart_with_upwards_trend:

PR #1004, opened 2015-03-19
Fixes #1003

### LINKED ISSUE #1003: Stats page is broken
Broken line: https://github.com/biocore/qiita/blob/master/qiita_pet/handlers/stats.py#L62

`Study.get_by_status` returns a set not a list, so that line no longer works. Issuing a PR soon.

josenavas:
Note that this is broken on master and on the live system


ElDeveloper:
:+1: granted that tests pass.

PR #1006, opened 2015-03-20
DataTables load using AJAX, Search by metadata on studies page

squirrelo:
This pull request lays the groundwork for selections by study on the study listing page. The following changes have happened:
* New Study.get_info() function to get all information for a given list of studies and, optionally, columns. This allows for orders of magnitude faster loading of study data for the listings tables.
* New SearchStudiesAJAX handler and page, for getting the DataTables json information to build the tables.
* Merged the public and user studies tables and pages into a single table that can be sorted by status, and allows for a single listing page that can be searched over.
* DataTables are now loaded from AJAX-derived json, meaning the page no longer has a wait logo and can reload the table as metadata searches are performed.
* Minor DB changes - Renaming the description and title columns on study satellite tables to allow for easier joins.


squirrelo:
SciPy build is failing, causing the tests to register as failed. Not sure what to do.

```
 Command "/home/travis/miniconda3/envs/env_name/bin/python -c "import setuptools, tokenize;__file__='/tmp/pip-build-baUMF0/scipy/setup.py';exec(compile(getattr(tokenize, 'open', open)(__file__).read().replace('\r\n', '\n'), __file__, 'exec'))" install --record /tmp/pip-t4sETC-record/install-record.txt --single-version-externally-managed --compile" failed with error code 1 in /tmp/pip-build-baUMF0/scipy
```

qiita_db/study.py
```diff
@@ -176,6 +182,28 @@ def get_by_status(cls, status):
         return {x[0] for x in conn_handler.execute_fetchall(sql, (status, ))}
 
     @classmethod
+    def get_info(cls, study_ids, info_cols=None):
```
ElDeveloper: Can you add a docstring?

qiita_db/study.py
```diff
@@ -176,6 +182,28 @@ def get_by_status(cls, status):
...
+        if info_cols is None:
+            info_cols = [s for s in cls._info_cols]
+        else:
+            info_cols = deepcopy(info_cols)
```
ElDeveloper: It doesn't seem like there's any need for this variable to be copied, am I missing something?

squirrelo: remnant of old way of doing something, thanks for catching.

qiita_db/study.py
squirrelo: done

squirrelo:
This is failing because of qiime install issues, but is ready for review.

josenavas:
Since I've already faced 2 database merge conflicts on my previous PR #900 please, review that one and merge that one first...

qiita_db/study.py
```diff
@@ -176,6 +182,41 @@ def get_by_status(cls, status):
...
+        Returns
+        -------
+        list of DictCursor
+            Table-like structure of metadata, one study per row. Can be
+            accessed as a list of dictionaries.
+        """
+        if info_cols is None:
+            info_cols = cls._info_cols
+
+        search_cols = ",".join(sorted(cls._info_cols.intersection(info_cols)))
```
adamrp: Might raise a warning if the user requests information columns that are not in `cls._info_cols`, since right now it silently ignores them, which could lead to bugs if/when there are typos or misspellings

qiita_db/test/test_study.py
```diff
@@ -177,6 +177,49 @@ def _make_sandbox(self):
...
+            ' the DNA sequenced. Roots from three independent plants of '
+            'different strains were examined. These roots were obtained '
+            'November 11, 2011 from plants that had been harvested in the '
+            'summer. Future studies will attempt to analyze the soils and '
+            'rhizospheres from the same location at different time points in '
+            'the plant lifecycle.', 'Cannabis Soils', 'Analysis of the '
+            'Cannabis Plant Microbiome', 1, 3, 'Identification of the '
+            'Microbiomes for Cannabis Soils', 'None', 1, None]
+        self.assertItemsEqual(obs[0].keys(), exp_keys)
+        self.assertItemsEqual(obs[0], exp_vals)
```
adamrp: Why not define your expected keys/values as an actual dict and use `self.assertEqual`? As of python 2.7, the correct equality method will automatically be called. This is not blocking, just thought it was a little unintuitive.

qiita_pet/handlers/study_handlers/listing_handlers.py
```diff
@@ -35,41 +39,78 @@ def _get_shared_links_for_study(study):
     return ", ".join(shared)
 
 
-def _build_study_info(studytype, user=None):
+def _build_study_info(studytype,  user, results=None):
```
adamrp: one extra space here

qiita_pet/handlers/study_handlers/listing_handlers.py
```diff
@@ -35,41 +39,78 @@ def _get_shared_links_for_study(study):
...
-                            'pmids owner status abstract')
+    # get list of studies for table
+    if studytype == "standard":
+        studylist = user.user_studies | Study.get_by_status('public')
+    elif studytype == "shared":
+        studylist = user.shared_studies
+    if results is not None:
+        studylist = studylist.intersection(results)
+    if not studylist:
+        # No studies lleft so no need to continue
```
adamrp: lleft -> left

qiita_pet/templates/list_studies.html
```diff
@@ -0,0 +1,266 @@
...
+              { "data": "id" },
+              { "data": "meta_complete" },
+              { "data": "num_samples" },
+              { "data": "num_raw_data" },
+              { "data": "shared" },
+              { "data": "pi" },
+              { "data": "pmid" },
+              { "data": "status" }
+            ],
+            order: [[10, "asc"], [ 1, "asc" ]],
```
adamrp: Does this sort first by status and then by title? I would think sorting just by title would be a user's expectation, but I am not sure

qiita_pet/test/test_study_handlers.py
```diff
@@ -12,46 +12,51 @@
...
+        }]
+        super(TestHelpers, self).setUp()
+
     def test_get_shared_links_for_study(self):
         obs = _get_shared_links_for_study(Study(1))
         exp = '<a target="_blank" href="mailto:shared@foo.bar">Shared</a>'
         self.assertEqual(obs, exp)
 
     def test_build_study_info(self):
+        self.maxDiff = None
```
adamrp: Is this supposed to be here, or was it just something that was useful while you were developing?

qiita_pet/test/test_study_handlers.py
```diff
@@ -12,46 +12,51 @@
...
-                'These roots were obtained November 11, 2011 from plants that '
-                'had been harvested in the summer. Future studies will '
-                'attempt to analyze the soils and rhizospheres from the same '
-                'location at different time points in the plant lifecycle.')]
-        self.assertEqual(obs, exp)
+        obs = _build_study_info('standard', User('test@foo.bar'))
+        self.assertEqual(obs, self.exp)
 
     def test_build_study_info_new_study(self):
+        self.maxDiff = None
```
adamrp: same question about maxDiff

qiita_pet/templates/list_studies.html
```diff
@@ -0,0 +1,266 @@
+{% extends sitebase.html %}
+{% block head %}
+<link rel="stylesheet" href="/static/vendor/css/jquery.dataTables.css" type="text/css">
+
+<script src="/static/vendor/js/jquery.dataTables.min.js"></script>
+<script src="/static/vendor/js/jquery.dataTables.plugin.natural.js"></script>
+
+<script type="text/javascript">
+var current_study;
+var ajaxURL = "/study/search/?&user={{current_user.id}}&sEcho=" + Math.floor(Math.random()*1001);
```
adamrp: Here you use  `*1001`, but later there are `sEcho` values of `1021` in your test code. Does this discrepancy matter? I admit, I don't fully understand the `sEcho` parameter here. Is it used to identify a particular request? If so, why do the requests need to be identified in this way? Lastly, if these are being used as identifiers, this does not allow for a tremendous number of simultaneous searches, why not increase 1001 to 1000001 or something?

adamrp:
A couple small comments, and a question, but the code looks good. Going to test now.

adamrp:
The behavior seems right to me, so :+1: when the comments are addressed

qiita_pet/templates/list_studies.html
squirrelo: sEcho is just a numeric value you need to pass and then echo back so DataTables will recognize it as a response handshake of sorts. Any number works, I just happened to choose 1021 randomly in the test code. It could be a static number and still work, but I wanted to add some variability to try and decrease attack odds.

qiita_pet/test/test_study_handlers.py
squirrelo: helpful for debugging, yes

qiita_pet/templates/list_studies.html
squirrelo: yes, status then title. Figured it was best to have all public then all your own stuff separated in some form at first glance.

qiita_pet/templates/list_studies.html
wasade: If the goal is to reduce the possibility of attack, and if this is a vulnerability, then I recommend using a cryptographically secure PRNG. Details [here](http://stackoverflow.com/questions/5651789/is-math-random-cryptographically-secure).

squirrelo: Ok, attack was wrong word. More just to add a bit of randomness rather than selecting a single number, that way we know the function is actually working and not just spitting back garbage.

qiita_db/study.py
```diff
@@ -143,6 +145,11 @@ class Study(QiitaStatusObject):
     _table = "study"
     # The following columns are considered not part of the study info
     _non_info = {"email", "study_status_id", "study_title"}
+    # The following tables are considered part of info
+    _info_cols = frozenset(chain(
```
wasade: None of this information will change at runtime, right? If so, I think this could be cached in `qiita_pet.__init__` and done once

wasade: ah! nevermind, it is at the static scope of the object. it is done once

qiita_db/study.py
```diff
@@ -176,6 +183,44 @@ def get_by_status(cls, status):
         return {x[0] for x in conn_handler.execute_fetchall(sql, (status, ))}
 
     @classmethod
+    def get_info(cls, study_ids, info_cols=None):
+        """Returns study data for a set of study_ids
+
+        Parameters
+        ----------
+        study_cols : list of ints
```
wasade: study_ids?

qiita_pet/handlers/study_handlers/listing_handlers.py
```diff
@@ -35,41 +39,78 @@ def _get_shared_links_for_study(study):
...
     """builds list of namedtuples for study listings"""
-    if studytype == "private":
-        studylist = user.user_studies
-    elif studytype == "shared":
-        studylist = user.shared_studies
-    elif studytype == "public":
-        studylist = Study.get_by_status('public')
-    else:
+    if studytype not in {"standard", "shared"}:
         raise IncompetentQiitaDeveloperError("Must use private, shared, "
```
wasade: The error message doesn't seem to correspond to the conditional?

qiita_pet/handlers/study_handlers/listing_handlers.py
```diff
@@ -35,41 +39,78 @@ def _get_shared_links_for_study(study):
...
                                              "or public!")
-
-    StudyTuple = namedtuple('StudyInfo', 'id title meta_complete '
-                            'num_samples_collected shared num_raw_data pi '
-                            'pmids owner status abstract')
+    # get list of studies for table
+    if studytype == "standard":
+        studylist = user.user_studies | Study.get_by_status('public')
```
wasade: if possible, it would be nice to stick with either operators (this line), or methods (line 53), to improve consistency within the method being worked on.

qiita_pet/handlers/study_handlers/listing_handlers.py
```diff
@@ -35,41 +39,78 @@ def _get_shared_links_for_study(study):
...
+        if info["number_samples_collected"] is None:
+            info["number_samples_collected"] = "None"
         shared = _get_shared_links_for_study(study)
-        infolist.append(StudyTuple(study.id, study.title,
-                                   info["metadata_complete"],
-                                   info["number_samples_collected"],
-                                   shared, len(study.raw_data()),
-                                   PI, pmids, owner, status,
-                                   info["study_abstract"]))
+        meta_complete = "ok" if info["metadata_complete"] else "remove"
```
wasade: remove the metadata if it is incomplete?

squirrelo: glyph icon styling code

qiita_pet/handlers/study_handlers/listing_handlers.py
```diff
@@ -35,41 +39,78 @@ def _get_shared_links_for_study(study):
...
+                 "data-target='#study-abstract-modal' "
+                 "onclick='fillAbstract(\"{0}-studies-table\", {1})'>"
+                 "<span class='glyphicon glyphicon-file' "
+                 "aria-hidden='true'></span></a> | "
+                 "<a href='/study/description/{2}' "
+                 "id='study{1}-title'>{3}</a>").format(
+                     studytype, str(row), str(study.id), info["study_title"])
+        meta_complete = "<span class='glyphicon glyphicon-%s'></span>" % \
+            meta_complete
+        shared = "Not Available" if info["status"] == 'public' else \
```
wasade: Could you break this apart please? The ternary operator is more difficult to interpret when its use exceeds the max line length

qiita_pet/handlers/study_handlers/listing_handlers.py
```diff
@@ -35,41 +39,78 @@ def _get_shared_links_for_study(study):
...
         PI = StudyPerson(info['principal_investigator_id'])
         PI = study_person_linkifier((PI.email, PI.name))
-        pmids = ", ".join([pubmed_linkifier([pmid])
-                           for pmid in study.pmids])
+        if info['pmid'] is not None:
+            pmids = ", ".join([pubmed_linkifier([p])
+                               for p in info['pmid']])
+        else:
+            pmids = ""
+        if info["number_samples_collected"] is None:
```
wasade: Is there a reason that `number_samples_collected` isn't pulled out into a variable while something like `metadata_complete` is?

qiita_pet/handlers/study_handlers/listing_handlers.py
```diff
@@ -169,3 +184,62 @@ def get(self):
...
+                self.clear()
+                self.set_status(400)
+                searchmsg = ''.join(e)
+                self.write(searchmsg)
+                return
+            except:
+                # catch any other error as generic server error
+                self.clear()
+                self.set_status(500)
+                self.write("Server error during search. Please try again "
```
wasade: Isn't this indicative of an actual problem that should be logged?

qiita_pet/handlers/study_handlers/listing_handlers.py
squirrelo: Styling reasons. We need the actual number in number_samples_collected, and this just cleans the None to actually be text if it is None, otherwise it is untouched. With the metadata_complete it's changed to something completely different for styling, so not reusing the same variable for that.

qiita_pet/handlers/study_handlers/listing_handlers.py
```diff
@@ -169,3 +184,62 @@ def get(self):
...
+
+    @authenticated
+    def get(self, ignore):
+        search_type = self.get_argument('type')
+        user = self.get_argument('user')
+        query = self.get_argument('query')
+        echo = int(self.get_argument('sEcho'))
+
+        res = None
+        if query != "":
```
wasade: `if query`

qiita_pet/handlers/study_handlers/listing_handlers.py
```diff
@@ -169,3 +184,62 @@ def get(self):
...
+            except:
+                # catch any other error as generic server error
+                self.clear()
+                self.set_status(500)
+                self.write("Server error during search. Please try again "
+                           "later")
+                return
+            if not res:
+                res = {}
+            info = _build_study_info(search_type, self.current_user,
```
wasade: this call only needs to be done once and can be done outside of the conditions if `res` is set to `None` in the `else` block

qiita_pet/handlers/study_handlers/listing_handlers.py
squirrelo: Yes. Wasn't logged in the code I copy-pasted to here but can add it.

qiita_pet/test/test_study_handlers.py
```diff
@@ -12,6 +12,37 @@
 class TestHelpers(TestHandlerBase):
     database = True
 
+    def setUp(self):
+        self.exp = [{
+            'status': 'public',
+            'checkbox': "<input type='checkbox' value='1' />",
+            'abstract': 'This is a preliminary study to examine the microbiota'
+            ' associated with the Cannabis plant. Soils samples from the bulk '
```
wasade: Aligning the quotes will improve the ability to differentiate between keys and values, for instance:

```python
self.exp = [{
    ...
    'abstract': 'foo bar '
                'more and more and more text '
                'even more text',
    'owner': 'stuff',
    ...
}]
```

qiita_pet/handlers/study_handlers/listing_handlers.py
```diff
@@ -169,3 +184,62 @@ def get(self):
...
+            search = QiitaStudySearch()
+            try:
+                res, meta = search(query, User(user))
+            except ParseException:
+                self.clear()
+                self.set_status(400)
+                self.write('Malformed search query. Please read "search help" '
+                           'and try again.')
+                return
+            except QiitaDBIncompatibleDatatypeError as e:
```
wasade: Under what types of situations would this arise? I didn't see a test case on it, so just curious

squirrelo: someone tries to search over both varchar and numeric on same column

qiita_db/support_files/patches/19.sql
```diff
@@ -0,0 +1,6 @@
+-- March 19, 2015
+-- Rename columns to be more descirptive and allow easier joins
+ALTER TABLE qiita.study_status RENAME COLUMN description TO status_description;
+ALTER TABLE qiita.portal_type RENAME COLUMN description TO portal_description;
+ALTER TABLE qiita.investigation RENAME COLUMN description TO investigation_description;
+ALTER TABLE qiita.investigation RENAME COLUMN name TO investigation_name;
```
adamrp: I'm surprised that these changes don't affect more code, but I haven't had a chance to look closely

qiita_pet/handlers/study_handlers/listing_handlers.py
adamrp: Like @wasade, it took me a minute to realize this while looking at the code; perhaps change the name of the variable to `metadata_complete_glyphicon` or `metadata_complete_glyph` or something similar

qiita_pet/handlers/study_handlers/listing_handlers.py
adamrp: please do :+1:

qiita_pet/handlers/study_handlers/listing_handlers.py
adamrp: :+1:

qiita_pet/templates/list_studies.html
adamrp: The documentation on this parameter is extremely sparse. Most users' experiences are that it must be specified, but that the value seems not to matter very much. [This is the official word](http://legacy.datatables.net/usage/server-side) on it, and it does mention XSS attacks, but how it helps, I don't know.

squirrelo: Yeah, that's pretty much it. I just have it as a random number so I can reasonably know the function is actually acting on the data sent to it, and not something else.

qiita_db/support_files/patches/19.sql
squirrelo: None of these columns are actually used anywhere, since investigations and portal types aren't implemented yet, and study_status description is never touched, only the status value itself.

qiita_pet/handlers/study_handlers/listing_handlers.py
squirrelo: woops, forgot to update. fixed.

qiita_pet/handlers/study_handlers/listing_handlers.py
squirrelo: will stick with methods, since they are more universal.

squirrelo:
ready for review/merge

qiita_db/study.py
```diff
@@ -199,6 +206,43 @@ def get_by_status(cls, status):
         return studies
 
     @classmethod
+    def get_info(cls, study_ids, info_cols=None):
```
josenavas: The name of this function is a bit confusing to me; I actually had to go and check how the instance method is called. Can we change it to something like `def get_info_by_study` and `study_ids` is optional (if not provided, it gives all the information for all the studies, probably useful from an admin point of view - that has access to everything).

qiita_db/support_files/patches/19.sql
```diff
@@ -31,4 +31,4 @@ UPDATE qiita.processed_data as pd
 -- We no longer need the status column on the study table
 ALTER TABLE qiita.study DROP CONSTRAINT fk_study_study_status;
 DROP INDEX qiita.idx_study_0;
-ALTER TABLE qiita.study DROP COLUMN study_status_id;
\ No newline at end of file
+ALTER TABLE qiita.study DROP COLUMN study_status_id;
```
josenavas: Can you undo this change?

qiita_db/study.py
```diff
@@ -199,6 +206,43 @@ def get_by_status(cls, status):
...
+        Returns
+        -------
+        list of DictCursor
+            Table-like structure of metadata, one study per row. Can be
+            accessed as a list of dictionaries.
+        """
+        if info_cols is None:
+            info_cols = cls._info_cols
+        elif not cls._info_cols.issuperset(info_cols):
+            warnings.warn("Non-info columns passed: %s" % ", ".join(
```
josenavas: Is this a warning or an error?

qiita_db/support_files/patches/20.sql
```diff
@@ -0,0 +1,6 @@
+-- March 19, 2015
+-- Rename columns to be more descriptive and allow easier joins
+ALTER TABLE qiita.processed_data_status RENAME COLUMN description TO status_description;
```
josenavas: `processed_data_status_description` so future changes do not break this (i.e. there are other status tables, and if they eventually get added, it will break this...)

qiita_db/test/test_study.py
```diff
@@ -179,6 +179,54 @@ def _change_processed_data_status(self, new_status):
             "UPDATE qiita.processed_data SET processed_data_status_id = %s",
             (id_status,))
 
+    def test_get_info(self):
```
josenavas: Can you add another test in which there is more than 1 study? You can use the study create to add a new study (note that if the creation fails, we already have tests for the create function that will fail....)

qiita_pet/handlers/study_handlers/listing_handlers.py
```diff
@@ -37,41 +43,80 @@ def _get_shared_links_for_study(study):
     return ", ".join(shared)
 
 
-def _build_study_info(studytype, user=None):
+def _build_study_info(studytype, user, results=None):
     """builds list of namedtuples for study listings"""
```
josenavas: Update the docstring (namedtuples no longer apply). Although this function is private, I think having a bit of numpy doc will help future developers...

qiita_pet/handlers/study_handlers/listing_handlers.py
```diff
@@ -37,41 +43,80 @@ def _get_shared_links_for_study(study):
     return ", ".join(shared)
 
 
-def _build_study_info(studytype, user=None):
+def _build_study_info(studytype, user, results=None):
```
josenavas: Actually, `studytype` do not follow pep8: `study_type`

qiita_pet/handlers/study_handlers/listing_handlers.py
```diff
@@ -37,41 +43,80 @@ def _get_shared_links_for_study(study):
...
     """builds list of namedtuples for study listings"""
-    if studytype == "private":
-        studylist = user.user_studies
+    if studytype not in {"standard", "shared"}:
+        raise IncompetentQiitaDeveloperError("Must use standard or shared!")
```
josenavas: I might be an `IncompetentQiitaDeveloper`; but I'll always raise this error when I use this function because I do not know the meaning of the study_type parameter (standard? shared?).

qiita_pet/handlers/study_handlers/listing_handlers.py
```diff
@@ -37,41 +43,80 @@ def _get_shared_links_for_study(study):
...
         shared = _get_shared_links_for_study(study)
-        infolist.append(StudyTuple(study.id, study.title,
-                                   info["metadata_complete"],
-                                   info["number_samples_collected"],
-                                   shared, len(study.raw_data()),
-                                   PI, pmids, owner, status,
-                                   info["study_abstract"]))
+        meta_complete_glyph = "ok" if info["metadata_complete"] else "remove"
+        # build the HTML elements needed for table cell
+        title = ("<a href='#' data-toggle='modal' "
```
josenavas: Is this bunch of html strictly needed here? or the template can be modified so this bit of html is included there? When I refactored the study pages I tried to remove the html from the code...

qiita_pet/handlers/study_handlers/listing_handlers.py
```diff
@@ -37,41 +43,80 @@ def _get_shared_links_for_study(study):
...
+                 "aria-hidden='true'></span></a> | "
+                 "<a href='/study/description/{2}' "
+                 "id='study{1}-title'>{3}</a>").format(
+                     studytype, str(row), str(study.id), info["study_title"])
+        meta_complete = "<span class='glyphicon glyphicon-%s'></span>" % \
+            meta_complete_glyph
+        if status == 'public':
+            shared = "Not Available"
+        else:
+            shared = ("<span id='shared_html_{0}'>{1}</span><br/>"
```
josenavas: Same here

qiita_pet/handlers/study_handlers/listing_handlers.py
```diff
@@ -82,47 +127,21 @@ def _check_owner(user, study):
...
-        self.write(self.render_string('waiting.html'))
-        self.flush()
-        user = self.current_user
-        user_studies = yield Task(self._get_private, user)
-        shared_studies = yield Task(self._get_shared, user)
         all_emails_except_current = yield Task(self._get_all_emails)
         all_emails_except_current.remove(self.current_user.id)
-        self.render('private_studies.html',
-                    user_studies=user_studies, shared_studies=shared_studies,
+        availmeta = SampleTemplate.metadata_headers() +\
```
josenavas: `availmeta` is not pep8 -> `avail_meta`

qiita_pet/handlers/study_handlers/listing_handlers.py
```diff
@@ -172,3 +191,64 @@ def get(self):
...
+        callback(_build_study_info("standard", user))
+
+    def _get_shared(self, user, callback):
+        """builds list of tuples for studies that are shared with user"""
+        callback(_build_study_info("shared", user))
+
+    @authenticated
+    def get(self, ignore):
+        search_type = self.get_argument('type')
+        user = self.get_argument('user')
```
josenavas: Why are you passing the user as an argument? It doesn't look like is needed as...

qiita_pet/handlers/study_handlers/listing_handlers.py
```diff
@@ -172,3 +191,64 @@ def get(self):
...
+        echo = int(self.get_argument('sEcho'))
+
+        if user != self.current_user.id:
+            raise HTTPError(403, 'Unauthorized search!')
+        res = None
+        if query:
+            # Search for samples matching the query
+            search = QiitaStudySearch()
+            try:
+                res, meta = search(query, User(user))
```
josenavas: just use self.current_user (less objs instantiations)

josenavas:
Just a general note. I do not agree with the fact that my private studies and the public studies are merged in the same list. A public study is not "mine" so that is not correct. I thought that what we agreed is to have the three lists in a single page, I do not recall saying anything about merging the public and the private list in a single list...

rob-knight:
I think it should be optional to view them separately or intermingled — the counter-argument is that it’s confusing to not see studies you know you have access to, and there are many reasons why you would want to sort them together in one list.

On Mar 27, 2015, at 11:53 AM, josenavas <notifications@github.com<mailto:notifications@github.com>> wrote:


Just a general note. I do not agree with the fact that my private studies and the public studies are merged in the same list. A public study is not "mine" so that is not correct. I thought that what we agreed is to have the three lists in a single page, I do not recall saying anything about merging the public and the private list in a single list...

josenavas:
Agree on being optional. In such case we have either 1 list or 3 lists, but I do not see the point on having 2 lists (private + public; and shared...). @squirrelo since you have the status property, you can use to put public, private or shared. The current configuration is the one that is misleading, at least IMOO.

qiita_db/test/test_study.py
squirrelo: It's tested indirectly in the study handler tests, but if you want one here I can do it.

squirrelo:
Single list is much easier to program, and would make the "refine by columns" make much more sense since searching in one box searches over all results, not just the ones in the given box.

qiita_db/study.py
squirrelo: A warning, as suggested by @adamrp.

qiita_pet/handlers/study_handlers/listing_handlers.py
squirrelo: This is all in here because it's returned as part of the json that is used to fill the table. Therefore you need the full column data to put in each cell.

qiita_pet/handlers/study_handlers/listing_handlers.py
squirrelo: same answer

qiita_pet/handlers/study_handlers/listing_handlers.py
squirrelo: line 211

josenavas:
I see the "easy to program" the single list, but I still would prefer to have the ability to have the 3 lists, as if I'm interested in my own study I don't have to search it over the twelve million public ones....

qiita_pet/handlers/study_handlers/listing_handlers.py
josenavas: That's what I'm saying... it's not needed, just use self.current_user...

qiita_db/test/test_study.py
josenavas: Something on qiita_db cannot be safely tested through something on qiita_pet... please add test here.

squirrelo:
Single list for now, Issue #1019 opened for later multi-table support.

qiita_pet/handlers/study_handlers/listing_handlers.py
squirrelo: It's to prevent URL hacking

josenavas:
Thanks @squirrelo, baby steps works, so PR are smaller. Assigned to you and to alpha.

josenavas:
BTW, this PR is not modifying the DB schema files neither the html. See [this section](https://github.com/biocore/qiita/blob/master/CONTRIBUTING.md#making-database-changes) on the contributing.md file for making database changes.

squirrelo:
Ready for review/merge

qiita_db/test/test_study.py
antgonza: You added a new test but it is still testing 1 single study, could you test multiple? `Study.get_info([1, 2])`

qiita_pet/handlers/study_handlers/listing_handlers.py
antgonza: I'm not sure this should raise an error. Let's have 2 studies with column Ph, one has only float values (created as float) and the other one has a combination of float/str (created as str). If a user searches for Ph = 0.4 it should return the samples that match that criteria from both studies, right?

antgonza:
Some comments on code. 

The GUI looks soooo much better, thanks! Anyway, when I try to search the button gets grayed out but nothing happens, not even python or JS errors, and never comes back. Also, not sure what the left checkbox column is for, there is no button to add or anything like that.

qiita_db/test/test_study.py
squirrelo: line 232

qiita_pet/handlers/study_handlers/listing_handlers.py
squirrelo: if it has a combination of float/str, the entire column will be varchar and we can't do proper numeric comparisons vs a string. This is a known issue with the search engine that will get fixed during the refactor. Before this explicit error raising the search would just 500 error if you do this type of search.

squirrelo:
Checkbox is a placeholder for sample selection, which will be added once all the initial bits are in place. Doesn't harm anything, and means there is less to change to get full functionality.

I'll take a look at the JS, sounds like browser compatibility issue.

qiita_db/test/test_study.py
antgonza: Got it, I was reading how it works wrongly, thanks.

qiita_pet/handlers/study_handlers/listing_handlers.py
antgonza: ok

antgonza:
What about adding the button(s) that will interact with the studies selected but when you click them you get a simple message like: "Not yet implemented" so we can review the distribution of the GUI.

BTW I'm using Chrome Version 41.0.2272.104 (64-bit)

squirrelo:
I'd rather wait and add the button when it works, that way we don't ever give the impression of being able to select studies on the page until you actually can. I don't want to confuse users.

squirrelo:
Also, if all goes well, there won't be a button. It will ajax-select the samples as the things rea clicked and unclicked.

antgonza:
That will be nice but note that having the checkboxes already gives
the wrong impression. What about removing or commenting that section
in the code so they do not get created and mislead users? Other option
is to have that functionality part of this PR but I can see the
benefits of having this already deployed.

squirrelo:
I guess I can comment out the code and just return a blank line for those cells.

antgonza:
Thanks, that works for me.

squirrelo:
Changes made, ready for merge.

antgonza:
Thanks. Tested and works fine except that if I search for the example:
env_matter includes soil AND (ph < 4 OR ph > 8) I get no results but
if I go to the metaanalysis page and do the same search I get results,
any ideas?

squirrelo:
Just tested and I get the same results on both pages. Not sure how you're getting no results as it's all the same search object setup.

antgonza:
Thanks for checking. Figured it out, when you search the "Narrow
search results by column data (Title, abstract, PI, etc):" field
doesn't get cleaned so it seems that there are no results but there
are ... could you clean on click?

squirrelo:
The checkboxes are too embedded in the code to safely hide them and re-add later. The textbox is cleared on search now though.

PR #1008, opened 2015-03-20
fixing new 454 issue

antgonza:
This solves an issue when you have 3 sff files, where 2 are from the same run (prefix aaa) and the 3rd one has a different prefix (prefix bbb). This is aaa1.sff, aaa2.sff, bbb.sff. In this case we need to split at the same time aaa1-2 and then bbb. 

Note that this is blocking some analysis so quick reviews are greatly appreciated. @biocore/qiita-dev

PR #1011, opened 2015-03-20
BLD/TST: Use conda to install scipy/mpl/etc

ElDeveloper:
This will speed-up the installation process and will (hopefully) avoid
conflicts with other packages and make the pip installation smoother.


.travis.yml
```diff
@@ -14,15 +14,14 @@ before_install:
...
-  - conda install --yes -n env_name python=$PYTHON_VERSION h5py>=2.3.1
+  # install a few of the dependencies that pip would otherwise try to install
+  # when intalling scikit-bio
+  - conda create --yes -n env_name python=$PYTHON_VERSION pip nose openpyxl=1.8.2 'pandas>=0.15' pyzmq flake8 'matplotlib>=1.1.0' 'ipython[all]==2.4.1' 'scipy>0.13.0' 'numpy>= 1.7' 'h5py>=2.3.1'
   - source activate env_name
-  - pip install coveralls cogent burrito qcli burrito-fillings ipython[all]==2.4.1
+  # upgrade click, older versions of click do not have __version__
+  - pip install -U click coveralls
   - pip install https://github.com/biocore/mustached-octo-ironman/archive/master.zip
-  - pip install https://github.com/biocore/qiime/archive/master.zip --no-deps
```
josenavas: I think this is removed because there is no point where we are actually executing qiime code so this is not needed, right?

josenavas:
Just one question @ElDeveloper otherwise :+1: given the tests pass

.travis.yml
ElDeveloper: Not really, this is removed because, as @gregcaporaso pointed out last week, the latest release version of QIIME is installed when we pip-install qiita a line below.

josenavas: Oh, thanks!

josenavas:
BTW, failure!

ElDeveloper:
I'm working on it, thanks.

ElDeveloper:
Please see my last commit message for an explanation of the changes.

.travis.yml
```diff
@@ -30,11 +28,8 @@ script:
   - qiita_env make --no-load-ontologies
   - if [ ${TEST_ADD_STUDIES} == "True" ]; then test_data_studies/commands.sh ; fi
   - if [ ${TEST_ADD_STUDIES} == "False" ]; then qiita_test_install ; fi
-  - if [ ${TEST_ADD_STUDIES} == "False" ]; then nosetests --with-doctest --with-coverage ; fi
+  - if [ ${TEST_ADD_STUDIES} == "False" ]; then nosetests -v --with-doctest --with-coverage ; fi
```
antgonza: Just wondering if we really need the -v?

antgonza: Saw your last commit message, any idea why? Note that IMOO the -v makes things less clear ... but I guess we can get used to the new output format

ElDeveloper: No idea, every single time when I removed that flag the suite would fail in a very *mysterious* way, see:

screen shot 2015-03-20 at 7 06 29 pm

----

Or see the build of the second to last commit.

ElDeveloper: Wait, now #1008 failed and it is showing the exact same behavior as
this, I'm uncertain of the cause for this problem. Will investigate
further as soon as I can.

PR #1014, opened 2015-03-25
Search stopgap

squirrelo:
This takes the current search engine output and formats it into what we expect the refactored search to spit out. This is needed because the processed data samples available may be a subset of the total samples available, and this needs to be reflected by the found samples in the search engine. This is also a stopgap because the search engine needs refactoring but there are database and logic changes blocking that, so this will sit on top for now to give us the output required until we can refactor the search engine.


qiita_ware/search.py
```diff
@@ -0,0 +1,96 @@
...
+        {study_id: {meta_col1: {value1: count, value2: count, ...}, ...}, ...}
+    """
+    def double_comprehension(results):
+        for samples in viewvalues(results):
+            for sample in samples:
+                yield sample
+
+    fullcount = {}
+    # rearrange all samples so that each metadata column found is its own list
+    meta_vals = zip(*double_comprehension(results))
```
wasade: might as well use `izip`

qiita_ware/search.py
```diff
@@ -0,0 +1,96 @@
...
+    for pos, cat in enumerate(meta_cols):
```
wasade: it isn't assured that `meta_cols` is not in the same order as `meta_vals`

qiita_ware/search.py
```diff
@@ -0,0 +1,96 @@
...
+    studycount = {}
+    for study_id in results:
+        studycount[study_id] = {}
+        # zip all samples for a given study so that each metadata column found
+        # is its own list
+        meta_vals = zip(*(sample for sample in results[study_id]))
+        for pos, cat in enumerate(meta_cols):
+            # use Counter object to count all metadata values for a column
+            # pos+1 so we skip the sample names list
+            studycount[study_id][cat] = Counter(meta_vals[pos + 1])
```
wasade: it would reduce indexing to build up the inner dict and then set, instead of creating an empty dict on line 42

qiita_ware/search.py
squirrelo: It is by the search object, not this function.

qiita_ware/search.py
```diff
@@ -0,0 +1,96 @@
...
+        # pos+1 so we skip the sample names list
+        fullcount[cat] = Counter(meta_vals[pos + 1])
+
+    # Now get metadata counts for each study, removing sample ids as before
+    studycount = {}
+    for study_id in results:
+        studycount[study_id] = {}
+        # zip all samples for a given study so that each metadata column found
+        # is its own list
+        meta_vals = zip(*(sample for sample in results[study_id]))
```
wasade: this is the second time this type of operation is being performed. It's going to likely be pretty expensive, and suggests that `results` is not formatted in as usable of a data structure as it should be?

qiita_ware/search.py
```diff
@@ -0,0 +1,96 @@
...
+            # use Counter object to count all metadata values for a column
+            # pos+1 so we skip the sample names list
+            studycount[study_id][cat] = Counter(meta_vals[pos + 1])
+
+    return fullcount, studycount
+
+
+def filter_by_processed_data(results, datatypes=None):
+    """Filters results to what is available in each processed data
+    Parameters
```
wasade: space please

qiita_ware/search.py
```diff
@@ -0,0 +1,96 @@
...
+    ----------
+    results : dict of lists of list
+        results in the format returned by the qiita_db search obj
+    datatypes : list of str, optional
+        Datatypes to selectively return. Default all datatypes available
+    Returns
```
wasade: space please

qiita_ware/search.py
wasade: This method is putting a lot of faith in its caller. You can see why this is a giant risk for nasty bugs going forward, right?

qiita_ware/search.py
```diff
@@ -0,0 +1,96 @@
...
+        study = Study(study_id)
+        samples_meta.update({s[0]: s[1:] for s in study_samples})
+        study_proc_ids[study_id] = defaultdict(list)
+        for proc_data_id in study.processed_data():
+            proc_data = ProcessedData(proc_data_id)
+            datatype = proc_data.data_type()
+            # skip processed data if it doesn't fit the given datatypes
+            if datatypes is not None and datatype not in datatypes:
+                continue
+            samps_available = proc_data.samples
```
wasade: Is this variable lookup friendly? E.g., `set` or `dict` or something comparable?

qiita_ware/search.py
```diff
@@ -0,0 +1,96 @@
...
+            proc_data_samples[proc_data_id] = [s[0] for s in study_samples
+                                               if s[0] in samps_available]
+            if len(proc_data_samples[proc_data_id]) == 0:
```
wasade: would be better to set this if its created, instead of setting empty first

wasade: oh, sorry, misunderstood what's going on

wasade: ...but still, why set in `proc_data_samples` if you're only going to delete?

qiita_ware/search.py
squirrelo: Yes, but this is a stopgap since we need it to go forward, and hopefully the refactor will come soon after. It's a standard assumption we've always made with the search though.

qiita_ware/search.py
squirrelo: the issue is that the info comes out as metadata value by sample_id, and it needs to get rotated 90 degrees to count it by metadata column. This is the fastest way I could think of to do that.

qiita_ware/search.py
squirrelo: izip breaks the Counter call below

qiita_ware/search.py
squirrelo: sounds right, done

qiita_ware/search.py
squirrelo: Don't know if all samples for the processed data gets filtered out or not when filling the key, so delete it if nothing is there. Can probably make the list, then attach though and be better.

qiita_ware/search.py
squirrelo: yes, it's a set.

qiita_ware/search.py
```diff
@@ -0,0 +1,101 @@
...
+from qiita_db.study import Study
+from qiita_db.data import ProcessedData
+
+
+def count_metadata(results, meta_cols):
+    """Counts the metadata found in a search, and returns these counts
+
+    Parameters
+    ----------
+    results : dict of lists of list
```
wasade: it is misleading to name an input parameter as `results`

qiita_ware/search.py
```diff
@@ -0,0 +1,101 @@
...
+        results in the format returned by the qiita_db search obj
```
wasade: this doesn't indicate what expectations there are about the keys and values, just types. if this method should only ever under any circumstance be called from the search object, then it should be a private member method of that object. otherwise, you have to assume that you do not know who the caller is. this is important both for the documentation here, and wide possibility for a bug on line 36

qiita_ware/search.py
```diff
@@ -0,0 +1,101 @@
...
+    fullcount : dict of dicts
+        counts for each found metadata value over all studies, in the format
+        {meta_col1: {value1: count, value2: count, ...}, ...}
+    studycount : dict of dict of dicts
+        counts for each found metadata value for each study, in the format
+        {study_id: {meta_col1: {value1: count, value2: count, ...}, ...}, ...}
+    """
+    def double_comprehension(results):
+        for samples in viewvalues(results):
+            for sample in samples:
```
wasade: is this actually an iterable of sample IDs or is it an iterable of metadata values?

squirrelo: Both. First position is sample_id, rest are metadata values.

qiita_ware/search.py
wasade: ah, I see. There really is an incredible amount of work being done with complex nested data structures in this method. It smells like there is quite a bit of redundant work. Given the inputs, returns, and the docstring, wouldn't the following suffice? It should require a lot less iteration and allocation

```python
n_cols = len(meta_cols)
studycount = {}
fullcount = {mc: defaultdict(int) for mc in meta_cols}
for study_id, meta_vals in viewitems(results):
    metadata_counts = [defaultdict(int) for i in range(n_cols)]
    for cat_idx, meta_col, meta_val in izip(range(n_cols), meta_cols, meta_vals):
        metadata_counts[cat_idx][meta_val] += 1
        fullcount[meta_col][meta_val] += 1
    studycount[study_id] = {mc: mcounts for mc, mcounts in izip(meta_cols, metadata_counts)}
```

The above is still placing extreme faith that `meta_cols` is in order with `meta_vals`, performing redundant work, and redundant representation. The latter two points are because `fullcount` really should be expressed as `reduce(add, studycount.values())`, however `dict` doesn't support that. What I'm wondering is actually if these data should be coming in as either `np.array` or `sp.sparse.spmatrix`?

qiita_ware/search.py
wasade: While the search method may state the values are in order on return, this method has no assurances about who is calling it and does not perform any validation on its inputs. What you're indicating is that this method ought to be a private static member method of the search object. That reduces the exposure, and potential for a bug in this situation, but does not remove it.

wasade: In addition, since `results` is a `dict`, no order is assured and it is mutable. What that means is that _IF_ the object is unmolested, then the order will remain unchanged. If _ANY_ modification happens, the order is not assured to remain the same. Since this is mutable, you cannot assume it will never be modified.

qiita_ware/search.py
```diff
@@ -0,0 +1,101 @@
...
+
+    return fullcount, studycount
+
+
+def filter_by_processed_data(results, datatypes=None):
+    """Filters results to what is available in each processed data
+
+    Parameters
+    ----------
+    results : dict of lists of list
```
wasade: same comment here, and on its description

qiita_ware/search.py
wasade: it would be more appropriate to describe this as metadata values then since a sample id is a metadata value for the SampleID category

qiita_ware/search.py
wasade: Nevermind, I see what's going on here now. The `list` of `list` contained in `results` is assured to remain in the same order. This is also a dense representation of a matrix as it isn't jagged. Numpy ops here probably would be much more natural.

qiita_ware/search.py
```diff
@@ -0,0 +1,101 @@
...
+    samples_meta : dict of lists
+        sample metadata in same order as the metadata given by search
+        Format {samp_id: [meta1, meta2, ...], ...}
+    """
+    study_proc_ids = {}
+    proc_data_samples = {}
+    samples_meta = {}
+    for study_id, study_samples in viewitems(results):
+        study = Study(study_id)
+        samples_meta.update({s[0]: s[1:] for s in study_samples})
```
wasade: this structure holds `{sample_id: [metadata_values]}` without any correspondence to the metadata category the value is associated with?

qiita_ware/search.py
squirrelo: The reason this is so expensive is because it's something tacked onto the search object and was not originally a function expected of it. As stated, this is kind of a stopgap so the results will be formatted how we expect the new search engine to work, without the actual re factoring that is going to happen once all the massive database changes stop happening. During that refactor we may move to numpy or some other holder, but for now we have this.

qiita_ware/search.py
```diff
@@ -0,0 +1,101 @@
...
+        study_proc_ids[study_id] = defaultdict(list)
+        for proc_data_id in study.processed_data():
+            proc_data = ProcessedData(proc_data_id)
+            datatype = proc_data.data_type()
+            # skip processed data if it doesn't fit the given datatypes
+            if datatypes is not None and datatype not in datatypes:
```
wasade: this conditional can be simplified if `datatypes` is made the empty `set` at the beginning of the method in the case where `datatypes is None`

qiita_ware/search.py
```diff
@@ -0,0 +1,101 @@
...
+        Samples available in each processed data id, in the format
+        {proc_data_id: [samp_id1, samp_id2, ...], ...}
+    samples_meta : dict of lists
+        sample metadata in same order as the metadata given by search
+        Format {samp_id: [meta1, meta2, ...], ...}
+    """
+    study_proc_ids = {}
+    proc_data_samples = {}
+    samples_meta = {}
+    for study_id, study_samples in viewitems(results):
```
wasade: is `study_samples` an iterable of the sample ids? it looks like it is the metadata values from its use

qiita_ware/search.py
squirrelo: ok, I can move it into the search object.

qiita_ware/search.py
```diff
@@ -0,0 +1,101 @@
...
+            proc_data = ProcessedData(proc_data_id)
+            datatype = proc_data.data_type()
+            # skip processed data if it doesn't fit the given datatypes
+            if datatypes is not None and datatype not in datatypes:
+                continue
+            samps_available = proc_data.samples
+            hold = [s[0] for s in study_samples if s[0] in samps_available]
+            if len(hold) == 0:
+                # all samples filtered so remove it as a result
+                del(proc_data_samples[proc_data_id])
```
wasade: There is no assurance that `proc_data_id` has been set in `proc_data_samples` from what I can see which would lead to a `KeyError`

qiita_ware/search.py
```diff
@@ -0,0 +1,101 @@
...
+        samples_meta.update({s[0]: s[1:] for s in study_samples})
+        study_proc_ids[study_id] = defaultdict(list)
+        for proc_data_id in study.processed_data():
+            proc_data = ProcessedData(proc_data_id)
+            datatype = proc_data.data_type()
+            # skip processed data if it doesn't fit the given datatypes
+            if datatypes is not None and datatype not in datatypes:
+                continue
+            samps_available = proc_data.samples
+            hold = [s[0] for s in study_samples if s[0] in samps_available]
```
wasade: so the goal here is to see if the intersection of the sample IDs in the study, and the processed/available samples, is not the empty set? What about:

```python
study_sample_ids = {s[0] for s in study_samples}
if samps_available.intersection(study_sample_ids):
    proc_data_samples[proc_data_id] = study_sample_ids
    study_proc_ids[study_id][datatype].append(proc_data_id)
```

qiita_ware/search.py
wasade: actually, that would not work based on the description in the docstring. You would instead want to set `datatypes = set(all_datatypes)`

qiita_ware/search.py
squirrelo: I had it that way originally but thought a list comprehension was faster. Testing now.

qiita_ware/search.py
```diff
@@ -0,0 +1,101 @@
...
+
+    Returns
+    -------
+    study_proc_ids : dict of dicts of lists
+        Processed data ids with samples for each study, in the format
+        {study_id: {datatype: [proc_id, proc_id, ...], ...}, ...}
+    proc_data_samples : dict of lists
+        Samples available in each processed data id, in the format
+        {proc_data_id: [samp_id1, samp_id2, ...], ...}
+    samples_meta : dict of lists
```
wasade: from the code below, the sample IDs contained in here will be a superset of the sample IDs described in `proc_data_samples`. Are those objects expected to be 1-1? If not, it would be useful to include a mention of this in the docstring

qiita_ware/test/test_search.py
```diff
@@ -0,0 +1,53 @@
...
+from qiita_db.user import User
+from qiita_ware.search import count_metadata, filter_by_processed_data
+
+
+@qiita_test_checker()
+class SearchTest(TestCase):
+    """Tests that the search helpers work as expected"""
+    def test_count_metadata(self):
+        search = QiitaStudySearch()
+        results, meta_cols = search('study_id = 1 AND ph > 0',
```
wasade: It would be much more useful if the inputs to `count_metadata` where explicitly described instead of fetching from the database. It is difficult to evaluate this test without knowing the inputs.

qiita_ware/test/test_search.py
```diff
@@ -0,0 +1,53 @@
...
+        fullcount, studycount = count_metadata(results, meta_cols)
+        expfull = {'study_id': {1: 27}, 'ph': {6.82: 10, 6.8: 9, 6.94: 8}}
+        expstudy = {1: {'study_id': {1: 27},
+                        'ph': {6.82: 10, 6.8: 9, 6.94: 8}}}
+        self.assertEqual(fullcount, expfull)
+        self.assertEqual(studycount, expstudy)
+
+    def test_filter_by_processed_data(self):
+        search = QiitaStudySearch()
+        results, meta_cols = search('study_id = 1', User('test@foo.bar'))
```
wasade: same comment here

qiita_ware/search.py
wasade: In general it is better to focus on the expressiveness of the code rather than performance. If performance was the concern for this method, then there are quite a few other things that should be addressed first.

squirrelo: fair point.

squirrelo:
completely refactored, should be good for merge

wasade:
:+1:

squirrelo:
Can someone do a final review/merge?

PR #1016, opened 2015-03-26
Add columns during edit

squirrelo:
Since this is needed for EBI submission, this should get reviewed and merged quickly.

This allows addition of columns during editing of a sample template. It also cleans up the add and remove category functions, as they were missing the addition of the column names to the study_sample_columns table.

### DIFF
--- a/qiita_db/metadata_template.py
+++ b/qiita_db/metadata_template.py
@@ -1696,11 +1696,11 @@ class SampleTemplate(MetadataTemplate):
         # add the column to the sample template
         sql = """ALTER TABLE qiita.{0}
             ADD COLUMN {1} {2}""".format(table_name, category, dtype)
+        sql_args = None
         if default is not None:
-            # mogrify the sql and add to end
-            with conn_handler.get_postgres_cursor() as cur:
-                sql = sql + cur.mogrify(" NOT NULL DEFAULT %s", [default])
-        conn_handler.add_to_queue(queue, sql)
+                sql = sql + " NOT NULL DEFAULT %s"
+                sql_args = [default]
+        conn_handler.add_to_queue(queue, sql, sql_args)
 
         # add the column to the column listings table
         conn_handler.add_to_queue(
--- a/qiita_db/metadata_template.py
+++ b/qiita_db/metadata_template.py
@@ -43,6 +43,7 @@ from os.path import join
 from time import strftime
 from functools import partial
 from os.path import basename
+from itertools import izip
 from future.utils.six import StringIO
 
 import pandas as pd
@@ -1530,7 +1531,6 @@ class SampleTemplate(MetadataTemplate):
         ------
         QiitaDBError
             If md_template and db do not have the same sample ids
-            If md_template and db do not have the same column headers
         """
         conn_handler = SQLConnectionHandler()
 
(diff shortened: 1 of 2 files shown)


josenavas:
@squirrelo besides the problems connecting to the redis server, it looks like there is a legitimate test failure!

squirrelo:
Seems like only redis failures now.

squirrelo:
ready for review

qiita_db/metadata_template.py
```diff
@@ -1672,16 +1686,29 @@ def add_category(self, category, samples_and_values, dtype, default):
...
+        queue = 'add_col_%d' % self._id
+        conn_handler.create_queue(queue)
+
+        # add the column to the sample template
+        sql = """ALTER TABLE qiita.{0}
+            ADD COLUMN {1} {2}""".format(table_name, category, dtype)
+        if default is not None:
+            # mogrify the sql and add to end
+            with conn_handler.get_postgres_cursor() as cur:
+                sql = sql + cur.mogrify(" NOT NULL DEFAULT %s", [default])
```
josenavas: Check the comment of the first answer on this [thread](http://stackoverflow.com/q/6775497/3746629). In short, mogrify should be used only for debugging, not for generating code. Anyways, I also find this solution more convoluted, so if you can just add the if and generate either of the sql in the specific case of the if would be more cleaner/maintainable.

qiita_db/metadata_template.py
```diff
@@ -1561,9 +1561,16 @@ def update(self, md_template):
                                % ', '.join(samples_diff))
         columns_diff = set(new_map.columns) - set(current_map.columns)
         if columns_diff:
-            raise QiitaDBError('The new sample template differs from what is '
-                               'stored in database by these columns names: %s'
-                               % ', '.join(columns_diff))
+            queue = 'add_cols_%d' % self._id
```
josenavas: A general comment: using the approach of multiple queues on a single function, we still can face the problem in which half of the function is successfully applied while the other half is not, leaving the system in a inconsistent state. Can you use a single queue in the entire function?

Thanks!

qiita_db/metadata_template.py
```diff
@@ -1561,9 +1561,16 @@ def update(self, md_template):
...
+            datatypes = dict(izip(new_map.columns, _get_datatypes(new_map)))
+            conn_handler.create_queue(queue)
+
+            for col in columns_diff:
+                self.add_category(col, new_map[col].to_dict(), datatypes[col])
```
josenavas: This is going to leverage a lot of connection openings/closing to the database, with the added danger of leveraging the db in an inconsistent state. You can either (a) duplicate the code of that function here so everything is executed in a single queue; or (b) modify the function signature so a conn_handler and a queue name can be passed, so it just gets added to the queue; or (c) do a private function to add things to a queue which is shared by both functions.

My vote is (c) as that pattern has been already done in other portions of the code (the data objects) and it is the one that maintains a consistency on the code and in the functions (i.e. each function only has a possible output w/o side effects depending on the input parameters)

squirrelo: This is how the old code did it (look at line 1587). If we want a full re-factor that is out of the scope of this pull request, we just need the functionality quickly for EBI submissions. Issue .#1020 created for a full refactor.

qiita_db/metadata_template.py
squirrelo: This is consistent with the existing codebase, and a refactor is out of scope. Issue #1020 created for the refactor.

qiita_db/metadata_template.py
josenavas: Thx

josenavas:
Closing in favor of #1021

PR #1018, opened 2015-03-27
BLD/TST: Add travis_retry to .travis.yml

ElDeveloper:
Prefix some installation commands that are known to spuriously fail.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,11 +17,11 @@ before_install:
 install:
   # install a few of the dependencies that pip would otherwise try to install
   # when intalling scikit-bio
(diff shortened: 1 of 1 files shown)


ElDeveloper:
I'm starting to believe the failures we see in other builds are due to the fact that more people uses Travis CI at those time of the day, unlike this time ... just sayin' :8ball:

josenavas:
Is there any statistics page? We can potentially check our failures and try to correlate it with the usage...

ElDeveloper:
Yes http://www.traviscistatus.com

Yoshiki Vázquez-Baeza

josenavas:
buff... it looks like from 6 to ~4 they're running the system as its maximum capacity! Impressive! It looks like there is still some more room on the container based ones, so I hope we get more lucky using those ones!

ElDeveloper:
Yeah, it seems like that's the case.

josenavas:
:+1:

Issue #781, opened 2015-01-07
Prep templates not deleting correctly


squirrelo:
We need to make the delete function (and really all the stuf like it) a transaction block. We are already seeing issues where everything but the actual reference in the `prep_template` table is deleted, causing issues with processing.

antgonza:
Can this be closed? I think this was addressed in https://github.com/biocore/qiita/pull/866 and https://github.com/biocore/qiita/pull/850, right? If not please describe what's missing.

josenavas:
Not really... I think the deletion of the prep templates should be also a transaction block. I used transaction blocks for template creation, but not deletion...

antgonza:
OK, thanks for checking.

Issue #680, opened 2014-11-24
refactor create_qiime_mapping_file


antgonza:
From @josenavas in https://github.com/biocore/qiita/pull/678:
It is a bit weird that we are relying on the filepaths to create the qiime mapping file (so we can parse them into pandas). I remember correctly, there is a function that @ElDeveloper wrote to transform a MetadataTemplate into a pandas DataFrame. I think the way to go is to make a function on the MetadataTemplate object called to_dataframe that returns a DataFrame with the template information.

@antgonza note: this will duplicate some code (as this will be similar to what we are doing now but retrieve the info from DB and generate the pandas dataframe) and then load in memory but is another option.

antgonza:
I _think_ this can be closed as everything here is implemented. Please reopen and add details if that's not the case.

Issue #465, opened 2014-10-18
Meta-analyses should use the last raw_data element


ElDeveloper:
We should not use the zero-th element as we are right now because that would be the oldest group of data that was uploaded, not the latest (which is the desirable case).

See the query in `Analysis._build_mapping_file`.

antgonza:
@squirrelo https://github.com/biocore/qiita/pull/865 addressed this issue, right?

squirrelo:
Yes

josenavas:
#865 was not merged

josenavas:
Solving as part of my latest PR: #1021

Issue #440, opened 2014-10-16
Simplify code on metadata_template


josenavas:
The create functions have a lot of duplicated code.

PR #1021, opened 2015-03-30
Relaxing metadata requirements

### LINKED ISSUE #941: re-defining the required fields for the study and prep templates to facilitate addition of new data
In working on [documentation of the sample and prep templates](https://github.com/biocore/qiita/wiki/Preparing-Qiita-template-files) @rob-knight, @adamrp, @wasade, @josenavas and I discussed updating the requirements for sample and prep template fields. 

These fields should no longer be required:
 - [ ] ``emp_status``
 - [ ] ``run_prefix``
 - [ ] ``sample_type``
 - [ ] ``description``
 - [ ] ``has_physical_specimen``
 - [ ] ``physical_location``
 - [ ] ``has_extracted_data``
 - [ ] ``host_subject_id``
 - [ ] ``barcode`` (this will not be relevant when users can start with demultiplexed data)
 - [ ] ``primer``
 - [ ] ``linker`` (this is not always used)
 - [ ] ``center_name``
 - [ ] ``platform``
 - [ ] ``library_construction_protocol``
 - [ ] ``experiment_design_description``

These fields should be removed all together:
 - [ ] ``required_sample_info_status``
 - [ ] ``center_project_name``

These fields should have their names changed:
 - [ ]  ``sample_name`` will be changed to ``SampleID``
 - [ ] ``has_physical_specimen`` should change to ``physical_specimen_remaining``  
 - [ ] ``physical_location`` should change to ``physical_specimen_location``
 - [ ] ``has_extracted_data``should change to ``dna_extracted``

Format/allowed value changes:
 - [ ] ``platform`` the allowed values need to be expanded in or linked from the [wiki document](https://github.com/biocore/qiita/wiki/Preparing-Qiita-template-files)

josenavas:
Fixes #941 
It also includes the code from @squirrelo in #1016 

I still have to push the changes on DB schema and de-activate the bits when metadata columns are not included, plus improve a little bit some parts of the code.

While implementing this change, I noticed a lot of small bugs; and I think that was a result of the complexity of the code. I think the best solution is a refactor, as started here, which divides the code in a easier-to understand structure.

Some of these bugs where caused due to improper testing. I noticed that some of the tests were copy-pasted directly from the python output (maybe from ipython) and just PEP8-ing them. This is not ok and as a remainder, I would like to say that tests should be executed "by hand" and manually checking that the result is correct; otherwise the tests are useless.

I am one day behind on this PR, I will have it ready by tomorrow (I was not able to work on this on Thursday giving the problems that we had on travis on PR #900 and I was working on solving them). Sorry about that.


josenavas:
This is ready for review @antgonza 

Fixes #465 but in a more intelligent way (it actually chooses the right prep template)

There is a bunch of other issues fixed here, I will compile the list tomorrow morning and list it here so they get closed... I am too asleep right now to do it.

This is another gigantic PR so 3 reviewers are more than welcome. Also these are sensitive changes so please review carefully.

Basically, the system will allow you to upload any sample/prep template, doesn't matter which columns are missing. However, it will disable some functionality based on those missing columns (preprocessing, EBI submission or processed data approval).

antgonza:
You have real pep8 failures. 

This is a huge PR, thanks for working on this. Do you think it will be worth splitting in 2 PRs? One with the split of qiita_db/metadata_template (merge first) and another with the code changes (this should allow us to find quickly the changes/modifications). I didn't think about how difficult it will be to review until I open this and it took a while to load, even without any comments or discussion. Sorry for this.

qiita_db/support_files/patches/python_patches/20.py
```diff
@@ -0,0 +1,204 @@
...
+             sample_data['description'],
+             sample_data['latitude'],
+             sample_data['longitude'],
+             sample_data['sample_id']))
+
+# Once all data has been moved, we can drop the columns from the
+# requried_sample_info table; and drop the required_sample_info_status table
+conn_handler.add_to_queue(
+    queue_name,
+    """ALTER TABLE qiita.required_sample_info
```
squirrelo: All these drops can be a single SQL statement
```SQL
alter table qiita.required_sample_info drop column collection_timestamp, drop column host_subject_id;
```
just keep adding to the list of drop column statements.

josenavas:
@antgonza I don't think we can split in 2 PR... the reason is that the changes on the metadata template need most of the other changes in order to pass the tests...

The actual code changes after this being solved (i.e. limiting the functionality of the interface) only include ~20 lines of code (it was way faster than I expected) so the first PR will still be huge, plus the master state will be inconsistent...

In order to make the review more smooth, I would recommend to review the `metadata_template/` directory, as it is where most of the changes happened. Once those files are well reviewed, the other portions of the code are simplifications based on the metadata_template changes, so it should be quicker.

Does anyone knows why flake8 is complaining about those lines? There are 2 other similar lines in that same file and it is not complaining... (is an __init__.py file for reference)

qiita_db/support_files/patches/python_patches/20.py
josenavas: Done

josenavas:
Fixes partially #247 - the mapping files generated for analysis include the prep info, unsure about the search, @squirrelo was that solved or the search still doesn't work over prep template info?

Fixes #781
Fixes #440 
Fixes #680 - I think the current solution is the best one giving the current system we have. I know that I said it is a bit weird that we rely on files, but since the files and the DB are sync, I don't think it is an issue. Also, we can avoid @antgonza point in code duplication.

squirrelo:
Search has the stopgap pull request #1014 in right now that needs merging. It will fix #247 until the full search refactor occurs.

josenavas:
:smile: green!

antgonza:
@josenavas what's the status of this PR?

josenavas:
This PR is being re-introduced with a bunch of small PR that I'm doing. So this PR can be closed as trying to fix the merge conflicts will be horrible. I'll be submitting a bunch of small PR today to a new branch in upstream so the entire functionality can make it.

antgonza:
OK, will close. Remember that we need to have things merged before 9am tomorrow so the sooner the PRs are ready, the better.

josenavas:
Yeah, I know, I'm still finding bugs on the metadata template that need to be addressed before the refactor comes in. Some of them are tricky to test...

PR #1025, opened 2015-03-31
Analysis cart creation

squirrelo:
This pull request creates the carts for each user and the cart viewing page. Major pieces are:
* Default user analyses added to act as carts
* `Analysis.get_user_default(User)` function added to get the default analysis for a user
* `processed_data.processing_info` added to get processing information needed for cart display.
* Cart viewing page added at http://localhost:21174/analysis/selected/   Logging in as test@foo.bar using the test database will prepopulate it with a few samples selected so you can view the page. Currently the page can only be accessed through direct URL. 

This is the final piece needed before tying everything together from the last few pull requests into a single unified sample selection system.


squirrelo:
Ready for review/merge

antgonza:
Note that https://github.com/biocore/qiita/pull/1006 also has a 20.sql. My guess is that this should be 21.sql, right?

squirrelo:
If that pull request is merged first, yes. I'll update once it's merged.

squirrelo:
Neither pull request depends on the other, so feel free to review and merge this one first as well.

qiita_db/analysis.py
```diff
@@ -79,6 +79,26 @@ def _status_setter_checks(self, conn_handler):
             raise QiitaDBStatusError("Can't set status away from public!")
 
     @classmethod
+    def get_user_default(cls, user):
```
josenavas: Uhm... from my point of view, is usually better to ask the user: give me your default cart. It will be a instance method rather than a classmethod, which is already a signal of better design...

squirrelo:
ready for review/merge

qiita_db/analysis.py
squirrelo: This allows you to return the actual object already instantiated, as opposed to just the ID and instantiating, saving a DB call. Do we care about the extra DB call?

josenavas: You are not saving any DB call...

squirrelo: The check if ID exists on Analysis instantiation isn't needed in this case. That's the DB call I'm referring to.

josenavas: That call is executed when you are doing `cls(aid)` (It is calling the __init__)

squirrelo: Oh, didn't realize that. Sure, can make that change then.

qiita_db/data.py
```diff
@@ -1303,12 +1303,54 @@ def data_type(self, ret_id=False):
...
-        return conn_handler.execute_fetchone(
-            "SELECT processed_date FROM qiita.{0} WHERE "
-            "processed_data_id=%s".format(self._table), (self.id,))[0]
+        sql = """SELECT processed_date, processed_params_table,
+            processed_params_id FROM qiita.{0}
+            WHERE processed_data_id=%s""".format(self._table)
+        static_info = conn_handler.execute_fetchone(sql, (self.id,))
+
+        # Get the info from the dynamic table, including reference used
+        sql = """SELECT * from qiita.{0}
```
josenavas: I'm getting confused in this SQL query, it looks like it is enough to do:

```sql
SELECT * FROM qiita.{0}
  JOIN qiita.reference USING (reference_id)
WHERE processed_params_id = {1}
```

I think this SQL query can return _only_ a single value, since there will be only one row with the given processed_params_id.

qiita_db/data.py
```diff
@@ -1303,12 +1303,54 @@ def data_type(self, ret_id=False):
...
+        dynamic_info = dict(conn_handler.execute_fetchone(sql))
+
+        # replace reference filepath_ids with full filepaths
+        # fiugre out what columns have filepaths and what don't
+        ref_fp_cols = {'sequence_filepath', 'taxonomy_filepath',
+                       'tree_filepath'}
+        fp_ids = [str(dynamic_info[col]) for col in ref_fp_cols
+                  if dynamic_info[col] is not None]
+        # Get the filepaths and create dict of fpid to filepath
+        sql = ("SELECT filepath_id, filepath FROM qiita.filepath WHERE "
```
josenavas: I think this query can be merged with the previous so a single query is done in this entire code, would you mind to take a look? I can potentially give it a shot later today if you can't.

qiita_db/support_files/patches/21.sql
```diff
@@ -0,0 +1,14 @@
+-- March 28, 2015
+-- Add default analyses for all existing users
+DO $do$
```
josenavas: I really like this patch

I think we can potentially use some of the SQL tricks to reduce some of the code in the qiita_db objects and we will improve the performance/quality of our code.

qiita_db/test/test_analysis.py
```diff
@@ -95,12 +100,12 @@ def test_create(self):
 
         new = Analysis.create(User("admin@foo.bar"), "newAnalysis",
                               "A New Analysis")
-        self.assertEqual(new.id, 3)
+        self.assertEqual(new.id, 7)
```
josenavas: Can we please avoid the use of magic numbers here? The easiest way to do this is using the function `get_count` to know how many are in the table already and then you just need to add 1 to know the new id. This will make future changes really easy :-)

josenavas: Just for completeness all the SQL queries/tests on this function will be parametrized.

qiita_db/test/test_analysis.py
```diff
@@ -245,18 +250,18 @@ def test_set_step(self):
         new = Analysis.create(User("admin@foo.bar"), "newAnalysis",
                               "A New Analysis", Analysis(1))
         new.step = 2
-        sql = "SELECT * FROM qiita.analysis_workflow WHERE analysis_id = 3"
+        sql = "SELECT * FROM qiita.analysis_workflow WHERE analysis_id = 7"
         obs = self.conn_handler.execute_fetchall(sql)
-        self.assertEqual(obs, [[3, 2]])
+        self.assertEqual(obs, [[7, 2]])
```
josenavas: Same here

qiita_db/test/test_analysis.py
```diff
@@ -245,18 +250,18 @@ def test_set_step(self):
...
     def test_set_step_twice(self):
         new = Analysis.create(User("admin@foo.bar"), "newAnalysis",
                               "A New Analysis", Analysis(1))
         new.step = 2
         new.step = 4
-        sql = "SELECT * FROM qiita.analysis_workflow WHERE analysis_id = 3"
+        sql = "SELECT * FROM qiita.analysis_workflow WHERE analysis_id = 7"
         obs = self.conn_handler.execute_fetchall(sql)
-        self.assertEqual(obs, [[3, 4]])
+        self.assertEqual(obs, [[7, 4]])
```
josenavas: And here

qiita_db/test/test_job.py
```diff
@@ -223,12 +223,12 @@ def test_create_exists_return_existing(self):
         self.conn_handler.execute(
             "INSERT INTO qiita.analysis_sample "
             "(analysis_id, processed_data_id, sample_id) VALUES "
-            "(3, 1, '1.SKB8.640193'), (3, 1, '1.SKD8.640184'), "
-            "(3, 1, '1.SKB7.640196'), (3, 1, '1.SKM9.640192'), "
-            "(3, 1, '1.SKM4.640180')")
+            "(7, 1, '1.SKB8.640193'), (7, 1, '1.SKD8.640184'), "
```
josenavas: Same here in terms of magic IDs

qiita_db/user.py
```diff
@@ -224,10 +224,21 @@ def create(cls, email, password, info=None):
...
-
+        queue = "add_user_%s" % email
+        conn_handler.create_queue(queue)
+        # crete user
         sql = ("INSERT INTO qiita.%s (%s) VALUES (%s)" %
                (cls._table, ','.join(columns), ','.join(['%s'] * len(values))))
-        conn_handler.execute(sql, values)
+        conn_handler.add_to_queue(queue, sql, values)
+        # create user default sample holder
+        sql = ("INSERT INTO qiita.analysis "
```
josenavas: This is personal opinion but I think this highly improves code readability and even when failures happen they're easier to read on the command line. Instead of using quotes ("), I like to user triple quotes (""") and then align the SQL in a easy to read way. In this example:

```python
sql = """INSERT INTO qiita.analysis
           (email, name, description, dflt, analysis_status_id)
         VALUES (%s, %s, %s, %s, 1)"""
```
The cool thing about this, is that if the query fails, it gets also formatted on the CLI. Also it is a more natural way of reading SQL as things align better (note the small indentation on the column names).

This is not blocking though...

qiita_db/user.py
```diff
@@ -224,10 +224,21 @@ def create(cls, email, password, info=None):
         # for sql insertion
         columns = info.keys()
         values = [info[col] for col in columns]
-
+        queue = "add_user_%s" % email
+        conn_handler.create_queue(queue)
+        # crete user
         sql = ("INSERT INTO qiita.%s (%s) VALUES (%s)" %
```
josenavas: I know you have not touch this code, but we have been (more or less) consistently using format for this type of changes. I would appreciate if you can change that. Non blocking...

qiita_db/user.py
```diff
@@ -360,8 +371,8 @@ def shared_studies(self):
     @property
     def private_analyses(self):
         """Returns a list of private analysis ids owned by the user"""
-        sql = ("Select analysis_id from qiita.analysis WHERE email = %s AND "
-               "analysis_status_id <> 6")
+        sql = ("Select analysis_id from qiita.analysis "
```
josenavas: Same comments as above. Also the capitalization of this SQL statement is weird (at the beginning I did not notice that it was SQL...)

qiita_pet/handlers/analysis_handlers.py
```diff
@@ -363,3 +363,20 @@ def validate_absolute_path(self, root, absolute_path):
                 root, absolute_path)
         else:
             raise QiitaPetAuthorizationError(user_id, absolute_path)
+
+
+class SelectedSamplesHandler(BaseHandler):
+    @authenticated
+    def get(self):
+        # Format seldata to get study IDs for the processed data
+        seldata = defaultdict(dict)
```
josenavas: please, use `_` for the variable names (e.g. `sel_data`) this is PEP8 so I think this should be blocking.

josenavas: I'm just going to mention it once, but it also applies to other variables.

josenavas:
Few comments @squirrelo 
I will pull it down and test it locally once these issues are solved.

Thanks!

qiita_db/data.py
squirrelo: The reason it's this way is because only sequence_filepath is a required column, so the other two may not exist. This leads to issues with JOINs that expect all three columns to exist.

qiita_db/test/test_analysis.py
squirrelo: I can do it here, but it will also need to be added as something to be done as part of the DB/NODB breakup issue so it is consistent across all tests. If that's agreeable, I'll make the change.

qiita_db/user.py
squirrelo: I can do this, but it is also something to put in contributing.md and needs to be consistently done across the entire codebase. Again, if that's agreeable I will make the change.

qiita_db/user.py
squirrelo: Yeah, I'll fix the capitalization.

qiita_db/test/test_analysis.py
josenavas: I don't think we need to make all changes at once. Since you're already modifying this test, please make the change so we don't have that burden on the future (it's just a small change and it will reduce any further change in the future). Thanks!

qiita_db/user.py
josenavas: Yeah, it's not on contributing so no worries if you don't want to change it. I'm doing it as I'm changing other parts of the code, but there is no documentation at this point.

qiita_db/data.py
squirrelo: Yeah, was being way to protective about the join screwing up. Changed.

qiita_db/data.py
josenavas: Oh true, thanks!

qiita_db/test/test_analysis.py
squirrelo: Oh I know, I'm just trying to make sure this is consistent across the code base and not a change that happens or not depending on who reviewed the code.

josenavas: Oh no, this is something that we are trying to remove, and @wasade has done strict comments on changing this, and I thing all @biocore/qiita-dev agree that this has to be done so we do not spend more time fixing magic numbers...

antgonza: Kill the magic numbers!!!

squirrelo: Cool, fixing it now.

ElDeveloper: :1234: :gun:

squirrelo:
All comments addressed, ready for another round of review.

qiita_db/test/test_analysis.py
wasade: it's going to be hard to determine what the other magic numbers are anyways
so I think working to reduce this burden overall is in everyones benefit

squirrelo:
Can this get a quick final review? It's holding up the remaining work needed to create the cart.

josenavas:
@squirrelo code looks good but the /study/description page is broken. Just log in with the default user and try to see the description of the test study. The page doesn't render. It works on master so it is something on this PR...

josenavas:
@squirrelo I just saw the traceback on the CLI:
```
ERROR:tornado.application:Future exception was never retrieved: Traceback (most recent call last):
  File "/Users/jose/.virtualenvs/qiita/lib/python2.7/site-packages/tornado/gen.py", line 197, in wrapper
    result = func(*args, **kwargs)
  File "/Users/jose/qiime_software/QiiTa/qiita_pet/handlers/study_handlers/description_handlers.py", line 641, in display_template
    prep_tab=prep_tab)
  File "/Users/jose/.virtualenvs/qiita/lib/python2.7/site-packages/tornado/web.py", line 664, in render
    html = self.render_string(template_name, **kwargs)
  File "/Users/jose/.virtualenvs/qiita/lib/python2.7/site-packages/tornado/web.py", line 771, in render_string
    return t.generate(**namespace)
  File "/Users/jose/.virtualenvs/qiita/lib/python2.7/site-packages/tornado/template.py", line 278, in generate
    return execute()
  File "study_description_html.generated.py", line 198, in _tt_execute
    _tt_tmp = _tt_modules.ProcessedDataTab(study, full_access)  # study_description.html:518 (via sitebase.html:192)
  File "/Users/jose/.virtualenvs/qiita/lib/python2.7/site-packages/tornado/web.py", line 1429, in render
    rendered = self._active_modules[name].render(*args, **kwargs)
  File "/Users/jose/qiime_software/QiiTa/qiita_pet/uimodules/processed_data_tab.py", line 24, in render
    study_id=study.id)
  File "/Users/jose/.virtualenvs/qiita/lib/python2.7/site-packages/tornado/web.py", line 2691, in render_string
    return self.handler.render_string(path, **kwargs)
  File "/Users/jose/.virtualenvs/qiita/lib/python2.7/site-packages/tornado/web.py", line 771, in render_string
    return t.generate(**namespace)
  File "/Users/jose/.virtualenvs/qiita/lib/python2.7/site-packages/tornado/template.py", line 278, in generate
    return execute()
  File "study_description_templates/processed_data_tab_html.generated.py", line 49, in _tt_execute
    _tt_tmp = _tt_modules.ProcessedDataInfoTab(study_id, pd)  # study_description_templates/processed_data_tab.html:23
  File "/Users/jose/.virtualenvs/qiita/lib/python2.7/site-packages/tornado/web.py", line 1429, in render
    rendered = self._active_modules[name].render(*args, **kwargs)
  File "/Users/jose/qiime_software/QiiTa/qiita_pet/uimodules/processed_data_tab.py", line 56, in render
    process_date = processed_data.processed_date
```

Probably useful to do a search over `processed_date` in the entire codebase

squirrelo:
Did, just missed that one. Good catch.

qiita_pet/templates/analysis_selected.html
```diff
@@ -0,0 +1,59 @@
...
+{% from future.utils import viewitems %}
+<h1>Selected Samples</h1>
+{% for sid, proc_datas in viewitems(sel_data) %}
+  {% set study = Study(sid) %}
+<div class="row" id="study{{sid}}">
+	<div class="col-md-12" style="border:2px solid #a1a1a1; border-radius: 15px;">
+	  <h2><a href="/study/description/{{sid}}">{{study.title}}</a></h2>
+	  <table class='table table-striped' id='study{{sid}}-table'>
+	  <tr>
+	  	<th></th><th>id</th><th>Datatype</th><th>Processed Date</th><th>Algorithm</th><th>Reference</th><th></th>
```
josenavas: Can we change the `id` by `Processed data id`? The issue is that in the actual page we don't know what is that id for, is not completely clear...

josenavas: Also, is it possible to add a modal that will show the full list of parameters used to process data processed data? I can see this really useful to further samples/processed data filtering

josenavas: Can we move the show/hide samples at the end of the table, is more natural IMOO

squirrelo: I had it there, but the samples come up on the very left so it's more natural to hit the link and then immediately scroll down.

squirrelo: I can add the rest of the changes though

josenavas: Sorry to disagree about your previous statement. What about this layout?
screen shot 2015-04-03 at 10 50 01 am

squirrelo:
Interface changes made, ready for another review.

josenavas:
I think you did not see my comment. Posting the layout here again:

screen shot 2015-04-03 at 10 50 01 am

squirrelo:
All set for another round.

josenavas:
Thanks @squirrelo one more comment. Instead of adding the info gif that you used, can you use glyphicon glyphicon-info-sign? I think it will be more consistent with other icons on the system

squirrelo:
Oh, yeah that makes sense.

squirrelo:
Closing to open as a pull request on a separate branch.

PR #1031, opened 2015-04-02
ENH: Minor enhancements to the CLI for qiita log

ElDeveloper:
Doing:

qiita log get

Is rather annoying and it should really just be:

qiita log

With an optional --n flag, like:

qiita log --n 11 # gets the 11 most recent entries

### DIFF
--- a/scripts/qiita
+++ b/scripts/qiita
@@ -79,11 +79,6 @@ def ware():
     pass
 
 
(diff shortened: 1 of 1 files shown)


antgonza:
:+1: Agree with the code and change, it was also getting annoying for me ...

PR #1032, opened 2015-04-02
Unify metadata creation

### LINKED ISSUE #440: Simplify code on metadata_template

josenavas:
The code that was adding SQL statements to the queue was duplicated between the PrepTemplate and the SampleTemplate. I moved this code to the base class so future changes in this code are centralized...

Fixes #440

### DIFF
--- a/qiita_db/metadata_template/base_metadata_template.py
+++ b/qiita_db/metadata_template/base_metadata_template.py
@@ -51,6 +51,7 @@ from qiita_db.util import (exists_table, get_table_cols,
                            convert_to_id,
                            get_mountpoint, insert_filepaths)
 from qiita_db.logger import LogEntry
+from .util import as_python_types, get_datatypes
 
 
 class BaseSample(QiitaObject):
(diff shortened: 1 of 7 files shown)


josenavas:
Do not merge this, I'm adding a test as per @antgonza request

josenavas:
This includes the commits from #1033. Once that one gets merged, I will pull from upstream/master and push again, so the commits will go away and only the new code is being shown here.

I've added a test for the `_add_common_creation_steps_to_queue` in both the SampleTemplate and PrepTemplate classes. This has been a good exercise as I was able to make sure that the SQL commands where correctly formatted directly, rather than relying on the database contests to assume that they're formatted correctly. I would suggest that we should do this when possible, as it will reduce potential issues with DB integrity and also it will help us to get a better view of how everything on the DB works :-)

antgonza:
:+1:

josenavas:
Ready for another review/merge if there are no comments!

qiita_db/metadata_template/base_metadata_template.py
```diff
@@ -519,6 +520,81 @@ def _check_special_columns(cls, md_template, obj):
...
+        queue_name : str
+            The queue where the SQL statements will be added
+        """
+        # Get some useful information from the metadata template
+        sample_ids = md_template.index.tolist()
+        num_samples = len(sample_ids)
+        headers = list(md_template.keys())
+
+        # Get the required columns from the DB
+        db_cols = get_table_cols(cls._table, conn_handler)
```
squirrelo: This looks looks like the function needs the "am I instantiating the base class" test.

qiita_db/metadata_template/base_metadata_template.py
```diff
@@ -519,6 +520,81 @@ def _check_special_columns(cls, md_template, obj):
...
+        # Remove the sample_id and _id_column columns
+        db_cols.remove('sample_id')
+        db_cols.remove(cls._id_column)
```
squirrelo: is there any way _id_column can be `sample_id`?

qiita_db/metadata_template/base_metadata_template.py
```diff
@@ -519,6 +520,81 @@ def _check_special_columns(cls, md_template, obj):
...
+        # Insert rows on *_columns table
+        headers = list(set(headers).difference(db_cols))
+        datatypes = get_datatypes(md_template.ix[:, headers])
+        # psycopg2 requires a list of tuples, in which each tuple is a set
+        # of values to use in the string formatting of the query. We have all
+        # the values in different lists (but in the same order) so use zip
+        # to create the list of tuples that psycopg2 requires.
+        values = [
+            v for v in zip([obj_id] * len(headers), headers, datatypes)]
+        conn_handler.add_to_queue(
```
squirrelo: Philosophical thing, but I'd add this to the queue after the table is made and filled.

squirrelo:
Few comments.

qiita_db/metadata_template/base_metadata_template.py
josenavas: No, `cls._id_column` is either `study_id` or `prep_template_id`

qiita_db/metadata_template/base_metadata_template.py
josenavas: I was thinking on that, but this is a private function and it should be only used in the `create` functions, that already have that test. If you feel strong about it, I can add it though...

qiita_db/metadata_template/base_metadata_template.py
josenavas: It's in a queue, all the changes happen or not, so I don't see the added value. Do you have strong feelings?

qiita_db/metadata_template/base_metadata_template.py
squirrelo: Would probably be good to add as a safety layer, since it will also mean a more interpretable error.

qiita_db/metadata_template/base_metadata_template.py
squirrelo: cool

qiita_db/metadata_template/base_metadata_template.py
squirrelo: No, not blocking, just philosophical.

josenavas:
@squirrelo, @antgonza already gave his :+1: if you think the code looks good can you merge? Thanks!

PR #1034, opened 2015-04-03
Delete process data

antgonza:
Adding Processed data deletion, both to the qiita_db and the GUI.


qiita_db/data.py
```diff
@@ -1257,6 +1257,60 @@ def create(cls, processed_params_table, processed_params_id, filepaths,
...
+        sql = ("DELETE FROM qiita.study_processed_data WHERE "
+               "processed_data_id = {0}".format(processed_data_id))
+        conn_handler.add_to_queue(queue, sql)
+
+        sql = ("DELETE FROM qiita.processed_data WHERE "
+               "processed_data_id = {0}".format(processed_data_id))
+        conn_handler.add_to_queue(queue, sql)
+
+        conn_handler.execute_queue(queue)
+
```
squirrelo: This also needs to delete the files on the filesystem, correct?

qiita_db/data.py
```diff
@@ -1257,6 +1257,60 @@ def create(cls, processed_params_table, processed_params_id, filepaths,
...
+        Raises
+        ------
+        QiitaDBUnknownIDError
+            If the processed data id doesn't exist
+        QiitaDBStatusError
+            If the processed data status is not sandbox
+        """
+        if cls(processed_data_id).status != 'sandbox':
+            raise QiitaDBStatusError(
+                "Illegal operation on public processed data")
```
squirrelo: It's not necessarily  public, so "...on non-sandbox ..." would probably make more sense here.

qiita_db/data.py
```diff
@@ -1257,6 +1257,60 @@ def create(cls, processed_params_table, processed_params_id, filepaths,
...
+
+        # delete
+        queue = "delete_processed_data_%d" % processed_data_id
+        conn_handler.create_queue(queue)
+
+        sql = ("DELETE FROM qiita.preprocessed_processed_data WHERE "
+               "processed_data_id = {0}".format(processed_data_id))
+        conn_handler.add_to_queue(queue, sql)
+
+        sql = ("DELETE FROM qiita.processed_filepath WHERE "
```
squirrelo: Also needs to remove the filepath table entries.

qiita_db/data.py
antgonza: Agree, will make the change.

qiita_pet/templates/study_description.html
```diff
@@ -222,6 +222,26 @@
   }
 }
 
+function delete_processed_data(processed_data_id) {
```
squirrelo: Outside the scope of this pull request, but we really should start using the AJAX POST capabilities instead of this form creation/submission we're currently doing.

squirrelo:
Few comments. Will test interface once addressed.

qiita_db/data.py
josenavas: This is something that we have not discussed as a group, but I will just make a note here. If you take a look, it doesn't even remove the row from qiita.filepath table. We are planning to run "purge_filepaths" in a cron job. That function, identifies which rows from qiita.filepath are not referenced and removes them from the table and from the filesystem. The advantage of this, is that the server is more free to answer requests rather than doing maintenance work, that can be done by a worker.

antgonza: For both comments: What we are currently doing is to leave the filepath entry (it doesn't affect anything) as deleting files could be pretty expensive. Additionally, the plan is to drop one of the already created functions once we have this in: https://github.com/biocore/qiita/issues/814

qiita_pet/templates/study_description.html
antgonza: OK, can you add an issue so we do not forget?

qiita_db/data.py
squirrelo: OK, can an issue be opened specifically for that then?

josenavas: See #814

squirrelo: One thing to note though with that way of doing things: If we want this to also have a single-user laptop mode, that way of cleanup will not work. If it's single user, it should delete the filepath entry and files immediately.

qiita_pet/templates/study_description.html
squirrelo: Issue #1036

qiita_db/data.py
antgonza: Not really, as the plan is that once you start the server: `qiita
webserver start`, the automatic job will also start. Please read the
issue and  your own comment.

qiita_pet/templates/study_description.html
antgonza: Thanks.

qiita_db/data.py
antgonza: BTW @josenavas, we discussed this decision as a group, search in your email for "deleting files and automatic jobs"

squirrelo: My question then is will that job be fired off as soon as the server starts or will there be a delay? Any delay means that the files could sit around indefinitely, as the user could start the server, quickly do something, and then shut it down again without that cleanup ever firing.

Also I do not have that email in any of my accounts.

josenavas: Sorry about that @antgonza I found the email. We should write those small parts on the Contributing.md file. Next week, I'm going to put a new version together and I will submit a PR, so we can all review it.

The reason why I thought it was not discussed is because that is handled inconsistently across the code base, so I thought that was not agreed. Thanks for pointing that out!

@squirrelo I forwarded the email to you for your reference.

antgonza: 1. Will you mind moving the discussion about how the job is gonna work
to the actual issue? That way we will ensure everybody's comments are
taken into account when that's resolved. Thanks!
2. Thread to which you replied:
https://groups.google.com/forum/#!searchin/qiita-dev/deleting$20files$20and$20automatic$20jobs/qiita-dev/X8pu1fHASuY/TT2_fARyS50J

squirrelo: Sure. Will just conclude by saying if this pull request is leveraging that expected functionality, that functionality should probably go in pretty quickly after this request.

antgonza:
Ready for another review.

qiita_db/data.py
```diff
@@ -1257,6 +1257,60 @@ def create(cls, processed_params_table, processed_params_id, filepaths,
...
+        conn_handler = SQLConnectionHandler()
+
+        analysis_ids = [str(_id[0]) for _id in conn_handler.execute_fetchall(
+            "SELECT DISTINCT analysis_id FROM qiita.analysis_sample WHERE "
+            "processed_data_id = {0} ORDER BY "
+            "analysis_id".format(processed_data_id))]
+
+        if analysis_ids:
+            raise QiitaDBError(
+                "Processed data %d is linked to (meta)analyses: %s" %
```
josenavas: Can we improve the error message here?
`Processed data %d cannot be removed because it is linked to the following (meta)analysis: %s`

qiita_db/data.py
josenavas: Not needed, see my comment below (about `purge_filepaths`)

qiita_db/data.py
```diff
@@ -1257,6 +1257,60 @@ def create(cls, processed_params_table, processed_params_id, filepaths,
...
+        """Removes the processed data with id processed_data_id
+
+        Parameters
+        ----------
+        processed_data_id : int
+            The processed data id
+
+        Raises
+        ------
+        QiitaDBUnknownIDError
```
josenavas: You are not raising this error, so no need to list here. In fact, there is no call to `exists` so probably useful to add.
Also, you're missing the description of the `QiitaDBError`.

qiita_db/data.py
squirrelo: analysis names would be better than IDs. There is no way to link an ID to an analysis in the UI

qiita_db/data.py
antgonza: That's actually needed (I think you are getting confused with some
table names), if not:
Error: update or delete on table "processed_data" violates foreign key
constraint "fk_processed_data_filepath" on table "processed_filepath"

DETAIL:  Key (processed_data_id)=(2) is still referenced from table
"processed_filepath".

qiita_db/test/test_data.py
```diff
@@ -785,6 +785,21 @@ def test_create(self):
         # preprocessed_data_id, processed_Data_id
         self.assertEqual(obs, [[1, 2]])
 
+    def test_delete(self):
+        """Correctly deletes a processed data and raises an error if it has
```
josenavas: In theory, these descriptions have to be one-liners, otherwise they do not show up correctly (or at least they did not show up before...)

josenavas: You're missing 1 test: the processed data cannot be removed because it is linked to a (meta)analysis.

qiita_db/data.py
squirrelo: Alternately, raise error without the IDs and use the Logger to store the IDs

josenavas:
Few more comments @antgonza. Thanks for putting this together!

qiita_db/data.py
antgonza: Good catch, I guess I erased the wrong ones in one of the iterations.

qiita_db/test/test_data.py
antgonza: OK

qiita_db/data.py
josenavas: @antgonza sorry for the confusion, I was answering to @squirrelo comment about deleting the entry on `qiita.filepath`. I guess I should have been more explicit :-)

qiita_db/data.py
squirrelo: I don't think this needs an exists call since line 1276 will do it for you, but I agree about removing the Raises text.

antgonza: ;) that's what I was doing

josenavas: Good call

qiita_db/data.py
antgonza: Going with adding names and QiitaDBError so the user can see the error.

squirrelo: One thing I just thought of if showing names: this will potentially show names of analyses that the user does not have access to. Probably not the end of the world, but could be a privacy issue.

antgonza: I think that's fine. The only way this could happen is if the owner
made it public, then move it to sandbox, and is trying to delete it.
Which will be a full other beast ...

qiita_db/data.py
```diff
@@ -1257,6 +1257,61 @@ def create(cls, processed_params_table, processed_params_id, filepaths,
...
+            The processed data id
+
+        Raises
+        ------
+        QiitaDBStatusError
+            If the processed data status is not sandbox
+        QiitaDBError
+            If the processed data has (meta)analyses
+        """
+        if cls(processed_data_id).status != 'sandbox':
```
josenavas: I'm just thinking about this. Should we limit the removal only to sandbox? I think we should be able to remove the processed data in any moment except if it is `public` (i.e. why we should not allow the removal if it is private?)

qiita_db/data.py
josenavas: And users should not be able to put it back to sandbox if it was public and
there was analysis using it.

2015-04-03 9:31 GMT-07:00 Antonio Gonzalez <notifications@github.com>:

> In qiita_db/data.py
> <https://github.com/biocore/qiita/pull/1034#discussion_r27738714>:
>
> > +            If the processed data status is not sandbox
> > +        """
> > +        if cls(processed_data_id).status != 'sandbox':
> > +            raise QiitaDBStatusError(
> > +                "Illegal operation on non sandbox processed data")
> > +
> > +
> > +        analysis_ids = [str(_id[0]) for _id in conn_handler.execute_fetchall(
> > +            "SELECT DISTINCT analysis_id FROM qiita.analysis_sample WHERE "
> > +            "processed_data_id = {0} ORDER BY "
> > +            "analysis_id".format(processed_data_id))]
> > +
> > +        if analysis_ids:
> > +            raise QiitaDBError(
> > +                "Processed data %d is linked to (meta)analyses: %s" %
>
> I think that's fine. The only way this could happen is if the owner made
> it public, then move it to sandbox, and is trying to delete it. Which will
> be a full other beast ...
>
> —
> <https://github.com/biocore/qiita/pull/1034/files#r27738714>.
>



-- 
Jose Navas

qiita_db/data.py
antgonza: Fine with me either way. However, remember that the process data is
the one that locks (sets the permissions of) everything (preprocess
data, pre template, sample ...) so allowing to only remove sandbox
kind of makes sense.

josenavas: Ok, fine with me. We can revisit this later and it is an easy change if we decide something different. I'm going to pull down the code to test.

josenavas:
:+1:

PR #1038, opened 2015-04-03
Cart analysis changes

squirrelo:
This pull request creates the carts for each user and the cart viewing page. Major pieces are:
* Default user analyses added to act as carts
* `Analysis.get_user_default(User)` function added to get the default analysis for a user
* `processed_data.processing_info` added to get processing information needed for cart display.
* Cart viewing page added at http://localhost:21174/analysis/selected/   Logging in as test@foo.bar using the test database will prepopulate it with a few samples selected so you can view the page. Currently the page can only be accessed through direct URL. 

This is the final piece needed before tying everything together from the last few pull requests into a single unified sample selection system.

This is also the first test of the separate branch for large changes system. This will be working on biocore/cart-branch and then merged in when finished.


squirrelo:
changed to new branch and ready for review

squirrelo:
ping @biocore/qiita-dev this needs review before I move forward on anything else.

qiita_db/data.py
```diff
@@ -1303,12 +1303,53 @@ def data_type(self, ret_id=False):
...
+            processed_params_id FROM qiita.{0}
+            WHERE processed_data_id=%s""".format(self._table)
+        static_info = conn_handler.execute_fetchone(sql, (self.id,))
+
+        # Get the info from the dynamic table, including reference used
+        sql = """SELECT * from qiita.{0}
+            JOIN qiita.reference USING (reference_id)
+            WHERE processed_params_id = {1}
+            """.format(static_info['processed_params_table'],
+                       static_info['processed_params_id'])
```
antgonza: Can these 2 sql be a single left (or right) join?

qiita_db/test/test_analysis.py
```diff
@@ -92,36 +92,36 @@ def test_has_access_no_access(self):
     def test_create(self):
         sql = "SELECT EXTRACT(EPOCH FROM NOW())"
         time1 = float(self.conn_handler.execute_fetchall(sql)[0][0])
-
+        new_id = get_count("qiita.analysis") + 1
```
antgonza: Nice! :smile:

josenavas:
:+1:

qiita_db/data.py
squirrelo: No, since it is dynamic table info so we need tha table name and row from the first to do the second query.

antgonza: OK, I think your method is way simpler even if it requires 2 queries.

PR #1041, opened 2015-04-06
Create analysis from cart

squirrelo:
This adds the `from_default` flag to the Analysis object, allowing you to create a new analysis from previously selected samples. This also modifies the analysis creation to use queues. 

~~This pull request relies on #1038 so please merge that one first.~~ Merged. Many of the file changes are from that, only `Analysis.create()` and the `test_analysis_from_default` test function were modified in this pull request.

### DIFF
--- a/qiita_db/analysis.py
+++ b/qiita_db/analysis.py
@@ -136,22 +136,23 @@ class Analysis(QiitaStatusObject):
                     WHERE analysis_id = {1};
                     RETURN aid;
                 END $do$""".format(cls._table, dflt_id)
-            a_id = conn_handler.execute_fetchone(
-                sql, (owner.id, name, description))[0]
+            a_id = conn_handler.add_to_queue(
+                queue, sql, (owner.id, name, description))
         else:
             # insert analysis information into table as "in construction"
             sql = ("""INSERT INTO qiita.{0} (email, name, description, "
                    "analysis_status_id) VALUES (%s, %s, %s, 1) "
                    "RETURNING analysis_id""".format(cls._table))
-            a_id = conn_handler.execute_fetchone(
-                sql, (owner.id, name, description))[0]
+            a_id = conn_handler.add_to_queue(
+                queue, sql, (owner.id, name, description))[0]
 
         # add parent if necessary
         if parent:
             sql = ("INSERT INTO qiita.analysis_chain (parent_id, child_id) "
-                   "VALUES (%s, %s)")
-            conn_handler.execute(sql, (parent.id, a_id))
+                   "VALUES (%s, {0}) RETURNING child_id")
+            conn_handler.add_to_queue(queue, sql, [parent.id])
 
+        a_id = conn_handler.execute_queue(queue)[0]
         return cls(a_id)
 
     # ---- Properties ----
--- a/qiita_db/analysis.py
+++ b/qiita_db/analysis.py
@@ -99,7 +99,7 @@ class Analysis(QiitaStatusObject):
         return {x[0] for x in conn_handler.execute_fetchall(sql, (status,))}
 
     @classmethod
-    def create(cls, owner, name, description, parent=None):
+    def create(cls, owner, name, description, parent=None, from_default=False):
         """Creates a new analysis on the database
 
         Parameters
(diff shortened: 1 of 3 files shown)


squirrelo:
Ready for review

qiita_db/analysis.py
```diff
@@ -112,23 +112,44 @@ def create(cls, owner, name, description, parent=None):
...
-               "analysis_status_id) VALUES (%s, %s, %s, 1) "
-               "RETURNING analysis_id".format(cls._table))
-        a_id = conn_handler.execute_fetchone(
-            sql, (owner.id, name, description))[0]
+        if from_default:
+            # insert analysis and move samples into that new analysis
+            dflt_id = owner.default_analysis
+            sql = """INSERT INTO qiita.{0}
+                    (email, name, description, analysis_status_id)
+                    VALUES (%s, %s, %s, 3)
```
josenavas: can we avoid hardcoding IDs? We never now when these can change. Using the `convert_to_id` function should make the trick.

qiita_db/analysis.py
```diff
@@ -112,23 +112,44 @@ def create(cls, owner, name, description, parent=None):
...
+                                                   description))
+            sql = """UPDATE qiita.analysis_sample
+                     SET analysis_id = %s
+                     WHERE analysis_id = %s RETURNING %s"""
+            conn_handler.add_to_queue(queue, sql, ['{0}', dflt_id, '{0}'])
+        else:
+            # insert analysis information into table as "in construction"
+            sql = """INSERT INTO qiita.{0}
+                  (email, name, description, analysis_status_id)
+                  VALUES (%s, %s, %s, 1)
```
josenavas: Same here.

josenavas:
Just a small comment.

qiita_db/analysis.py
squirrelo: convert_to_id doesn't currently work with status tables, hence issue #292. I have updated the `convert_to_id` function to work around the issue for now though.

squirrelo: Aaaand that update breaks all the things. Using a specific SQL statement here and tagged with issue ID.

josenavas: Thank you.

squirrelo:
Changes made and ready for another review.

qiita_db/analysis.py
```diff
@@ -112,23 +112,53 @@ def create(cls, owner, name, description, parent=None):
...
+            dflt_id = owner.default_analysis
+            sql = """INSERT INTO qiita.{0}
+                    (email, name, description, analysis_status_id)
+                    VALUES (%s, %s, %s, %s)
+                    RETURNING analysis_id""".format(cls._table)
+            conn_handler.add_to_queue(queue, sql, (owner.id, name,
+                                                   description, status_id))
+            # MAGIC NUMBER 3: command selection step
+            # needed so we skip the sample selection step
+            sql = """INSERT INTO qiita.analysis_workflow
```
josenavas: I just noticed this. Why is the `analysis_workflow` table needed? Is it only handling a column called step, and there is no further control on it, so it looks like that column can be included in the analysis object???

josenavas:
@squirrelo just one comment that should not block the PR, but I'd like to get an answer to check if we actually need to open an issue for that, as it looks like that table is not needed at all.
Thanks!

qiita_db/analysis.py
squirrelo: The table was supposed to be a housekeeping table to track what step the analysis is on and any other things that needed tracking during analysis creation. I guess it turns out we didn't need as much housekeeping as originally thought.

josenavas: Yeah, I think we are keeping track of everything else in the multiple tables that build up the analysis and the step just limits what can be done on the analysis. I've created a new issue https://github.com/biocore/qiita/issues/1052 and I assigned to beta. Later we can decide to exactly which alpha release it belongs, but I don't think it is either 0.1 or 0.2. Thanks for the feedback!

josenavas:
:+1:

squirrelo:
Can this get a second reviewer?

ElDeveloper:
I'll do this in a bit.

PR #1042, opened 2015-04-07
Delete preprocess data

antgonza:


qiita_db/data.py
```diff
@@ -807,6 +807,72 @@ def create(cls, study, preprocessed_params_table, preprocessed_params_id,
...
+            The preprocessed data id
+
+        Raises
+        ------
+        QiitaDBStatusError
+            If the preprocessed data status is not sandbox or if the
+            preprocessed data EBI and VAMPS submission is not in a valid state
+            ['not submitted', 'failed']
+        QiitaDBError
+            If the preprocessed data has (meta)analyses
```
josenavas: This is not accurate: `If the preprocessed data has been processed`

josenavas:
Just a documentation comment. Once it is fixed, :+1:

ElDeveloper:
I tried deleting a Preprocessed data object that had already generated a processed data object and I got the following error:

`Couldn't remove preprocessed data 111: Illegal operation due to EBI submission status ('false') or VAMPS submission status ('not submitted')`

And I honestly don't understand what that means.

qiita_db/data.py
```diff
@@ -807,6 +807,72 @@ def create(cls, study, preprocessed_params_table, preprocessed_params_id,
...
+            preprocessed data EBI and VAMPS submission is not in a valid state
+            ['not submitted', 'failed']
+        QiitaDBError
+            If the preprocessed data has been processed
+        """
+        valid_submission_states = ['not submitted', 'failed']
+        ppd = cls(ppd_id)
+        if ppd.status != 'sandbox':
+            raise QiitaDBStatusError(
+                "Illegal operation on non sandbox preprocessed data")
```
ElDeveloper: Since these messages (832 and 836) are likely to be seen by users on the GUI, do you think you could rephrase them?

qiita_db/data.py
```diff
@@ -807,6 +807,72 @@ def create(cls, study, preprocessed_params_table, preprocessed_params_id,
...
+        elif (ppd.submitted_to_vamps_status() not in valid_submission_states or
```
squirrelo: Can you break this into two if statements? That way we know which is actually the error, VAMPS or INSDC status.

qiita_db/test/test_data.py
```diff
@@ -478,6 +478,38 @@ def test_create_data_type_only(self):
...
 
+    def test_delete(self):
+        """Correctly deletes a preprocessed data"""
+        # testing regular delete
+        ppd = PreprocessedData.create(
+            self.study, self.params_table,
+            self.params_id, self.filepaths, prep_template=self.prep_template,
+            ebi_submission_accession=self.ebi_submission_accession,
+            ebi_study_accession=self.ebi_study_accession)
+        PreprocessedData.delete(ppd.id)
```
squirrelo: This should have some SQL tests to make sure the delete actually removed all the rows expected from the DB.

squirrelo:
Couple of comments.

qiita_db/data.py
ElDeveloper: :+1:

antgonza: OK

qiita_db/test/test_data.py
antgonza: I can add but I see them as useless as the test just after this one is to check that the deletion fails cause this object doesn't exist anymore, do you agree?

squirrelo: This is to make sure that the lines are all deleted out of the 4 or 5 tables, since the failure is only due to one row in one table.

antgonza: Well, the lines are in a queue so if one fails, all fail, right?

squirrelo: It's for future-proofing to make sure the lines are always deleted no matter the code changes. If we change the code, we still need to make sure the functionality remains.

antgonza: I will argue that if somebody changes the way queues work we will have a bigger issue than a delete, could you make a better argument of why do you feel so strongly about this?

josenavas: I'd suggest that it will be enough using the function `get_count` on these tables and make sure that their count has been reduced.

antgonza: Again, not sure what's the benefit ...

josenavas: @ElDeveloper any thoughts on this? I'm fine with the code as it is. The reasons why I do not think that more checks are needed are:
 - Everything is in a queue. If one statement fails, all of them fail.
 - If one of the other tables are not removed, SQL will complain due to FK restriction.

I was just proposing an alternative solution, to see if @antgonza and @squirrelo agree on something intermediate. @squirrelo given these arguments, are you still feeling strong on put some extra checks in place?

squirrelo: My feeling is that, since this is a test of functionality, it should
explicitly test the functionality. Since the code is supposed to delete
rows from multiple tables, the deletion of those rows should be explicitly
checked. This also means that if any future code does anything to deletion,
we can still make sure those rows are deleted.

ElDeveloper: @josenavas is right, the FK restriction and the usage of queues should keep this sane, @squirrelo can you still see a scenario where we would not be safe?

However I don't think a check for "trying to delete" again should be the way to verify things work, the test below should check it can't instantiate the object. And for completeness there should be another test trying to delete an object that doesn't exist (maybe this same one).

josenavas: :+1: on the changes proposed by @ElDeveloper

antgonza: That makes a bit more sense. Nevertheless, the first thing we do when trying to delete any object is to check that the object exists by instantiate it, which we do cause we need to check different status to allow delete. In fact, if you see the test is that it raises a QiitaDBUnknownIDError, which is raised by the instantiation.

josenavas: I see your point @antgonza. However, this won't work in the case that we move the "delete" function from a classmethod to an instance method. This is something that I've been thinking and I was planning to propose to the group, as I think it makes way more sense to have the delete function as an instance method rather than a classmethod.

antgonza: If you feel like that's what we should do, IMOO and following the "rules" we discussed last week, that should be a separate issue. Anyway, the concern about having it as instance method is:
```
pd.delete()
pd.whatever = 55
```

josenavas: Yes, I agree that it should be a separate issue. BTW, that concern is not solved with the current solution:

```python
PreprocessedData.delete(ppd.id)
ppd.whatever = 55
```

I moved the discussion here: https://github.com/biocore/qiita/issues/1045

josenavas: BTW, I think that @ElDeveloper enhancements should be implemented, as it is safer.

squirrelo: Agree with @ElDeveloper enhancements as well.

qiita_db/data.py
ElDeveloper: Thanks for the changes. I think it should be "sandboxed" not "sandbox".

qiita_db/test/test_data.py
```diff
@@ -478,6 +478,44 @@ def test_create_data_type_only(self):
...
+
+        # testing that it raises an error if ID doesn't exist
+        with self.assertRaises(QiitaDBUnknownIDError):
+            PreprocessedData.delete(ppd.id)
+
+        # testing that we can not remove cause the preprocessed data != sandbox
+        with self.assertRaises(QiitaDBStatusError):
+            PreprocessedData.delete(1)
+
+    def test_delete_advance(self):
```
ElDeveloper: advance -> advanced

ElDeveloper:
Testing on my system.

ElDeveloper:
Looks good from my testing the only outstanding thing is that, it gave me the following message even though I never submitted to EBI:

screen shot 2015-04-08 at 10 38 19 am

I attach the whole screenshot, because it may be that the EBI status it has is messing with the check? Not sure why is this the case though.

ElDeveloper:
To the rest of the qiita devs, the reason why this is failing for some studies is because 'false' is not a valid status for the EBI submission. The [`load_preprocessed`](https://github.com/biocore/qiita/blob/0a047797cea88567d88ffda259465235c1aa916c/scripts/qiita#L138) command is the culprit of this inconsistency as it sets the value to `true` or `false`. AFAIK, `false` and `true` are not accepted statuses, but if I'm wrong, then this PR should accept `true`/`false` as a valid status, can someone else confirm?

PR #1043, opened 2015-04-07
Unify template validation

josenavas:
Moves the function _clean_validate_template to the base class in order to remove code duplication. Tests added to both subclasses.


josenavas:
It looks like one of my previous PR introduced a potential stochastic failure in the system; since it is a test that relies in a dict order. I'm going to solve it as a separate PR and then I'll pull here.

josenavas:
:boom: ready for review!

qiita_db/metadata_template/base_metadata_template.py
```diff
@@ -520,6 +523,82 @@ def _check_special_columns(cls, md_template, obj):
...
+
+        # Prefix the sample names with the study_id
+        prefix_sample_names_with_id(md_template, study_id)
+
+        # In the database, all the column headers are lowercase
+        md_template.columns = [c.lower() for c in md_template.columns]
+
+        # Check that we don't have duplicate columns
+        if len(set(md_template.columns)) != len(md_template.columns):
+            raise QiitaDBDuplicateHeaderError(
```
ElDeveloper: Hey, just noticed that this error message is not going to be very information (and it is not your fault, the code was like this already). If possible can you change to something along the lines of: `'The following header names are duplicated: %s' % ', '.join(find_duplicates(md_template.columns))`

ElDeveloper:
Looks really good, just one comment, and if another person can have a look, this should be ready to be merged.

ElDeveloper:
Also, just noticed this has conflicts now, probably my bad :-\

qiita_db/metadata_template/base_metadata_template.py
ElDeveloper: Nix that, QiitaDBDuplicateHeaderError already takes care of that formatting.

josenavas:
Thanks @ElDeveloper , conflicts solved

ElDeveloper:
Thanks! This one needs one more reviewer and should be ready to go! It's a really quick one :clock1:

josenavas:
ping @biocore/qiita-dev I would appreciate a review on all my PR, but specially this one. The changes are starting creating merge conflicts between them.

qiita_db/metadata_template/base_metadata_template.py
```diff
@@ -585,6 +588,82 @@ def _check_special_columns(cls, md_template, obj):
...
+            If the sample names in md_template contains invalid names
+        QiitaDBDuplicateHeaderError
+            If md_template contains duplicate headers
+        QiitaDBColumnError
+            If md_template is missing a required column
+        """
+        cls._check_subclass()
+        invalid_ids = get_invalid_sample_names(md_template.index)
+        if invalid_ids:
+            raise QiitaDBColumnError("The following sample names in the sample"
```
squirrelo: ...in the template... (remove sample)

qiita_db/metadata_template/base_metadata_template.py
```diff
@@ -585,6 +588,82 @@ def _check_special_columns(cls, md_template, obj):
             cls._check_template_special_columns(md_template, obj))
 
     @classmethod
+    def _clean_validate_template(cls, md_template, study_id, obj,
```
squirrelo: How do you validate that the passed study_id is actually attached to this template?

squirrelo:
Couple questions/comments.

qiita_db/metadata_template/base_metadata_template.py
josenavas: That's a good point. If the template being tested is a prep template, the process will fail later because the samples are not found in the sample template. If it is the study_id, there is no way, since the object still doesn't exist. But 2 things: (1) this was already like that in the previous implementation; and (2) this is a private function that should not be called from outside the class, so if somebody is messing up with this function they should know what they're doing (and this is completely pythonic).

josenavas:
@squirrelo addressed/commented

qiita_db/metadata_template/base_metadata_template.py
squirrelo: Can we get another dev to chime in on this? It's a little disconcerting to me to just take it on faith the study is correct and wait until it fails to see otherwise.

squirrelo: Discussed offline and :+1: will merge when tests pass

PR #1044, opened 2015-04-07
Moving setitem to the base class

josenavas:
I just move the code to the base class and updated the SQL to make use of the class variables to fill the sql correctly.


qiita_db/metadata_template/test/test_prep_template.py
```diff
@@ -220,9 +220,11 @@ def test_get_none(self):
...
     """Tests the PrepSample class"""
     def test_setitem(self):
-        """setitem raises an error (currently not allowed)"""
-        with self.assertRaises(QiitaDBNotImplementedError):
-            self.tester['barcodesequence'] = 'GTCCGCAAGTTA'
+        with self.assertRaises(QiitaDBColumnError):
+            self.tester['column that does not exist'] = 0.3
+        self.assertEqual(self.tester['center_name'], 'ANL')
+        self.tester['center_name'] = "FOO"
+        self.assertEqual(self.tester['center_name'], "FOO")
```
squirrelo: This also needs a test for when a column has the wrong datatype, triggering the ValueError.

qiita_db/metadata_template/base_metadata_template.py
```diff
@@ -308,17 +309,79 @@ def handler(x):
...
+        column : str
+            The column to update
+        value : str
+            The value to set. This is expected to be a str on the assumption
+            that psycopg2 will cast as necessary when updating.
+
+        Raises
+        ------
+        QiitaDBColumnError
+            If the column does not exist in the table
```
squirrelo: Can you add the ValueError raised when value and database column types are not compatible here?

squirrelo:
Couple comments.

qiita_db/metadata_template/test/test_prep_template.py
josenavas: This is a great point! When I wrote the test I noticed that there is a "bug" on the code (it is just preventing a useful message, therefore quoted). I'm fixing it right now.

qiita_db/metadata_template/base_metadata_template.py

qiita_db/metadata_template/base_metadata_template.py
```diff
@@ -308,17 +309,79 @@ def handler(x):
...
+                conn_handler.execute(sql, (value, self._id))
+            except Exception as e:
+                column_type = conn_handler.execute_fetchone(
+                    """SELECT data_type
+                       FROM information_schema.columns
+                       WHERE column_name=%s AND table_schema='qiita'
+                    """, (column,))[0]
+                value_type = type(value).__name__
+
+                if column_type != value_type:
```
squirrelo: This seems to be comparing SQL types (column_type) and python type (value_type). Is that kosher?

josenavas:
:boom: ready for another round of review!

ElDeveloper:
:+1:

PR #1047, opened 2015-04-07
Removing the remove_category function

josenavas:
I don't think the `remove_category` function is needed, so I'm removing it from the code base.

The reasons why I think this function should be removed:
 - It is not used in any place in the code base
 - It has a comment that says "This operation may invalidate another user's perspective on the table", but no checks o safe mechanism is put in place.
 - It is buggy: it is dropping a column from the dynamic table, but is not updating the study-columns table.
 - Leaves the filesystem in an inconsistent state: once the column is removed, the sample template file and the qiime mapping file is not being re-generated, so the files do not represent the data in the DB
 - Before removing any column in the metadata templates, multiple checks has to be done (i.e. it has been used in an analysis).

Since it is clear that a lot of though has to be put in this function, I think it is safer to just remove it for now and put it back once the functionality is needed and we actually have a design for all the mechanisms to notify users once a template has been changed in this way. Note that IMOO, adding new columns is not a big deal, however, removing columns is another story.

### DIFF
--- a/qiita_db/metadata_template/sample_template.py
+++ b/qiita_db/metadata_template/sample_template.py
@@ -481,32 +481,6 @@ class SampleTemplate(MetadataTemplate):
                     if '_qiime_' not in basename(fp):
                         pt.create_qiime_mapping_file(fp)
 
-    def remove_category(self, category):
-        """Remove a category from the sample template
-
-        Parameters
-        ----------
-        category : str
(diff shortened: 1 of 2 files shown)


ElDeveloper:
:+1: on the basis that it is not being used anywhere and its use could lead to an inconsistent state of the system.

PR #1048, opened 2015-04-08
BUG: Fix bug with Safari and MOI

### LINKED ISSUE #946: Server web sockets don't work on Safari
Running on OS X 10.10.

ElDeveloper:
Assigning a function to document.ready doesn't seem to be supported in Safari,
and thus the call to initialize MOI would never occurr (giving the impression
that WebSocket never started). Using jQuery's proxy for the ready callback
seems to resolve this problem.

Fixes #946


josenavas:
:+1:

PR #1050, opened 2015-04-08
Unify update category

josenavas:
I moved the `update_category` function to the base class. This PR includes the commits in #1044 so review/merge that one first.

I just moved the code around and created the tests for the prep template class.

PR #1054, opened 2015-04-09
Fix inconsistency on failure updating

josenavas:
~~This PR depends on #1050, review and merge that one first.~~

Main motivation of the PR: the `update_category` function can leave the DB in an inconsistent state because the updates to the samples are done independently, so if one fails, the previous ones are already modified. By moving the `__setitem__` queries to a function that adds them to a queue, we can use that function in `update_category` to perform everything in a single transaction block.

Also, in a previous PR, @squirrelo commented that the way types in the error were checked was probably not correct (for some reason I received that notification from github today). I tested it and @squirrelo was right, so I fixed it here. It is less elegant, but at least is correct.

### DIFF
--- a/qiita_db/metadata_template/base_metadata_template.py
+++ b/qiita_db/metadata_template/base_metadata_template.py
@@ -36,7 +36,7 @@ Methods
 # -----------------------------------------------------------------------------
 
 from __future__ import division
-from future.utils import viewitems
+from future.utils import viewitems, viewvalues
 from os.path import join
 from functools import partial
 
@@ -45,7 +45,8 @@ import pandas as pd
 from qiita_core.exceptions import IncompetentQiitaDeveloperError
 from qiita_db.exceptions import (QiitaDBUnknownIDError,
                                  QiitaDBNotImplementedError,
-                                 QiitaDBColumnError)
+                                 QiitaDBColumnError,
+                                 QiitaDBExecutionError)
 from qiita_db.base import QiitaObject
 from qiita_db.sql_connection import SQLConnectionHandler
 from qiita_db.util import (exists_table, get_table_cols,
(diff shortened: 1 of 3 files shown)


squirrelo:
One small comment.

qiita_db/metadata_template/base_metadata_template.py
```diff
@@ -363,27 +364,51 @@ def __setitem__(self, column, value):
...
             # column type or something else
-            column_type = conn_handler.execute_fetchone(
-                """SELECT data_type
-                   FROM information_schema.columns
-                   WHERE column_name=%s AND table_schema='qiita'
-                """, (column,))[0]
-            value_type = type(value).__name__
-
-            if column_type != value_type:
+            if "invalid input syntax for" in str(e):
```
squirrelo: Something like [this](https://github.com/biocore/qiita/blob/master/qiita_db/search.py#L277) from search.py would probably be cleaner, since it directly connects the python datatype to the SQL datatype and the invalid syntax could be for other reasons besides type.

qiita_db/metadata_template/base_metadata_template.py
```diff
@@ -1101,12 +1126,44 @@ def update_category(self, category, samples_and_values):
...
             sample = self[k]
-            sample[category] = v
+            sample.add_setitem_queries(category, v, conn_handler, queue_name)
+
+        try:
+            conn_handler.execute_queue(queue_name)
+        except QiitaDBExecutionError as e:
+            # catching error so we can check if the error is due to different
+            # column type or something else
+            if "invalid input syntax for" in str(e):
```
squirrelo: same as above

squirrelo:
One comment.

josenavas:
This should be ready for final review/merge!

josenavas:
ping @biocore/qiita-dev

qiita_db/metadata_template/base_metadata_template.py
```diff
@@ -313,8 +315,8 @@ def handler(x):
                            " in template %d" %
                            (key, self._id, self._md_template.id))
 
-    def __setitem__(self, column, value):
-        r"""Sets the metadata value for the category `column`
+    def add_setitem_queries(self, column, value, conn_handler, queue):
+        """Adds the SQL queries needed to set and item to the queue
```
ElDeveloper: Can you rephrase this docstring, I don't think it makes sense.

ElDeveloper:
Looks good, but I think this might be missing some tests?

qiita_db/metadata_template/base_metadata_template.py

josenavas:
Thanks @ElDeveloper 
I've added tests!

ElDeveloper:
Thanks @josenavas!

PR #1055, opened 2015-04-09
Removing add_category function

josenavas:
Reasons for it's removal:
 - It is not used anywhere in the code base
 - It has a bug, it is not modifying the `study_columns` table.
 - It is repetitive. The same functionality can be accomplished with the `extend` function.


ElDeveloper:
:+

ElDeveloper:
1:

PR #1057, opened 2015-04-13
Websockets on cart page

squirrelo:
Requires #1040 so merge that first.

This adds websocket-based removal of samples and processed data on the selected samples cart page. It also moves the analysis creation to that page as well.


squirrelo:
Woops, pointed at master.

Issue #993, opened 2015-03-18
WebSockets will not work over HTTPS


ElDeveloper:
If the server is being accessed through HTTPS, anything that relies on WebSockets will fail silently and only show an error on the JavaScript command line. The most important feature that's unavailable when this happens is being able to submit jobs.

```javascript
[blocked] The page at 'https://qiita.ibdmdb.org/analysis/wait/1' was loaded over HTTPS, but ran insecure content from 'ws://qiita.ibdmdb.org/moi-ws/': this content should also be loaded over HTTPS. moi.js:77
Uncaught SecurityError: Failed to construct 'WebSocket': An insecure WebSocket connection may not be initiated from a page loaded over HTTPS.
```

For more information, see [this issue in moi](https://github.com/biocore/mustached-octo-ironman/issues/16) and this [StackOverflow question](http://stackoverflow.com/q/25496039/379593).

squirrelo:
The fix seems to just be to use `wss://` but we need to know when to use that vs the standard.

ElDeveloper:
I tried by changing moi.js and compute_waiting.html to use wss but no
luck either.

wasade:
Some info if you google "tornado was". Are the certs etc setup in
webserver.py?

ElDeveloper:
I tried this out, but it didn't work either.

ElDeveloper:
I was able to verify that the websockets in fact work using Safari 8.xxxx. It seems that the problem is arising from the fact that the interface is not being properly updated and gives a misleading impression of nothing happening. My guess is that the creation of the HTML elements is not working.

squirrelo:
Needs to be verified, but may have been solved by #1058

ElDeveloper:
How can that be? I don't recall any changes to the webserver's configuration to take in a certificate.

squirrelo:
Oh, I thought you had the certs up and it was just a ws:// vs wss:// issue. My bad.

ElDeveloper:
Np, it's really both things. We should add a parameter to the config
file such that you can set what the certificates, etc you want to use
are.

antgonza:
Now we are using https and we don't have this issue, @ElDeveloper can we close?

ElDeveloper:
Yes, I believe this is the case, since we now have the system work with HTTPS.

Allan-Nava:
[...]

PR #1056, opened 2015-04-11
Integrate select samples

squirrelo:
~~This pull request depends on #1039 so merge that one first~~ Merged

This pull request slightly changes the analysis object's add_samples feature to bring it in line with the current sample passing convention of dict of list in form {pid: [samp1, samp2, ...], ...}

It also adds websocket-based sample selection to the studies listing page, allowing the selection of samples in real time without having to click a button.

NOTE: this pull request seems large because it is also ripping out the old selection mechanism.

PR #1058, opened 2015-04-13
Websockets on cart page

squirrelo:
~~Requires #1041 so merge that first.~~Merged
~~Depends on #1056 so merge that first~~Merged

This adds websocket-based removal of samples and processed data on the selected samples cart page. It also moves the analysis creation to that page as well.


squirrelo:
Ready for review

ElDeveloper:
argh, conflicts, can you pull from upstream master and upstream cart-branch.

squirrelo:
Done, tests running now.

squirrelo:
@biocore/qiita-dev Tests pass, ready for review!

ElDeveloper:
Ok, now I'm reviewing this one.

qiita_pet/handlers/websocket_handlers.py
```diff
@@ -74,7 +74,24 @@ def on_close(self):
...
 
+class SelectedSocketHandler(WebSocketHandler, BaseHandler):
+    """Websocket for removing samples on default analysis display page"""
+    @authenticated
+    def on_message(self, msg):
+        # When the websocket receives a message from the javascript client,
+        # parse into JSON
+        msginfo = loads(msg)
+        default = Analysis(self.current_user.default_analysis)
+        if 'samples' not in msginfo:
```
ElDeveloper: I think you can simplify this if/else block by using msginfo.get('samples'), which would return a `None` when `'samples'` is not part of the dictionary. And perhaps, this is the way both arguments should be retrieved such that in case nothing was sent remove_samples would raise an exception (QiitaIncompetentDeveloper, I think).

ElDeveloper:
This looks great!! Such an awesome improvement overall, a few things:

- If you (1) select samples for an analysis, then (2) go to "view selected samples", then (3) remove all the samples and (4) try to create the analysis. You'll get a 405 error. Similarly if no samples are selected and you click "view selected samples", you'll get the following traceback:

```python
  File "/Users/yoshikivazquezbaeza/.virtualenvs/qiita/lib/python2.7/site-package
s/tornado/web.py", line 1141, in _when_complete
    callback()

  File "/Users/yoshikivazquezbaeza/.virtualenvs/qiita/lib/python2.7/site-package
s/tornado/web.py", line 1162, in _execute_method

  File "/Users/yoshikivazquezbaeza/.virtualenvs/qiita/lib/python2.7/site-package
s/tornado/web.py", line 2297, in wrapper

  File "/Users/yoshikivazquezbaeza/git_sw/qiita/qiita_pet/handlers/analysis_hand
lers.py", line 220, in get
    proc_info=proc_data_info)

  File "/Users/yoshikivazquezbaeza/.virtualenvs/qiita/lib/python2.7/site-packages/tornado/web.py", line 538, in render
    html = self.render_string(template_name, **kwargs)

  File "/Users/yoshikivazquezbaeza/.virtualenvs/qiita/lib/python2.7/site-packages/tornado/web.py", line 645, in render_string
    return t.generate(**namespace)

  File "/Users/yoshikivazquezbaeza/.virtualenvs/qiita/lib/python2.7/site-packages/tornado/template.py", line 273, in generate
    return execute()

  File "analysis_selected_html.generated.py", line 276, in _tt_execute
    for pid in proc_datas:  # analysis_selected.html:79 (via sitebase.html:192)

UnboundLocalError: local variable 'proc_datas' referenced before assignment
```

qiita_pet/templates/analysis_selected.html
```diff
@@ -63,6 +72,10 @@
 </div>
 {% end %}
 
+<button type="button" class="btn btn-primary btn-lg" data-toggle="modal" data-target="#create-analysis-modal-view">
```
ElDeveloper: This button would be better ( I think ) at the top.

ElDeveloper: And probably colored in "green"

ElDeveloper:
When I select a few samples and try to create the analysis, I get a 405 error.

ElDeveloper:
Last thing, now that I've been able to list several studies under the "/selected/" handler, I think it would be better if the studies were presented inside a data table. And the table shouldn't use the full width of the window, it looks weird.

qiita_pet/templates/analysis_selected.html
```diff
@@ -1,8 +1,15 @@
 {% extends sitebase.html %}
 {% block head %}
 <script type="text/javascript">
+var ws = new WebSocket('ws://' + window.location.host + '/analysis/selected/socket/');
```
wasade: It would be really nice if it were possible to reuse the moi init here since it is already doing browser checking.

qiita_pet/templates/analysis_selected.html
```diff
@@ -1,8 +1,15 @@
...
+ws.onmessage = function(evt) { if(evt.data != 'true') {ws.close();} };
```
wasade: Wouldn't jquery be better here, or at least equivalent? I ask because the websocket is created and torn down per message and isn't persistent.

qiita_pet/handlers/websocket_handlers.py
wasade: if the object sent over the line always includes `samples` then the message is consistent. In this branch, the message would just need to have `samples` associated with `[]`

wasade: or None

qiita_pet/templates/list_studies.html
```diff
@@ -11,6 +11,11 @@
 
 <script type="text/javascript">
 var current_study;
+var ws = new WebSocket('ws://' + window.location.host + '/analysis/selected/socket/');
```
wasade: same here on construction. if not moi's init, the construction needs to be centralized somehow otherwise boilerplate like browser checks will become tedious

qiita_pet/templates/analysis_selected.html
squirrelo: The connection is persistent, this is just so the websocket is closed if there is an error server side and the expected response isn't given. Then it closes the connection so the error is shown.

qiita_pet/templates/analysis_selected.html
squirrelo: I have it at the bottom to force people to look at what they selected before hitting create. I can make it green though.

qiita_pet/templates/analysis_selected.html
squirrelo: Can do.

qiita_pet/templates/analysis_selected.html
wasade: Isn't that implicit already?

squirrelo: Maybe? I may be being overly cautious, but at this point my thinking is better safe than sorry.

qiita_pet/templates/analysis_selected.html
squirrelo: Added an `init_websocket` function for taking care of boilerplate code.

qiita_pet/templates/analysis_selected.html
wasade: Well, better to understand than guess...

squirrelo: So it will close if there is an error raised during the action in the socket, but not if something goes wrong silently. This will take care of those silent issues.

squirrelo:
@ElDeveloper I looked at the datatables idea and, to me, it doesn't look like it will make this more usable. The issue is the show/hide samples, and how the datatables Jquery will interact with the removal code.

squirrelo:
Ready for another review.

ElDeveloper:
Ok, I was proposing that idea, because out of the box it organizes large
data sizes, but if you think it's not a good idea, then I trust you.
However, I think the current view could be better.

squirrelo:
@biocore/qiita-dev Can someone give this another review?

ElDeveloper:
@wasade, can you check if your comments were addressed, after that, this should be ready for merge.

josenavas:
I've not gone over the code but I was testing this branch as @squirrelo suggested that this is the last PR on this branch.

One of the things that act weirdly is that I can select samples, go to the "View selected samples" page; and, when I go back to the study listing page, I have no feedback of which samples have been already selected. I would suggest that the checkbox should be checked if you've already chosen those samples. Alternatively,  I would suggest changing the checkbox to a `+` sign (glyphicon); and use a `-` sign when that study has been already selected. Other options are using a color scheme, but I think this can look like a Christmas tree. Anyways, I think some kind of feedback is needed in the study selection page.

josenavas:
Also, I cannot create an analysis. I added a new study, completely processed it, I only choose it's samples. When I click create analysis on the modal that appears, nothing happens...

squirrelo:
Fixed the issue with creation of new analyses, the modal was submitting to the wrong page. 

As for previously selected samples, I agree that something needs to be done, but it's something that needs a lot of discussion and thought. For now, I think what we have will work, and it can be expanded upon when we come to a conclusion for how previously selected samples are shown. The big issue is that you aren't selecting by sample, but by processed data.

josenavas:
Thanks for the comments @squirrelo 
However, I don't think that we should merge all this branch into master (it can go to it's own branch). Based on the new Contributing.md document, this functionality is not complete, and leaving the system in a somewhat inconsistent state (it can be really misleading to the user - it was for me...), so we should discuss this and being able to make it for the alpha 0.1 release...

squirrelo:
This is good to go as is for now though, since it's still on it's own branch. In light of that, can we get this a final review and merge?

josenavas:
As @ElDeveloper suggests, I will be more comfortable if @wasade can review this, as he has more experience with websockets and he already had some comments on this PR.

squirrelo:
This only changed a single number, so the tests will pass fine. It's good to review.

qiita_pet/templates/analysis_selected.html
```diff
@@ -1,8 +1,15 @@
 {% extends sitebase.html %}
 {% block head %}
 <script type="text/javascript">
+var ws = init_websocket(window.location.host + '/analysis/selected/socket/');
+ws.onmessage = function(evt) { if(evt.data != 'true') {ws.close();} };
+ws.onerror = function(evt) { $('#ws-error').html("<b>Server communication error. Sample removal will not be recorded. Please try again later.</b>"); };
+ws.onclose = function(evt) { $('#ws-error').html("<b>Server communication error. Sample removal will not be recorded. Please try again later.</b>"); };
```
wasade: Closing a web socket should be a normal operation, right? Why is this considered an error?

qiita_pet/templates/analysis_selected.html
```diff
@@ -57,13 +66,7 @@
...
 		</table>
 		</td></tr>
-	{% end %}
-	</table>
-	</div>
-</div>
-{% end %}
-
-{% for pid in proc_datas%}
+		{% for pid in proc_datas%}
```
wasade: Please add a space before the second `%`

wasade:
What is the reason for not reusing the moi client-side websocket code? It would require only a minor modification to moi.js (allow specifying an onerror method at construction, and the specific handler the websocket should use). This is an opportunity to reduce js code replication and centralize websocket handling client-side. 
.

qiita_pet/templates/analysis_selected.html
squirrelo: n this case, it should stay open until you leave the page in order to capture sample removals. If it closes prematurely that’s bad and should be treated as an error.

squirrelo:
The MOI.js is specific to what MOI is doing. I can make it generalizable, but then it’s no longe the same asr MOI.js in the MOI repo and the changes will need to get moved there as well. This is a fast solution that allows us to have qiita-specific websocket code and MOI specific web socket code coexist.

qiita_pet/templates/analysis_selected.html
squirrelo: added

wasade:
It really isn't specific. Is a PR for the minor changes against moi not possible? It literally does the exact same thing, is just more flexible. 

That argument is analogous to writing a custom fasta parser every time we need to parse a fasta file. It's quick, allows for special casing whatever random garbage is in the comment field, but hurts us in the long run.

squirrelo:
Ok, broke out the web socket initialization and in the process may have fixed #993. If the MOI code looks good I’ll open a pull request to add it directly into MOI.

wasade:
Great, thank you

squirrelo:
This is now fully MOI-ified so it should be ready for merge. I will add the cart info stuff to the top corner in a subsequent pull request, since this one is already growing rather large.

wasade:
:+1:

PR #1059, opened 2015-04-13
delete sample template

antgonza:


qiita_db/metadata_template/base_metadata_template.py
```diff
@@ -761,21 +761,36 @@ def delete(cls, id_):
...
+        conn_handler.add_to_queue(
+            queue,
+            "DELETE FROM qiita.sample_template_filepath WHERE study_id = %s",
+            (id_, ))
+
+        conn_handler.add_to_queue(
+            queue,
+            "DROP TABLE IF EXISTS qiita.{0}".format(table_name))
+
+        conn_handler.add_to_queue(
```
squirrelo: Drop the table twice? Also it should probably just be `drop table`, that way we get an error if the table doesn't exist. This will point out that the DB is in an inconsistent state.

qiita_pet/handlers/study_handlers/description_handlers.py
```diff
@@ -640,6 +640,33 @@ def display_template(self, study, user, msg, msg_level, full_access,
...
+        """
+        sample_template_id = int(self.get_argument('sample_template_id'))
+
+        try:
+            SampleTemplate.delete(sample_template_id)
+            msg = ("Sample template %d has been deleted from study: "
+                   "<b><i>%s</i></b>" % (sample_template_id, study.title))
+            msg_level = "success"
+        except Exception as e:
+            msg = "Couldn't remove %d raw data: %s" % (sample_template_id,
```
squirrelo: raw data?

squirrelo:
Couple comments. Also there is no test code for this delete function.

qiita_db/metadata_template/base_metadata_template.py
antgonza: Thanks!

qiita_pet/handlers/study_handlers/description_handlers.py
antgonza: thanks.

antgonza:
Comments addressed on last commit but not sure what you mean about tests, can you clarify? Note that the base-delete is tested both in there and the sample-template-delete.

squirrelo:
I meant qiita_pet test, sorry for not being clear.

antgonza:
Added test for this and other delete funcs. Ready for another pass.

qiita_pet/test/test_study_handlers.py
```diff
@@ -382,5 +382,36 @@ class TestEBISubmitHandler(TestHandlerBase):
     pass
 
 
+class TestDelete(TestHandlerBase):
+    def test_delete_sample_template(self):
+        response = self.post('/study/description/1',
+                             {'sample_template_id': 1,
+                              'action': 'delete_sample_template'})
+        self.assertEqual(response.code, 200)
```
squirrelo: Can you also make sure that the sample template is actually deleted here? Same goes for all other delete tests. Just want to make sure the functionality is actually happening along with the page loading fine.

squirrelo:
One comment.

qiita_pet/test/test_study_handlers.py
antgonza: Not sure this is the right place for that as it already in qiita_db but I
can add if you feel really strong about it. Note that this will change the
current code and add some complexity, for example: you can not erase a
study template that has prep template so you need to cascade the deletion
...

squirrelo: Oh no, this isn't to make sure the delete function works, just that the delete page actually calls delete and the items are actually deleted when you hit the buttons on the page.

antgonza: Perhaps I'm missing something, what's the difference (an example will be great)?

squirrelo: Just because the delete function works in qiita_db, doesn't mean the call works in qiita_pet. The delete function may change, breaking qiita_pet, or some other thing may change. If you test in qiita pet that the deletion actually happens, you know the functionality always works. As an example, in test_delete_raw_data, you can make sure that the QiitaDBIDError is raised when instantiating RawData(1) to make sure it was deleted by the page call.

squirrelo: Concrete example [here](https://github.com/biocore/qiita/blob/master/qiita_pet/test/test_analysis_handlers.py#L86) Where the database change called for in the page is explicitly tested that it happens.

antgonza:
I have added the extra tests but while doing this I realized that it will be better to move the delete method from base_metadata_template.py to sample_template.py as the error raised by the method is specific to sample template. Note that the prep template has its own method. Ready for another review.

josenavas:
That is the _easy_ solution for the delete. I will refactor it as a part of my work to get all changes from #1021 in the system...

antgonza:
Cool, can you also review/merge this PR?

qiita_pet/test/test_study_handlers.py
```diff
@@ -382,5 +382,57 @@ class TestEBISubmitHandler(TestHandlerBase):
     pass
 
 
+class TestDelete(TestHandlerBase):
```
josenavas: I think this tests should include `database = True`; since they can potentially modify the database, in case of malfunction; making a lot of tests to fail and harder to find the original problem.

josenavas:
IMOO, it is very misleading that the button for removing the sample template is next to "Study information", since we are not removing the study information. Also it is the only tab on that group that has the `x` button to remove. I would suggest adding a `glyphicon glyphicon-trash` button in red next to the "Add samples to study". For other devs, this is how it currently looks like:

screen shot 2015-04-15 at 11 21 55 am

josenavas:
:+1: @squirrelo is your comment about the tests solved? If so, can you merge once the tests pass?

PR #1062, opened 2015-04-15
Creating generate_files function

josenavas:
There was a lot of code duplication across the sample/prep template classes to generate the files; and in case of the sample template, there was a bug, since the extend function was modifying the contents of the template but it was not re-generating the qiime mapping file. There is another bug on the extend function, but I'll fix it in a different PR.

That being said, I just created a new function to encapsulate all the files creation and I removed the duplicated code and substitute it for a call to the new function.

### DIFF
--- a/qiita_db/metadata_template/base_metadata_template.py
+++ b/qiita_db/metadata_template/base_metadata_template.py
@@ -1026,6 +1026,17 @@ class MetadataTemplate(QiitaObject):
 
         return result
 
+    def generate_files(self):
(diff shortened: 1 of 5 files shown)


josenavas:
The CI is failing due to the stochastic failure in which the "longitude" columns magically disappears. I've restarted the build.

josenavas:
Yay green! Can I have this reviewed? Thanks!

antgonza:
:+1:

Issue #1045, opened 2015-04-07
Should `delete` be an instance method instead of a classmethod?


josenavas:
I think it makes more sense to have it as an instance method rather than a class method.

@antgonza concern is:
```
pd.delete()
pd.whatever = 55
```

However, this is not solved with the current solution:
```
PreprocessedData.delete(ppd.id)
ppd.whatever = 55
```

IMOO, I think it will be safer being an instance method because in such case we can invalidate the `.id` attribute by setting it to `-1` and then do some decorator magic to check that value before any function is executed and raise a useful error.

What do others think?

ElDeveloper:
From the code examples it is not immediately clear, what the problem
is. Would you mind clarifying?

josenavas:
I think the most immediate problem is that we can have an object that is referencing an object in the DB that does not exist:
```python
o = DB_Object(id)
# Do some stuff with that instance here
DB_Object.delete(o.id)
# Do some other stuff with the instance here e.g.:
var = o.attribute
```
And the error that will be raised is dependent of what we are doing with that instance.

If we move it from a classmethod to an instance method, we can always raise a consistent error if we put some decorators on the functions/properties:
```python
o = DB_Object(id)
# Do some stuff with that instance here
o.delete()
# Do some other stuff with the instance here e.g.:
var = o.attribute
IncompetentQiitaDeveloperError: the object has been deleted!
```

The other issue that I see was pointed out by @wasade a long time ago. We can call delete with a random identifier:
```python
pd = ProcessedData(1)
RawData.delete(pd.id)
Study.delete(np.random.randint(low=0, high=10))
```
but if it is an instance method, you are exactly aware of which object you're deleting:
```python
pd = ProcessedData(1)
pd.delete()
rd = RawData(1)
rd.delete()
```

And finally, I think that philosophically, the delete function makes more sense to be in the instance rather than in the class. I think this was an oversight in the original design, and it is not too late to change.

ElDeveloper:
Hmmm, got it thanks a lot for the explanation. Would it make sense then
to use the __del__ method? That way you can no longer access that
object and we avoid having to decorated things, etc. A problem though is
that since doing `del pd` only decrements the reference count of `pd` by
1, we might encounter situations where things are not actually deleted
i.e. a case where another object still keeps a reference to the
instantiated object (which shouldn't be the case), but ... maybe.

josenavas:
[...]

PR #1068, opened 2015-04-16
Delete study

antgonza:


ElDeveloper:
This was an installation problem, not a problem with the PR.

qiita_db/study.py
```diff
@@ -347,6 +348,56 @@ def create(cls, owner, title, efo, info, investigation=None):
...
+
+        Parameters
+        ----------
+        id_ : integer
+            The object identifier
+
+        Raises
+        ------
+        QiitaDBError
+            If the sample_(_id) table exists == a sample template exists
```
ElDeveloper: It took me a while to understand this comment, would you mind rephrasing it:

`_id` -> `id_`
`==` -> `means`

qiita_db/study.py
```diff
@@ -347,6 +348,56 @@ def create(cls, owner, title, efo, info, investigation=None):
...
+        """
+        cls._check_subclass()
+
+        conn_handler = SQLConnectionHandler()
+        if exists_table('sample_%d' % id_, conn_handler):
+            raise QiitaDBError('Study "%s" can not be erased because it has a '
```
ElDeveloper: `can not` -> `cannot`.

qiita_pet/templates/study_description.html
```diff
@@ -548,6 +568,7 @@
         <a class="btn btn-default glyphicon glyphicon-edit" href="/study/edit/{{study.id}}" title="Edit the study information" style="margin:5px; word-spacing: -10px;"> Edit</a>
       {% end %}
       <a href="/study/upload/{{study.id}}" class="btn btn-default glyphicon glyphicon-upload" title="Upload study files" style="margin:5px; word-spacing: -10px;"> Upload</a>
+      <a class="btn btn-danger glyphicon glyphicon-trash" data-toggle="modal" data-target="#delete-study"></a>
```
ElDeveloper: Since there are two "trash" icons (when viewing the sample template), I think this button should have a "Delete Study" label associated with it.

qiita_pet/templates/study_description.html
```diff
@@ -580,4 +601,21 @@
   {% end %}
 </div>
 
+
+<div class="modal fade delete-study" tabindex="-1" role="dialog" id="delete-study">
+  <div class="modal-dialog modal-sm">
+    <div class="modal-content">
+      <div class="modal-header">
+        <h3>Deleting:<br/></h3><h4>{{study_title}}</h4>
+        You will only be able to delete a study that has no data associated with it.
```
ElDeveloper: I think this could be moved to the "modal-body" section.

qiita_pet/templates/study_description.html
```diff
@@ -580,4 +601,21 @@
...
+      </div>
+      <div class="modal-body">
```
ElDeveloper: And if you follow the comment above, this should be moved to the "modal-footer" section.

qiita_pet/templates/study_description.html
```diff
@@ -580,4 +601,21 @@
...
+        To continue you need to write the alias name of the study:
+          <input type="text" name="study-alias" id="study-alias">
+          <a class="btn btn-danger glyphicon glyphicon-trash" onclick="delete_study()"></a>
```
ElDeveloper: A comment that @josenavas made is that we are not being careful about what the HTML elements can and can't do, and this is one of those cases. If there's no text in the text box, then the button should not be enabled.

ElDeveloper:
I think this looks good and all the comments are rather minor, can you add either a close or cancel button to the modal view?

antgonza:
None of the modals have a close/cancel button ...

ElDeveloper:
That's not true, the "create analysis" modal has a `x` to close the modal.

antgonza:
I can add and we should add to all of them, including those to create
meta-analysis.

ElDeveloper:
Thanks, aside from those comments, the functionality seems to work just fine! Thanks!

qiita_pet/templates/study_description.html
antgonza: I tested your 2 suggestions and I do not like them.

qiita_db/study.py
```diff
@@ -347,6 +348,56 @@ def create(cls, owner, title, efo, info, investigation=None):
...
+        id_ : integer
+            The object identifier
+
+        Raises
+        ------
+        QiitaDBError
+            If the sample_(id_) table exists means a sample template exists
+        """
+        cls._check_subclass()
+
```
josenavas: Need to check if the study with id 'id_' exists.

Off topic: This is another reason why I think `delete` should be an instance method.

qiita_db/study.py
```diff
@@ -347,6 +348,56 @@ def create(cls, owner, title, efo, info, investigation=None):
...
+        conn_handler.add_to_queue(
+            queue,
+            "DELETE FROM qiita.study_environmental_package WHERE study_id = "
+            "%s", (id_, ))
+
+        conn_handler.add_to_queue(
+            queue,
+            "DELETE FROM qiita.study WHERE study_id = %s", (id_, ))
+
+        conn_handler.execute_queue(queue)
```
josenavas: I'm surprised that this is actually working. It should fail since you are not removing the row from the `study_users` and `investigation_study` table. Does this mean that we are missing a FK on these tables?

qiita_db/study.py
```diff
@@ -347,6 +348,56 @@ def create(cls, owner, title, efo, info, investigation=None):
...
+
+        Raises
+        ------
+        QiitaDBError
+            If the sample_(id_) table exists means a sample template exists
+        """
+        cls._check_subclass()
+
+        conn_handler = SQLConnectionHandler()
+        if exists_table('sample_%d' % id_, conn_handler):
```
josenavas: A question about the future. Is a user going to be able to add raw data, preprocessed data or processed data (otu tables) w/o a sample template? If so, we should check if the study has something else...

qiita_pet/handlers/study_handlers/description_handlers.py
```diff
@@ -640,6 +640,30 @@ def display_template(self, study, user, msg, msg_level, full_access,
...
+            The current user object
+        callback : function
+            The callback function to call with the results once the processing
+            is done and it fails
+        """
+        study_id = study.id
+
+        try:
+            Study.delete(study_id)
+            self.redirect('/study/list/')
```
josenavas: Can we redirect and show a message saying that the study has been successfully removed? There is no feedback if it actually was successfully removed or not.

josenavas:
Few comments. Some of them could have raised a higher problem on the DB and we should check those issues before moving forward...

qiita_db/study.py
antgonza: The way it works right now is that you need a raw data to have a preprocessed data, a preprocessed data to have a processed data, which I think is fine and give us the possibility of "play" with the options. For example, when we allow to upload a process data we can link it to the same default raw data for faster searches.

qiita_db/study.py
antgonza: See https://github.com/biocore/qiita/issues/1078. For the time being, I added a instantiation of `Study(id_)` to validate this case and a test just to be sure.

qiita_db/study.py
antgonza: Added https://github.com/biocore/qiita/issues/1079

qiita_pet/handlers/study_handlers/description_handlers.py
antgonza: Done.

qiita_db/study.py
josenavas: Thanks for adding the issue. However you still should remove those rows here, otherwise the DB is going to be in an inconsistent state.

qiita_db/study.py
```diff
@@ -347,6 +348,59 @@ def create(cls, owner, title, efo, info, investigation=None):
...
+
+        Raises
+        ------
+        QiitaDBError
+            If the sample_(id_) table exists means a sample template exists
+        """
+        cls._check_subclass()
+
+        # checking that the id_ exists
+        cls(id_)
```
josenavas: Can you use `cls._check_id(id_)` instead? Thanks.

josenavas:
2 more comments @antgonza 
I've also checked this on my machine and I like the interface. Thanks!

qiita_db/study.py
antgonza: Just to confirm, what you are asking is for me to solve
https://github.com/biocore/qiita/issues/1078 within this PR and use
that function?

josenavas: No, but you are only instantiating the object to check if the `id` exists. I think that is unnecessary given that we have the function `_check_id`. i.e. `_check_id` is called in `__init__` (among other things) so you avoid all these other things by using the function that you really want.

josenavas: @antgonza and I discussed this offline. The `_check_id` method is not a `classmethod` so it cannot be called here.

As a note, I think this brings up the discussion I started in #1045, so we can avoid all these checks...

PR #1072, opened 2015-04-17
Fix extend functionality

josenavas:
Some background:

The extend functionality was first introduced to add new samples to a sample template. However, if there was a new column, that new column was also added to the database. The problem is that at the beginning of the function, the existing samples were filtered from the metadata template so the existing samples didn't get the values for the new column.

This PR includes:
- the bug described above is fixed in a way that now extend can be used to add new columns, new samples or both.
 - moving part of the extend functionality to the base class, in preparation for adding it to the prep template (it will be needed in the refactor).
- the tests of the extend function were quite limited (they just tested that the sample id was there). I created a bunch of new tests, so the functionality is better tested in all the possible paths the function can be executed.

### DIFF
--- a/qiita_db/metadata_template/base_metadata_template.py
+++ b/qiita_db/metadata_template/base_metadata_template.py
@@ -44,18 +44,18 @@ from copy import deepcopy
 
 import pandas as pd
 from skbio.util import find_duplicates
+import warnings
 
 from qiita_core.exceptions import IncompetentQiitaDeveloperError
 
 from qiita_db.exceptions import (QiitaDBUnknownIDError, QiitaDBColumnError,
                                  QiitaDBNotImplementedError,
-                                 QiitaDBExecutionError,
+                                 QiitaDBExecutionError, QiitaDBWarning,
                                  QiitaDBDuplicateHeaderError)
 from qiita_db.base import QiitaObject
 from qiita_db.sql_connection import SQLConnectionHandler
-from qiita_db.util import (exists_table, get_table_cols,
-                           convert_to_id,
-                           get_mountpoint, insert_filepaths)
+from qiita_db.util import (exists_table, get_table_cols, convert_to_id,
+                           get_mountpoint, insert_filepaths, scrub_data)
 from qiita_db.logger import LogEntry
 from .util import (as_python_types, get_datatypes, get_invalid_sample_names,
                    prefix_sample_names_with_id)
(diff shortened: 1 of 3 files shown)


josenavas:
yay!!!!!!!

ElDeveloper:
I'm going to start reviewing this.

qiita_db/metadata_template/base_metadata_template.py
```diff
@@ -772,6 +773,123 @@ def _add_common_creation_steps_to_queue(cls, md_template, obj_id,
...
+            The queue where the SQL statements will be added
+
+        Raises
+        ------
+        QiitaDBError
+            If no new samples or new columns are present in `md_template`
+        """
+        # Check if we are adding new samples
+        sample_ids = md_template.index.tolist()
+        sql = """SELECT sample_id FROM qiita.{0}
```
ElDeveloper: I'm kinda surprised we don't have an `ids` method for the templates.

qiita_db/metadata_template/base_metadata_template.py
```diff
@@ -772,6 +773,123 @@ def _add_common_creation_steps_to_queue(cls, md_template, obj_id,
...
+                 WHERE {1}=%s""".format(self._table, self._id_column)
+        curr_samples = set(
```
ElDeveloper: I was curious and decided to try this out, turns out there's a faster way:

```python

In [4]: import numpy as np

In [9]: x = np.random.randint(0, 100, size=1000)

In [12]: %timeit h = {a for a in x}
10000 loops, best of 3: 102 µs per loop

In [13]: %timeit h = set(a for a in x)
10000 loops, best of 3: 128 µs per loop
```

Can you change this and all the other similar instances in this PR?

qiita_db/metadata_template/base_metadata_template.py
```diff
@@ -772,6 +773,123 @@ def _add_common_creation_steps_to_queue(cls, md_template, obj_id,
...
+            s[0] for s in conn_handler.execute_fetchall(sql, (self._id,)))
+        existing_samples = curr_samples.intersection(sample_ids)
+        new_samples = set(sample_ids).difference(existing_samples)
+
+        # Check if we are adding new columns
+        table_name = self._table_name(self._id)
+        # Get the required columns from the DB
+        db_cols = sorted(get_table_cols(self._table, conn_handler))
```
ElDeveloper: Why do these need to be sorted?

qiita_db/metadata_template/base_metadata_template.py
```diff
@@ -772,6 +773,123 @@ def _add_common_creation_steps_to_queue(cls, md_template, obj_id,
...
+        QiitaDBError
+            If no new samples or new columns are present in `md_template`
+        """
+        # Check if we are adding new samples
+        sample_ids = md_template.index.tolist()
+        sql = """SELECT sample_id FROM qiita.{0}
+                 WHERE {1}=%s""".format(self._table, self._id_column)
+        curr_samples = set(
+            s[0] for s in conn_handler.execute_fetchall(sql, (self._id,)))
+        existing_samples = curr_samples.intersection(sample_ids)
```
ElDeveloper: @josenavas, please correct me if I'm wrong but I believe these two lines can be simplified with the symmetric_difference operation.

qiita_db/metadata_template/base_metadata_template.py
```diff
@@ -772,6 +773,123 @@ def _add_common_creation_steps_to_queue(cls, md_template, obj_id,
...
+            for category, dtype in zip(new_cols, datatypes):
+                conn_handler.add_to_queue(
+                    queue_name, sql_cols, (self._id, category, dtype))
+                conn_handler.add_to_queue(
+                    queue_name, sql_alter.format(table_name, category, dtype))
+
+            if existing_samples:
+                warnings.warn(
+                    "The new columns '%s' have been added to the existing "
+                    "samples '%s'. Any other value of these samples have been "
```
ElDeveloper: I think the second sentence of this message is much more important than the first one, and users can miss it if they are updating a large number of columns. Also, I don't think the second sentence is very clear.

ElDeveloper:
Looks good, just a couple questions.

ElDeveloper:
And comments.

qiita_db/metadata_template/base_metadata_template.py
josenavas: that's what an oversight on my end... we have `self.keys()`, changed.

qiita_db/metadata_template/base_metadata_template.py
josenavas: I replaced this line with the call to `keys`. However this is good to know for future PR!

qiita_db/metadata_template/base_metadata_template.py
josenavas: You're correct, however I need both variables (`existing_samples` and `new_samples`). With symmetric_difference I only get `new_samples`.

qiita_db/metadata_template/base_metadata_template.py
josenavas: Good catch, probably some leftover code...

qiita_db/metadata_template/base_metadata_template.py
josenavas: Changed!

qiita_db/metadata_template/base_metadata_template.py
```diff
@@ -772,6 +773,119 @@ def _add_common_creation_steps_to_queue(cls, md_template, obj_id,
...
+        QiitaDBError
+            If no new samples or new columns are present in `md_template`
+        """
+        # Check if we are adding new samples
+        sample_ids = md_template.index.tolist()
+        curr_samples = set(self.keys())
+        existing_samples = curr_samples.intersection(sample_ids)
+        new_samples = set(sample_ids).difference(existing_samples)
+
+        # Check if we are adding new columns
```
squirrelo: This is an excessive number of comments, to the point it makes the code hard to read.

qiita_db/metadata_template/base_metadata_template.py
```diff
@@ -772,6 +773,119 @@ def _add_common_creation_steps_to_queue(cls, md_template, obj_id,
...
+        table_name = self._table_name(self._id)
+        # Get the required columns from the DB
```
squirrelo: This is an excessive number of comments, to the point it makes the code hard to read.

squirrelo:
One comment, non-blocking.

qiita_db/metadata_template/base_metadata_template.py
```diff
@@ -772,6 +773,119 @@ def _add_common_creation_steps_to_queue(cls, md_template, obj_id,
...
+                    "No values have been modified for samples '%s'. However, "
+                    "the following columns have been added to them: '%s'"
+                    % (", ".join(existing_samples), ", ".join(new_cols)),
+                    QiitaDBWarning)
+                # The values for the new columns are the only ones that get
+                # added to the database. None of the existing values will be
+                # modified (see update for that functionality)
+                min_md_template = md_template[new_cols].loc[existing_samples]
+                values = as_python_types(min_md_template, new_cols)
+                values.append(existing_samples)
```
squirrelo: Can you add an explanation for this and the following line? It's a bit confusing.

squirrelo:
:+1:

PR #1073, opened 2015-04-17
Relax db

josenavas:
Note: this PR is made against the new `relax-md-req`.

I'm only changing the SQL layout and the populate_test_db.sql, so people can check if there is something terribly wrong.

I'm not changing any code, so tests are expected to fail.
In the next PR, (which will be build on top of this one) I'll start fixing the metadata_template objects, so some tests start passing with the new layout :-)

PR #1074, opened 2015-04-17
Fix sample obj

josenavas:
This PR is build on top of ~~#1072~~ (merged) and #1073. At least #1072 should be merged in master before merging this one, so the commits from there (quite a few) disappear here.

This PR fixes the BaseSample, Sample and PrepSample objects and their tests. The tests for these objects are expected to pass.


squirrelo:
#1073 was just merged, can you pull upstream on this?

josenavas:
@squirrelo Done!

squirrelo:
:+1:

adamrp:
Looks good. @squirrelo @josenavas is there a way to test this, or are we merging everything into the `relax-md-req` first, and saving the testing for later?

josenavas:
@adamrp all the functionality is not in place after a couple more PR. I think the best path is to review the code to see if there is anything horribly wrong, and then wait for the PR that everything is in place to make sure that everything is correct. Is really hard to test until everything is in place, but if the code looks good, I can fix the remaining issues later. Does everybody agrees with that?

ElDeveloper:
@josenavas, yes that's a good strategy moving forward.

adamrp:
great! Thanks.

PR #1075, opened 2015-04-17
Fix metadata obj

josenavas:
This PR is built on top of #1074 

This PR makes all the needed modifications in the metadata template objects to get all the tests under `qiita_db/metadata_template/tests` to pass. The PR right now is a bit big, but as the other PR in which this one depends gets merged, the size will be reduced.


qiita_db/metadata_template/constants.py
```diff
@@ -6,10 +6,68 @@
...
+from future.utils import viewkeys, viewvalues
+
+Restriction = namedtuple('Restriction', ['columns', 'error_msg'])
+
+# A dict containing the restrictions that apply to the sample templates
+SAMPLE_TEMPLATE_COLUMNS = {
+    'EBI': Restriction(columns={'collection_timestamp': 'timestamp',
+                                'physical_specimen_location': 'varchar'},
+                       error_msg="EBI submission disabled"),
+    'Qiita_main': Restriction(columns={'sample_type': 'varchar',
```
antgonza: I'm not sure I like the casing. This is not an actual issue but looks weird and perhaps all should be lower case ...

qiita_db/metadata_template/constants.py
```diff
@@ -6,10 +6,68 @@
...
 
 TARGET_GENE_DATA_TYPES = ['16S', '18S', 'ITS']
-REQUIRED_TARGET_GENE_COLS = {'barcodesequence', 'linkerprimersequence',
-                             'run_prefix', 'library_construction_protocol',
-                             'experiment_design_description', 'platform'}
-RENAME_COLS_DICT = {'barcode': 'barcodesequence',
-                    'primer': 'linkerprimersequence'}
+
+PREP_TEMPLATE_COLUMNS_TARGET_GENE = {
+    'Demultiplex': Restriction(
```
antgonza: Same here: casing.

ElDeveloper:
Can you pull from upstream?

josenavas:
Done!

ElDeveloper:
@josenavas, from looking at the tests in Travis, it seems like there's a test case that's raising a warning that's not being catched, this might or not be intentional, but would be good to update the tests so there's no warnings being raised, see this for example:

```
/home/travis/build/biocore/qiita/qiita_db/metadata_template/base_metadata_template.py:583: QiitaDBWarning: Some functionality will be disabled due to missing columns:
	EBI submission disabled: primer.
Check https://github.com/biocore/qiita/wiki/Preparing-Qiita-template-files for a description of these fields.
  QiitaDBWarning)
```

josenavas:
Ok, I'll update the tests so no warning is raised. There are quite a few instances of those... will ASAP.

qiita_db/metadata_template/base_metadata_template.py
```diff
@@ -559,22 +533,13 @@ def _clean_validate_template(cls, md_template, study_id, obj,
...
-            Any extra object needed by the template to perform any extra check
+        restriction_dict : dict of {str: Restriction}
+            A dictionary with the restrictions that apply to the metadata
 
         Returns
         -------
         md_template : DataFrame
             Cleaned copy of the input md_template
-
-        Raises
```
ElDeveloper: Can you add the warning being raised when there are missing columns?

antgonza:
After these changes go in, do we need to update ftp://ftp.microbio.me/pub/qiita/sample_prep_template_examples.tgz ?

qiita_db/metadata_template/base_metadata_template.py
```diff
@@ -641,44 +601,27 @@ def _add_common_creation_steps_to_queue(cls, md_template, obj_id,
...
         sample_ids = md_template.index.tolist()
-        num_samples = len(sample_ids)
-        headers = list(md_template.keys())
-
-        # Get the required columns from the DB
-        db_cols = sorted(get_table_cols(cls._table, conn_handler))
-        # Remove the sample_id and _id_column columns
-        db_cols.remove('sample_id')
-        db_cols.remove(cls._id_column)
+        headers = sorted(md_template.keys().tolist())
```
ElDeveloper: Do we no longer have to check the column names are unique? Or is this guaranteed by `.keys`?

qiita_db/metadata_template/base_metadata_template.py
```diff
@@ -1060,38 +984,16 @@ def to_file(self, fp, samples=None):
...
-            # First write the headers
-            f.write("sample_name\t%s\n" % '\t'.join(headers))
-            # Write the values for each sample id
-            for sid, d in sorted(metadata_map.items()):
-                values = [str(d[h]) for h in headers]
-                values.insert(0, sid)
-                f.write("%s\n" % '\t'.join(values))
+            df = df.loc[samples]
+
+        # Apply some sorting to the dataframe
```
ElDeveloper: Why is this sorting needed?

josenavas:
@antgonza we probably will need to update those, good call. I will take a note to update those once the entire branch is ready to be merged in master.

I just found a small bug on the code and I'm fixing it. However, it requires modifying multiple tests. Will ping when I update everything (and including the comments made so far).

qiita_db/metadata_template/constants.py
josenavas: lowercased!

qiita_db/metadata_template/constants.py
josenavas: done!

qiita_db/metadata_template/base_metadata_template.py
josenavas: Done, I was unsure if the warnings were documented in the same way :-)

qiita_db/metadata_template/constants.py
ElDeveloper: What I don't like about this name is that "qiita main" isn't really meaningful.

Additionally the meaning of each of these keys should be documented somewhere.

qiita_db/metadata_template/base_metadata_template.py
josenavas: At this point is it assumed that the md_template has been validated and clean (I think it's a safe assumption since this is a private function). Those checks are done in `_clean_validate_template`. That sounds reasonable?

qiita_db/metadata_template/constants.py
```diff
@@ -6,10 +6,68 @@
...
+                                       'description': 'varchar',
+                                       'physical_specimen_remaining': 'bool',
+                                       'dna_extracted': 'bool',
+                                       'latitude': 'float8',
+                                       'longitude': 'float8',
+                                       'host_subject_id': 'varchar'},
+                              error_msg="Processed data approval disabled")
+}
+
+# A dict containing the restrictions that apply to the sample templates
```
ElDeveloper: sample -> prep

qiita_db/metadata_template/prep_template.py
```diff
@@ -456,19 +426,31 @@ def create_qiime_mapping_file(self, prep_template_fp):
...
         pt_sample_names = set(pt.index)
 
         if not pt_sample_names.issubset(st_sample_names):
             raise ValueError(
-                "Prep template is not a sub set of the sample template, files:"
-                "%s %s - samples: %s" % (sample_template_fp, prep_template_fp,
-                                         str(pt_sample_names-st_sample_names)))
+                "Prep template is not a sub set of the sample template, files"
+                "%s - samples: %s" % (sample_template_fp,
+                                      str(pt_sample_names-st_sample_names)))
```
ElDeveloper: Can you join this set with ', '?

qiita_db/metadata_template/base_metadata_template.py
josenavas: The sorting is not needed for functionality. However, since I'm testing that the actual file is correctly written, I need a way to make sure that the file is always written in the same way. If we remove these sorting, we can potentially have another stochastic failure on travis...

qiita_db/metadata_template/prep_template.py
```diff
@@ -456,19 +426,31 @@ def create_qiime_mapping_file(self, prep_template_fp):
...
 
         mapping = pt.join(st, lsuffix="_prep")
         mapping.rename(columns=rename_cols, inplace=True)
 
+        # We cannot ensure that the QIIME-required columns are present in the
```
ElDeveloper: This comment should be documented in the Notes of the docstring.

qiita_db/metadata_template/prep_template.py
```diff
@@ -456,19 +426,31 @@ def create_qiime_mapping_file(self, prep_template_fp):
...
+        # metadata map. However, we have to generate a QIIME-compliant mapping
+        # file. Since the user may need a QIIME mapping file, but not these
+        # QIIME-required columns, we are going to create them here and
+        # populate them with the value XXQIITAXX
+        index = mapping.index
+        placeholder = ['XXQIITAXX'] * len(index)
+        for val in viewvalues(rename_cols):
```
ElDeveloper: Do you think it would make sense to raise a warning in this case? The warning would make a note of the columns that would have place-holders.

qiita_db/metadata_template/constants.py
josenavas: What do you recommend for this? The restrictions under the `qiita_main` key are restrictions that only apply to our official Qiita main site, but they will not apply on a single user environment...

Re documentation: do you think it's enough to add a comment here?

qiita_db/metadata_template/constants.py
josenavas: Done

qiita_db/metadata_template/base_metadata_template.py
ElDeveloper: Cool, thanks!

qiita_db/metadata_template/prep_template.py
josenavas: Done

qiita_db/metadata_template/base_metadata_template.py
ElDeveloper: Would it make more sense to have a test that doesn't care about order instead? Or have the test sort as needed? By no means is this a blocking comment, but if the sorting is only needed for tests, then it's pointless to have it here.

qiita_db/metadata_template/prep_template.py
josenavas: Good call, moved to notes and reduced the comment here.

qiita_db/metadata_template/prep_template.py
josenavas: Good call, adding.

qiita_db/metadata_template/base_metadata_template.py
josenavas: I guess I could parse the file using pandas and check at the parsed pandas dataframe level. Does this sound good to everybody? Note that the original intent of the test is to check the contents of the file, and I'm unsure if we are loosing this check in this way...

qiita_db/metadata_template/constants.py
ElDeveloper: Hmmm, then the restriction only applies to non-single-user systems right? As for better documenting this, I think we need something more than just a comment in the "constants" file. However I cannot think of a good way to do this.

qiita_db/metadata_template/base_metadata_template.py
ElDeveloper: Oh, ok ok ok, this is more of a "for consistency" through serializations. This is useful outside of the scope of testing, so +1 to keep as is, and just change the comment "apply some sorting ...", to describe the reason why we sort the dataframe.

qiita_db/metadata_template/constants.py
josenavas: Re documentation: I will add a comment here and I've created a new issue: https://github.com/biocore/qiita/issues/1111

I think I used the wrong words on my previous comment, the idea is that those restrictions only apply to our system. Other labs that may want to have a qiita installation locally with multiple users shouldn't have these restrictions either... Is one of those cases in which our system is more specialized than the rest of the users.

Thinking a bit more about this. I think that having this in a configuration file will allow us to change these restrictions dynamically, and we can have them documented there. What do you think?

josenavas:
@antgonza @ElDeveloper this should be ready for another review, thanks guys!

qiita_db/metadata_template/base_metadata_template.py
```diff
@@ -599,29 +572,23 @@ def _clean_validate_template(cls, md_template, study_id, obj,
...
+            if missing:
+                warning_msg.append(
+                    "%s: %s" % (restriction.error_msg, ', '.join(missing)))
 
-        # Retrieve the headers of the metadata template
-        headers = list(md_template.keys())
+        if warning_msg:
+            warnings.warn(
+                "Some functionality will be disabled due to missing "
+                "columns:\n\t%s.\nCheck https://github.com/biocore/qiita/wiki"
```
antgonza: Should this be a link so the user can access it?

qiita_db/metadata_template/constants.py
ElDeveloper: Got it, yeah it makes more sense that way. Otherwise it is one of those
seemingly arbitrary restrictions of the system. If these are only for us
then they should be parametrized, although we do want to provide Qiita
as a service, the software should not be specific to our needs in any
way.

qiita_db/metadata_template/base_metadata_template.py
josenavas: How should I make it a link? Note that this error is raised in qiita_db. I'm hesitant to add specific html here since this warning can be seen from the CLI...

ElDeveloper: I was also thinking about this, but it's going to be tricky since this
can also be raised from a terminal and you would get an HTML formatted
error message, which is not appropriate. Ideas?

qiita_db/metadata_template/constants.py
josenavas: Yeah, I agree the idea of making this parametrized. Also, we don't know
when EBI decides to change their requirements. However, in interests of
time, I think we should create an issue and I can solve it as a part of
Alpha 0.2, since that will require some refactor. Does that sound good to
you?

2015-04-27 15:24 GMT-07:00 Yoshiki Vázquez Baeza <notifications@github.com>:

> In qiita_db/metadata_template/constants.py
> <https://github.com/biocore/qiita/pull/1075#discussion_r29197966>:
>
> > @@ -6,10 +6,68 @@
> >  # The full license is in the file LICENSE, distributed with this software.
> >  # -----------------------------------------------------------------------------
> >
> > +from collections import namedtuple
> > +from future.utils import viewkeys, viewvalues
> > +
> > +Restriction = namedtuple('Restriction', ['columns', 'error_msg'])
> > +
> > +# A dict containing the restrictions that apply to the sample templates
> > +SAMPLE_TEMPLATE_COLUMNS = {
> > +    'EBI': Restriction(columns={'collection_timestamp': 'timestamp',
> > +                                'physical_specimen_location': 'varchar'},
> > +                       error_msg="EBI submission disabled"),
> > +    'Qiita_main': Restriction(columns={'sample_type': 'varchar',
>
> Got it, yeah it makes more sense that way. Otherwise it is one of those
> seemingly arbitrary restrictions of the system. If these are only for us
> then they should be parametrized, although we do want to provide Qiita as a
> service, the software should not be specific to our needs in any way.
> … <#14cfcfaf015149db_>
> On (Apr-27-15|15:12), josenavas wrote: > @@ -6,10 +6,68 @@ > # The full
> license is in the file LICENSE, distributed with this software. > #
> -----------------------------------------------------------------------------
> > > +from collections import namedtuple > +from future.utils import
> viewkeys, viewvalues > + > +Restriction = namedtuple('Restriction',
> ['columns', 'error_msg']) > + > +# A dict containing the restrictions that
> apply to the sample templates > +SAMPLE_TEMPLATE_COLUMNS = { > + 'EBI':
> Restriction(columns={'collection_timestamp': 'timestamp', > +
> 'physical_specimen_location': 'varchar'}, > + error_msg="EBI submission
> disabled"), > + 'Qiita_main': Restriction(columns={'sample_type':
> 'varchar', Re documentation: I will add a comment here and I've created a
> new issue: https://github.com/biocore/qiita/issues/1111 I think I used
> the wrong words on my previous comment, the idea is that those restrictions
> only apply to our system. Other labs that may want to have a qiita
> installation locally with multiple users shouldn't have these restrictions
> either... Is one of those cases in which our system is more specialized
> than the rest of the users. Thinking a bit more about this. I think that
> having this in a configuration file will allow us to change these
> restrictions dynamically, and we can have them documented there. What do
> you think? --- Reply to this email directly or view it on GitHub:
> https://github.com/biocore/qiita/pull/1075/files#r29197135
>
> —
> <https://github.com/biocore/qiita/pull/1075/files#r29197966>.
>



-- 
Jose Navas

qiita_db/metadata_template/prep_template.py
```diff
@@ -432,6 +404,16 @@ def create_qiime_mapping_file(self, prep_template_fp):
...
+
+        Notes
+        -----
+        We cannot ensure that the QIIME-required columns are present in the
+        metadata map. However, we have to generate a QIIME-compliant mapping
+        file. Since the user may need a QIIME mapping file, but not these
+        QIIME-required columns, we are going to create them and
+        populate them with the value XXQIITAXX.
         """
         rename_cols = {
```
antgonza: Do we still need to do this? I'm not sure but it seems unnecessary at this point.

qiita_db/metadata_template/base_metadata_template.py
antgonza: What's the issue with CLI? I think is fine ...

qiita_db/metadata_template/constants.py
ElDeveloper: Yeah, this sounds good we should resolve this before too long though.
Can you open an issue about this?

qiita_db/metadata_template/prep_template.py
```diff
@@ -456,19 +438,38 @@ def create_qiime_mapping_file(self, prep_template_fp):
...
+                "%s - samples: %s"
+                % (sample_template_fp,
+                   ', '.join(pt_sample_names-st_sample_names)))
 
         mapping = pt.join(st, lsuffix="_prep")
         mapping.rename(columns=rename_cols, inplace=True)
 
+        # Pre-populate the QIIME-required columns with the value XXQIITAXX
+        index = mapping.index
+        placeholder = ['XXQIITAXX'] * len(index)
```
ElDeveloper: You probably thought about this, but why not just setting an empty string?

ElDeveloper:
All my comments have been addressed, I only have a question, but that's not a blocking comment. @antgonza can you merge once your comments/questions are addressed?

qiita_db/metadata_template/base_metadata_template.py
josenavas: @ElDeveloper you're the CLI guru, what do you think? I prefer not to add HTML code in the CLI error/warning output, but if more people agrees I'm happy to add it...

qiita_db/metadata_template/constants.py
josenavas: Done https://github.com/biocore/qiita/issues/1113

2015-04-27 15:32 GMT-07:00 Yoshiki Vázquez Baeza <notifications@github.com>:

> In qiita_db/metadata_template/constants.py
> <https://github.com/biocore/qiita/pull/1075#discussion_r29198597>:
>
> > @@ -6,10 +6,68 @@
> >  # The full license is in the file LICENSE, distributed with this software.
> >  # -----------------------------------------------------------------------------
> >
> > +from collections import namedtuple
> > +from future.utils import viewkeys, viewvalues
> > +
> > +Restriction = namedtuple('Restriction', ['columns', 'error_msg'])
> > +
> > +# A dict containing the restrictions that apply to the sample templates
> > +SAMPLE_TEMPLATE_COLUMNS = {
> > +    'EBI': Restriction(columns={'collection_timestamp': 'timestamp',
> > +                                'physical_specimen_location': 'varchar'},
> > +                       error_msg="EBI submission disabled"),
> > +    'Qiita_main': Restriction(columns={'sample_type': 'varchar',
>
> Yeah, this sounds good we should resolve this before too long though. Can
> you open an issue about this?
> … <#14cfd0256077b3f3_>
> On (Apr-27-15|15:26), josenavas wrote: > @@ -6,10 +6,68 @@ > # The full
> license is in the file LICENSE, distributed with this software. > #
> -----------------------------------------------------------------------------
> > > +from collections import namedtuple > +from future.utils import
> viewkeys, viewvalues > + > +Restriction = namedtuple('Restriction',
> ['columns', 'error_msg']) > + > +# A dict containing the restrictions that
> apply to the sample templates > +SAMPLE_TEMPLATE_COLUMNS = { > + 'EBI':
> Restriction(columns={'collection_timestamp': 'timestamp', > +
> 'physical_specimen_location': 'varchar'}, > + error_msg="EBI submission
> disabled"), > + 'Qiita_main': Restriction(columns={'sample_type':
> 'varchar', Yeah, I agree the idea of making this parametrized. Also, we
> don't know when EBI decides to change their requirements. However, in
> interests of time, I think we should create an issue and I can solve it as
> a part of Alpha 0.2, since that will require some refactor. Does that sound
> good to you? 2015-04-27 15:24 GMT-07:00 Yoshiki Vázquez Baeza <
> notifications@github.com>: > In qiita_db/metadata_template/constants.py >
> <https://github.com/biocore/qiita/pull/1075#discussion_r29197966>: > > >
> @@ -6,10 +6,68 @@ > > # The full license is in the file LICENSE,
> distributed with this software. > > #
> -----------------------------------------------------------------------------
> > > > > +from collections import namedtuple > > +from future.utils import
> viewkeys, viewvalues > > + > > +Restriction = namedtuple('Restriction',
> ['columns', 'error_msg']) > > + > > +# A dict containing the restrictions
> that apply to the sample templates > > +SAMPLE_TEMPLATE_COLUMNS = { > > +
> 'EBI': Restriction(columns={'collection_timestamp': 'timestamp', > > +
> 'physical_specimen_location': 'varchar'}, > > + error_msg="EBI submission
> disabled"), > > + 'Qiita_main': Restriction(columns={'sample_type':
> 'varchar', > > Got it, yeah it makes more sense that way. Otherwise it is
> one of those > seemingly arbitrary restrictions of the system. If these are
> only for us > then they should be parametrized, although we do want to
> provide Qiita as a > service, the software should not be specific to our
> needs in any way. > … <#14cfcfaf015149db_> > On (Apr-27-15|15:12),
> josenavas wrote: > @@ -6,10 +6,68 @@ > # The full > license is in the file
> LICENSE, distributed with this software. > # >
> -----------------------------------------------------------------------------
> > > > +from collections import namedtuple > +from future.utils import >
> viewkeys, viewvalues > + > +Restriction = namedtuple('Restriction', >
> ['columns', 'error_msg']) > + > +# A dict containing the restrictions that
> > apply to the sample templates > +SAMPLE_TEMPLATE_COLUMNS = { > + 'EBI': >
> Restriction(columns={'collection_timestamp': 'timestamp', > + >
> 'physical_specimen_location': 'varchar'}, > + error_msg="EBI submission >
> disabled"), > + 'Qiita_main': Restriction(columns={'sample_type': >
> 'varchar', Re documentation: I will add a comment here and I've created a >
> new issue: https://github.com/biocore/qiita/issues/1111 I think I used >
> the wrong words on my previous comment, the idea is that those restrictions
> > only apply to our system. Other labs that may want to have a qiita >
> installation locally with multiple users shouldn't have these restrictions
> > either... Is one of those cases in which our system is more specialized >
> than the rest of the users. Thinking a bit more about this. I think that >
> having this in a configuration file will allow us to change these >
> restrictions dynamically, and we can have them documented there. What do >
> you think? --- Reply to this email directly or view it on GitHub: >
> https://github.com/biocore/qiita/pull/1075/files#r29197135 > > — > Reply
> to this email directly or view it on GitHub > <
> https://github.com/biocore/qiita/pull/1075/files#r29197966>. > -- Jose
> Navas --- Reply to this email directly or view it on GitHub:
> https://github.com/biocore/qiita/pull/1075/files#r29198175
>
> —
> <https://github.com/biocore/qiita/pull/1075/files#r29198597>.
>



-- 
Jose Navas

qiita_db/metadata_template/prep_template.py
josenavas: Empty strings are easy to miss. On previous discussions, @rob-knight suggested that any value that we populate in the templates should be make clear that it is wrong and needs to be fixed.

qiita_db/metadata_template/base_metadata_template.py
ElDeveloper: First of all, thanks for the compliment, but I am no such thing :P That
being said, this is code that's specifically part of the Model, not of
the View. The View should be concerned with the most convenient
rendering of URLs and in that spirit, we should probably have JS code to
recognize and do the appropriate formatting of URLs. Does that make
sense to everyone?

qiita_db/metadata_template/prep_template.py
ElDeveloper: Now I remember, thanks!

qiita_db/metadata_template/base_metadata_template.py
antgonza: I think this should be a separate conversation/issue. BTW IMOO the parsing should be in python/templates and not in JS.

ElDeveloper: That works!

josenavas: So that means that this should stay as it is and we'll improve the qiita_pet error parsing?

antgonza: yup, can you open an issue?

josenavas: Sure! Done: https://github.com/biocore/qiita/issues/1115

2015-04-27 17:04 GMT-07:00 Antonio Gonzalez <notifications@github.com>:

> In qiita_db/metadata_template/base_metadata_template.py
> <https://github.com/biocore/qiita/pull/1075#discussion_r29204247>:
>
> >
> > -        # Retrieve the headers of the metadata template
> > -        headers = list(md_template.keys())
> > +        if warning_msg:
> > +            warnings.warn(
> > +                "Some functionality will be disabled due to missing "
> > +                "columns:\n\t%s.\nCheck https://github.com/biocore/qiita/wiki"
>
> yup, can you open an issue?
>
> —
> <https://github.com/biocore/qiita/pull/1075/files#r29204247>.
>



-- 
Jose Navas

josenavas:
@antgonza can you merge if there are no outstanding issues?

qiita_db/metadata_template/prep_template.py
antgonza: ping! I do not think we need to patch this way for this specific fields, right?

josenavas: Oooops! Good call, we still need it, but not as extensive. Instead of:
```python
rename_cols = {
    'barcode': 'BarcodeSequence',
    'barcodesequence': 'BarcodeSequence',
    'primer': 'LinkerPrimerSequence',
    'linkerprimersequence': 'LinkerPrimerSequence',
    'description': 'Description',
}
```
We need to use:
```python
rename_cols = {
    'barcode': 'BarcodeSequence',
    'primer': 'LinkerPrimerSequence',
    'description': 'Description',
}
```
The reason is that we are using different names on the database than on QIIME. Does this seams reasonable? The other solution is to require the user to introduce those values directly. This will require changing the restrictions. What do you think?

antgonza: The reduce version makes more sense.

josenavas: Thanks, while changing this I noticed that the prep template already created in the database is not fully correct. I'm changing it and it will probably require a couple of more changes in the rest of the tests.

josenavas:
This is ready to go if there are no more comments!

ElDeveloper:
LGTM.

PR #1096, opened 2015-04-23
Cart corner info

squirrelo:
This adds the cart information dropdown and link to the top right corner of the page next to logout. It uses AJAX to keep the counts constantly correct.

### DIFF
--- a/qiita_pet/templates/sitebase.html
+++ b/qiita_pet/templates/sitebase.html
@@ -157,12 +157,12 @@
               <a href="/auth/logout/">Log Out</a>
             </li>
             <li class="dropdown">
-                <a href="/analysis/selected/" id="dflt-sel-info" data-toggle="dropdown" class="dropdown-toggle"><span class="glyphicon glyphicon-list-alt"></span> Selected</a>
-                <ul class="dropdown-menu" style="padding: 5px">
-                  <li> <span id="dflt-study-count"></span> Studies</li>
-                  <li><span id="dflt-pd-count"></span> Processed Data</li>
-                  <li><span id="dflt-samp-count"></span> Samples</li>
-                </ul>
+              <a href="/analysis/selected/" id="dflt-sel-info" data-toggle="dropdown" class="dropdown-toggle"><span class="glyphicon glyphicon-list-alt"></span> Selected</a>
+              <ul class="dropdown-menu" style="padding: 5px">
+                <li> <span id="dflt-study-count"></span> Studies</li>
+                <li><span id="dflt-pd-count"></span> Processed Data</li>
+                <li><span id="dflt-samp-count"></span> Samples</li>
+              </ul>
             </li>
           </ul>
           <!-- otherwise show the login form -->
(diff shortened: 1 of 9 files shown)


squirrelo:
This is ready to go

qiita_pet/templates/sitebase.html
```diff
@@ -138,13 +149,21 @@
...
+          <ul class="nav navbar-nav navbar-right">
             <li>
               <a href="/profile/">Welcome {{user.id}}</a>
             </li>
-            <li class="pull-right">
+            <li>
               <a href="/auth/logout/">Log Out</a>
             </li>
+            <li class="dropdown">
+                <a href="/analysis/selected/" id="dflt-sel-info" data-toggle="dropdown" class="dropdown-toggle"><span class="glyphicon glyphicon-list-alt"></span> Selected</a>
```
ElDeveloper: This block is over-indented.

ElDeveloper:
Looks good @squirrelo! I think all it is missing is a "remove" button for the studies, but aside from that, it looks good.

squirrelo:
I'm hesitant to add a remove button for studies straight up, since we can remove processed datas and when they're all gone the study is removed.

ElDeveloper:
Ok, so you can only remove a study from the /selected/ page?

squirrelo:
Yup

ElDeveloper:
As long as everyone is ok with this, I'm ok as well.

:+1:, I restarted your build (stochastic failure related to the lat/long
test).

antgonza:
Nice, some comments:
- "Add" button is not clear enough about what it does. Perhaps "Add to
meta-analysis" but might be too long. Other option could be glyphicon
glyphicon-plus or glyphicon glyphicon-share-alt ?
- Could you move the "Add" button to be near the expand button?
(either side is fine)
- Could you change the color of the "Selected" drop down menu so we
know when there is something and when there is not?
- What about using glyphicon glyphicon-shopping-cart for "Selected"?
- Add should not be present for studies that have no process data
- Fine with removing particular samples/studies in the /selected/ page
but we should have a "Clear" button.

ElDeveloper:
This is a good idea, but instead of meta-analysis, let's just start
using the word analysis.

squirrelo:
I'll take a look at the stuff I don't mention here, but quickly:
* I am not cool with using a shopping cart icon. It implies that someone is going to pay for whatever is in the cart, and that's not the case.
* The add button seems clear enough to me, since this is already a crowded page so adding that extra text is going to further crowd things.

squirrelo:
Can we merge this as is and do another pull request to get the tiny detail things?

antgonza:
Merged but we will not merge to master until those details are fixed ...

ElDeveloper:
Sounds like a good idea.

josenavas:
I think this is the best approach for PR that goes to their own branch :-)

2015-04-23 14:17 GMT-07:00 Yoshiki Vázquez Baeza <notifications@github.com>:

> Sounds like a good idea.
>
> On (Apr-23-15|14:15), Antonio Gonzalez wrote:
> > Merged but we will not merge to master until those details are fixed ...
> >
> >
> > ---
> > Reply to this email directly or view it on GitHub:
> > https://github.com/biocore/qiita/pull/1096#issuecomment-95720621
>
> —
> <https://github.com/biocore/qiita/pull/1096#issuecomment-95721444>.
>



-- 
Jose Navas

PR #1098, opened 2015-04-23
Update patch 22

squirrelo:
This adds deletion of duplicate selections in the analysis_sample table before adding the PK


squirrelo:
Ready for review

antgonza:
I thought we agreed we will just erase those duplicated rows. If so, I do not understand why we need to create a copy of the full table, could you comment?

squirrelo:
Since there is no current key or way to differentiate rows, you need to create that copy with the counter as an added row. Then you delete all and reinsert the rows with just one copy (the rn = 1).

antgonza:
:+1:

adamrp:
looks good to me, but I'll wait until tests pass :+1:

PR #1099, opened 2015-04-23
Fix _almost_ all qiita db tests

josenavas:
Built on top of #1075 

Fixes most of the qiita_db tests. The only tests that are not fixed are the analysis.py tests. The reason is that fixing those tests requires a bit more work and I do not want to blow-up the size of this PR.


squirrelo:
I thought we agreed that you were going to fix each test file in separate pull requests. This is way too large to review.

josenavas:
This is including all the changes in the previous PR. The actual changes here are minimal...

squirrelo:
That was not made clear anywhere in the pull request. Please remember to add that information for future pull requests.

josenavas:
Will do, sorry about that. I'll update the description to reflect that.

josenavas:
Just pulled from upstream!

qiita_db/search.py
```diff
@@ -167,8 +167,8 @@ def __repr__(self):
 class QiitaStudySearch(object):
     """QiitaStudySearch object to parse and run searches on studies."""
 
-    # column names from required_sample_info table
-    required_cols = set(get_table_cols("required_sample_info"))
+    # column names from study_sample table
+    required_cols = set(get_table_cols("study_sample"))
```
squirrelo: This check isn't required any more, since all we have in the table now is study id and sample id, correct?

qiita_db/search.py
```diff
@@ -310,7 +310,7 @@ def _parse_study_search_string(self, searchstr,
                     meta_header_type_lookup[header] = 'varchar'
 
         # create the study finding SQL
-        # remove metadata headers that are in required_sample_info table
+        # remove metadata headers that are in study_sample table
         meta_headers = tuple(meta_headers.difference(
```
squirrelo: I think you can just directly drop the sample column now if needed, otherwise this is not required

qiita_db/search.py
```diff
@@ -348,7 +348,7 @@ def _parse_study_search_string(self, searchstr,
             else:
                 header_info.append("sa.%s" % meta)
         # build the SQL query
-        sample_sql = ("SELECT r.sample_id,%s FROM qiita.required_sample_info "
+        sample_sql = ("SELECT r.sample_id,%s FROM qiita.study_sample "
                       "r JOIN qiita.sample_{0} sa ON sa.sample_id = "
```
squirrelo: I know this isn't your code, but could you update this to use the `USING` convention? Thanks.

qiita_db/test/test_search.py
```diff
@@ -28,7 +28,7 @@ def test_parse_study_search_string(self):
                       "lower(column_name) = lower('altitude') and column_type "
                       "in ('integer', 'float8')")
         exp_samp_sql = ("SELECT r.sample_id,sa.altitude FROM "
-                        "qiita.required_sample_info r JOIN qiita.sample_{0} sa"
+                        "qiita.study_sample r JOIN qiita.sample_{0} sa"
                         " ON sa.sample_id = r.sample_id JOIN qiita.study st ON"
```
squirrelo: Same for all of these.

squirrelo:
Couple of comments.

qiita_db/test/test_search.py
```diff
@@ -84,12 +84,14 @@ def test_parse_study_search_string(self):
         st_sql, samp_sql, meta = \
             self.search._parse_study_search_string(
                 'host_subject_id includes "Chicken little"')
-        exp_st_sql = "SELECT study_id FROM qiita.study_sample_columns"
-        exp_samp_sql = ("SELECT r.sample_id,r.host_subject_id FROM "
-                        "qiita.required_sample_info r JOIN qiita.sample_{0} sa"
+        exp_st_sql = ("SELECT study_id FROM qiita.study_sample_columns "
+                      "WHERE lower(column_name) = lower('host_subject_id') "
```
ElDeveloper: There's no need to have `lower('host_subject_id')` is there?

ElDeveloper: i.e. there's no need to call lower on that string.

qiita_db/util.py
```diff
@@ -1008,9 +952,19 @@ def get_processed_params_tables():
 
 def get_lat_longs():
     conn = SQLConnectionHandler()
-    sql = """select latitude, longitude
-             from qiita.required_sample_info"""
-    return conn.execute_fetchall(sql)
+    sql = """SELECT DISTINCT table_name
```
ElDeveloper: Minor and non-blocking, but would you mind adding a docstring in this function?

qiita_db/test/test_search.py
```diff
@@ -28,7 +28,7 @@ def test_parse_study_search_string(self):
                       "lower(column_name) = lower('altitude') and column_type "
                       "in ('integer', 'float8')")
         exp_samp_sql = ("SELECT r.sample_id,sa.altitude FROM "
-                        "qiita.required_sample_info r JOIN qiita.sample_{0} sa"
+                        "qiita.study_sample r JOIN qiita.sample_{0} sa"
```
ElDeveloper: Very minor but the letter r in this query no longer makes sense, same goes for all the queries you updated in this file, perhaps worth updating it?

qiita_db/test/test_search.py
squirrelo: That is actually built programmatically, and lowers everything it gets so that search is case-insensitive.

ElDeveloper:
@josenavas, looks in really good shape, I only had two rather minor comments :+1: on my end granted that @squirrelo's comments are addressed.

qiita_db/search.py
josenavas: You're correct. Removing! Your feedback on this code is really useful, as I had some problems understanding it...

qiita_db/search.py
josenavas: By sample column you mean the sample id columns right?

squirrelo: yes

josenavas: Done!

qiita_db/search.py
josenavas: Done!

qiita_db/test/test_search.py
josenavas: Done!

qiita_db/util.py

josenavas:
Thanks guys, ready for another round!

ElDeveloper:
:+1:

ElDeveloper:
@squirrelo, would you mind merging if all looks good to you?

squirrelo:
I'g getting "Internal Server Error, please try again later" When using search now. Can you look into this and fix it?

squirrelo:
Nevermind, local system issue. Merging.

PR #1101, opened 2015-04-24
Pretty-fying sample cart and selection alerts

squirrelo:
This adds the color change and the floating div.

### DIFF
--- a/qiita_pet/templates/list_studies.html
+++ b/qiita_pet/templates/list_studies.html
@@ -44,7 +44,7 @@ function sel_proc_data(row, pid) {
   moi.send('sel', proc_data);
 }
 
-function show_alert(data) { bootstrapAlert(data + ' samples selected.', "success", 2000); }
+function show_alert(data) { bootstrapAlert(data + ' samples selected.', "success", 2000); $('#dflt-sel-info').css('color', 'rgb(0, 160, 0)');}
 
 function fillAbstract(table, row) {
     $('#title-text-area').text($('#' + table).find('#study' + row + "-title").text());
--- a/qiita_pet/templates/list_studies.html
+++ b/qiita_pet/templates/list_studies.html
@@ -2,6 +2,12 @@
 {% block head %}
 <link rel="stylesheet" href="/static/vendor/css/jquery.dataTables.css" type="text/css">
 <style>
+.alert {
+  position: fixed;
+  top: 50px;
+  left: 40px;
+  z-index:100;
+}
 td.details-control {
   cursor: pointer;
 }
--- a/qiita_pet/templates/list_studies.html
+++ b/qiita_pet/templates/list_studies.html
@@ -44,7 +44,10 @@ function sel_proc_data(row, pid) {
   moi.send('sel', proc_data);
 }
 
-function show_alert(data) { bootstrapAlert(data + ' samples selected.', "success", 2000); $('#dflt-sel-info').css('color', 'rgb(0, 160, 0)');}
+function show_alert(data) {
+  bootstrapAlert(data + ' samples selected.', "success", 2000);
+   $('#dflt-sel-info').css('color', 'rgb(0, 160, 0)');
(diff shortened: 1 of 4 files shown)


squirrelo:
This is ready to go

qiita_pet/templates/list_studies.html
```diff
@@ -38,7 +44,7 @@
   moi.send('sel', proc_data);
 }
 
-function show_alert(data) { bootstrapAlert(data + ' samples selected.', "success", 2000); }
+function show_alert(data) { bootstrapAlert(data + ' samples selected.', "success", 2000); $('#dflt-sel-info').css('color', 'rgb(0, 160, 0)');}
```
ElDeveloper: Seems rather strange that you have to have these two lines in the same line.

ElDeveloper:
:+1: this looks great, though a minor thing: I think it would look better on the right side of the screen, to hint at the fact that it refers to the "selected" menu.

squirrelo:
Issue with right side of screen is that the div is position:absolute, so I would need to dynamically adjust the x position with that. As is, I use x=0 so it's much easier and screen-size-independent.

qiita_pet/templates/list_studies.html
squirrelo: I don't, but javascript doesn't care about line breaks and it saves some transmit info.

wasade: A byte? There are likely other larger areas of connection overhead that
would be more effective to worry about if this is a problem. We can do .min
versions later which would resolve extraneous whitespace, likely better to
worry about readability right now

squirrelo: Eh, ok, I was lazy. Fixed.

antgonza:
Some suggestions:
- If a study/processed-data has already been selected either change
the "Add" button or when adding change the message so it reflects that
the study/processed-data was already selected and nothing changed
- You need to add something that lets the user know that a study was
already selected. In the past you had a checkmark. Perhaps combining
this suggestion with the previous one will be a good idea
- Can you add a clear button? Remove everything from the analysis.
- In the expanded version there is a lot of space between "Processed
Date" and "Algorithm", not sure why.

squirrelo:
In my opinion, your first two suggestions are going to confuse users. They are actually selecting the samples every time, it's just whether they are already registered or not on the back end that matters. Users will most likely not care about previously selected counts and all that, just that the ones they are currently selecting are registered.

squirrelo:
Took care of last 2 suggestions. Good to go again.

antgonza:
I see your point but if a user stops adding studies, goes away, comes
back, they will not know immediately from the GUI what has been
selected. Thus, adding a way to know quickly what they have already
selected.

squirrelo:
So tehy look at the selected studies page?

antgonza:
Not convenient, neither quick.

squirrelo:
Neither is what you are proposing. It's inconvenient and confusing.

squirrelo:
This is something we need to leave to user testing in my opinion. I think it's good as is and, if in user testing, it comes out as needed I will add it. This is the first alpha after all.

josenavas:
@squirrelo just a general comment in terms of the UI. Note that the first users of the interface are the reviewers. If the reviewers already find something that is confusing to them, is a read flag that the interface is going to be confusing to the users. Also you're basing your arguments on your brain image of the interface, while @antgonza is already seeing in your current interface that something does now work. Probably, his proposal is not the best one, but is already saying that there is something wrong with the interface. Check the new Contributing.md to see why this is important.

@antgonza given that it looks like there is no agreement on the final UI (although I agree with you that there is a problem), it would be useful if you can mock up something to show an improvement. In the past, I downloaded the final HTML from chrome and I did quick modifications there so I can see how the interface is going to look like (but is not functional). I think this will help to @squirrelo to understand that his UI is missing something.

PR #1106, opened 2015-04-25
Fix analysis tests

josenavas:
Built on top of #1099 

The test database was incomplete. We were missing the prep template file and the qiime mapping file for the prep template that was already in the database. I've added those and I fixed the tests for the analysis object.

After this PR, all tests in the qiita_db folder should be passing. However, there will be still some errors on qiita_ware, which I will fix in another PR.


josenavas:
just pushed from upstream

ElDeveloper:
Thanks @josenavas, I'm gonna start reviewing right now. @squirrelo, quick reminder that you should also review this one.

qiita_db/analysis.py
```diff
@@ -719,81 +720,61 @@ def _build_mapping_file(self, samples, conn_handler=None):
...
-        # prep headers, making sure they follow mapping file format rules
-        all_headers = list(all_headers - {'linkerprimersequence',
-                           'barcodesequence', 'description', 'sample_id'})
-        all_headers.sort()
-        all_headers = ['BarcodeSequence', 'LinkerPrimerSequence'] + all_headers
-        all_headers.append('Description')
-
-        # write mapping file out
+            if len(samples) != len(set(samples)):
+                duplicates = [s for s in samples if samples.count(s) > 1]
```
ElDeveloper: You can use find_duplicates here: http://scikit-bio.org/docs/latest/generated/skbio.util.find_duplicates.html?highlight=duplicates#skbio.util.find_duplicates

qiita_db/analysis.py
```diff
@@ -719,81 +720,61 @@ def _build_mapping_file(self, samples, conn_handler=None):
...
+                raise QiitaDBError("Duplicate sample ids found: %s"
+                                   % ', '.join(duplicates))
+            # Get the QIIME mapping file
+            qiime_map_fp = conn_handler.execute_fetchall(sql, (pid,))[0][1]
+            # Parse the mapping file
+            qiime_map = pd.read_csv(
```
ElDeveloper: Loading the mapping file with Pandas may not be the best idea here, namely the datatype of a few columns may be inferred and this in turn will cause some problems down the road. The simplest example is a column where the values are 1001, 0001, 0002 and 0003; pandas will cast these values to ints and turn them into 1001, 1, 2 and 3. If there's any way where we can load only the string values and then do all the needed transformations, that would be ideal.

qiita_db/analysis.py
```diff
@@ -719,81 +720,61 @@ def _build_mapping_file(self, samples, conn_handler=None):
...
+        merged_map = pd.concat(to_concat)
+
+        cols = merged_map.columns.values.tolist()
+        cols.remove('BarcodeSequence')
+        cols.remove('LinkerPrimerSequence')
+        cols.remove('Description')
+        new_cols = ['BarcodeSequence', 'LinkerPrimerSequence']
+        new_cols.extend(cols)
+        new_cols.append('Description')
+        merged_map = merged_map[new_cols]
```
ElDeveloper: Ah, this is super awesome! :sparkles:

ElDeveloper:
@josenavas, this looks in really good shape.

qiita_db/analysis.py
squirrelo: I'm personally still wary about using the files vs using the database values, but that's a separate conversation.

squirrelo:
Good to go after @ElDeveloper comments addressed.

qiita_db/analysis.py
josenavas: Done!

qiita_db/analysis.py
josenavas: @ElDeveloper I think I'm going to pull out a wrapper function to pandas so all the parsing is always done consistently.

@squirrelo With the changes that I've done in the metadata template, the files and the DB are always sync'd. If we are not going to trust our files, why to store them?

ElDeveloper: @josenavas, as long as we can guarantee that all values are loaded as strings, then that should work just fine. Searching on the intertubes I came across this: http://stackoverflow.com/questions/12101113/prevent-pandas-from-automatically-infering-type-in-read-csv

josenavas: I've been thinking more about this (because we already have the template parsing centralizes and it is smarter). Thanks for the link but I think it is an overkill. I think the easier way is to do use the converters parameters like this: `converters=defaultdict)lambda: str)` so it is always parsing all columns as str, which is what we want here :smile:

josenavas:
Ready for another round!

qiita_db/analysis.py
ElDeveloper: Got it that makes sense, if that's the case then we don't need to infer the datetime or parse dates, would you mind setting those to False or just removing them altogether.

josenavas: Done!

ElDeveloper:
Thanks @josenavas

PR #1107, opened 2015-04-25
Fix qiita all tests

josenavas:
Build on top of #1106 

Fixes all the tests in qiita_ware. Everything expected to pass now.


ElDeveloper:
@antgonza, quick reminder, it would be great if we can get this reviewed ASAP.  Tests should be running now and this time they should all pass.

antgonza:
:+1: if tests pass.

ElDeveloper:
Same here :+1: if tests pass.

ElDeveloper:
@josenavas, there's a few legitimate failures, would you mind having a look?

antgonza:
Also, will it be possible to remove all QiitaDBWarning?

josenavas:
Thanks guys, I know the source of the problem. What I'm going to do is merge #1109 in here, as the code in there will help me to resolve this issue. Given that the 2 PR are small, I don't think it will blow up the size of the PR :-)

ElDeveloper:
Sounds like a good idea, would you mind closing 1109 then?

josenavas:
Tests should pass now. I'm currently working on removing all the warnings.

josenavas:
This should be good to go. This is the last PR os the functionality is in place. Note that the interface is still not limited, i.e. it still allows to submit to EBI although columns are missing. AFAIK, we decided that this will go in in alpha 0.2.

qiita_db/metadata_template/prep_template.py
```diff
@@ -524,3 +524,25 @@ def status(self):
         pd_statuses = conn_handler.execute_fetchall(sql, (self._id,))
 
         return infer_status(pd_statuses)
+
+    @property
+    def qiime_map_fp(self):
+        """The QIIME mapping file path of the prep template
```
antgonza: This is not really clear, could you rephrase?

qiita_ware/processing_pipeline.py
```diff
@@ -22,51 +22,56 @@
 
 def _get_qiime_minimal_mapping(prep_template, out_dir):
     """Generates a minimal QIIME-compliant mapping file for split libraries
-
     The columns of the generated file are, in order: SampleID, BarcodeSequence,
```
antgonza: Should this also add the reverselinkerprimer?

antgonza: Obviously, if it exists.

qiita_ware/processing_pipeline.py
```diff
@@ -22,51 +22,56 @@
 
 def _get_qiime_minimal_mapping(prep_template, out_dir):
     """Generates a minimal QIIME-compliant mapping file for split libraries
-
```
ElDeveloper: Why did you remove this line? numpy docs allow them, same for the next two deleted lines.

qiita_ware/processing_pipeline.py
```diff
@@ -22,51 +22,56 @@
...
-    # the run_prefix column
-    output_fps = []
     path_builder = partial(join, out_dir)
-    for prefix, df in pt.groupby('run_prefix'):
-        df = df[cols]
-        out_fp = path_builder("%s_MMF.txt" % prefix)
-        output_fps.append(out_fp)
+    if 'run_prefix' in qiime_map:
+        # The study potentially has more than 1 lane, so we should generate a
+        # qiita MMF for each of the lanes. We know how to split the prep
```
ElDeveloper: qiita -> qiime ... right?

qiita_ware/processing_pipeline.py
```diff
@@ -22,51 +22,56 @@
...
+        # template based on the run_prefix column
+        output_fps = []
+        for prefix, df in qiime_map.groupby('run_prefix'):
+            df = df[cols]
+            # Sorting for consistency between serialization
```
ElDeveloper: This comment seems to be outdated.

qiita_ware/processing_pipeline.py
```diff
@@ -22,51 +22,56 @@
...
-    # If the study has more than 1 lane, we should generate a qiita MMF for
-    # each of the lanes. We know how to split the prep template based on
-    # the run_prefix column
-    output_fps = []
     path_builder = partial(join, out_dir)
-    for prefix, df in pt.groupby('run_prefix'):
-        df = df[cols]
-        out_fp = path_builder("%s_MMF.txt" % prefix)
-        output_fps.append(out_fp)
+    if 'run_prefix' in qiime_map:
```
ElDeveloper: I don't recall from the stack of pull requests that have been coming in, but there are tests for both of these two branches right? i.e. cases with run_prefix and cases without it.

qiita_db/metadata_template/prep_template.py
josenavas: Done

qiita_ware/processing_pipeline.py
josenavas: Is the reverselinkerprimer a column used in QIIME split_libraries? I don't think it is required...

qiita_ware/processing_pipeline.py
josenavas: No idea, I think it has been a copy-paste issue from my other PR.... adding them back.

qiita_ware/processing_pipeline.py
josenavas: right

qiita_ware/processing_pipeline.py
josenavas: indeed

qiita_ware/processing_pipeline.py
antgonza: Yes it is, is not required but we should use it if it exists in the mapping file.

qiita_ware/processing_pipeline.py
josenavas: Good catch, there was not. Adding right now.

ElDeveloper: Awesome, thanks!

josenavas:
Done, I needed to modify `create_qiime_mapping_file` to include the reverse linker primer as a controlled column too. Should be ready for another review.

That would mean that we will need to modify the restrictions at some point to include that column to disable some processing. We should discuss this tomorrow, but I don't think it should block the release since we are not using it in the current pipeline.

ElDeveloper:
It seems like there's some legitimate failures.

ElDeveloper:
@biocore/qiita-admin is everyone ok with this:

> Done, I needed to modify create_qiime_mapping_file to include the reverse linker primer as a controlled column too. Should be ready for another review.

> That would mean that we will need to modify the restrictions at some point to include that column to disable some processing. We should discuss this tomorrow, but I don't think it should block the release since we are not using it in the current pipeline.

antgonza:
I think this is a real bug as we should be doing this for all sff files that have the reverselinkerprimer ... not sure how difficult/easy is to fix

josenavas:
@ElDeveloper that was a complete WTF error hahaha it looks I removed some characters in the code while I was trying to type in another window.

@antgonza we currently do not support this type of processing. Actually, the fact of adding the ReverseLinkerPrimer has generated new QiitaDBWarnings that are going to require me be a bit smarter in the code, as that column is not always required by QIIME.

In my opinion, you're currently biased given the discussions at today's meeting. However, I think that processing should be discussed tomorrow at the meeting, as I don't think that we have time to make it to the release... Additionally, the new issues that you recently added are related to this and they've been added to the beta milestone. I think we can solve it for the alpha 0.2, but I don't think we are going to be able to add it, let's say, in the following 2 hours...

antgonza:
That fine, could you create an issue about this and assign it to the next release ?

josenavas:
Done: https://github.com/biocore/qiita/issues/1121

antgonza:
Thanks! However, there are still errors.

josenavas:
I just pushed the fix for them!!! You're 1 minute faster...

ElDeveloper:
Thanks @josenavas

PR #1109, opened 2015-04-27
Add qiime map func

josenavas:
Built on top of #1107 

Adds the qiime_map_fp property to the prep template.


josenavas:
Closing in favor of #1107 (it already includes the commits in here)

PR #1110, opened 2015-04-27
Update INSTALL.md for pgbouncer

squirrelo:
Add pgbouncer install instructions.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -91,7 +91,7 @@ Once libevent is installed, you can install pgbouncer. Source code is available 
 ```
 ./configure --prefix=/usr/local --with-libevent=/usr/local/libevent
 ```
-Then make and make install as usual. Configuration and usage instructions are at [pgbouncer's homepage](https://pgbouncer.github.io/).
+Then `make` and `make install` as usual. Note: the `make install` command may fail in OSX. In this case, move the pgbouncer executable from the base source folder to your bin file manually. Configuration and usage instructions are at [pgbouncer's homepage](https://pgbouncer.github.io/).
 
 ## Troubleshooting installation on non-Ubuntu operating systems
 
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -82,6 +82,17 @@ It might be necessary to restart postgresql: `sudo service postgresql restart`.
 
 Furthermore, the `pg_hba.conf` file can be modified to change authentication type for local users to trust (rather than, e.g., md5) but we haven't tested this solution.
 
+Installing optional qiita dependencies
+--------------------------------------
+
+For large scale systems, pgbouncer is preferable to a straight postgres connection as it removes postgresql connection overhead. To install pgbouncer, you will need libevent. Installation instructions for OSX are available [here](http://mac-dev-env.patrickbougie.com/libevent/).
+
+Once libevent is installed, you can install pgbouncer. Source code is available [here](https://pgbouncer.github.io/downloads/). If libevent was installed using the instructions above, run config using 
+```
+./configure --prefix=/usr/local --with-libevent=/usr/local/libevent
+```
+Then make and make install as usual. Configuration and usage instructions are at [pgbouncer's homepage](https://pgbouncer.github.io/).
+
 ## Troubleshooting installation on non-Ubuntu operating systems
 
 ### xcode


squirrelo:
Stupid longitude error in tests. Good to go.

antgonza:
Could you add an example of how to configure it on a local machine? I do not think every user should need to read a huge config page and explanation so they can start using pgbouncer with their local install. Similar to the [nginx](https://github.com/biocore/qiita/blob/master/qiita_pet/nginx_example.conf) example.

squirrelo:
I am against having to keep documentation up to date for a project we do not control or work on. There is an example minimal configuration at the bottom of https://pgbouncer.github.io/config.html  I can point to that if we absolutely need an explicit example.

wasade:
We also don't provide install information for Centos, how to configure your
firewall, etc. This really is site specific and there is no aspect of qiita
that depends on it.

adamrp:
I also wanted to make a note of [this question in the pgbouncer FAQ](https://pgbouncer.github.io/faq.html#should-pgbouncer-be-installed-on-webserver-or-database-server%3F), particularly this part:

> Installing on webserver is good when short-connections are used, then the connection setup latency is minimised - TCP requires couple of packet roundtrips before connection is usable. Installing on database server is good when there are many different hosts (eg. webservers) connecting to it, then their connections can be optimised together

For our purposes, it seems like we want to install pgbouncer on the webserver (and not the database server).

antgonza:
Note that IMOO this is a completely different issue from @wasade examples. I think this is different because this fixes (at least we think) the speed problem in loading studies. Note that this will be an issue to every single user that wants to develop using the EMP as an initial dataset, which take ~9 secs to load.

squirrelo:
Yeah, I talked to Jeff briefly before the lab meeting about setting it up
on the webserver. I or someone else should follow up.

wasade:
We do have quite a few connections coming from ipengines and AG, but don't
have practical experience to know where the balance point is

squirrelo:
pgbouncer runs on a per-database level as well it looks like, so we can have it specifically for qiita on the qiita webserver.

wasade:
@antgonza, if the inclusion criteria is based on performance, then should we include tips on optimizing postgres, hardware choices, filesystem tuning, kernel tuning, etc? 

The document describes this as "for large-scale systems" which implies that there is someone who has sysadmin experience on the deployment end. While this is a piece of documentation, we need to minimize the chance of it being out of sync with a project that we have no involvement in, otherwise it will become an incorrect piece of documentation. 

I think @squirrelo's suggestion of linking out is acceptable here.

antgonza:
I think making users and admins life easier is a good idea. If you have suggestions to make qiita run better I'm all about creating a section for this. 

BTW I completely disagree that this install is only for large-scale systems. If you remember the reason we catch this before deploy is because I have the EMP studies in my laptop, which IMOO should be the regular deploy environment.

wasade:
It only makes their lives easier if the documentation is ensured to be
correct.

As the PR is presently written, it is describing pgbouncer for use in
large-scale systems.

On Tue, Apr 28, 2015 at 10:01 AM, Antonio Gonzalez <notifications@github.com

antgonza:
Agree on correctness, and that's a commitment we have to make on all our tools.

@squirrelo could you change your PR so it reflects @wasade's point about large-scale systems? Which is incorrect.

squirrelo:
I think you are optimizing for an edge case here, since yes, it's for systems with a large amount of studies in them. If we are saying the normal install is on a laptop, that means most likely for working on your personal studies. I think it's safe to leave it as large scale for now, or change the wording to "hundreds of studies" and revisit later if necessary.

wasade:
It's also why we historically have not done things like describing site
specific optimizations...

I wasn't suggesting changing it away from large-scale, but that's just my
opinion.


On Tue, Apr 28, 2015 at 10:05 AM, Antonio Gonzalez <notifications@github.com

antgonza:
@wasade: I think qiita is a different piece of software of what we have done before and that it needs way more documentation for regular users. Thus, parting from what we have done before so we can reduce possible user and developer bad experiences is the way to go.

@squirrelo: this is an issue even with the test dataset (1-2 studies), as it take ~3 secs to load. Also remember that the plan is to allow users to perform meta-analysis, at least with the EMP.

josenavas:
My 2 cents:

I completely agree with @antgonza. I can see questions coming from users like: why qiita takes so long to load my study page? The EMP studies are only 100, and I can see people easily reaching this numbers. I've not been involved in too many analysis and I can tell that I would already have ~25 studies.

I think we should include at least the minimal instructions required to make Qiita usable. If somebody has a special machine, we can then point the actual pgbouncer documentation.

We should be realistic and understand that we are going to deal with non-computer experts users, so adding this documentation will make our lives easier in the long run.

wasade:
It took ~3 seconds for me to retrieve all the public studies on
qiita.microbio.me, I didn't do that locally or on a test dataset.





On Tue, Apr 28, 2015 at 10:13 AM, Antonio Gonzalez <notifications@github.com

adamrp:
@josenavas @antgonza I think the point that @wasade is making (which I agree with) is that we should reference pgbouncer's own documentation (which is going to be far more detailed, comprehensive, up-to-date, and accurate than we can hope to duplicate) rather than re-iterate the same instructions in our own words. We have to trust that our users can follow others' instructions as well as they would be able to follow our own.

squirrelo:
Also note, if we are going to write this for non-expert computer users (biologists, biochemists, etc) Then the source code, make, make install, all of it is way too technical for a standard user. In fact, a lot of the documentation is daunting when looked at by a non-technical computer user.

antgonza:
Wanna have a call?

wasade:
@antgonza, the profiling data yesterday wasn't for loading studies

wasade:
I can talk briefly

josenavas:
I can talk too

squirrelo:
From the call, it sounded like we no longer need this pull request?

antgonza:
yup

PR #1112, opened 2015-04-27
ENH: Meta Analysis -> Analysis

ElDeveloper:
Users have frequently pointed out that naming this tab as "Meta
Analysis" is misleading and that a better name would be simply Analysis.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,7 +6,7 @@ Qiita
 [![Build Status](https://travis-ci.org/biocore/qiita.png?branch=master)](https://travis-ci.org/biocore/qiita)
 [![Coverage Status](https://coveralls.io/repos/biocore/qiita/badge.png?branch=master)](https://coveralls.io/r/biocore/qiita)
 
-Qiita (canonically pronounced *cheetah*) is a meta-analysis environment for microbiome (and other "comparative -omics") datasets. You can find the public Qiita server at [qiita.microbio.me](http://qiita.microbio.me).
+Qiita (canonically pronounced *cheetah*) is an analysis environment for microbiome (and other "comparative -omics") datasets. You can find the public Qiita server at [qiita.microbio.me](http://qiita.microbio.me).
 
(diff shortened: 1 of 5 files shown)


josenavas:
:+1:  I did not test locally but I will be very surprised if this broke something...

antgonza:
I think you should replace all cases of meta*analys*s ...

ElDeveloper:
@antgonza, sounds good will do.

I also just noticed that we have two ways of getting to the "selected samples page" (through the selected button on the far right and through the "view selected samples" under the Analysis drop down menu). What do you guys think about:

- Remove the selected button on the far right.
- Add another option to the dropdown menu "create from selected samples", that brings up the form that lets you enter an analysis name.

Note these changes are not necessary and there are other things that are higher priority ... but I'm putting this out there just in case.

ElDeveloper:
@antgonza, I was surprised to find this but that's the only occurrence:

```bash
yoshikivazquezbaeza:qiita@master$ grep -ri 'meta analysis' qiita_pet
qiita_pet/templates/sitebase.html:              <a href="#" data-toggle="dropdown" class="dropdown-toggle">Meta Analysis<b class="caret"></b></a>
yoshikivazquezbaeza:qiita@master$ grep -ri 'metaanalysis' qiita_pet
yoshikivazquezbaeza:qiita@master$ grep -ri 'metaanalysis' qiita_db/
yoshikivazquezbaeza:qiita@master$ grep -ri 'metaanalysis' qiita_ware/
yoshikivazquezbaeza:qiita@master$ grep -ri 'meta analysis' qiita_db/
yoshikivazquezbaeza:qiita@master$ grep -ri 'meta analysis' qiita_ware/
```

antgonza:
Try:
grep -iR 'meta.analy' ./

ElDeveloper:
Thanks, that helped. Pushed.

antgonza:
:+1: @josenavas could you give it another pass?

PR #1123, opened 2015-04-29
BUG: Fix bug in moi.init

ElDeveloper:
The position of the group_id argument was moved and the rest of the code
was not updated to reflect this. This change set moves group_id so it is
the first argument.

### DIFF
--- a/qiita_pet/static/vendor/js/moi.js
+++ b/qiita_pet/static/vendor/js/moi.js
@@ -66,15 +66,15 @@ var moi = new function () {
      * websocket. On construction, this method will send a message over the
      * socket to get all known job information associated with this client.
      *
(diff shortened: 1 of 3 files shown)


ElDeveloper:
This problem would cause the analysis waiting page to be blank.

wasade:
I'd say to open a PR upstream, but I know your take on moi...

ElDeveloper:
@wasade, I went over there first, but I saw this: https://github.com/biocore/mustached-octo-ironman/pull/19/files ... I've added a note in the code :beer:

ElDeveloper:
Tests are passing but it seems like coveralls has choked, anyone wants to have a look? This is blocking for the release.

PR #1131, opened 2015-04-29
The single conn: forks with corks in the eye

wasade:
To a single connection. Should support both admin types, regular user, and support the existing method of instantiation.

### DIFF
--- a/qiita_core/util.py
+++ b/qiita_core/util.py
@@ -80,14 +80,7 @@ def qiita_test_checker():
             @reset_test_database
             def tearDown(self):
                 super(DecoratedClass, self).tearDown()
-                if self.conn_handler._user_conn is not None:
-                    self.conn_handler._user_conn.close()
-
-                if self.conn_handler._admin_conn is not None:
-                    self.conn_handler._admin_conn.close()
-
-                if self.conn_handler._admin_nodb_conn is not None:
-                    self.conn_handler._admin_nodb_conn.close()
+                self.conn_handler.close()
 
         return DecoratedClass
     return class_modifier
--- a/qiita_core/util.py
+++ b/qiita_core/util.py
@@ -80,7 +80,6 @@ def qiita_test_checker():
             @reset_test_database
             def tearDown(self):
                 super(DecoratedClass, self).tearDown()
-                self.conn_handler.close()
 
         return DecoratedClass
     return class_modifier
--- a/qiita_core/util.py
+++ b/qiita_core/util.py
@@ -80,7 +80,14 @@ def qiita_test_checker():
             @reset_test_database
             def tearDown(self):
                 super(DecoratedClass, self).tearDown()
-                del self.conn_handler
+                if self.conn_handler._user_conn is not None:
+                    self.conn_handler._user_conn.close()
+
+                if self.conn_handler._admin_conn is not None:
+                    self.conn_handler._admin_conn.close()
+
+                if self.conn_handler._admin_nodb_conn is not None:
+                    self.conn_handler._admin_nodb_conn.close()
 
         return DecoratedClass
     return class_modifier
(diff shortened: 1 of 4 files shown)


antgonza:
:+1:

ElDeveloper:
@josenavas @squirrelo can you both have a look?

squirrelo:
I can delete, add, do things with known queues so it seems to be working.

squirrelo:
Also, it is FAST :dash: :fast_forward:

ElDeveloper:
Analysis can be created and executed? What about creating a study and
processing an OTU table?

squirrelo:
Creating study and analysis yes, not sure about OTU table.

ElDeveloper:
What is the airspeed velocity ... :P

josenavas:
Not sure where to comment this on the code:

The connection cannot be created and left it open all the time. If you check the tests, you will see a failure because it cannot drop the database because there are 1 open connection. I would suggest connecting to the DB the first time that it tries to connect... I'm not sure if that will solve the issue though...

josenavas:
i.e. my last comment: basically the qiita_env drop command is broken now

squirrelo:
Maybe an explicit close_connection() function for when it's needed? The code should then fire up a new connection as needed.

wasade:
Will add, ~10 for commit

wasade:
test harness updated.

wasade:
bah. ya, should be a static method to do that. 1 more min

wasade:
That _should_ do it

squirrelo:
:+1:  if tests pass

adamrp:
reviewing now

squirrelo:
Now it seems to be opening, and keeping open, a connection on every function test.

josenavas:
Well this errors looks concerning...
```
RuntimeError: An OperationalError with the following message occured
	FATAL:  sorry, too many clients already
 For more information, review `INSTALL.md` in the Qiita installation base directory.
```

wasade:
Yup. The subset of tests executed passed locally. Will investigate in 20,
currently off laptop

josenavas:
Correct. Now that I'm thinking a bit more on this, that is exactly what is happening and that is going to keep happening as we are instantiating those objects. The correct approach is to do `conn_handler = SQLConnectionHandler()` in the sql_connection.py file and import that conn_handler in the rest of the code. That is, doing the singleton pattern.

Alternatively, the connection can be open outside of the SQLConnectionHandler object, and each new instance just uses the already created one. This will reduce the number of code modifications, as we'll only need to change that on the sql_connection.py file.

adamrp:
@josenavas I may be misunderstanding, but I think @wasade's vision here is that one doesn't actually instantiate `SQLConnectionHandler` -- the psycopg2 connection is an attribute of the *class* and not instances of the class.

wasade:
Shouldn't be necessary as the connections are static members. Issue most
likely is line 252 of env manager. The del is leaving dangling objects

Correct. Now that I'm thinking a bit more on this, that is exactly what is
happening and that is going to keep happening as we are instantiating those
objects. The correct approach is to do conn_handler = SQLConnectionHandler()
in the sql_connection.py file and import that conn_handler in the rest of
the code. That is, doing the singleton pattern.

Alternatively, the connection can be open outside of the
SQLConnectionHandler object, and each new instance just uses the already
created one. This will reduce the number of code modifications, as we'll
only need to change that on the sql_connection.py file.

—
<https://github.com/biocore/qiita/pull/1131#issuecomment-97619972>.

wasade:
Instantiation is fine, but correct about class attrs

wasade:
We're green

squirrelo:
:+1:

adamrp:
:+1: for [corks on forks](http://i.imgur.com/US4zF30.gif)

wasade:
heh

PR #1137, opened 2015-04-30
Use URL messages for messaging.

squirrelo:
Needed for failed email messages. Now when URL has level and message in it, the message is rendered correctly.


antgonza:
:+1:

PR #1140, opened 2015-05-01
Fix 1136

antgonza:
Removed all unnecessary conn_handler as parameter. Note that queues need to have the conn_handler and the queue_name, thus no removing.


josenavas:
:+1: 
@antgonza thanks for working on this! The code in here looks good. This PR is also a good exercise, since I noticed in how many places the queues are needed and we are not using them! That is a separate issue though, and we already have an issue to keep track of that, so this is good to merge :-)

PR #1141, opened 2015-05-01
Adding check_restrictions functionality

josenavas:
This adds the `check_restrictions` functionality to the metadata template objects. This functionality is needed to check the restrictions from the GUI and disable functionality as needed.

Should be easy to review!

### DIFF
--- a/qiita_db/metadata_template/base_metadata_template.py
+++ b/qiita_db/metadata_template/base_metadata_template.py
@@ -1171,9 +1171,7 @@ class MetadataTemplate(QiitaObject):
         set of str
             The missing columns
         """
-        def _col_iter():
-            for restriction in restrictions:
-                for col in restriction.columns:
-                    yield col
+        cols = {col for restriction in restrictions
+                for col in restriction.columns}
 
-        return set(_col_iter()).difference(self.categories())
+        return cols.difference(self.categories())
--- a/qiita_db/metadata_template/base_metadata_template.py
+++ b/qiita_db/metadata_template/base_metadata_template.py
@@ -1157,3 +1157,24 @@ class MetadataTemplate(QiitaObject):
                     % (category, value_str, value_types_str, column_type))
 
             raise e
+
+    def check_restrictions(self, restrictions):
+        """Checks if the template fulfills the restrictions
+
+        Parameters
+        ----------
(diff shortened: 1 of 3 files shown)


antgonza:
Your friend pep(e)8 is calling ...

qiita_db/metadata_template/base_metadata_template.py
```diff
@@ -1157,3 +1157,24 @@ def update_category(self, category, samples_and_values):
...
+        -------
+        set of str
+            The missing columns
+        """
+        def _col_iter():
+            for restriction in restrictions:
+                for col in restriction.columns:
+                    yield col
+        categories = self.categories()
+        missing = {col for col in _col_iter() if col not in categories}
```
antgonza: Will it be better to do this in a set operation?

josenavas:
Thanks @antgonza 
Comments solved!

antgonza:
:+1:

qiita_db/metadata_template/base_metadata_template.py
```diff
@@ -1157,3 +1157,23 @@ def update_category(self, category, samples_and_values):
...
+        Returns
+        -------
+        set of str
+            The missing columns
+        """
+        def _col_iter():
+            for restriction in restrictions:
+                for col in restriction.columns:
+                    yield col
+
```
wasade: Why not just do `{col for restriction in restrictions for col in restriction.columns}`? The generator is immediately consumed so it isn't clear what the benefit of lazy eval is here

wasade:
Just one comment, :+1: otherwise

qiita_db/metadata_template/base_metadata_template.py
josenavas: I was just avoiding the double list comprehension, as sometimes it has been criticized. I think it is not too complicated, so I'll change it.

wasade: makes sense. in that case though, it would be better to just create the set
using two for loops. eg

cols = set()
for restriction in restrictions:
    for col in restriction.columns:
        cols.add(col)

josenavas:
@antgonza @wasade any remaining comment in here?

PR #1143, opened 2015-05-02
adding to an investigation and sharing with a user. Fix #1079

### LINKED ISSUE #1079: missing FK in study_users and investigation_study
@josenavas noticed while reviewing https://github.com/biocore/qiita/pull/1068/files#r28648682 that we don't need to remove links to study_users and investigation_study table for a study to be deleted. This means that we do not have FK constrains on those tables.

antgonza:
@josenavas I think you are right and just adding those lines are enough. However, I think this is easier.

### DIFF
--- a/qiita_db/test/test_study.py
+++ b/qiita_db/test/test_study.py
@@ -442,7 +442,11 @@ class TestStudy(TestCase):
 
     def test_delete(self):
         title = "Fried chicken microbiome"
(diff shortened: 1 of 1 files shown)


josenavas:
:+1:

wasade:
Should this be decomposed out into a separate test, like `test_deleted_shared`? Otherwise :+1:

josenavas:
I don't think it is needed, as this tests is more extensive.

Issue #1091, opened 2015-04-21
Delete meta-analysis


josenavas:
So if a user wants to remove a private study that has meta-analyses, he can delete first the meta-analyses and then remove all the other study items.

antgonza:
I think the plan is to allow to remove any analysis, right? The current constrains about deleting a process data will not allow to delete it as long as it has an analysis ...

josenavas:
Yes, any analysis. At this point we do not allow to create an analysis from another one (i.e. the analysis chain table) and we don't have a way to share analysis or make them public (although this is a goal at the end), so we should allow to delete any analysis right now.

PR #1144, opened 2015-05-02
Fix 947

antgonza:
This adds a nicer visualization of dropped samples in an analysis. This is the base to add https://github.com/biocore/qiita/issues/1091, which will be in another PR.

### DIFF
--- a/qiita_pet/handlers/analysis_handlers.py
+++ b/qiita_pet/handlers/analysis_handlers.py
@@ -143,7 +143,7 @@ class AnalysisResultsHandler(BaseHandler):
                                              jobject.results))
 
         dropped_samples = analysis.dropped_samples
-        dropped = {}
+        dropped = defaultdict(list)
         if dropped_samples:
             for proc_data_id, samples in viewitems(analysis.dropped_samples):
                 if not samples:
@@ -151,10 +151,8 @@ class AnalysisResultsHandler(BaseHandler):
                 proc_data = ProcessedData(proc_data_id)
                 data_type = proc_data.data_type()
                 study = proc_data.study
-                if data_type not in dropped:
-                    dropped[data_type] = []
-                dropped[data_type].append({'study': Study(study).title,
-                                           'samples': samples})
+                dropped[data_type].append((Study(study).title, len(samples),
+                                           ', '.join(samples)))
 
         self.render("analysis_results.html",
                     jobres=jobres, aname=analysis.name, dropped=dropped,
--- a/qiita_pet/handlers/analysis_handlers.py
+++ b/qiita_pet/handlers/analysis_handlers.py
@@ -30,6 +30,7 @@ from qiita_db.util import (get_db_files_base_dir,
                            check_access_to_analysis_result,
                            filepath_ids_to_rel_paths)
 from qiita_db.exceptions import QiitaDBUnknownIDError
+from qiita_db.study import Study
 
 SELECT_SAMPLES = 2
 SELECT_COMMANDS = 3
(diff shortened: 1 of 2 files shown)


qiita_pet/handlers/analysis_handlers.py
```diff
@@ -141,14 +142,19 @@ def get(self, analysis_id):
             jobres[jobject.datatype].append((jobject.command[0],
                                              jobject.results))
 
-        dropped = {}
         dropped_samples = analysis.dropped_samples
+        dropped = {}
```
josenavas: You can use `defaultdict(list)` and remove the `if` statement in lines 154-155

qiita_pet/templates/analysis_results.html
```diff
@@ -11,44 +11,60 @@
...
+          </div>
+          <div id="dropped-samples" class="panel-collapse collapse">
+              <div class="panel-body">
+                  {% for data_type, studies in viewitems(dropped) %}
+                    <b><u>{{data_type}}</u></b>
+                    <br/><br/>
+                    {% for s in studies %}
+                      <b>{{s['study']}}:</b>
+                      <br/>
+                      Total dropped: {{len(s['samples'])}}
```
josenavas: Can you compute the length of the list and do the `', '.join` in the handler? In previous iterations of the code we avoid to add as much python code in templates as possible.

qiita_pet/templates/analysis_results.html
```diff
@@ -11,44 +11,60 @@
...
+        <div class="panel panel-default">
+          <div class="panel-heading">
+            <h3 class="panel-title"><a data-toggle="collapse" data-parent="#accordion" href="#dropped-samples">Dropped Samples</a></h3>
+          </div>
+          <div id="dropped-samples" class="panel-collapse collapse">
+              <div class="panel-body">
+                  {% for data_type, studies in viewitems(dropped) %}
+                    <b><u>{{data_type}}</u></b>
+                    <br/><br/>
+                    {% for s in studies %}
```
josenavas: I think it is better to use tuples rather than dictionaries in here:
```
{% for title, num_samples, samples in studies %}
```
this helps to reduce also the amount of python code.

josenavas:
Thanks @antgonza this looks way better! Few comments though...

josenavas:
:+1:

qiita_pet/handlers/analysis_handlers.py
```diff
@@ -141,14 +142,17 @@ def get(self, analysis_id):
             jobres[jobject.datatype].append((jobject.command[0],
                                              jobject.results))
 
-        dropped = {}
         dropped_samples = analysis.dropped_samples
+        dropped = defaultdict(list)
         if dropped_samples:
-            for proc_data_id, samples in viewitems(dropped_samples):
+            for proc_data_id, samples in viewitems(analysis.dropped_samples):
```
adamrp: Why did you change this line? `dropped_samples` was already set to `analysis.dropped_samples` on line 145

adamrp: In fact, the `if` statement on line 147 is not needed here, since this `for` loop will not run if `dropped_samples` is empty

adamrp:
A couple small comments

qiita_pet/handlers/analysis_handlers.py
antgonza: 1. analysis.dropped_samples -> dropped_samples: will change, left behind by mistake
2. if statement is necessary as dropped_samples is either a dict or None.

Thanks!

qiita_pet/handlers/analysis_handlers.py
```diff
@@ -141,14 +142,17 @@ def get(self, analysis_id):
             jobres[jobject.datatype].append((jobject.command[0],
                                              jobject.results))
 
-        dropped = {}
         dropped_samples = analysis.dropped_samples
+        dropped = defaultdict(list)
         if dropped_samples:
```
josenavas: @adamrp I think you suggested that this `if` statement can be removed. However, `analysis.dropped_samples` returns `None` if there are no dropped samples, instead of an empty list. This will make the call to `viewitems` to fail because of an AttributeError.

I think this should not block this PR and we should open a new issue for modifying `analysis.dropped_samples` so it returns consistent types (list). Do you agree?

adamrp: Agreed -- thanks for pointing that out, I didn't look carefully at `analysis.dropped_samples` before commenting.

josenavas: Cool! New issue created: https://github.com/biocore/qiita/issues/1152

PR #1155, opened 2015-05-05
Modify DB to support SortMeRNA dbs

### LINKED ISSUE #811: sortmerna preindexed references
Currently we are using the fna reference, which means that every single picking indexes the reference. We should be using the already indexed files.

josenavas:
This adds the first bits to fix #811 

This is modifying the database and the reference.py object to support that.

Remaining bits needed for this to be fully functional:
- [ ] Modify the CLI interface to accept the SortMeRNA indexed database
- [ ] Create a qiita_ware command to index the DB (done in the cluster)
- [ ] Add a 24.py patch that will index the database for the already existing DBs using the previous command.
- [ ] Modify the processing pipeline to use the reference database if it already exists

Note that this is going to it's own branch. Since this include DB changes, I will appreciate fast reviews so I can move forward faster and avoid what happened with the latest release.


josenavas:
This is ready for review @biocore/qiita-dev

qiita_db/reference.py
```diff
@@ -6,15 +6,71 @@
...
+            % smr_db)
+
+    new_fp = join(smr_db_dir, "%s_%s.stats" % (db_name, db_version))
+    move(old_fp, new_fp)
+
+    suff = ["pos", "bursttrie", "kmer"]
+
+    for s in suff:
+        g_re = "%s.%s_[0-9]*.dat" % (smr_db, s)
+        s_re = "(?<=%s_)[0-9]*(?=.dat)" % s
```
adamrp: In the lookahead, the `.` needs to be escaped with a `\` in front of it, otherwise it will match any character. Also, I *think* the lookbehind should have a `\.` in it also (assuming the filename is always of the form "something.pos_##.dat"):
```python
s_re = "(?<=\.%s_)[0-9]*(?=\.dat)" % s
```
You could also simplify this RE (get rid of the lookbehind and lookahead) if you wanted
```python
s_re = "\.%s_(\d+?)\.dat)" % s
```
And then the number you want would be accessible at `result.group(1)` after doing the `re.search`

qiita_db/reference.py
```diff
@@ -6,15 +6,71 @@
...
+
+        old_fps = glob(g_re)
+
+        if not old_fps:
+            raise QiitaDBError(
```
wasade: if this exception is raised the filesystem will be inconsistent as the move on line 49 will not have been reverted. Recommend queuing the moves, and do all at once if possible

qiita_db/reference.py
wasade: since the names are predictable, why is the search necessary?

qiita_db/support_files/patches/24.sql
```diff
@@ -0,0 +1,6 @@
+ALTER TABLE qiita.reference ADD sortmerna_indexed_db_filepath bigint  ;
+
+CREATE INDEX idx_reference_2 ON qiita.reference ( sortmerna_indexed_db_filepath ) ;
```
adamrp: Is there a need for this index? I.e., will we ever be looking through the reference table looking for a particular filepath?

qiita_db/reference.py
wasade: is there an assurance that databases won't be combined? e.g., `foo.idx.bursttrie_0.dat` and `bar.idx.bursttrie_0.dat`?

qiita_db/test/test_reference.py
```diff
@@ -7,19 +7,22 @@
...
+from qiita_db.exceptions import QiitaDBWarning, QiitaDBError
+from qiita_db.reference import Reference, _rename_sortmerna_indexed_db_files
 from qiita_db.util import get_mountpoint, get_count
 
 
 @qiita_test_checker()
 class ReferenceTests(TestCase):
     def setUp(self):
-        self.name = "Fake Greengenes"
+        self.name = "Fake_Greengenes"
```
adamrp: What changed here? Something I'm not seeing, or github messing up?

qiita_db/reference.py
```diff
@@ -23,6 +79,7 @@ class Reference(QiitaObject):
     sequence_fp
     taxonomy_fp
     tree_fp
+    sortmerna_indexed_db
```
wasade: I don't believe it is correct to assume that all sequence databases will (or can) have a sortmerna index. For instance, I do not believe `indexdb_rna` will work with protein sequence.

wasade: This comment is associated also with the property on like 249

qiita_db/test/test_reference.py
```diff
@@ -70,6 +77,60 @@ def test_create(self):
                [tree_id, exp_tree, 12, '0', 1, 6]]
         self.assertEqual(obs, exp)
 
+    def test_create_with_smr_indexed_db(self):
+        fp_count = get_count('qiita.filepath')
+
+        # We need to create a sortmerna db
+        suffixes = ['.bursttrie_0.dat', '.kmer_0.dat', '.pos_0.dat', '.stats']
+        smr_dir = mkdtemp()
```
adamrp: According to [the documentation](https://docs.python.org/2/library/tempfile.html#tempfile.mkdtemp), the creator is responsible for removing the temporary directory.

qiita_db/reference.py
adamrp: @wasade I thought about that, too, but this class is also assuming that it will have a taxonomy and tree, so I figured if it is assumed to have those things, it can be assumed to have a sortmerna indexed db, too, right? In any case, if it isn't passed, a warning is raised (not an error), but maybe it shouldn't even be a warning?

qiita_db/reference.py
adamrp: @wasade the search is necessary because you don't know how many numbers will follow that first underscore, but it should be at least 1 (hence the use of `+` instead of `*` in the regex)

Not sure what you mean by databases being combined

qiita_db/reference.py
wasade: @adamrp, taxonomy and tree are still valid for a focused protein database, though it does suggest that we need to rethink the concept of a sequence reference here as we need to support multi locus dbs. Note also that the property on like 249 will raise not warn

qiita_db/reference.py
wasade: Ya, I retract the comment on the dbs combining as the regex and glob is anchored. 

Under what situations are there multiple files produced, and what is the max that can be produced? I'm not seeing it in the documentation. I indexed the 99% Greengenes 13_8 OTUs, which span about 200k sequences, and it still only has a single set of files.

What I'm wondering is whether it is feasible to be explicit.

qiita_db/reference.py
adamrp: True it will raise on that property -- I was thinking about the `create` warning. I think the property should neither raise nor warn, just return None if it doesn't exist.

adamrp: :+1: to having different reference classes for 16S, 18S, proteins, etc.

qiita_db/reference.py
adamrp: Ah I see -- I'm not sure about that, I have never looked closely at the index files generated by SortMeRNA

josenavas: Thanks for the tips ~~regex whisperer~~ @adamrp 
@wasade I think is possible that multiple files are produced. It will be a point that if the reference DB grows enough, multiple files will be generated. In any case, I'd prefer to not limit the system, and just support everything just in case. Better be safe than sorry.

josenavas: @adamrp I cannot get you regex to work. When I'm using what you're proposed, it didn't work and I had to remove the last parenthesis.. When I do that, the search matches this: `'.pos_0.dat'` which is not what I was looking for, although I can use that also and simplify the creation of the new filepath...

qiita_db/reference.py
josenavas: Good point. Will do

qiita_db/reference.py
adamrp: Ah yeah sorry, that last parenthesis is a typo -- remove it (I edited my comment above to be correct). And are you using group(0) or group(1)? group(0) will be the entire matched expression, whereas group(1) will be the thing that's inside the capturing parentheses (i.e., just the number)

qiita_db/reference.py
josenavas: I don't know if we really want to have different reference classes for 16S, 18S, proteins, etc...
I think I just need to change the property to return None if that is not used. The taxonomy_fp and the tree_fp are also optional (e.g. ITS does not have a tree), so they should also return None if that doesn't exist. That is a bug on the code and I'm going to fix it (right now it fails if they're not present). I'll also add tests for them.

qiita_db/support_files/patches/24.sql
josenavas: This index was created by DBSchema, when I added the foreign key. If you're sure that is not needed. I can remove it w/o a problem :-)

qiita_db/test/test_reference.py
josenavas: I just removed the space character and I used an underscore

qiita_db/test/test_reference.py
josenavas: Thanks for pointing this out. I also just noticed that this entire test suite is not cleaning the environment after the tests. Dammit @josenavas from the past hahaha

qiita_db/reference.py
josenavas: :astonished: I was using group(0)... thanks!

josenavas:
Thanks @adamrp and @wasade I think I've addressed your comments!

josenavas:
Nope, wait, I forgot one!

qiita_db/test/test_reference.py
adamrp: Haha I guess my brain just interprets underscores as spaces these days, I truly honestly did not even see it. Thanks, was just curious :)

josenavas:
@adamrp @wasade Now this is ready for another review. The comment that I was missing was to clean up the files and directories created by the tests.

ElDeveloper:
@adamrp & @wasade would you guys mind having a look?

josenavas:
@adamrp, @wasade don't worry about this PR. It got out of date and is going to conflict with my other PR.

adamrp:
@josenavas okay, thanks for the heads up. Should this one be closed?

josenavas:
Yeah, it's probably the best.

PR #1158, opened 2015-05-06
Fixes #1151

### LINKED ISSUE #1151: Listing the samples that are not modified by the extend functionality was a bad idea
In a study with a lot of samples, the entire screen just get populated with all the sample names, which is useless for the user. We should just put the number of samples, and it will be equally useful.

josenavas:
Fixes #1151 by not printing all the sample IDs. I just changed the warning message to provide a count.


josenavas:
@biocore/qiita-dev same here, if this PR can make it for today's update, it will be good for improving the visualization and the work for @ackermag

PR #1159, opened 2015-05-06
Fix #1150

### LINKED ISSUE #1150: Change the "Add samples" button to "extend"
Now the extend functionality has to be used to add more sample and to add more columns, so that button is no longer accurate.

josenavas:


josenavas:
@biocore/qiita-dev The system will be updated tonight. It will be useful if this PR can make it, as this is generating some confusion to the users (including @ackermag)

squirrelo:
:+1:

PR #1166, opened 2015-05-07
Fix 1061 - analysis.jobs always returns list

squirrelo:
Does what it says on the tin. Also checks for jobs existing before running an analysis.


qiita_ware/analysis_pipeline.py
```diff
@@ -28,6 +28,9 @@ def _build_analysis_files(analysis, r_depth=None, **kwargs):
     r_depth : int, optional
         Rarefaction depth for biom table creation. Default None
     """
+    if not analysis.jobs:
+        raise RuntimeError("Analysis %d has no jobs attached!" % analysis.id)
```
josenavas: How is this error shown to the user? Or it will never be shown to them?

squirrelo: Should never be shown to them. This is more for CLI interface we were talking about where you might try and run an analysis with no jobs.

qiita_db/analysis.py
```diff
@@ -432,7 +432,7 @@ def jobs(self):
                "analysis_id = %s".format(self._table))
         job_ids = conn_handler.execute_fetchall(sql, (self._id, ))
         if job_ids == []:
-            return None
+            return []
```
wasade: This is a small can of worms... but can the docstring be updated to indicate what to expect when the job isn't found?

Also, if `job_ids` comes back as `[]`, then the if statement is redundant. As in, the following should still be valid:

```python
return [job_id[0] for job_id in job_ids]
```

squirrelo: Good point. I can update.

josenavas:
@squirrelo 

I just tried this in my end. If I create an analysis and I don't choose any job, the analysis moves forward and tries to execute. I'm redirected to the compute wait page. I'm then redirected to the analysis results page and I cannot see anything:

screen shot 2015-05-07 at 11 23 18 am

However, if I go to the error page as an admin, I can see the error there:
screen shot 2015-05-07 at 11 22 25 am

And as the user I can see that the analysis failed:
screen shot 2015-05-07 at 11 24 57 am

However as a user I don't have any clue why it failed... Is it possible to add a simple check on the Select Commands page that does not allow the user to submit jobs if he doesn't select anything?

squirrelo:
I thought that check already existed. Good to catch it now I guess. I'll add the javascript.

squirrelo:
Added and ready to  go

qiita_pet/templates/select_commands.html
```diff
@@ -2,7 +2,17 @@
 {% autoescape None %}
 
 {%block head%}
+
 <script>
+  function check_selection(){
+    if($('.command:checkbox:checked').length > 0){
+       return true;
+   }else {
+        $('#error').text('Please select at least one command!');
```
josenavas: Don't add the `!` sign at the end, it's not nice to yell at the users :smiley:

qiita_ware/analysis_pipeline.py
josenavas: Can you update the docstring to reflect that the function raises an error? Thanks!

qiita_pet/templates/select_commands.html
squirrelo: But they're being bad! :P

josenavas:
:+1: thanks @squirrelo

antgonza:
Looks fine, could you also remove this if https://github.com/biocore/qiita/blob/master/qiita_pet/handlers/analysis_handlers.py#L147? Thanks.

josenavas:
@squirrelo ping on this. @antgonza 's comment is really easy to address.

squirrelo:
Done and ready.

josenavas:
:+1: @antgonza can you merge once the tests pass if you are happy with the code?

antgonza:
Real failures.

squirrelo:
Passing again and ready for merge

PR #1167, opened 2015-05-07
Add ON UPDATE CASCADE to the sample ids FK

josenavas:
This adds the ON UPDATE CASCADE to the FK that points to the sample ids.

While developing the patch I noticed that the dynamic tables did not had the Foreign Key referencing the sample ids. The first part of the patch is adding that FK, while the second one is adding the ON UPDATE CASCADE to everything that references to sample id.

I've also updated the code so when the dynamic tables are created, the FK with ON UPDATE CASCADE constraint is also added to them.

@ElDeveloper @adamrp @wasade @squirrelo I would appreciate a review on this since this is blocking the update of an EMP study, and there is some people waiting on this to be able to update the study to EBI.

### DIFF
--- a/qiita_db/support_files/patches/24.sql
+++ b/qiita_db/support_files/patches/24.sql
@@ -1,0 +1,57 @@
+-- May 7, 2015
+-- This patch adds the ON UPDATE CASCADE constraint to all the FK
+-- that are referencing the sample ids
+
+DO $do$
+DECLARE
+    dyn_t varchar;
+    fk_vals RECORD;
+BEGIN
+
+-- The dynamic tables do not have a FK set on their sample ID
+-- We need to find the dynamic tables existing in the system and we add the
+-- FK constraint to them.
+FOR dyn_t IN
+    SELECT DISTINCT table_name
+    FROM information_schema.columns
+    WHERE (SUBSTR(table_name, 1, 7) = 'sample_'
+           OR SUBSTR(table_name, 1, 5) = 'prep_')
+        AND table_schema = 'qiita'
+        AND table_name NOT IN ('prep_template',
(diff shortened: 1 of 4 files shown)


josenavas:
@antgonza this will need top be deployed in the live system. Do you have any notes on what are you doing to update the system, so I can do it once this is merged?

Thanks!

squirrelo:
So if I'm reading this right, the FK will be tested by the create tests that currently exist, correct? Other than that question, :+1:

antgonza:
I do not have notes but I should be able to update the system once this is merged and create some ...

josenavas:
@squirrelo I'm not understanding your question...

squirrelo:
Are there tests that make sure the foreign key is created and valid?

josenavas:
There is a test that checks that the SQL contains the creation of the FK.

qiita_db/support_files/patches/24.sql
```diff
@@ -0,0 +1,57 @@
...
+    fk_vals RECORD;
+BEGIN
+
+-- The dynamic tables do not have a FK set on their sample ID
+-- We need to find the dynamic tables existing in the system and we add the
+-- FK constraint to them.
+FOR dyn_t IN
+    SELECT DISTINCT table_name
+    FROM information_schema.columns
+    WHERE (SUBSTR(table_name, 1, 7) = 'sample_'
```
wasade: won't this match `sample_template_filepath`?

wasade: ...oh, just saw the additional conditionals :)

adamrp:
This looks good to me :+1:

squirrelo:
:+1:

PR #1168, opened 2015-05-07
Removing any instance of required_sample_info and common_prep_info. …

josenavas:
…Also adding a test for metadata_headers

@wasade noticed that there was still a couple of references to the `required_sample_info` and `common_prep_info` tables. I'm removing any reference to those here.

The function in which `required_sample_info` was used did not had a test. When I added a test, I noticed that the populate_test_db.sql was not consistent which what our code represents (i.e. we are not allowing to have uppercase column names) so I modified it to be consistent with the rest of our code.


wasade:
:+1:

PR #1169, opened 2015-05-07
DO NOT MERGE - Stop sending empty files to EBI

### LINKED ISSUE #1070: Modify EBI code so that it does not upload empty files
These cause error messages to be thrown from EBI's end

adamrp:
Still need to test this out, but I think this will do the trick

Fix #1070

### DIFF
--- a/qiita_ware/ebi.py
+++ b/qiita_ware/ebi.py
@@ -15,15 +15,10 @@ from future.utils import viewitems
 from skbio.util import safe_md5
 
 from qiita_core.qiita_settings import qiita_config
-
 from qiita_db.logger import LogEntry
 from qiita_db.ontology import Ontology
 from qiita_db.util import convert_to_id
-
-
-class EmptyFileWarning(Warning):
-    """Warning that is raised when an empty file is encountered"""
-    pass
+from .exceptions import EmptyFileWarning
 
 
 class InvalidMetadataError(Exception):
--- a/qiita_ware/ebi.py
+++ b/qiita_ware/ebi.py
@@ -4,7 +4,8 @@ from subprocess import call
 from shlex import split as shsplit
 from glob import glob
 from os.path import basename, exists, join, split
-from os import environ, close
+from os import environ, close, stat
+from os.path import exists
 from datetime import date, timedelta, datetime
 from xml.etree import ElementTree as ET
 from xml.dom import minidom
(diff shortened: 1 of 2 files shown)


qiita_ware/ebi.py
```diff
@@ -20,6 +21,11 @@
 from qiita_db.util import convert_to_id
 
 
+class EmptyFileWarning(Warning):
```
josenavas: I recommend to add this warning in the exceptions class and add a filter so the Warning is always raised. There is an example in qiita_db.exceptions.

josenavas:
Is this really the best way of checking for empty files? I think it will be better if when we are generating the per sample fastq we detect that a sample do not have sequences and we don't write that file out. Does that makes sense?

qiita_ware/ebi.py
```diff
@@ -404,6 +410,17 @@ def add_sample_prep(self, sample_name, platform, file_type, file_path,
                              "%s (case insensitive)" % (platform,
                                                         ', '.join(platforms)))
 
+        with open(file_path) as fp:
```
wasade: what about:

```python
import os
if not os.path.exists(file_path) or os.stat(file_path).st_size == 0:
    # warn, return
```

wasade:
@josenavas, it may depend on how that code is written as to if it is easier

adamrp:
@josenavas yes, I think that would be better, but I thought the per-sample fastq files were being generated by QIIME.

@wasade I think using `os.stat(file_path).st_size` would be fine.

qiita_ware/ebi.py
adamrp: I'll move the warning to `qiita_ware.exceptions` and import it here. What is the purpose of the filter? i.e., in what circumstances would the warning not be raised if the `warnings.simplefilter("always")` were not there?

josenavas:
@adamrp if you don't add the filter, in the live system the warning will be raised only once, (the first time), since this is the default behavior for warnings in python.

adamrp:
Okay, thanks! I added the filter.

adamrp:
Actually, checking st_size == 0 is not sufficient, since a gzipped empty file is not zero bytes:

```
(python276)adam:test@ebi_empty_files-issue-1070$ touch hello
(python276)adam:test@ebi_empty_files-issue-1070$ du -h hello
  0B	hello
(python276)adam:test@ebi_empty_files-issue-1070$ gzip hello
(python276)adam:test@ebi_empty_files-issue-1070$ du -h hello.gz
4.0K	hello.gz
```

adamrp:
And not all empty files have the same md5 (the name of the file is encoded):

```
(python276)adam:test@ebi_empty_files-issue-1070$ touch test test2
(python276)adam:test@ebi_empty_files-issue-1070$ gzip test test2
(python276)adam:test@ebi_empty_files-issue-1070$ md5 test.gz test2.gz
MD5 (test.gz) = 6976ece38ea13638c2c4724da1fdd640
MD5 (test2.gz) = a84c25cf07c3c8fbd22e6c6d69c3f615
(python276)adam:test@ebi_empty_files-issue-1070$ xxd test.gz
0000000: 1f8b 0808 684a 4d55 0003 7465 7374 0003  ....hJMU..test..
0000010: 0000 0000 0000 0000 00                   .........
(python276)adam:test@ebi_empty_files-issue-1070$ xxd test2.gz
0000000: 1f8b 0808 684a 4d55 0003 7465 7374 3200  ....hJMU..test2.
0000010: 0300 0000 0000 0000 0000                 ..........
```
So I think the best we can do is open (or gzip.open) the file, and check if calling `read` on it returns the empty string

wasade:
Perhaps it is worth checking if its easy to just not produce the files?

adamrp:
Thought that was coming from QIIME, but I might be mistaken. I'll have a look.

wasade:
I thought we were generating from the demux files?

adamrp:
Yep, just found that in qiita_ware/commands.py. Workig on it

adamrp:
I think it will be easier just to change that code, so I'm going to close this PR.

PR #1170, opened 2015-05-08
Fix demux to_ascii for fasta files

adamrp:
Previously, this function would fail with fasta files because the call to qual.astype would fail (because qual is None when the file is fasta)


josenavas:
:+1: @wasade do you mind taking a look?

wasade:
:+1:

adamrp:
Merge?

PR #1171, opened 2015-05-08
Remove unused files

adamrp:
As far as I can tell, these fastq files are not used anywhere in the codebase. The `__init__.py` file should not be needed here since nothing in this directory is importable.


adamrp:
never mind, I think they are used in test_ebi

PR #1172, opened 2015-05-09
Stop writing empty files and improve EBI tests

### LINKED ISSUE #1070: Modify EBI code so that it does not upload empty files

adamrp:
Fix #1070


josenavas:
@adamrp it looks like there are legitimate test failures

adamrp:
Getting the same build failures that are over in #1175. Not sure *exactly* why this just started to be a problem, but the error message is related to pandas, which had an update on May 11, so I'm suspecting something there. The only thing I could find in the [changelog](http://pandas.pydata.org/pandas-docs/stable/whatsnew.html) related to this warning being raised is [from an 8 Oct 2014 release (version 0.15)](http://pandas.pydata.org/pandas-docs/stable/whatsnew.html#whatsnew-0150-deprecations).

Anyways, I think the issue is that because the warning we're looking for isn't the *first* warning, it's erroring (according to the [nympy documentation on assert_warns](http://docs.scipy.org/doc/numpy/reference/generated/numpy.testing.assert_warns.html)), so we might have to take a different approach, perhaps something that borrows from [this SO post](http://stackoverflow.com/questions/3892218/how-to-test-with-pythons-unittest-that-a-warning-has-been-thrown).

ElDeveloper:
Although less desirable, we could also just stick with an older version
of Pandas i.e. the one that was working just fine.

squirrelo:
I think I've traced the issue to [here](https://github.com/biocore/qiita/blob/master/qiita_db/metadata_template/util.py#L115), but I don't know enough about Pandas to fix the issue.

adamrp:
Ready for review

qiita_ware/test/test_ebi.py
```diff
@@ -26,15 +25,12 @@
 
 class TestEBISubmission(TestCase):
     def setUp(self):
-        self.path = path.dirname(path.abspath(__file__)) + '/test_data'
+        self.path = path.dirname(path.abspath(__file__)) + \
```
josenavas: Can you use `join` instead of string concatenation?

josenavas: you can even use the `partial` function, so it will simplify the following 3 lines.

josenavas:
Thanks @adamrp , just a small comment.

adamrp:
Thanks @josenavas, done!

josenavas:
:+1:

PR #1175, opened 2015-05-10
Delete analysis

antgonza:
Adding the possibility of deleting analysis. The first step to be able to delete an analysis is to make sure it exists, thus also adding the exist function to the Analysis object.


qiita_db/analysis.py
```diff
@@ -165,6 +169,71 @@ def create(cls, owner, name, description, parent=None, from_default=False):
         a_id = conn_handler.execute_queue(queue)[0]
         return cls(a_id)
 
+    @classmethod
+    def delete(cls, _id):
```
josenavas: @antgonza there is still another table in which the analysis can be present: `analysis_chain`. Right now we are not using that table for anything, but we can easily forget about it on deletion once we start using it. There are 2 possible solutions:
 * Open an issue, so we don't keep track of it
 * We add that table here. This implies that you also need to add a check: a study cannot be deleted if it has 'child' analyses.

I'm fine with adding an issue, since currently we are not using that and a lot of other changes will be needed once that gets in place.

qiita_pet/handlers/analysis_handlers.py
```diff
@@ -154,21 +154,54 @@ def get(self, analysis_id):
...
+            msg_level = "success"
+        except Exception as e:
+            msg = ("Couldn't remove <b><i>%s</i></b> analysis: %s" % (
+                analysis_name, str(e)))
+            msg_level = "danger"
+
+        # redirecting to list of analysis but also passing messages and
+        # we need to change the request.method to GET
+        self.request.method = 'GET'
+        ShowAnalysesHandler(self.application, self.request)._execute(
```
josenavas: Is this safe? You are calling a private method of the handler, which already smells fishy... I don't think this is the correct way of doing this, you should be using the `redirect` call.

josenavas:
Few comments @antgonza

qiita_db/analysis.py
antgonza: I happy adding the check and the delete of those rows ... shouldn't be a problem.

qiita_pet/handlers/analysis_handlers.py
antgonza: That's the only way I found to "redirect" with parameters. This is, a simple redirect will not send the full list of parameters (for example, in this case: msg or msg_level). Any suggestions?

josenavas: I think you should be able to encode the parameters in the URL: `self.redirect('/analysis/show/?msg=%s&msg_level=%s')`. Sorry, I can't test right now, but there is a couple of examples in the code base (`qiita_pet//handlers/preprocessing_handlers.py` or `iita_pet//handlers/auth_handlers.py`).

antgonza: I tested that and it didn't work. Simply try: http://qiita.ucsd.edu/?msg=Jose&msg_level=danger

josenavas: I just take a closer look and I think the bug is in the `ShowAnalysesHandler`. Instead of expecting the parameters in the `get` method, you have to ask for them in the code by using `self.get_argument('message', '')`.

This is how it is done in the preprocessing_handlers.py and it works in there, and I do remember having some issue like that. Can you give it a shot?

qiita_db/analysis.py
antgonza: Decided to leave it as an issue https://github.com/biocore/qiita/issues/1176 so the solution is complete when implemented.

qiita_pet/handlers/analysis_handlers.py
antgonza: those changes actually worked.

qiita_db/analysis.py
josenavas: Thanks @antgonza just one minor thing: can you add a comment with a reference to the issue? Like `# TODO: issue #1176`, so we have a pointer in the code and we don't forget about it.

josenavas:
@antgonza I see some errors in your build. After a quick look, they don't look like it's related to your changes, but we will need to investigate more.

I pulled down your code and it looks good and works awesome!

adamrp:
@antgonza @josenavas seeing the same thing over in #1172.

josenavas:
Thanks for pointing that out. I'm wondering if one of our dependencies changed something, so that's why we are seeing that behavior...

adamrp:
@antgonza There are merge conflicts here, which is preventing auto-merge, which is preventing the fixes in #1186 from being applied before the test suite runs. Fixing the merge conflicts should resolve the issue.

antgonza:
Ready for another review ...

josenavas:
:+1:

qiita_pet/handlers/analysis_handlers.py
```diff
@@ -153,21 +153,51 @@ def get(self, analysis_id):
...
+
+        try:
+            Analysis.delete(analysis_id)
+            msg = ("Analysis <b><i>%s</i></b> has been deleted." % (
+                analysis_name))
+            level = "success"
+        except Exception as e:
+            msg = ("Couldn't remove <b><i>%s</i></b> analysis: %s" % (
+                analysis_name, str(e)))
+            level = "danger"
```
adamrp: Can you log the error?

adamrp:
One comment, but looks good otherwise

PR #1177, opened 2015-05-14
Issue 1129 - Add messaging for analyses with changed underlying files

squirrelo:
This adds two tables to the DB for logging changes to prep and sample templates. These are linked to analyses by analysis_samples table, and used to store and retrieve changes made that relate to the sample and prep templates. This allows us to not only keep track of how and when changes were made, but also to display these changes to the users so they can decide whether or not to re-run an analysis. This also starts a backbone for us to do the twitter feed.


squirrelo:
Passing and ready for first review.

qiita_db/metadata_template/prep_template.py
```diff
@@ -75,6 +75,27 @@ class PrepTemplate(MetadataTemplate):
     _sample_cls = PrepSample
     _fp_id = convert_to_id("prep_template", "filepath_type")
     _filepath_table = 'prep_template_filepath'
+    _log_table = "prep_template_edit"
+
+    @staticmethod
+    def _update_analyses(pt_id):
```
josenavas: Why is this a staticmethod? You're only applying it to a single prep template, so it is, by definition of the method, an instance method.

qiita_db/metadata_template/prep_template.py
```diff
@@ -244,11 +265,19 @@ def delete(cls, id_):
...
+            "DELETE FROM qiita.{0} where {1} = %s".format(cls._log_table,
+                                                          cls._id_column),
+            (id_,))
+
         # Remove the row from prep_template
         conn_handler.execute(
             "DELETE FROM qiita.prep_template where "
             "{0} = %s".format(cls._id_column), (id_,))
 
+        cls._update_analyses(id_)
```
josenavas: I think this line is not needed. A prep template cannot be deleted if it's data has been used in an analysis.

qiita_db/metadata_template/sample_template.py
```diff
@@ -70,6 +70,28 @@ class SampleTemplate(MetadataTemplate):
     _sample_cls = Sample
     _fp_id = convert_to_id("sample_template", "filepath_type")
     _filepath_table = 'sample_template_filepath'
+    _log_table = "sample_template_edit"
+
+    @staticmethod
```
josenavas: Lot's of code duplication in here. Possible to move to the base class and just do some smart thing to get the sql query?

qiita_db/metadata_template/base_metadata_template.py
```diff
@@ -755,6 +756,7 @@ def _add_common_extend_steps_to_queue(self, md_template, conn_handler,
                 table_name, ", ".join(headers),
                 ', '.join(["%s"] * len(headers)))
             conn_handler.add_to_queue(queue_name, sql, values, many=True)
+        return new_cols, new_samples
```
josenavas: Please, update documentation to reflect this change.

qiita_db/analysis.py
```diff
@@ -628,6 +629,40 @@ def remove_samples(self, proc_data=None, samples=None):
...
+        Returns
+        -------
+        dict of list
+            Changes in the form {study_title: [change1,change2, ...]...}
+        """
+        if self.status != "altered_data":
+            return {}
+        conn_handler = SQLConnectionHandler()
+        studydict = defaultdict(list)
+        sql = """SELECT DISTINCT study_title, change, timestamp
```
josenavas: timestamp is a reserver word by SQL. Can you change the name of the column so it doesn't use the reserved word?

qiita_db/analysis.py
```diff
@@ -628,6 +629,40 @@ def remove_samples(self, proc_data=None, samples=None):
...
+        JOIN qiita.prep_template_preprocessed_data USING (prep_template_id)
+        JOIN qiita.preprocessed_processed_data USING (preprocessed_data_id)
+        JOIN qiita.study_processed_data USING (processed_data_id)
+        JOIN qiita.study study_title USING (study_id)
+        JOIN qiita.analysis_sample USING (processed_data_id)
+        WHERE analysis_id = %s ORDER BY timestamp DESC"""
+        for title, change in conn_handler.execute_fetchall(sql, [self._id]):
+            studydict[title].append(change)
+
+        sql = """SELECT DISTINCT study_title, change, timestamp
```
josenavas: Same here.

qiita_db/analysis.py
```diff
@@ -628,6 +629,40 @@ def remove_samples(self, proc_data=None, samples=None):
...
+        studydict = defaultdict(list)
+        sql = """SELECT DISTINCT study_title, change, timestamp
+        FROM qiita.prep_template_edit
+        JOIN qiita.prep_template_preprocessed_data USING (prep_template_id)
+        JOIN qiita.preprocessed_processed_data USING (preprocessed_data_id)
+        JOIN qiita.study_processed_data USING (processed_data_id)
+        JOIN qiita.study study_title USING (study_id)
+        JOIN qiita.analysis_sample USING (processed_data_id)
+        WHERE analysis_id = %s ORDER BY timestamp DESC"""
+        for title, change in conn_handler.execute_fetchall(sql, [self._id]):
```
josenavas: How is this working if the SQL query is returning 3 values?

josenavas: I guess there is a test missing in here.

qiita_db/metadata_template/sample_template.py
josenavas: Same here re static method.

qiita_db/metadata_template/sample_template.py
```diff
@@ -172,11 +194,18 @@ def delete(cls, id_):
...
+            (id_,))
+
+        conn_handler.add_to_queue(
+            queue,
             "DELETE FROM qiita.{0} where {1} = %s".format(cls._column_table,
                                                           cls._id_column),
             (id_,))
 
         conn_handler.execute_queue(queue)
+        cls._update_analyses(id_)
```
josenavas: same here. A sample template cannot be removed if there is data using it. No need for this line.

qiita_db/metadata_template/test/test_sample_template.py
```diff
@@ -1528,6 +1540,29 @@ def test_extend_add_samples(self):
         md_ext = pd.DataFrame.from_dict(md_dict, orient='index')
 
         st.extend(md_ext)
+        analysis = Analysis(1)
```
josenavas: Can you actually create a different test for this? It would be easier to identify which functionality is missing.

qiita_db/support_files/patches/25.sql
```diff
@@ -0,0 +1,29 @@
+-- May 12, 2015
+-- Add in analysis status marking that undelying data has changed.
+-- Add logging tables for changes to sample and prep templates
+INSERT INTO qiita.analysis_status (status) VALUES ('altered_data');
+
+-- Add logging tables for template edits
+CREATE TABLE qiita.sample_template_edit (
+	study_id   bigint  NOT NULL,
+	change               varchar  NOT NULL,
+	timestamp            timestamp DEFAULT current_timestamp NOT NULL,
```
josenavas: timestamp is a reserved word, please use a different column name

qiita_pet/templates/analysis_results.html
```diff
@@ -12,6 +12,19 @@
...
+  {% if status == "altered_data" %}
+    <h2 style="color:red">UNDERLYING DATA FOR THIS ANALYSIS HAS CHANGED. CONSIDER RE-RUNNING THIS ANALYSIS.</h2>
+    {% for study, change in changes %}
+    <p><b>study</b><br/>
+      {% for c in change %}
+        {{c}}<br/>
+      {% end %}
+      </p>
+    {% end %}
+    {% if changes == {} %}
```
josenavas: This will never happen. If it does, we have a huge problem in another part of the code base.

qiita_pet/templates/analysis_results.html
```diff
@@ -12,6 +12,19 @@
 
 <div class="row">
   <h1>&nbsp;Analysis: <u>{{aname}}</u></h1>
+  {% if status == "altered_data" %}
```
josenavas: Probably useful to use an accordion-like visualization here. The problem is that if a lot of thing change, it will be really hard to go over the list.

qiita_db/metadata_template/sample_template.py
```diff
@@ -221,9 +250,15 @@ def extend(self, md_template):
...
-        self._add_common_extend_steps_to_queue(md_template, conn_handler,
-                                               queue_name)
+        new_cols, new_samples = self._add_common_extend_steps_to_queue(
+            md_template, conn_handler, queue_name)
+
         conn_handler.execute_queue(queue_name)
+        self._update_analyses(self.id)
+        if new_cols:
+            self.log_change("Columns added: %s" % ', '.join(new_cols))
+        if new_samples:
```
josenavas: Not concerned about new samples, as the ProcessedData that the user is using will never be updated to include the new samples, to this will not affect any analysis.

josenavas:
Few comments @squirrelo 

General comment. The logging tables are going to be growing and growing and there is no way to clean them up. We need to have another method to clean rows from those tables if no analysis are on 'altered_data' status pointing to the prep templates. I'm happy of adding an issue for now, but it should be resolved at some point.

Another general comment: this is logging all the changes independently if the prep/sample template has been used in an analysis. I would suggest to log the changes if and only if an analysis uses its data, there is no need to store that information otherwise.

qiita_db/metadata_template/prep_template.py
squirrelo: because delete is a class method, which means it needs to be a static method in order to update on delete. Is a good argument for moving the delete to be an instance method.

qiita_db/metadata_template/sample_template.py
squirrelo: Still good for the user to see, since they may want to then re-create the analysis with those new samples.

qiita_db/metadata_template/prep_template.py
josenavas: See my other comment about the delete. So no need for a staticmethod

qiita_db/metadata_template/sample_template.py
josenavas: The user cannot do anything if a new processed data has been generated and has been make public. It is NOT changing any underlying data used by the user, is it creating a new one. Please remove

squirrelo:
I'm thinking long term for the logging tables. It will also come in handy for the twitter feed, as I can then update everyone when a shared study's prep or sample template changes. Hence logging every change. It also keeps a nice record of what has changed for any given prep/sample template if you want to dig into the history of your or someone else's study.

josenavas:
@biocore/qiita-dev Can someone weight in @squirrelo latest comment? I don't think we need to keep track of this detailed changelog for the templates. I'm also specially concern about IRB issues, in case we are shown a change needed by IRB, it will still be visible by all users...

Another concern about that approach is the size of that table, since it will keep increasing and increasing, where we can find ourselves with an extremely huge table...

Is somebody else seeing an advantage in keeping that changelog?

qiita_db/support_files/patches/25.sql
squirrelo: it actually works fine as a column name:
```sql
qiita_test=# select timestamp from qiita.sample_template_edit;                  
 timestamp 
-----------
(0 rows)
```

qiita_pet/templates/analysis_results.html
squirrelo: it did for delete updates, but since that's not a thing any more this check is no longer needed.

qiita_db/metadata_template/sample_template.py
squirrelo: They should be able to generate it themselves, though. We agreed that by getting access to a study through public or sharing, you can generate new processed data to work with.

josenavas: That's the first notice that I have for this. If the study is public, you cannot generate new ProcessedData. Note that what we are making public is the ProcessedData not the study... so there is no way of doing this...

adamrp:
While I doubt the change log tables would grow to be problematically large (it is *possible* but unlikely), I do not immediately see the benefit in keeping these details in the database. We are already keeping the old templates on the filesystem.

Also, there are merge conflicts at the moment

wasade:
having detailed logging information about usage is great for doing
subsequent analysis of what features are used, how long things take, etc. i
agree it is unlikely the table will grow to be unwieldy (as in, like 100s
of billions of rows), and agree that having information about user actions
(submitted job of type X, uploaded data of size Y, etc) could be really
cool later.

josenavas:
@wasade note that this table is only storing which changes are performed in the sample/prep templates not in the entire system.

wasade:
Were storing logging information in multiple places?

josenavas:
We were not storing these logging information. @squirrelo is proposing adding a table to store all the changes that are done in the templates. This is useful for flagging the users when the underlying data used in their analysis changes. That's why I don't think that storing all the changes in the templates always is useful.

Storing something as you are proposing might be useful at early stages of development. However, I would suggest that your proposal should be a different table and we should decorate the functions that we want to log. But that is a different discussion.

antgonza:
@squirrelo conflicts + you need to address latest comments.

squirrelo:
After discussion with @josenavas closing this pull request to implement all of this in redis, not postgres.

squirrelo:
While looking at this in redis, found an issue. If a user does not log in for over 3 months (or whatever redis timeout we decide on), they will never see changes list. This pull request does not have that issue, and also allows for change tracking for later stats usage as @wasade suggested.

qiita_db/metadata_template/sample_template.py
squirrelo: How do we allow people to use data from a public study but pick against a new reference database then?

qiita_db/metadata_template/base_metadata_template.py
squirrelo: done

antgonza:
Can you open an issue about this (redis timeout)? Also, there are real errors but I'm not sure if they are due to changes here or other dependency errors, could you take a look?

wasade:
is that an issue: do we want the changes in the list to exist forever? if yes, then those data should be placed in a persistent store. if no, then we just need to decide on a reasonable time frame.

squirrelo:
Shit, the issue here is that I need to import redis connection for testing, but that starts importing MOI before the ipython is spooled up. Everyone OK with starting a new redis connection for testing only?

antgonza:
I think a split of environment_manager.py into 2 should be fine (ipython and redis), perhaps a folder with two files. @wasade, what do you think?

JWDebelius:
If I can weigh in, it's easy for me to imagine that someone might be an occasional user and may not log in within the 3 month timeframe, but might still need to be kept aware of the changes. It's not hard to imagine that one last experiment needed before you can re-analyze your data taking way longer than expected, and suddenly, there being changes to the data four months later that might entirely change your results.

You may also want to plan for groups who need the long-term ability to track what is happening with their data. Losing change logs might affect people using analyses for Intellectual Property applications, the forensics community, or hopefully eventually, if Qiita gets adapted to be part of medical records. If metadata is as critical to analyses as seems to be the case, data integrity means that changes to metadata need to be logged carefully, and that log should be maintained.

josenavas:
I'm not against doing some general bookkeeping of all the changes, but if we do that, it has to be a well though centralized system were we are keeping track of all the changes.

What is happening here is some duct tape that works for some parts of the system but not for everything. Personally, I'm already heating ourselves from the past given all the code that I've been rewriting lately, and although I'm getting exited with the new Qiita as it is getting way more robust, I'm really hesitant to introduce more duct tape code in it...

PR #1180, opened 2015-05-14
ENH: Bring email validation closer to specs

### LINKED ISSUE #1179: Email validation is too strict
Just saw this error on the server (actual email censored, but it contains undrescores):
```

    callback()

  File "/home/qiita/qiita_main/qiita_pet/handlers/auth_handlers.py", line 39, in post
    created = User.create(username, password, info)

  File "/home/qiita/qiita_main/qiita_db/user.py", line 196, in create
    raise IncorrectEmailError("Bad email given: %s" % email)

IncorrectEmailError: Bad email given: email_with_underscores@hotmail.com
```
This is because of [`User.validate_email`](https://github.com/biocore/qiita/blob/master/qiita_db/user.py#L459-L487), which disallows underscores, even though they should be allowed.

Additionally, line 487
```python
return True if match(pattern, email) is not None else False
```
can be simplified to
```python
return match(pattern, email) is not None
```

adamrp:
Fix #1179

This doesn't fully account for all valid email addresses (doing so is too
much of a headache to implement, at least until it becomes necessary...), but
it's closer, and will pass for the majority of email addresses.

### DIFF
--- a/qiita_db/test/test_user.py
+++ b/qiita_db/test/test_user.py
@@ -180,7 +180,7 @@ class UserTest(TestCase):
     def test_exists_notindb(self):
         self.assertFalse(User.exists("notexist@foo.bar"))
 
-    def test_exists_invaid_email(self):
+    def test_exists_invalid_email(self):
         with self.assertRaises(IncorrectEmailError):
             User.exists("notanemail.@badformat")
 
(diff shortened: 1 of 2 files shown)


wasade:
test failures, I think :+1: otherwise tho

adamrp:
I think the only failures are the same two that are popping up on all our PRs since pandas was updated. I have a PR in for one of them #1182, but not sure about the other one :frowning:

squirrelo:
:+1:  once the failures are worked out

adamrp:
Merge?

PR #1181, opened 2015-05-14
BUG: Check that result is not None before indexing

### LINKED ISSUE #1178: User.verify_code throws exception when user does not exist
[these lines](https://github.com/biocore/qiita/blob/master/qiita_db/user.py#L272-L275) select the verification code that's stored in the database for the user so that it can be compared with the verification code supplied by the user on a password reset or user creation. Line 275 assumes that the user exists and throws an exception if it does not; we are seeing this error on the server:
```

    callback()

  File "/home/qiita/qiita_main/qiita_pet/handlers/auth_handlers.py", line 66, in get
    if User.verify_code(email, code, "create"):

  File "/home/qiita/qiita_main/qiita_db/user.py", line 275, in verify_code
    db_code = conn_handler.execute_fetchone(sql, (email,))[0]

TypeError: 'NoneType' object has no attribute '__getitem__'
```
We should check whether the `execute_fetchone` actually returned anything before trying to access index 0.

adamrp:
Fix #1178

### DIFF
--- a/qiita_db/user.py
+++ b/qiita_db/user.py
@@ -272,7 +272,15 @@ class User(QiitaObject):
         sql = ("SELECT {1} from qiita.{0} where email"
                " = %s".format(cls._table, column))
         conn_handler = SQLConnectionHandler()
-        db_code = conn_handler.execute_fetchone(sql, (email,))[0]
+        db_code = conn_handler.execute_fetchone(sql, (email,))
+
+        # If the query didn't return anything, then there's no way the code
+        # can match
+        if db_code is None:
+            return False
+
+        db_code = db_code[0]
+
         if db_code == code and code_type == "create":
             # verify the user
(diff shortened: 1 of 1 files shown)


josenavas:
:+1:

qiita_db/user.py
```diff
@@ -272,7 +272,13 @@ def verify_code(cls, email, code, code_type):
         sql = ("SELECT {1} from qiita.{0} where email"
                " = %s".format(cls._table, column))
         conn_handler = SQLConnectionHandler()
-        db_code = conn_handler.execute_fetchone(sql, (email,))[0]
+        db_code = conn_handler.execute_fetchone(sql, (email,))
+
+        if db_code is None:
+            return False
+        else:
```
wasade: `else` isn't necessary, but non blocking

adamrp:
Should be good to go

PR #1182, opened 2015-05-14
TST: Fix warnings

adamrp:
The message of the warning could also be checked, if desired, but previously
it was not being checked explicitly.

### DIFF
--- a/qiita_db/metadata_template/test/test_prep_template.py
+++ b/qiita_db/metadata_template/test/test_prep_template.py
@@ -12,6 +12,7 @@ from tempfile import mkstemp
 from os import close, remove
 from os.path import join, basename
 from collections import Iterable
+import warnings
 
 import numpy.testing as npt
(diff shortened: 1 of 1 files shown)


adamrp:
There are other places where `npt.assert_warns` is used, but so far none of those places is posing a problem (since the "desired" warning is the first one that is raised, apparently). Do we want to go ahead and replace all the calls to `npt.assert_warns` with a construct similar to what's here?

ElDeveloper:
I'm not sure I like this approach, in general our code should not
produce warnings and if we are testing for a warning in specific, then
that's the only warning that should be raised in that test case.

adamrp:
You're right, I'll try to figure out if it's something in our code that's
eliciting the futurewarning from pandas. But in the meantime, perhaps we
should consider this aproach as a stop gap so that our builds pass and we
can gain added confidence merging new code?

PR #1183, opened 2015-05-15
Cast index to array to avoid pandas warning

adamrp:
This fixes the pandas warning in a more elegant way than #1182

### DIFF
--- a/qiita_db/metadata_template/util.py
+++ b/qiita_db/metadata_template/util.py
@@ -113,7 +113,7 @@ def prefix_sample_names_with_id(md_template, study_id):
         # Create a new column on the metadata template that includes the
         # metadata template indexes prefixed with the study id
         md_template['sample_name_with_id'] = (study_ids + '.' +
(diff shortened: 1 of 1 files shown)


qiita_db/metadata_template/util.py
```diff
@@ -113,7 +113,7 @@ def prefix_sample_names_with_id(md_template, study_id):
         # Create a new column on the metadata template that includes the
         # metadata template indexes prefixed with the study id
         md_template['sample_name_with_id'] = (study_ids + '.' +
-                                              md_template.index)
+                                              np.array(md_template.index))
```
ElDeveloper: `md_template.index.values` should just return the underlying numpy array, without having to create a copy.

squirrelo:
:+1:

squirrelo:
Pulled into  #1186 so we can make sure all tests pass and have full green before merge

adamrp:
Thanks @squirrelo

PR #1186, opened 2015-05-15
Cast index to array to avoid pandas warning, and fix file checksum

squirrelo:
The expected file had messy floats, but the newest pandas fixes that, so the test file was changed to reflect this, and the checksum updated and verified on the test file.


qiita_db/metadata_template/test/test_prep_template.py
```diff
@@ -1311,10 +1311,10 @@ def test_check_restrictions(self):
...
                               self.test_study, self.data_type)
 
-        obs = pt.check_restrictions(
-            [PREP_TEMPLATE_COLUMNS['EBI'],
-             PREP_TEMPLATE_COLUMNS_TARGET_GENE['demultiplex']])
-        self.assertEqual(obs, {'primer'})
+        # obs = pt.check_restrictions(
+        #     [PREP_TEMPLATE_COLUMNS['EBI'],
+        #      PREP_TEMPLATE_COLUMNS_TARGET_GENE['demultiplex']])
+        # self.assertEqual(obs, {'primer'})
```
adamrp: I don't think you want these changes in here

adamrp:
:+1: if tests pass

ElDeveloper:
:+1: ... though I'm kinda worried that this is only going to work for newer versions of pandas, should we also change setup.py and qiita_test_install to force to require this version? That way we shouldn't encounter this problem ... I hope.

adamrp:
Good point @ElDeveloper. Yes, I think we should.

squirrelo:
Weirdly, I just went back to pandas 0.15.2 and the tests passed. I have no idea why, but it does.

squirrelo:
GREEN!

adamrp:
Sasquatch bug... I think we should probably merge this now, but investigate what the minimum version of pandas is that we should require.

antgonza:
My guess is that is passes because you also fixed the messed up numbers ...

Issue #1161, opened 2015-05-06
Custom parameters on BIOM table


ElDeveloper:
When users upload a BIOM table, they should be able to indicate what parameters were used to process the BIOM table, in the case where this parameters are not available in the database, should the new parameters be added to the database by the user? The downside to this is that those parameters would become available for users to process new sequences, is this OK or do we want to keep this separate?

Related to #1085.

adamrp:
When we discussed this previously, we decided that we were going to provide and curate the processing parameters that users could use to process (or preprocess, for that matter) their data. I think allowing the parameter sets to be added to the database by arbitrary users circumvents this regulation.

IMHO, users should be able to upload their biom tables for comparison with other data in Qiita, but such data should only be sandbox. This would provide a convenient way for users to look at their data in the context of other studies and data in Qiita, but we would maintain the standardization of processing parameters for studies that we curate (i.e., public and private studies).

ElDeveloper:
> IMHO, users should be able to upload their biom tables for comparison with other data in Qiita, but such data should only be sandbox. This would provide a convenient way for users to look at their data in the context of other studies and data in Qiita, but we would maintain the standardization of processing parameters for studies that we curate (i.e., public and private studies).

You are correct, this is exactly what the different stages (sandboxed,
private, etc) allows users to do.

josenavas:
We should be careful with this. Are we talking in terms of our own site or Qiita in general? Note the issues we had in the code sprint and the feedback that we got from @gregcaporaso . We want the system to be installed by the users and use it locally. In order to do that, the system should allow to add any arbitrary set of parameters. However, if we don't want this behavior in our system, we should put other mechanisms in place, limiting the general functionality, rather than limiting the entire system to our specific use case...

adamrp:
@josenavas you are right. This is getting pretty complex, can we enumerate the actual use cases that we are trying to support?

1. **Single user, local installation**:  In this case, the fact that everything is "sandbox" does not matter, AFAIK
2. **Multi-user, "server" installation**: E.g., our deployment, the Broad's deployment, etc.
3. **Multi-user, local installation**: Is this *actually* a use case? I don't really know why one would want to do this.

[...]

PR #1188, opened 2015-05-18
DO NOT MERGE: fix #1026

### LINKED ISSUE #1026: Allow (pre)processing multiple times
If you have preprocessed a prep template or processed a preprocessed data, you cannot re-(pre)process them. This is a limitation on the GUI, not in the database.

antgonza:
This allows the user to process the same data multiple times. Note that: 
(1) This allows to pick against the same reference multiple times, not sure if we should prevent this.
(2) Added the processing info into the preprocessing page. Not really happy with the display so open to suggestions.

### DIFF
--- a/qiita_pet/uimodules/preprocessed_data_tab.py
+++ b/qiita_pet/uimodules/preprocessed_data_tab.py
@@ -6,13 +6,15 @@
 # The full license is in the file LICENSE, distributed with this software.
 # -----------------------------------------------------------------------------
 
-from qiita_db.data import PreprocessedData
+from qiita_db.data import PreprocessedData, ProcessedData
 from qiita_db.metadata_template import PrepTemplate
 from qiita_db.ontology import Ontology
 from qiita_db.util import convert_to_id
 from qiita_db.parameters import ProcessedSortmernaParams
 from .base_uimodule import BaseUIModule
 from qiita_pet.util import generate_param_str, STATUS_STYLER
+
+from future.utils import viewitems
 
 
 class PreprocessedDataTab(BaseUIModule):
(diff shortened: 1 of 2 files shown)


qiita_pet/uimodules/preprocessed_data_tab.py
```diff
@@ -40,7 +42,11 @@ def render(self, study_id, preprocessed_data):
         is_local_request = self._is_local()
         show_ebi_btn = user.level == "admin"
         processing_status = preprocessed_data.processing_status
-        processed_data = preprocessed_data.processed_data
+        processed_data = [(id_,
```
wasade: this is pretty confusing to read, possible to break it up?

qiita_pet/uimodules/preprocessed_data_tab.py
```diff
@@ -40,7 +42,11 @@ def render(self, study_id, preprocessed_data):
...
+                           '<br/>'.join(['<b>%s:</b> %s' % (k, v) for k, v in
```
wasade: minor, but the formatting will implicitly split up the tuple so the expansion of `(k, v)` isn't necessary

qiita_pet/uimodules/preprocessed_data_tab.py
antgonza: I can, do you have suggestions? I decided to left it like this to avoid multiple separate loops ...

wasade: hows something like this?

```python
infos = []
fmt = '<b>%s:</b %s'
for id_ in preprocessed_data.processed_data:
    info = ProcessedData(id_).processing_info
    infos.append((id_, '<br/>'.join(fmt % kv for kv in view_items(info)))
```

josenavas:
Thanks for working on this @antgonza 

I don't agree on having the processing information in there. Note that such information belongs to the processed data, not to the preprocessed data, so it is confusing having it in there. The idea of having the processed data id to be a link was to direct the user to the tab so they can check that information in there.

I would suggest moving that information to the processed data tab, and leave the preprocessed data tab just with a list of links. Note that this is the approach used in the prep template.

Also, how are you storing the processing status? If I understand correctly, the database only allows to store 1 processing status at a time. The user can potentially start 2 different processings (let's say against GG and against Silva) but only one status will be stored. Furthermore, if they fail, the user will be only able to see one of the errors. I think this requires some changes to the DB, so this issue is bigger that we thought at the beginning.

Re your first point: I would suggest that we need to prevent processing two times using the exact same parameters set, as it won't change the results, so we are wasting compute time. However, if the user tries to process using different parameters agains the same reference, then we should allow this (I think right now we have only one parameter set, but see #1161)

PR #1191, opened 2015-05-19
DOC/ENH: Add base documentation setup

### LINKED ISSUE #179: Create user manual
A document that explains how to use the system.

ElDeveloper:
This pull request adds an initial setup of Sphinx for Qiita.

To build the documentation, make sure you have installed `sphinx` and
`sphinx-bootstrap-theme`, if you haven't, proceed to:

```bash
pip install sphinx sphinx-bootstrap-theme

make -C doc html

open doc/build/html/index.html
```

This PR mainly ports over the existing documents that were scattered between
wiki entries, google docs, etc. In the future we should formalize what needs to
be contained in these documents and maybe even sprint on documentation
additions.

The documentation is now built as part of the travis build.

Fixes #179


antgonza:
This looks great, thanks for putting together. Now, is the plan to completely remove the wiki?

ElDeveloper:
I think that would be a good way to go. FWIW, we can continue using the
wiki for workshop or help-specific topics and if these documents grow
into something that was useful for the general audience we can easily
migrate them into rst using pandoc.

The main idea (I think), is that with this documentation, we can add new
features to the system, and add new sections in the documentation that
explain how to use these new features. That way these documentation
additions can also be reviewed as part of the PR process. We probably
should talk more about this at our next qiita meeting.

antgonza:
Could you remove any meta-analysis mention? Remember that we decided to only say Analysis. Also, could you add an issue so we add this section?

antgonza:
Sorry, forgot to mention. Could you add in the contribution.md file that if you are modifying the interface you should also update the documentation? Thanks.

ElDeveloper:
@antgonza, I think I've addressed both of your suggestions now. Note
that the README.md in doc, is heavily based on the one that exists in
scikit-bio/doc.

antgonza:
Thanks! :+1:

ElDeveloper:
The only outstanding thing remaining with this PR, is to figure out how do we want to serve the documentation to users. Alternatives:

- Serve through read the docs.
- Serve through github pages.
- Serve from the static folder.
- Something else ...

adamrp:
I can see this is the user documentation, but I can't tell at a quick glance if this also includes developer documentation (i.e., nicely rendered numpy-doc strings in a browsable format, etc.)

I think we should definitely serve user documentation from the static folder, but I'm not sure about developer documentation.

ElDeveloper:
This should be ready for review/merge. Be sure to try it out on your own system!

josenavas:
@ElDeveloper checking the output of travis, the actual building of the documentation failed, so I'm unsure why travis is reporting that the build was successful. Can you take a look?

scripts/qiita
```diff
@@ -59,9 +62,24 @@ def pet():
     pass
 
 
-@pet.group()
-def webserver():
-    pass
+@pet.group(invoke_without_command=True)
```
josenavas: Should this instead be part of the 'start' command? IMO it's a bit weird that we are actually calling a group rather than the command.

ElDeveloper: The reason why I didn't make it part of the start command, is because
you would not be able to just build the documentation if you wanted to,
you would be forced to start the server every time you wanted to build
the docs. I'm open to suggestions on this end though.

josenavas: Just to make sure. If I then run `qiita pet webserver start` will build the docs and start the server? and then `qiita pet webserver start --no-build-docs` will only start the server?

ElDeveloper: Yes, you are correct!

ElDeveloper:
> @ElDeveloper checking the output of travis, the actual building of the documentation failed, so I'm unsure why travis is reporting that the build was successful. Can you take a look?

Good catch, thanks I just pushed a fix!

scripts/qiita
josenavas: Ok, in that case I like this approach. click is actually really awesome!

ElDeveloper: Yeah, it is extremely flexible!

ElDeveloper:
Tests are finally passing :checkered_flag: should be good to review/merge now!

josenavas:
It built in my system without a problem! Thanks for working on this!

I took a really quick look into it and I don't see any major issues. :+1:

ElDeveloper:
Excellent, before I forget!!!!!! I noticed that if you install Qiita in
a location where you don't have permissions, to build the documentation
you are going to have to do `sudo qiita pet webserver` which might get
annoying, but I really don't see a way out of this.

josenavas:
I think that is fine.

squirrelo:
Concern with running `sudo qiita pet webserver` is that you are then running the webserver as root, which is a big security concern.

josenavas:
we can ask the user to run `sudo qiita pet webserver` to create the documentation. And then use `qiita pet webserver start` to start the webserver not as root.

josenavas:
What I meant is that should be in the installation instructions.

antgonza:
Yes, I agree on having those instructions. Note that this is an specific case where you install qiita in a certain restricted location but the user that is starting the service can not write in there, which can occur but is strange ...

josenavas:
@ElDeveloper can you add `qiita_pet/static/doc/` to the .gitignore? This way it doesn't get added to the repo by mistake...

ElDeveloper:
I just added some notes to INSTALL.md

josenavas:
Thanks @ElDeveloper :+1:

antgonza:
Everything seems fine, thanks for working on this. However, there is still one thing missing. The system was updated (how a prep template, raw data is created and the preprocessing screenshots) and this documentation doesn't reflect those changes, could you fix?

ElDeveloper:
Good catch, should be done now!

ElDeveloper:
@antgonza I've your comments.

antgonza:
Errors ...

ElDeveloper:
Should be fixed now.

PR #1192, opened 2015-05-19
Add mode to call to open_file

adamrp:


josenavas:
:+1: Thanks @adamrp !!

adamrp:
One more reviewer? Should be quick. If you want to test, you can use a sample template (or prep template; any template that is loaded with `load_template_to_dataframe`) that uses carriage returns (byte 0x0d) instead of linefeeds (0x0a) for the newline representation.

PR #1198, opened 2015-05-20
Check for non-utf8 characters. Fix #1197

### LINKED ISSUE #1197: UTF-8 encoding errors
When non UTF-8 readable characters are inserted into the prep or sample template, the SQL insertion of the data goes haywire and returns an ugly SQL error.

squirrelo:
Does what it says on the tin. Also raises meaningful error with location of cells that have non-utf8 characters.

### DIFF
--- a/qiita_db/metadata_template/util.py
+++ b/qiita_db/metadata_template/util.py
@@ -9,7 +9,6 @@
 from __future__ import division
 from future.utils import PY3
 from future.utils.six import StringIO
-from csv import reader
 
 import pandas as pd
 import numpy as np
--- a/qiita_db/metadata_template/util.py
+++ b/qiita_db/metadata_template/util.py
@@ -141,13 +141,15 @@ def load_template_to_dataframe(fn, strip_whitespace=True):
     Raises
     ------
     ValueError
-        Empty file passed, or non UTF-8 characters are found in the file.
+        Empty file passed
     QiitaDBColumnError
         If the sample_name column is not present in the template.
         If there's a value in one of the reserved columns that cannot be cast
         to the needed type.
     QiitaDBWarning
         When columns are dropped because they have no content for any sample.
+    QiitaDBError
+        When non UTF-8 characters are found in the file.
 
     Notes
     -----
@@ -242,8 +244,8 @@ def load_template_to_dataframe(fn, strip_whitespace=True):
                     cell.encode('utf-8')
                 except UnicodeError:
                     errors.append('row %d, header %s' % (row, headers[col]))
-        raise ValueError('Non UTF-8 characters found at ' +
-                         '; '.join(errors))
+        raise QiitaDBError('Non UTF-8 characters found at ' +
+                           '; '.join(errors))
 
     # let pandas infer the dtypes of these columns, if the inference is
     # not correct, then we have to raise an error
--- a/qiita_db/metadata_template/util.py
+++ b/qiita_db/metadata_template/util.py
@@ -9,6 +9,7 @@
 from __future__ import division
 from future.utils import PY3
 from future.utils.six import StringIO
+from csv import reader
 
 import pandas as pd
 import numpy as np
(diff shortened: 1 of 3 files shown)


qiita_db/metadata_template/util.py
```diff
@@ -215,21 +216,33 @@ def load_template_to_dataframe(fn, strip_whitespace=True):
...
+        # Find row number and col number for utf-8 encoding errors
+        errors = []
+        for row, line in enumerate(holdfile):
+            for col, cell in enumerate(line.split('\t')):
+                try:
+                    cell.encode('utf-8')
+                except UnicodeError:
+                    errors.append('Non-unicode value in cell at '
+                                  'row %d col %d' % (row+1, col+1))
+        raise ValueError(', '.join(errors))
```
josenavas: Can you rise one of the QiitaDB errors?

josenavas: And update the documentation of the method to reflect this change?

josenavas:
Just a small comment @squirrelo

qiita_db/metadata_template/util.py
squirrelo: None of the QiitaDB errors make much sense here though. Which would you suggest?

antgonza:
Code looks good, could you add a test?

squirrelo:
Oh, yup, good call.

qiita_db/metadata_template/util.py
josenavas: Just use QiitaDBError, at least we know that the errors is generated by us.

qiita_db/metadata_template/util.py
```diff
@@ -215,21 +215,33 @@ def load_template_to_dataframe(fn, strip_whitespace=True):
...
+                                   # collection_timestamp is not added here
+                                   'host_subject_id': str,
+                                   'description': str,
+                                   # common prep template information
+                                   'center_name': str,
+                                   'center_projct_name': str})
+    except UnicodeDecodeError:
+        # Find row number and col number for utf-8 encoding errors
+        errors = []
+        for row, line in enumerate(holdfile):
```
wasade: suggest using `enumerate(holdfile, 1)` given the use of the index below. Same on the nested forloop

qiita_db/metadata_template/util.py
```diff
@@ -215,21 +215,33 @@ def load_template_to_dataframe(fn, strip_whitespace=True):
...
+            for col, cell in enumerate(line.split('\t')):
+                try:
+                    cell.encode('utf-8')
+                except UnicodeError:
+                    errors.append('Non-unicode value in cell at '
```
wasade: This text is redundant and doesn't add anything. The message could just be stated once when raising the specific exception below

qiita_db/metadata_template/util.py
```diff
@@ -215,21 +215,33 @@ def load_template_to_dataframe(fn, strip_whitespace=True):
...
+                                  'row %d col %d' % (row+1, col+1))
```
wasade: the col number in particular will be difficult for someone using excel, what about presenting the user `(sample_name, header)`?

qiita_db/metadata_template/util.py
squirrelo: Nice, didn't know about that. thanks.

qiita_db/metadata_template/util.py
squirrelo: Done

qiita_db/metadata_template/util.py
squirrelo: Done

squirrelo:
All comments taken care of. Ready for another round

antgonza:
:+1:

squirrelo:
Hold on, checking something

squirrelo:
Had to revert to ValueError because raising the QiitaDBError caused a 500 error on the interface side.

josenavas:
Is the QiitaDBError not catched on the interface? If that's the case, I would suggest catching it rather than rely on ValueError (I know, more work, but core more robust...)

squirrelo:
Done

josenavas:
:+1: @wasade can you review this again?

ElDeveloper:
Yargh, just noticed this PR is missing tests .... :sob:

squirrelo:
It does: qiita_db/metadata_template/test/test_util.py

ElDeveloper:
I entirely missed that. :eyes:

Issue #369, opened 2014-10-13
EBI submission scripts should be able to source from the database


wasade:
Currently expect prep/sample templates to be files. Also need support for HDF5 demux

adamrp:
To take this a step further, I think we should remove the ability to submit to EBI directly from files, and instead require that all information needed for an EBI submission be properly linked in the database. Note that the only reason it allowed submission from files in the first place was due to some urgent submissions that needed to go out before pertinent database code was in place.

josenavas:
:+1:

antgonza:
Fixed in #1451. Note that you still can submit from CLI but you can only pass the preprocessed_data_id, the action and if you really want to send.

PR #1199, opened 2015-05-20
Fixes 1194

### LINKED ISSUE #1194: TAXON_ID is EBI required
We should add this to the constraints.

We can probably try to target this jointly with #1113.

josenavas:
Fixes #1194 

"taxon_id" and "scientific_name" are required columns by EBI.
The EBI submission code now requires them instead of allowing them to be nullable.

EBI submission was uploading all the samples in the sample template. It now filters the samples to only upload the samples that are in the prep template, which is the correct thing to do (the other samples in the prep template may get uploaded in another EBI submission - i.e. different PreprocessedData).

### DIFF
--- a/qiita_db/test/test_commands.py
+++ b/qiita_db/test/test_commands.py
@@ -725,19 +725,19 @@ SAMPLE_TEMPLATE = (
     "sample_name\trequired_sample_info_status\tcollection_timestamp\t"
     "sample_type\tphysical_specimen_remaining\tphysical_specimen_location\t"
     "dna_extracted\thost_subject_id\tTreatment\tDOB\tlatitude\tlongitude"
-    "\tDescription\n"
+    "\ttaxon_id\tscientific_name\tDescription\n"
     "PC.354\treceived\t2014-06-18 16:44\ttype_1\tTrue\tLocation_1\tTrue\t"
     "HS_ID_PC.354\tControl\t20061218\t1.88401499993\t56.0003871552\t"
-    "Control_mouse_I.D._354\n"
+    "9606\thomo sapiens\tControl_mouse_I.D._354\n"
     "PC.593\treceived\t2014-06-18 16:44\ttype_1\tTrue\tLocation_1\tTrue\t"
     "HS_ID_PC.593\tControl\t20071210\t35.4079458313\t83.2595338611\t"
-    "Control_mouse_I.D._593\n"
+    "9606\thomo sapiens\tControl_mouse_I.D._593\n"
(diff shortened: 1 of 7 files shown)


qiita_db/test/test_commands.py
```diff
@@ -725,19 +725,19 @@ def test_update_preprocessed_data_from_cmd_ppd(self):
...
-    "Control_mouse_I.D._593\n"
+    "9606\thomo sapiens\tControl_mouse_I.D._593\n"
     "PC.607\treceived\t2014-06-18 16:44\ttype_1\tTrue\tLocation_1\tTrue\t"
     "HS_ID_PC.607\tFast\t20071112\t18.3175615444\t91.3713989729\t"
-    "Fasting_mouse_I.D._607\n"
+    "9606\thomo sapiens\tFasting_mouse_I.D._607\n"
     "PC.636\treceived\t2014-06-18 16:44\ttype_1\tTrue\tLocation_1\tTrue\t"
     "HS_ID_PC.636\tFast\t20080116\t31.0856060708\t4.16781143893\t"
-    "Fasting_mouse_I.D._636")
+    "9606\thomo sapiens\tFasting_mouse_I.D._636")
```
antgonza: It's fine but a homo sapiens that is a fasting mouse?

antgonza:
:+1: if tests pass ...

qiita_db/test/test_commands.py
josenavas: hahaha I know. That is the taxon id that was used in the EBI example and I didn't wanted to check in NCBI to check which one is for the mouse, in interest of time.

qiita_ware/ebi.py
```diff
@@ -280,18 +280,20 @@ def generate_study_xml(self):
...
 
         Parameters
         ----------
         sample_name : str
             Unique identifier for the sample
-        taxon_id : str, optional
-            Defaults to ``None``. If not provided, the `empty_value` will be
-            used for the taxon ID
-        description : str, optional
+        taxon_id : str
```
wasade: isn't this an int?

josenavas: EBI submission is based on the template files, rather than on the data in the DB, and everything is parsed as a string.

We can refactor the EBI submission code to actually get the information from the DB, and I think this is something that we should do. In fact we already have an issue for it: https://github.com/biocore/qiita/issues/369. I would like to work on that one, so I assigned to myself to visit it after the ASM.

wasade: okay

josenavas:
@wasade any other comment? If not, can you merge?

PR #1200, opened 2015-05-20
Using regex rather than directly check for indices in the sql args

josenavas:
The previous implementation will fail with empty strings, and other issues can be found. The current implementation is more robust.

### DIFF
--- a/qiita_db/sql_connection.py
+++ b/qiita_db/sql_connection.py
@@ -155,6 +155,8 @@ class SQLConnectionHandler(object):
                  'admin_with_database': '_admin_args',
                  'admin_without_database': '_admin_nodb_args'}
 
+    _regex = re.compile("{(\d+)}")
+
     """Encapsulates the DB connection with the Postgres DB
 
     Parameters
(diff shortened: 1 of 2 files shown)


antgonza:
Do we need a test for this?

josenavas:
@antgonza added

antgonza:
:+1: if tests pass.

josenavas:
~~regex whisperer~~ @adamrp @squirrelo @wasade any of you available for a quick review?

qiita_db/sql_connection.py
```diff
@@ -362,11 +363,12 @@ def execute_queue(self, queue):
                 if sql_args is not None:
                     for pos, arg in enumerate(sql_args):
                         # check if previous results needed and replace
-                        if isinstance(arg, str) and \
-                                arg[0] == "{" and arg[-1] == "}":
-                            result_pos = int(arg[1:-1])
-                            sql_args[pos] = results[result_pos]
-                            clear_res = True
+                        if isinstance(arg, str):
+                            result = re.search("{(\d+)}", arg)
```
squirrelo: This can probably be compiled at the class level and then called from there for speed.

squirrelo:
One comment

qiita_db/sql_connection.py

PR #1201, opened 2015-05-24
1084 db changes

josenavas:
Tests are expected to fail. Note that this is going to the branch for fixing 1084.

Makes the needed DB changes.
The DB changes in the patch could be done in SQL. However, before I can execute those changes I need to perform some DB and filesystem cleaning (i.e. moving files to the upload folder again), so that's why those changes are performed in the python patch.


qiita_db/support_files/patches/python_patches/25.py
```diff
@@ -0,0 +1,121 @@
+# May 19, 2015
+# We attach the prep template directly to the study. The raw data is no longer
+# attached to the study directly, the prep template point to them. This will
```
antgonza: point -> points

qiita_db/support_files/patches/python_patches/25.py
```diff
@@ -0,0 +1,121 @@
...
+# Retrieve all those raw data ids
+sql = """SELECT raw_data_id
+         FROM qiita.raw_data
+         WHERE raw_data_id NOT IN (
+            SELECT DISTINCT raw_data_id FROM qiita.prep_template);"""
+rd_ids = [x[0] for x in conn_handler.execute_fetchall(sql)]
+
+# We will delete those RawData. However, if they have files attached, we should
+# move them to the uploads folder of the study
+sql_detach = """DELETE FROM qiita.study_raw_data
```
antgonza: Why not use the already build unlink and delete functions?

qiita_db/support_files/patches/python_patches/25.py
josenavas: Dome

qiita_db/support_files/patches/python_patches/25.py
josenavas: The reason is that the patch will be executed with the new code. In the new code, the RawData does not know to which studies it belongs unless it is connected to a PrepTemplate. The RawData objects that are here, are "duplicating" the existing code, but once all the pieces are in place, that code is different. Does this make sense to you?

josenavas:
Thanks @antgonza! Comments addressed.

qiita_db/support_files/patches/python_patches/25.py
ElDeveloper: So this is accounting for the changes that are coming in the other open PRs right? Such that the patch will succeed to execute with all the changes to the codebase.

qiita_db/support_files/patches/python_patches/25.py
```diff
@@ -0,0 +1,121 @@
...
+for rd_id in rd_ids:
+    rd = RawData(rd_id)
+    filepaths = rd.get_filepaths()
+    studies = sorted(rd.studies)
+    if filepaths:
+        # we need to move the files to a study. We chose the one with lower
+        # study id. Currently there is no case in the live database in which a
+        # RawData with no prep templates is attached to more than one study,
+        # but I think it is better to normalize this just in case
+        move_files.append((min(studies), filepaths))
```
ElDeveloper: Just to confirm, there's no possible way we would get to this line if there are no studies right? i.e. is there any way that studies could be an empty list? If so, then `min` would error.

qiita_db/support_files/patches/python_patches/25.py
```diff
@@ -0,0 +1,121 @@
...
+    # To delete the RawData we first need to unlink all the files
+    conn_handler.add_to_queue(queue, sql_unlink, (rd_id,))
+
+    # Then, remove the raw data from all the studies
+    for st_id in studies:
+        conn_handler.add_to_queue(queue, sql_detach, (rd_id, st_id))
+
+    conn_handler.add_to_queue(queue, sql_delete, (rd_id,))
+
+# We can now perform all changes in the DB. Although this changes can be
```
ElDeveloper: this -> these

qiita_db/support_files/patches/python_patches/25.py
```diff
@@ -0,0 +1,121 @@
...
+# done in an SQL patch, they are done here because we need to execute the
+# previous clean up in the database before we can actually execute the SQL
+# patch.
+sql = """CREATE TABLE qiita.study_prep_template (
```
ElDeveloper: I'm assuming the first part of this SQL string was generated by DBSchema?

ElDeveloper:
:+1:, very minor comments.

qiita_db/support_files/patches/python_patches/25.py
josenavas: Exactly. That's why I didn't do the PR until I get the changes done in the codebase. I then just break up the entire work in multiple PR for easy review.

qiita_db/support_files/patches/python_patches/25.py
josenavas: Nope, there will be at least one study.

qiita_db/support_files/patches/python_patches/25.py
josenavas: Done

qiita_db/support_files/patches/python_patches/25.py
josenavas: Yes, you're correct, I just copied here.

josenavas:
Thanks @ElDeveloper, I've addressed the comments.

ElDeveloper:
Thanks @josenavas, :+1:

josenavas:
@antgonza >

PR #1202, opened 2015-05-24
1084 populate db

josenavas:
Built on top of #1201 (review-merge that one first).

With the database changes, populate_test_db.sql was creating an inconsistent DB: there was raw data not associated with any prep template.

Since it was a lot of work to add prep templates to add those raw data, and those raw data where mainly used in the qiita_ware tests, I removed those raw data so populate is easier to manage.

Tests are still expected to fail, since I still haven't changed any code.


antgonza:
:+1: as we can not really review, I would merge but rather wait to see someone else opinion ...

PR #1203, opened 2015-05-24
1084 prep template changes

josenavas:
Built on top of #1202 (review/merge that one first).

This PR modifies the prep template object to reflect the changes in the DB. It also fixes the tests for the prep template object.

All the other tests expected to fail.

### DIFF
--- a/qiita_db/metadata_template/prep_template.py
+++ b/qiita_db/metadata_template/prep_template.py
@@ -77,16 +77,13 @@ class PrepTemplate(MetadataTemplate):
     _filepath_table = 'prep_template_filepath'
 
     @classmethod
-    def create(cls, md_template, raw_data, study, data_type,
-               investigation_type=None):
+    def create(cls, md_template, study, data_type, investigation_type=None):
         r"""Creates the metadata template in the database
 
         Parameters
         ----------
         md_template : DataFrame
             The metadata template file contents indexed by samples Ids
-        raw_data : RawData
-            The raw_data to which the prep template belongs to.
         study : Study
             The study to which the prep template belongs to.
         data_type : str or int
@@ -111,7 +108,7 @@ class PrepTemplate(MetadataTemplate):
 
         # Get a connection handler
         conn_handler = SQLConnectionHandler()
-        queue_name = "CREATE_PREP_TEMPLATE_%d" % raw_data.id
+        queue_name = "CREATE_PREP_TEMPLATE_%d" % study.id
         conn_handler.create_queue(queue_name)
 
         # Check if the data_type is the id or the string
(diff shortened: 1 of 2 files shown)


qiita_db/metadata_template/prep_template.py
```diff
@@ -111,7 +108,7 @@ def create(cls, md_template, raw_data, study, data_type,
 
         # Get a connection handler
         conn_handler = SQLConnectionHandler()
-        queue_name = "CREATE_PREP_TEMPLATE_%d" % raw_data.id
+        queue_name = "CREATE_PREP_TEMPLATE_%d" % study.id
```
antgonza: As a study can have several prep templates, this could cause collisions, could you change? Perhaps study.id + some random string.

qiita_db/metadata_template/test/test_prep_template.py
```diff
@@ -1308,14 +1308,34 @@ def test_check_restrictions(self):
...
+        with self.assertRaises(QiitaDBError):
+            self.tester.raw_data = rd
+
+    def test_raw_data_setter(self):
+        rd = RawData(1)
+        pt = PrepTemplate.create(self.metadata, self.test_study,
+                                 self.data_type_id)
+        self.assertEqual(pt.raw_data, None)
+        pt.raw_data = rd
+        self.assertEqual(pt.raw_data, 1)
```
antgonza: Should this be `pt.raw_data, rd.id)`?

antgonza:
A couple of comments.

qiita_db/metadata_template/prep_template.py
wasade: These are just the queues though. Given that tornado is single threaded, is it possible for this block of code to be executed twice without the first queue draining?

antgonza: Good point but this will not be the case if we allow for multi-threaded tornado or if we move this section to run as a separate IPython job. Better safe than sorry, no?

wasade: What would the benefit be of using a thread pool vs. just running multiple web servers? Threading adds a _lot_ of complexity, and I doubt qiita is thread-safe anyway. IPython jobs do not have shared memory, so I do not understand your concern there. Doing something random because it "seems" safer does not inherently make it safer.

antgonza: If you do not feel that is important and @josenavas thinks is fine, let's leave it as is ...

wasade: It isn't a feeling as there is a logical conclusion here, right?

josenavas: Agree that is not likely to happen, however it is really easy to make it unique, so fixing because I think it can be a pain to try to debug an issue with this:

```python
queue_name = "CREATE_PREP_TEMPLATE_%d_%d" % (study.id, id(md_template))
```

qiita_db/metadata_template/test/test_prep_template.py
josenavas: Note that the RawData id is hard-coded above: `rd = RawData(1)`. However, if this ever changes, by using rd.id it will simplify future changes, so changing. Thanks!

josenavas:
Thanks @antgonza , addressed your comments. One more reviewer so this can get merged?

antgonza:
:+1:

ElDeveloper:
:+1:

PR #1204, opened 2015-05-24
1084 data changes

josenavas:
~~Built on top of #1203 (review/merge that one first)~~

This PR modifies the RawData object to reflect the changes in the DB. It also fixes the tests for the RawData object.

All the other tests are still expected to fail.


qiita_db/data.py
```diff
@@ -288,29 +288,48 @@ class RawData(BaseData):
...
+        sql = """SELECT EXISTS(
+                    SELECT *
+                    FROM qiita.prep_template
+                    WHERE prep_template_id IN ({})
+                        AND raw_data_id IS NOT NULL)""".format(
+            ', '.join(['%s'] * len(prep_templates)))
+        exists = conn_handler.execute_fetchone(
+            sql, [pt.id for pt in prep_templates])[0]
+        if exists:
+            raise IncompetentQiitaDeveloperError(
```
antgonza: Just checking, is it really an IncompetentQiitaDeveloperError or a QiitaError? I'm thinking than a user can hit this error from IPython.

antgonza:
1 minor comment so :+1:

qiita_db/data.py
```diff
@@ -288,29 +288,48 @@ class RawData(BaseData):
...
         A new instance of `cls` to access to the RawData stored in the DB
+
+        Raises
+        ------
+        IncompetentQiitaDeveloperError
+            If any of the passed prep templates already have a raw data id
         """
-        # Add the raw data to the database, and get the raw data id back
         conn_handler = SQLConnectionHandler()
+        # We first need to check if the passed prep templates doesn't have
```
ElDeveloper: doesn't -> don't

qiita_db/data.py
```diff
@@ -576,40 +583,6 @@ def clear_filepaths(self):
         # self.studies should only have one element, thus self.studies[0]
         move_filepaths_to_upload_folder(self.studies[0], filepaths)
 
-    def remove_filepath(self, fp):
```
ElDeveloper: Was this method not used anywhere else? That's surprising :eyes:

ElDeveloper:
Very minor comments, should be good to go after.

qiita_db/data.py

qiita_db/data.py
josenavas: Good call, done

qiita_db/data.py
josenavas: Yup, it was not used, sorry I forgot to comment that on the PR. I checked and it was not used. This function was intended to remove a single filepath from the raw data, but we are not supporting right now because it gets tricky: we are checking if the raw data has the same number of barcode files and forward reads, so removing a single file will break this checks. Thus, we are allowing just removing all using `clear_filepaths`

Also, this function was using `purge_filepaths` instead of `move_filepaths_to_upload_folder`.

Given that it was not used, and I don't see a use for it soon; and it's a developer burden to maintain it, I think is better to remove it and we will re-add it if we actually need it.

josenavas:
Comments should be addressed @antgonza @ElDeveloper

ElDeveloper:
Got it, thanks! :+1: ready for merge.

PR #1205, opened 2015-05-24
1084 qiita db

josenavas:
Built on top of #1204 

Makes qiita_db to pass al tests. Still failures expected on qiita-ware and qiita-pet.

### DIFF
--- a/qiita_db/commands.py
+++ b/qiita_db/commands.py
@@ -153,7 +153,7 @@ def load_sample_template_from_cmd(sample_temp_path, study_id):
     return SampleTemplate.create(sample_temp, Study(study_id))
 
 
-def load_prep_template_from_cmd(prep_temp_path, raw_data_id, study_id,
+def load_prep_template_from_cmd(prep_temp_path, study_id,
                                 data_type):
     r"""Adds a prep template to the database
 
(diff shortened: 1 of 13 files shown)


qiita_db/commands.py
```diff
@@ -153,7 +153,7 @@ def load_sample_template_from_cmd(sample_temp_path, study_id):
     return SampleTemplate.create(sample_temp, Study(study_id))
 
 
-def load_prep_template_from_cmd(prep_temp_path, raw_data_id, study_id,
+def load_prep_template_from_cmd(prep_temp_path, study_id,
                                 data_type):
```
antgonza: you can bring this line up, no?

antgonza:
1 minor comment :+1:

josenavas:
Done!

qiita_db/test/test_study.py
```diff
@@ -586,26 +585,20 @@ def test_retrieve_data_types_none(self):
         self.assertEqual(new.data_types, [])
 
     def test_retrieve_raw_data(self):
-        self.assertEqual(self.study.raw_data(), [1, 2, 3, 4])
+        self.assertEqual(self.study.raw_data(), [1])
```
ElDeveloper: There used to be four raw datas associated to this study but now there's only one? How would that happen?

ElDeveloper:
1 question, but :+1: (for the most part :100: ).

qiita_db/test/test_study.py
josenavas: I removed them in populate_test_db.sql. Keeping them would require adding 3 more prep templates there, which is a lot of work...

ElDeveloper: Thanks!

PR #1206, opened 2015-05-24
1084 qiita ware

josenavas:
Built on top of #1205 

Fixes all tests for qiita-ware. Failures still expected on qiita-pet.

Issue #1193, opened 2015-05-19
All samples in the sample template are shown when a processed data is public


josenavas:
Once a processed data of a study becomes public, the entire sample template becomes public. In order to keep the data privacy settings of the users, only the samples present on the prep templates made public should be presented to the user.

This is not a quick fix, as multiple things will need to be performed on the fly to solve the issue. I don't think this has to be done by ASM, but shortly after, so I'm going to assign it to myself as I already have an idea on how to solve it, but I'm going to work on it after ASM.

Thanks @squirrelo for finding the issue!

Issue #1145, opened 2015-05-02
Move artifact creation/move away form the handlers


josenavas:
No processing should be done in the handlers. Instead, we should create functions that returns the parameters for rendering the interface. This will allow to add tests for those lines.

ElDeveloper:
:+1:

wasade:
:+1:

antgonza:
@josenavas can we close this?

josenavas:
No, there is still a lot of work happening on the handlers. This come up in a recent discussion with @squirrelo regarding his new refactor, but I couldn't find the issue then. Even if with the new refactor the work is reduced, there is still a lot of other handlers that will not be modified and need to be refactored to reduce the amount of work.

antgonza:
@josenavas, has this been solved?

josenavas:
Not fully - I think there are still places where the work can be reduced.

antgonza:
OK, can you list them so we know how close are we? Thanks!

antgonza:
At this point, I think the only processing still happening on handlers is the artifact creation/move. Changing the name.

antgonza:
This was solved some time ago, closing.

PR #1207, opened 2015-05-24
1084 qiita pet

josenavas:
Built on top of #1206 

This PR refactors the interface so now the user has to upload first the prep template before he can add the raw data. This is how the database works and this is the normal workflow when using Qiita.

This is still a work in progress PR. I had to completely rewrite some parts of the interface code, and I've still not removed the old code, as it was useful for reference.

It would be great if people can pull down this branch and check which things are missing in the interface, and which things can be improved. I'll be able to work on it tonight and tomorrow morning, so the entire interface is consistent with the new DB layout. After that, I will be able to finally add the QIIME mapping file parsing, which shouldn't impact the interface and it's only a backend change!


josenavas:
Tests should pass now in this PR. Ready for review once all the previous PR have been reviewed and merged.

josenavas:
@biocore/qiita-dev In this PR I've changed the interface a lot. I would suggest first downloading the code and check the interface before actually doing the code review. The idea is that if the interface needs to change, we don't waste time reviewing code that is going to change.

ElDeveloper:
Thanks for the heads up, will do!

qiita_db/test/test_commands.py
```diff
@@ -563,7 +563,7 @@ def test_update_preprocessed_data_from_cmd(self):
         # We need to sort the list returned from the db because the ordering
         # on that list is based on db modification time, rather than id
         obs_fps = sorted(ppd.get_filepaths())
-        self.assertEqual(obs_fps, exp_fps)
+        self.assertEqual(obs_fps, sorted(exp_fps))
```
squirrelo: 'assertItemsEqual' would be better here, since it would eliminate the sorting.

qiita_db/test/test_data.py
```diff
@@ -545,7 +550,7 @@ def test_get_filepaths(self):
                 "preprocessed_fastq"),
                (5, join(self.db_test_ppd_dir, '1_seqs.demux'),
                 "preprocessed_demux")]
-        self.assertEqual(obs, exp)
+        self.assertEqual(sorted(obs), sorted(exp))
```
squirrelo: Same as above

qiita_pet/templates/study_description_templates/prep_template_info_tab.html
```diff
@@ -0,0 +1,196 @@
...
+              , <a href="#" onClick="display_tab('preprocessed_data_tab', {{pdi}});">{{pdi}}</a>
+            {% end %}
+          {% end %}
+        </td>
+      </tr>
+    {% else %}
+      <!-- There is no raw data associated with the prep template, show options for adding raw data -->
+      <tr>
+        <td style="padding: 5px;">
+          <button class="btn btn-primary glyphicon glyphicon-plus" type="button" data-toggle="collapse" data-target="#add-raw-data-{{pt_id}}" aria-expanded="false" aria-controls="add-raw-data-{{pt_id}}" style="word-spacing: -10px;">Add raw data</button>
```
squirrelo: Can there be more than one association of raw data per prep template? I thought we agreed it was a 1 to 1 thing, so why have this button here? It should just show the accordion of add new or add from existing, no need for this extra step.

qiita_pet/templates/study_description_templates/raw_data_info.html
```diff
@@ -0,0 +1,56 @@
+<div>
+  <div class="panel-group" id="accordion-rd-info-{{rd_id}}-{{prep_template_id}}" role="tablist" aria-multiselectable="true">
+    <div class="panel panel-default">
+      <div class="panel-heading" role="tab" id="rd-info-{{rd_id}}-{{prep_template_id}}">
+        <h4 class="panel-title">
+          <a class="collapsed" data-toggle="collapse" data-parent="#accordion-rd-info-{{rd_id}}-{{prep_template_id}}" href="#collapse-rd-info-{{rd_id}}-{{prep_template_id}}" aria-expanded="false" aria-controls="collapse-rd-info-{{rd_id}}-{{prep_template_id}}">
+            Raw data - {{rd_filetype}} (ID: {{rd_id}})
```
squirrelo: This really doesn't need to be an accordion, we can just show the files in a table by type.

squirrelo:
Few comments.

ElDeveloper:
The code seems to work just fine i. e. I was able to create a study and process the sequences up to an OTU table.

qiita_db/test/test_data.py
```diff
@@ -261,6 +261,11 @@ def test_delete(self):
             RawData.delete(rd.id, self.pt1.id)
 
         # Clear the files so we can actually remove the RawData
+        study_id = rd.studies[0]
+        path_for_removal = join(get_mountpoint("uploads")[0][1], str(study_id))
+        self._clean_up_files = [join(path_for_removal,
```
ElDeveloper: This should be `self._clean_up_files.extend([....`.

qiita_db/test/test_util.py
```diff
@@ -575,7 +575,10 @@ def test_move_upload_files_to_trash(self):
 
         # create file to move to trash
         fid, folder = get_mountpoint("uploads")[0]
-        open(join(folder, '1', test_filename), 'w').write('test')
+        test_fp = join(folder, '1', test_filename)
+        open(test_fp, 'w').write('test')
```
ElDeveloper: I know the garbage collector is supposed to get rid of these unclosed file handles, but I rather have this handle be closed explicitly (using with open maybe?).

qiita_pet/handlers/compute.py
```diff
@@ -28,6 +30,49 @@ def get(self, job_id):
...
+            return x.split(',') if x else []
+        filepaths, fps = [], []
+        fps.append((_split(barcodes_str), 'raw_barcodes'))
+        fps.append((_split(fasta_str), 'raw_fasta'))
+        fps.append((_split(qual_str), 'raw_qual'))
+        fps.append((_split(forward_reads_str), 'raw_forward_seqs'))
+        fps.append((_split(reverse_reads_str), 'raw_reverse_seqs'))
+        fps.append((_split(sff_str), 'raw_sff'))
+
+        for _, f in get_mountpoint("uploads", retrieve_all=True):
```
ElDeveloper: Would you mind adding a comment describing this block, it took me a while to understand it :eyeglasses:

qiita_pet/templates/study_description.html
```diff
@@ -111,71 +111,158 @@
...
+      .attr("type", "hidden")
+      .attr("name", "reverse")
+      .attr("value", reverse.join()));
     $("body").append(form);
     form.submit();
+  }
+}
 
+function add_raw_data(prep_id) {
+  raw_data_id = $("#previous-raw-data-" + prep_id).val()
```
ElDeveloper: Minor, but you are missing a semicolon at the end of this line.

antgonza:
This looks really nice and like @ElDeveloper said it works.

Sore general comments:
- Do we really need an '+Add raw data' button? Why not always show the 2 frames?
- If we keep the previous button, add a space between the + and the Add raw data
- In "Use existing raw data" add new lines between the dropbox and the "Add" button 
- Will it be possible to make the space bigger in the grey-add-raw-data section?

Gonna check the code now.

qiita_pet/uimodules/prep_template_tab.py
```diff
@@ -0,0 +1,273 @@
...
+    lambda: filepath_types, SFF=['sff'], FASTA=['fasta', 'qual'],
+    FASTQ=['barcodes', 'forward seqs', 'reverse seqs'])
+
+
+def get_accessible_raw_data(user):
+    """Retrieves a tuple of raw_data_id and the last study title for that
+    raw_data
+    """
+    d = {}
+    for sid in user.user_studies:
```
ElDeveloper: I would almost prefer that this loop plus the other loop were done in a single SQL query.

ElDeveloper: Thoughts?

qiita_pet/uimodules/prep_template_tab.py
```diff
@@ -0,0 +1,273 @@
...
+
+filepath_types = [k.split('_', 1)[1].replace('_', ' ')
+                  for k in get_filepath_types()
+                  if k.startswith('raw_')]
+fp_type_by_ft = defaultdict(
+    lambda: filepath_types, SFF=['sff'], FASTA=['fasta', 'qual'],
+    FASTQ=['barcodes', 'forward seqs', 'reverse seqs'])
+
+
+def get_accessible_raw_data(user):
```
ElDeveloper: This and the other helper functions should probably be prepended with an underscore in the name as they are only used in this module and not intended to be used anywhere else.

wasade: Possible to move this method, `get_raw_data`, and `get_prep_templates` to under `qiita_db`, or possibly `qiita_ware`?

qiita_pet/uimodules/prep_template_tab.py
```diff
@@ -0,0 +1,273 @@
...
+    list of PrepTemplate
+    """
+    return [PrepTemplate(pt_id) for pt_id in sorted(pt_ids)]
+
+
+class PrepTemplateTab(BaseUIModule):
+    def render(self, study, full_access):
+        files = [f for _, f in get_files_from_uploads_folders(str(study.id))]
+        data_types = sorted(viewitems(get_data_types()), key=itemgetter(1))
+        prep_templates_info = [
```
ElDeveloper: Just wanted to highlight that this list comprehension is very busy, still valid PEP8 and adheres to our coding guidelines, but it's probably too much.

qiita_pet/uimodules/prep_template_tab.py
```diff
@@ -0,0 +1,273 @@
...
+        filetypes = sorted(
+            ((ft, ft_id, fp_type_by_ft[ft])
+             for ft, ft_id in viewitems(get_filetypes())),
+            key=itemgetter(1))
+        files = [f for _, f in get_files_from_uploads_folders(str(study.id))]
+
+        other_studies_rd = sorted(viewitems(
+            get_accessible_raw_data(user)))
+
+        # A prep template can be modified if it's status is sanbdox
```
ElDeveloper: it's -> its

ElDeveloper:
The code looks good to me, :+1: after the comments are addressed.

qiita_db/test/test_commands.py
josenavas: Other checks in this tests depend on these 2 lists being in the same order. Leaving the sort.

qiita_db/test/test_data.py
josenavas: Done

qiita_pet/templates/study_description_templates/prep_template_info_tab.html
josenavas: Done

qiita_pet/templates/study_description_templates/raw_data_info.html
josenavas: As you said, it doesn't need to be an accordion. However it is convenient for multiple reasons:
 - Conceptual division of the data. This accordion makes clear that such piece of information belongs to the RawData, not necessarily to the prep tempalte
 - The user can collapse it if he doesn't want to see the RawData information
 - Last but not least, if a RawData consists of multiple files, the 'preprocess' button and any other information below the raw data can end up on the bottom of the page, so the user will need to scroll a lot to get into them (imagine a RawData in which the user has the per sample fastq...)

I think it doesn't harm anything and we get a lot of advantages of having it as an accordion, so I prefer to leave it as it is.

qiita_db/test/test_data.py
josenavas: Good catch, done

qiita_db/test/test_util.py
josenavas: Done

ElDeveloper:
When showing the `Add new raw data` and `Use existing raw data` accordions, they resize a bit oddly:

minor

Would be nice if the size was fixed.

---------

The `Add` button should probably be green instead of blue. It is also a bit confusing that the name of the button is `Add`, but the instructions are `Choose the files to link to the raw data`. However, I think this is ok for the `Use existing raw data` section.

------

This should probably just be a bootsrap table:

screen shot 2015-05-28 at 5 05 21 pm

-----

For some extra Yoshiki-points, we probably want to convert the dropdown that's shown in `Use existing raw data` into a chosen dropdown.

-----

Looks good otherwise!

qiita_pet/templates/study_description_templates/raw_data_info.html
ElDeveloper: I like the accordions, don't see why they should not be there.

qiita_pet/handlers/compute.py
josenavas: Done

qiita_pet/templates/study_description.html
josenavas: Good catch, done!

qiita_pet/uimodules/prep_template_tab.py
josenavas: I agree that we should move this away from here. However, I'm afraid of spending time in something that will be removed soon. There are some issues that make me think that:
 - One is https://github.com/biocore/qiita/issues/1145 which englobes changes like this.
 - We should provide finer control on the study elements ownership and visibility. I sent this idea in an email but we've not been discussing it. Related to https://github.com/biocore/qiita/issues/1215 and https://github.com/biocore/qiita/issues/1193.

I agree that we should improve all the code as we move forward, but I think it is not worth to spend time in moving this code around (it was already like this) if this is going to be refactored.

qiita_pet/uimodules/prep_template_tab.py
josenavas: See my comment above. If we introduce finer control over the study visibility, this kind of functions will be added as User attributes.

qiita_pet/uimodules/prep_template_tab.py
josenavas: Changed by a generator function. It actually get's rid of the `get_prep_templates` function.

qiita_pet/uimodules/prep_template_tab.py
josenavas: Done

josenavas:
@ElDeveloper thanks for the suggestions!

I want all the extra Yoshiki-points (worth :beers: ??). Weirdly enough, the `Use existing raw data` dropdown is already a chosen dropdown:
```html
<select class="chosen-select" id="previous-raw-data-{{pt_id}}" data-placeholder=" ">
```
And I cannot find why is not working properly...

qiita_pet/uimodules/prep_template_tab.py
ElDeveloper: Got it.

qiita_pet/uimodules/prep_template_tab.py
ElDeveloper: Sounds good, at least for the time being they are prepended with an
underscore, which should highlight that these are helper functions
others shouldn't depend on.

qiita_pet/uimodules/prep_template_tab.py
```diff
@@ -0,0 +1,274 @@
...
+
+filepath_types = [k.split('_', 1)[1].replace('_', ' ')
+                  for k in get_filepath_types()
+                  if k.startswith('raw_')]
+fp_type_by_ft = defaultdict(
+    lambda: filepath_types, SFF=['sff'], FASTA=['fasta', 'qual'],
+    FASTQ=['barcodes', 'forward seqs', 'reverse seqs'])
+
+
+def _get_accessible_raw_data(user):
```
ElDeveloper: Another thing ... there's no tests for these functions, which I think we are ok as this is a transitive step, but I wanted to highlight this in case someone else has stronger feelings about it.

ElDeveloper:
:beers: indeed, thanks for making the changes!

I think you need to activate the dropdown with JavaScript:
https://harvesthq.github.io/chosen/#setup

ElDeveloper:
BTW, I'm :+1: for the most part now. @squirrelo, do you have any other comments, or others? Would be great to get this merged ASAP.

squirrelo:
I'm still meh on the accordion for Raw Data, but for the sake of having this functionality in :+1: on this code and we can do interface changes later.

antgonza:
RE accordion: I actually like it.

PR #1219, opened 2015-05-28
1084 qiime map

josenavas:
Built on top of #1207 

This PR adds the ability to upload a QIIME mapping file trough the interface. When uploading a QIIME mapping file, a SampleTemplate and a PrepTemplate is created.

### DIFF
--- a/qiita_db/metadata_template/test/test_util.py
+++ b/qiita_db/metadata_template/test/test_util.py
@@ -17,7 +17,8 @@ from qiita_db.exceptions import (QiitaDBColumnError, QiitaDBWarning,
                                  QiitaDBError)
 from qiita_db.metadata_template.util import (
     get_datatypes, as_python_types, prefix_sample_names_with_id,
-    load_template_to_dataframe, get_invalid_sample_names)
+    load_template_to_dataframe, get_invalid_sample_names,
+    looks_like_qiime_mapping_file)
 
 
 class TestUtil(TestCase):
@@ -218,6 +219,45 @@ class TestUtil(TestCase):
             # prevent flake8 from complaining
             str(obs)
 
+    def test_looks_like_qiime_mapping_file(self):
+        obs = looks_like_qiime_mapping_file(
+            StringIO(EXP_SAMPLE_TEMPLATE))
+        self.assertFalse(obs)
+
+        obs = looks_like_qiime_mapping_file(
+            StringIO(QIIME_TUTORIAL_MAP))
+        self.assertTrue(obs)
+
+    def test_looks_like_qiime_mmapping_file_error(self):
+        with self.assertRaises(QiitaDBError):
+            looks_like_qiime_mapping_file(StringIO())
+
+
+QIIME_TUTORIAL_MAP = (
+    "#SampleID\tBarcodeSequence\tLinkerPrimerSequence\tTreatment\tDOB\t"
+    "Description\n"
+    "#Example mapping file for the QIIME analysis package.  These 9 samples "
+    "are from a study of the effects of exercise and diet on mouse cardiac "
+    "physiology (Crawford, et al, PNAS, 2009).\n"
+    "PC.354\tAGCACGAGCCTA\tYATGCTGCCTCCCGTAGGAGT\tControl\t20061218\t"
+    "Control_mouse_I.D._354\n"
+    "PC.355\tAACTCGTCGATG\tYATGCTGCCTCCCGTAGGAGT\tControl\t20061218\t"
+    "Control_mouse_I.D._355\n"
+    "PC.356\tACAGACCACTCA\tYATGCTGCCTCCCGTAGGAGT\tControl\t20061126\t"
+    "Control_mouse_I.D._356\n"
+    "PC.481\tACCAGCGACTAG\tYATGCTGCCTCCCGTAGGAGT\tControl\t20070314\t"
+    "Control_mouse_I.D._481\n"
+    "PC.593\tAGCAGCACTTGT\tYATGCTGCCTCCCGTAGGAGT\tControl\t20071210\t"
+    "Control_mouse_I.D._593\n"
+    "PC.607\tAACTGTGCGTAC\tYATGCTGCCTCCCGTAGGAGT\tFast\t20071112\t"
+    "Fasting_mouse_I.D._607\n"
+    "PC.634\tACAGAGTCGGCT\tYATGCTGCCTCCCGTAGGAGT\tFast\t20080116\t"
+    "Fasting_mouse_I.D._634\n"
+    "PC.635\tACCGCAGAGTCA\tYATGCTGCCTCCCGTAGGAGT\tFast\t20080116\t"
+    "Fasting_mouse_I.D._635\n"
+    "PC.636\tACGGTGAGTGTC\tYATGCTGCCTCCCGTAGGAGT\tFast\t20080116\t"
+    "Fasting_mouse_I.D._636\n"
+)
 
 EXP_SAMPLE_TEMPLATE = (
     "sample_name\tcollection_timestamp\tdescription\thas_extracted_data\t"
--- a/qiita_db/metadata_template/test/test_util.py
+++ b/qiita_db/metadata_template/test/test_util.py
@@ -65,6 +65,17 @@ class TestUtil(TestCase):
         exp.index.name = 'sample_name'
         assert_frame_equal(obs, exp)
 
+    def test_load_template_to_dataframe_qiime_map(self):
+        obs = load_template_to_dataframe(StringIO(QIIME_TUTORIAL_MAP_SUBSET),
+                                         index='#SampleID')
+        exp = pd.DataFrame.from_dict(QIIME_TUTORIAL_MAP_DICT_FORM)
+        exp.index.name = '#SampleID'
+        obs.sort_index(axis=0, inplace=True)
+        obs.sort_index(axis=1, inplace=True)
+        exp.sort_index(axis=0, inplace=True)
+        exp.sort_index(axis=1, inplace=True)
+        assert_frame_equal(obs, exp)
+
     def test_load_template_to_dataframe_duplicate_cols(self):
         obs = load_template_to_dataframe(
             StringIO(EXP_SAMPLE_TEMPLATE_DUPE_COLS))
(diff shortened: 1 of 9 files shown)


antgonza:
Works fine. Except that the validation doesn't accept comments (lines that start with #). Also, I know is not part of this issue but it will be great if you could fix the warnings when adding a template. The issue is that if you have warnings about EBI, processing, etc all of them are in a single line, could you put them in different ones (add `</br>`)?  Gonna take a look a the code.

qiita_db/metadata_template/util.py
```diff
@@ -315,3 +320,38 @@ def get_invalid_sample_names(sample_names):
...
+    some other different column.
+    """
+    first_line = None
+    with open_file(fp, mode='U') as f:
+        first_line = f.readline()
+    if not first_line:
+        raise QiitaDBError('Empty file passed!')
+
+    first_col = first_line.split()[0]
+    return first_col == '#SampleID'
```
antgonza: Should you also check for LinkerPrimer, BarcodeSequence, and ReverseBarcodeSequence (optional)?

qiita_db/metadata_template/test/test_util.py
```diff
@@ -685,6 +719,19 @@ def test_invalid_lat_long(self):
...
      'str_column': {'2.Sample1': 'NA', '2.Sample2': 'NA', '2.Sample3': 'NA'}}
 
+QIIME_TUTORIAL_MAP_DICT_FORM = {
+    'BarcodeSequence': {'PC.354': 'AGCACGAGCCTA',
+                        'PC.607': 'AACTGTGCGTAC'},
+    'LinkerPrimerSequence': {'PC.354': 'YATGCTGCCTCCCGTAGGAGT',
+                             'PC.607': 'YATGCTGCCTCCCGTAGGAGT'},
+    'Treatment': {'PC.354': 'Control',
+                  'PC.607': 'Fast'},
+    'DOB': {'PC.354': 20061218,
```
ElDeveloper: Strictly speaking, QIIME mapping files do not do any sort of data type inferences, but I guess it's OK in this context as we need these data types to insert the information into the DB?

antgonza:
Code looks fine, only one comment. However, not sure if we should have a qiime map loading reimplementation here or use qiime. I think having it here could help us deprecate the other one for qiime 2.

qiita_pet/handlers/study_handlers/description_handlers.py
```diff
@@ -177,25 +178,33 @@ def process_sample_template(self, study, user, callback):
             # The file does not exist, fail nicely
             raise HTTPError(404, "This file doesn't exist: %s" % fp_rsp)
 
+        # Define here the message and message level in case of success
```
ElDeveloper: this comment is outdated.

qiita_pet/templates/study_description_templates/study_information_tab.html
```diff
@@ -12,15 +12,25 @@
   <hr>
 
   {% if show_select_sample %}
-    Select your sample template <br/>
+    Select your sample template or, alternatively, a QIIME mapping file <br/>
     (only files with the "txt" and "tsv" file extensions will be displayed here):
     <select id="sample_template">
       {% for f in files %}
-        {% if f.endswith(('txt', 'tsv')) %}
```
ElDeveloper: Why did you remove this check?

qiita_ware/test/test_metadata_pipeline.py
```diff
@@ -0,0 +1,95 @@
...
+        for fp in self._clean_up_files:
+            if exists(fp):
+                remove(fp)
+
+    def test_create_templates_from_qiime_mapping_file(self):
+        new_pt_id = get_count('qiita.prep_template') + 1
+        obs_st, obs_pt = create_templates_from_qiime_mapping_file(
+            StringIO(QIIME_MAP), self.new_study, "16S")
+
+        # Be green: clean the environment
```
ElDeveloper: :sunglasses:

qiita_ware/metadata_pipeline.py
```diff
@@ -0,0 +1,76 @@
...
+    -------
+    (SampleTemplate, PrepTemplate)
+        The templates created from the QIIME mapping file
+    """
+    qiime_map = load_template_to_dataframe(fp, index='#SampleID')
+
+    # There are a few columns in the QIIME mapping file that are special and
+    # we know how to deal with them
+    rename_cols = {
+        'BarcodeSequence': 'barcode',
```
ElDeveloper: What happens when none of these columns are present? Would an empty prep template be generated? Should we error in this case?

ElDeveloper:
Code looks great, I'm gonna test the system now.

@biocore/qiita-admin would be good if others could test this.

ElDeveloper:
As for having two parsers, I almost want to re-use the code in QIIME,
having multiple parsers for the same file-format is a nightmare and
prone to error. What do you guys think, I know there's the whole GPL
limitation, etc ... but then again, I think we should think this
through.

josenavas:
do you want to have a call about the 2 parsers thing? I think it will be faster than discussing over here.

ElDeveloper:
Sure, anyone else who would like to join?

qiita_db/metadata_template/util.py
josenavas: This is just a quick and fast check to see if the file is a mapping file. Note that templates do not have any # sign at the beginning, not even for comments. The fact that the file starts with '#SampleID' is enough to know that we are not parsing a template, but a mapping file.

qiita_db/metadata_template/test/test_util.py
josenavas: Yes, we need the type inference to be able to search correctly over metadata. Also, the idea is that the DB will not know that this data came from a mapping file.

qiita_pet/handlers/study_handlers/description_handlers.py
josenavas: It looks like I end up removing two lines that I shouldn't. I've re-added the lines that define the message and message level in case of success.

qiita_pet/templates/study_description_templates/study_information_tab.html
josenavas: I move this check to the handler, so less python is executed on the template.

qiita_ware/metadata_pipeline.py
josenavas: Good point. I was assuming "correct" QIIME mapping files, but it is possible that the users are not uploading this information. I've added a check.

josenavas:
@ElDeveloper @antgonza this should be ready to review!

qiita_pet/templates/study_description_templates/study_information_tab.html
adamrp: :+1:

qiita_pet/handlers/study_handlers/description_handlers.py
```diff
@@ -177,25 +178,35 @@ def process_sample_template(self, study, user, callback):
...
+        msg = "The sample template '%s' has been added" % sample_template
+        msg_level = "success"
+        is_mapping_file = looks_like_qiime_mapping_file(fp_rsp)
```
adamrp: I like the escalation from "looks like" to "is" :laughing:

qiita_db/metadata_template/util.py
```diff
@@ -315,3 +331,115 @@ def get_invalid_sample_names(sample_names):
...
+    This is not doing a validation of the QIIME mapping file. It simply checks
+    the first line in the file and it returns true if the line starts with
+    '#SampleID', since a sample/prep template will start with 'sample_name' or
+    some other different column.
+    """
+    first_line = None
+    with open_file(fp, mode='U') as f:
+        first_line = f.readline()
+    if not first_line:
+        raise QiitaDBError('Empty file passed!')
```
adamrp: Why raise an error here instead of just returning `False`?

qiita_pet/handlers/study_handlers/description_handlers.py
josenavas: hahaha, yeah it was way shorter....

qiita_db/metadata_template/util.py
josenavas: That is actually a very good point. An empty file does not look like a QIIME mapping file, so the right thing to do is return False. The process will raise in another step though, but this will reduce the amount of code duplication that we have.

qiita_pet/templates/study_description_templates/study_information_tab.html
ElDeveloper: :+1:

qiita_ware/metadata_pipeline.py
```diff
@@ -0,0 +1,87 @@
...
+        'LinkerPrimerSequence': 'primer',
+        'Description': 'description',
+    }
+
+    if 'ReverseLinkerPrimer' in qiime_map:
+        rename_cols['ReverseLinkerPrimer'] = 'reverselinkerprimer'
+
+    missing = set(rename_cols).difference(qiime_map.columns)
+    if missing:
+        raise QiitaWareError(
```
ElDeveloper: Can you add this exception to this [try/except](https://github.com/biocore/qiita/blob/master/qiita_pet/handlers/study_handlers/description_handlers.py#L192).

ElDeveloper:
PS - You also have some PEP8 failures:

```
qiita_db/metadata_template/util.py:400:13: E731 do not assign a lambda expression, use a def
qiita_db/metadata_template/util.py:403:13: E731 do not assign a lambda expression, use a def
qiita_db/metadata_template/util.py:407:13: E731 do not assign a lambda expression, use a def
qiita_db/metadata_template/util.py:410:13: E731 do not assign a lambda expression, use a def
```

qiita_ware/metadata_pipeline.py
josenavas: Done

josenavas:
Thanks @ElDeveloper . I don't understand why the flake8 in travis is different from the one on my machine... I think the one in conda has something different.

ElDeveloper:
Thanks @josenavas. Yeah, conda is sometimes a bit outdated compared to the pypi packages ...

ElDeveloper:
I hereby officially declare this as :+1:! If someone else wants to go ahead and merge, that would be awesome.

antgonza:
```bash
In nomine Patris, et Filii, et Spiritus Sancti. Amen
```

PR #1220, opened 2015-05-29
Add searching gif and feedback when search running. Fix #1069

### LINKED ISSUE #1069: Search on list_studies.html gives no visual feedback
Currently, if I do a search, there is no explicit visual feedback. There *might* be implicit feedback (if the search results in changes to the list of studies being displayed), but explicit feedback would be great (e.g., a very obvious/colorful/bold statement that indicates "Only showing studies that comprise samples hit by the following criteria: XXXXXXX").

Another improvement would be to show on this page how many of each study's samples are selected by the criteria specified.

squirrelo:
Does what it says on the tin.

### DIFF
--- a/qiita_pet/templates/list_studies.html
+++ b/qiita_pet/templates/list_studies.html
@@ -45,7 +45,7 @@ function sel_proc_data(row, pid) {
 }
 
 function show_alert(data) {
-  bootstrapAlert(data + ' samples selected.', "success", 2000);
+  bootstrapAlert(data + ' samples selected.', "success", 1000);
    $('#dflt-sel-info').css('color', 'rgb(0, 160, 0)');
 }
 
@@ -71,6 +71,7 @@ $(document).ready(function() {
         }
 
         $('#studies-table').dataTable({
+            "deferRender": true,
             "columns": [
               {"className": 'details-control', "orderable": false, "data": null, "defaultContent": '<span class="glyphicon glyphicon-chevron-down"></span>'},
               { "orderable": false},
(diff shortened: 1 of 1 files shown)


antgonza:
When I search on the study page I can't see the wait logo (even when doing large searches), where should it be? Note that I had to pull this to branch 1084-qiita-pet cause I have already applied the database changes.

squirrelo:
It should be showing up under the search bar

antgonza:
:disappointed: can't see it, have you tested with that branch?

squirrelo:
not with that branch, no. I'll take a look

squirrelo:
Should be good to go

antgonza:
Thanks, it's working now. However, the search completed message is a bit misleading as the query finished but the rendering on the website is still not done. Not sure how to fix this, perhaps simply adding "Search complete (finishing rendering): query" .

squirrelo:
That should not happen, as the render call happens before that text is added to the page.

ElDeveloper:
I think this is working well, though I don't have enough studies to see the animated gif.

The only thing I wanted to bring up, is the fact that there's not enough feedback when two queries that retrieve no results are subsequently entered, see for example this:

feedback

It is only clear that something changed when you enter a query that retrieves studies, otherwise it's a bit hard to catch.

Perhaps using some sort of jQuery animations or highlighting that info box somehow else would do the trick.

antgonza:
If you see @ElDeveloper's gif, he has the same behavior that I saw. The text under the text box says: "Search Completed: study_title includes soil" but the list still says "No studies found". In the next frame the studies appear.

squirrelo:
His gif is actually working correctly, he just has no results found. I can add `(X studies found)` to the end of the search if it makes it clearer.

squirrelo:
Now the page grays out when you search so that you know when the search is running and when you have results. Minimum of 400ms gray so no matter how fast the search completes you still see the gray out.

ElDeveloper:
Awesome, thanks!

ElDeveloper:
:+1:

PR #1223, opened 2015-05-30
IPython section was missing from example config

adamrp:


adamrp:
Actually it's not missing, just missing some arguments.. don't merge this yet

adamrp:
Actually I guess this section is not used any more? I had it lingering in my personal configuration file. My mistake.

wasade:
It's part of the MOI config now, qiita proper doesn't need to know about
ipython

Issue #913, opened 2015-03-09
small interface suggestions from qiita install fest at biocore 15.3


gregcaporaso:
``qiita-admin`` makes more sense to me as a command name than ``qiime_env`` (and it's easier to type, b/c ``-`` only requires one key and ``_`` requires two).

gregcaporaso:
Related: when running:

```
qiita_env start_cluster qiita_general
```

Could ``qiita_general`` be the default? And/or can it be called ``qiita-default``?

gregcaporaso:
@jairideout and I were both confused with this error message:

```
$ qiita_env start_cluster qiita_demo
Usage: qiita_env start_cluster [OPTIONS] CLUSTERS...

Error: Invalid value for "clusters": invalid choice: qiita_demo. (choose from qiita_general)
```

I think this error message would be better if it looked like:

```
Error: Unknown cluster choice: qiita_demo

Valid choices are: 
 qiita_general
```

wasade:
The choices formatting is left to click and out of our control

wasade:
:+1: on others (-admin and - in cluster names)

adamrp:
:+1: to dashes in place of underscores in cluster names. No strong feelings about `qiita_env` vs. `qiita-admin`, although if we keep `qiita_env`, I would :+1: changing it to `qiita-env`

Some other thoughts:
We currently have these subgroups of the top-level `qiita` click group:
```
db
ebi
log
maintenance
ware
webserver
```

What does @biocore/qiita-dev think of the following organizational changes:
- Move `log` into the `maintenance` group
- Move `ebi` into the `ware` group.
- Create a `pet` group as a subgroup directly under `qiita` (same level as `db` and `ware`)
- Move `webserver` into the new `pet` group

ElDeveloper:
I like this new organization!

josenavas:
:+1: @adamrp

josenavas:
And I actually like more `qiita-admin`

PR #1224, opened 2015-05-30
Reorganize click commands

adamrp:
See #913


josenavas:
Thanks for working on this @adamrp I commented late on the issue. Is it possible to change to qiita-admin if other people agree?

adamrp:
I was wondering about whether or not to do that, and I decided against it, but would like to hear others' opinions. The reason I didn't make the change is that *all* of the CLI commands assume admin access, so it seemed a little weird to me to have an explicit `qiita-admin` script -- after all, many of the commands in the regular `qiita` script are largely administrative (`qiita maintenance` commands, `qiita pet webserver` command(s), etc.)

adamrp:
Note that the original reason for having these two scripts was out of necessity -- the commands that created/destroyed the environment (`qiita_env`) couldn't be in the same file as the commands that manipulated an existing environment (`qiita`) because many of the default arguments to parameters in the `qiita` script assume an existing environment. So, the `qiita_env` script really was (at least originally) intended to get around that problem

josenavas:
That's a really good point @adamrp :+1: on leaving as it is.

adamrp:
This should be good to go -- tests pass, and coverage remained the same (not sure why the github interface isn't being populated with the coveralls results...)

antgonza:
Could you also change the md files to reflect these changes?

adamrp:
Good catch @antgonza! I think I caught all occurrences.

antgonza:
:+1:

ElDeveloper:
Can you add a mention to the changelog?

Yoshiki Vázquez-Baeza

adamrp:
Good idea, @ElDeveloper, thanks. Added.

adamrp:
Still not sure what's up with coveralls here, but this is ready for another round of review

ElDeveloper:
> Is it possible to change to qiita-admin if other people agree?

:+1: I like that better than `qiita-env`, `env` makes a lot of sense to us because that's the term we've been using to refer to our development environments, but it doesn't make sense when put in the context of a production server (I think).

adamrp:
@ElDeveloper currently the commands in `qiita-env` are:
```
Commands:
  clean_test     Cleans the test database environment.
  drop           Drops the database specified in config
  job_details    Pretty print moi job details
  make           Creates the database specified in config
  patch          Patches the database schema based on the...
  start_cluster  Start a compute environment
  stop_cluster   Stop a compute environment
  test           Test the environment Check to make sure that...
```
All of which are indeed related to the environment. `qiita-admin` doesn't make more sense to me than `qiita-env` for this reason, as well as the other reasons I mentioned in my previous comments. So far it's 2 :+1:s for keeping it the same, and 1 :-1: for changing to `qiita-admin`. Any other votes?

ElDeveloper:
` ¯\_(ツ)_/¯`

adamrp:
Any further comments on this?

antgonza:
:+1:

PR #1225, opened 2015-06-01
fixing a template issue

antgonza:
I looked for all occurrences of `.status()` and this was the only one not in JS, which should be fine.

### DIFF
--- a/qiita_pet/uimodules/prep_template_tab.py
+++ b/qiita_pet/uimodules/prep_template_tab.py
@@ -62,9 +62,10 @@ def _template_generator(study, full_access):
         the PrepTemplate object and a tuple with 3 strings for the style of
         the prep template status icons
     """
+
     for pt_id in study.prep_templates():
         pt = PrepTemplate(pt_id)
-        if full_access or pt.status() == 'public':
+        if full_access or pt.status == 'public':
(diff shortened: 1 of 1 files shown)


wasade:
:+1: once travis is content

adamrp:
Same, :+1: when tests pass

josenavas:
:+1:

ElDeveloper:
Is this an indicator that this handler does not have a test for this case? Code looks good otherwise.

antgonza:
Yup, your intuition is right. Note that I'm not sure how that will look like.

josenavas:
Re: tests. We need to move the code out of the handlers. I think the best thing that we can do is to create a function that the handlers call and that function returns all the parameters that the template need. this way will be able to test every single thing in the handlers. This functions should live in a module in qiita pet.

I will start a branch with a proof of concept (i.e. one handler) so other devs can take handlers and do PR against that one. I think this will be the best way to have all the code moved fast. In the main PR, I will create a list of the handlers that need to be taken care of, so devs can write their name next to it. Does this sound like a good plan?

antgonza:
Yup, :+1:

ElDeveloper:
:+1: that sounds like a good plan, though I think we now need to focus on the "portal types" functionality, as it is our next milestone (Sloan meeting). We should discuss about this at the next Qiita meeting.

PR #1233, opened 2015-06-04
Allow hypens, but not at beginning or end

### LINKED ISSUE #1232: valid email not being accepted
A user reported not being able to create an account and only getting a 500 error. 

After checking the logs (note that I'm replacing the actual email for: name@a.b-c.d, which is the general format of the original email):
ERROR:
Bad email given: name@a.b-c.d
TRACE:

callback()

File "/home/qiita/qiita_main/qiita_pet/handlers/auth_handlers.py", line 39, in post
created = User.create(username, password, info)

File "/home/qiita/qiita_main/qiita_db/user.py", line 196, in create
raise IncorrectEmailError("Bad email given: %s" % email)

IncorrectEmailError: Bad email given: name@a.b-c.d

adamrp:
Fix #1232

### DIFF
--- a/qiita_db/test/test_user.py
+++ b/qiita_db/test/test_user.py
@@ -41,16 +41,20 @@ class SupportTests(TestCase):
         valid1 = 'foo@bar.com'
         valid2 = 'asdasd.asdasd.asd123asd@stuff.edu'
         valid3 = 'w00t@123.456.789.com'
+        valid4 = 'name@a.b-c.d'
         invalid1 = '@stuff.com'
         invalid2 = 'asdasdásd@things.com'
         invalid3 = '.asdas@com'
+        invalid4 = 'name@a.b-c.d-'
 
         self.assertTrue(validate_email(valid1))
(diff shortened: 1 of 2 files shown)


antgonza:
:+1: if tests pass

josenavas:
:+1: 

I'm really surprised that we are having so many issues checking emails... This is something that is constantly done in the intertubes and I don't think people has that many problems.

Probably we shouldn't be restrictive at all (just check that it looks like <something>@<something>.<something>). If the user provides a bad email he will never be able to use the system. We can then remove the emails that have not been validated for more than 2 weeks...

adamrp:
The regex for identifying valid email addresses is [surprisingly complex](http://www.regular-expressions.info/email.html) and adherence to the standards varies. I am not against relaxing the checking, but I'd be surprised if we continue to see more false negatives here.

josenavas:
Thanks for the link @adamrp! I think now it makes more sense why it is so complex...

Issue #597, opened 2014-11-13
Be able to (re)name prep templates


antgonza:
Even so that the id is useful, it will be better if we could name raw data. For example, if an experiment has 2 raw datas be able to name them V2 and V4 or whatever.

antgonza:
When this issue was created renaming the raw data made sense, now it should be the prep template. Renaming the issue.

josenavas:
As discussed in here https://github.com/biocore/qiita/pull/1241#issuecomment-109338432 we should do this for any object of the study:
- [ ] PrepTemplate
- [ ] RawData
- [ ] PreprocessedData
- [ ] ProcessedData

antgonza:
Solved in: https://github.com/biocore/qiita/pull/1738

PR #1241, opened 2015-06-05
fix #1015

### LINKED ISSUE #1015: tabs in study description
In most use cases, based on my last few week tests, the user will be mainly working on the last object added into a working tab. Thus, when clicking on a tab in the study description (let's say: Processed data) the focus should be in the last tab within the new tabs vs. the first one. This should make it more user friendly.

antgonza:
Fixing #1015 and making sure that when displaying the tabs are always in ascending order by id.

### DIFF
--- a/qiita_pet/templates/study_description.html
+++ b/qiita_pet/templates/study_description.html
@@ -601,6 +601,11 @@ function submit_preprocess(id) {
 };
 
 function display_tab(top_tab, sub_tab, prep_tab){
+  // making sure the last tab is always the default
+  $('#prep_template_nav_tabs a:last').tab('show');
+  $('#preprocessed_data_tab li:eq(1) a').tab('show');
+  $('#processed_data_tab li:eq(1) a').tab('show');
+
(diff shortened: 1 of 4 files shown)


josenavas:
Thanks @antgonza :+1: 

Just to throw an idea, and let's see what you and others think. I think it will be useful to add a 'name' property to the PrepTemplate/RawData/PreprocessedData/ProcessedData objects, and we can use that to name the tabs. We can probably limit that to ~10 chars or so. It will make live easier to the users to identify the object that they are working with.

antgonza:
Agree, basically expand: https://github.com/biocore/qiita/issues/597

josenavas:
Thanks @antgonza, i've commented on that issue

ElDeveloper:
:+1: works like a charm, question. Should this be added to the ChangeLog? We kinda need to know what things to describe there and what things not to.

antgonza:
I do not feel this change is ChangeLog worth ... IMOO we should add only things that are huge improvements, bugs, or GUI changes.

PR #1242, opened 2015-06-05
fix #1130

### LINKED ISSUE #1130: Allow update prep template through UI
Some thought: disable if some functionality (i.e. preprocessing/EBI submission) has been already done on the prep template.

antgonza:
In this PR:
- Moved the update code from sample_template.py to base_metadata_template.py & generalized for prep template updates
- Fixed an issue with sanbdox -> sandbox
- Fix #1130 (add update to the prep template)


qiita_db/metadata_template/prep_template.py
```diff
@@ -75,6 +75,7 @@ class PrepTemplate(MetadataTemplate):
     _sample_cls = PrepSample
     _fp_id = convert_to_id("prep_template", "filepath_type")
     _filepath_table = 'prep_template_filepath'
+    _columns = PREP_TEMPLATE_COLUMNS
```
josenavas: This is not correct. In case that the prep template is of a target gene datatype, additional columns are requested. My suggestion is, instead of using a class variable as done here, create a function call `get_column_restrictions` that each subclass (`SampleTemplate` and `PrepTemplate`) implements. The `SampleTemplate` implementation will return the `SAMPLE_TEMPLATE_COLUMNS` while the `PrepTemplate` implementation will return a new dictionary with the contents of `PREP_TEMPLATE_COLUMNS` + `PREP_TEMPLATE_COLUMNS_TARGET_GENE` in case of being of a `TARGET_GENE_DATA_TYPE`, or `PREP_TEMPLATE_COLUMNS` otherwise. Does this make sense? There is an example of this in the create method of the prep template.

josenavas:
Thinking about this a bit more, is there any metadata column that we don't want the user to update? One of the examples that pops in my mind is barcode. Although it is useful to update the barcode in case that the user messed it up, if there is a PreprocessedData generated from it, we shouldn't allow to update the PrepTemplate. The main reason is that one of the PreprocessedData will be generated with an old version of it, and the information in the PrepTemplate is critical for the contents of the PreprocessedData. Thus, I think a bit more thought should be imposed when updating a prep template.

An option to prevent this is: create a function called `can_be_updated`. The SampleTemplate implementation always returns True, while the PrepTemplate one checks if there is any PreprocessedData generated. If a PreprocessedData is generated, it then checks which values are being updated. If it is one of the columns controlled by `PREP_TEMPLATE_COLUMNS_TARGET_GENE` it doesn't allow the update, while if it is any other column, this changed is allowed. The reason why I want to allow this other changes is because some of those updates may be needed for EBI submission, and those changes will not impact the contents of the PreprocessedData.

Sorry for the long comment, but I think is better to be explicit than implicit.

antgonza:
This is ready for another review ...

qiita_db/metadata_template/base_metadata_template.py
```diff
@@ -1094,6 +1095,79 @@ def categories(self):
...
+            The metadata template file contents indexed by samples Ids
+
+        Raises
+        ------
+        QiitaDBError
+            If md_template and db do not have the same sample ids
+            If md_template and db do not have the same column headers
+            If self.can_be_updated is not True
+        """
+        self._check_subclass()
```
josenavas: I don't think this check is needed. There is no way to instantiate the base class, so this check should only be needed in the classmethods. Since this is an instance method, this check is "already" performed.

qiita_db/metadata_template/prep_template.py
```diff
@@ -289,6 +289,45 @@ def data_type(self, ret_id=False):
...
+
+        Returns
+        -------
+        bool
+            If the template can be updated
+        """
+        if not self.preprocessed_data:
+            return True
+
+        tg_columns = {e for v in viewvalues(PREP_TEMPLATE_COLUMNS_TARGET_GENE)
```
josenavas: I think this test should only be performed if the prep template is of a TARGET GENE type.
I'm unsure if other data types will have different requirements at this level, so it is better to not check this always.

qiita_db/metadata_template/prep_template.py
```diff
@@ -289,6 +289,45 @@ def data_type(self, ret_id=False):
...
+        Parameters
+        ----------
+        columns : set
+            A set of the names of the columns to be updated
+
+        Returns
+        -------
+        bool
+            If the template can be updated
+        """
```
josenavas: Not blocking, but it might be useful to add a "Notes" section in which it describes the cases in which the prep template can be updated.

qiita_pet/templates/study_description_templates/prep_template_info_tab.html
```diff
@@ -188,5 +188,19 @@ <h4 class="panel-title">
         </td>
       </tr>
       {% end %}
+      {{is_editable}}
```
josenavas: I think this is not needed?

qiita_pet/templates/study_description_templates/prep_template_info_tab.html
```diff
@@ -188,5 +188,19 @@ <h4 class="panel-title">
...
+    {% if is_editable %}
+      <tr><td><hr></td></tr>
+      <tr>
+        <td>
+          <select id="update_prep_template_{{pt_id}}">
+            {% for f in files %}
```
josenavas: This is currently showing all files. Probably useful to filter them and show only the ones than end with txt or tsv, as is done in other places. (This needs to be changed in the PrepTemplateInfoTab class)

josenavas:
I was trying to update a template and some error happened. However, all that is shown is this, so it's hard to debug...

screen shot 2015-06-07 at 3 15 30 pm

qiita_db/metadata_template/base_metadata_template.py
wasade: Agree

Doesn't that imply the classmethods shouldn't actually be classmethods as they depend on instantiated state?

antgonza:
ready for another pass ...

josenavas:
:+1:

antgonza:
@squirrelo a reminder about reviewing this one.

qiita_db/metadata_template/base_metadata_template.py
```diff
@@ -1094,6 +1095,77 @@ def categories(self):
 
         return cols
 
+    def update(self, md_template):
+        r"""Update values in the sample template
```
squirrelo: in the template, not just sample template.

qiita_db/metadata_template/base_metadata_template.py
```diff
@@ -1094,6 +1095,77 @@ def categories(self):
...
+        current_map = self._transform_to_dict(conn_handler.execute_fetchall(
+            "SELECT * FROM qiita.{0} WHERE {1}=%s".format(self._table,
+                                                          self._id_column),
+            (self.id,)))
+        dyn_vals = self._transform_to_dict(conn_handler.execute_fetchall(
+            "SELECT * FROM qiita.{0}".format(self._table_name(self.id))))
+
+        for k in current_map:
+            current_map[k].update(dyn_vals[k])
+            current_map[k].pop('study_id', None)
```
squirrelo: why pop every loop?

qiita_db/metadata_template/prep_template.py
```diff
@@ -289,6 +289,54 @@ def data_type(self, ret_id=False):
...
+        or the prep template data-type is not part of TARGET_GENE_DATA_TYPES,
+        (2) if is part of TARGET_GENE_DATA_TYPES then we will only update if
+        the columns being updated are not part of
+        PREP_TEMPLATE_COLUMNS_TARGET_GENE
+        """
+        if (not self.preprocessed_data or
+           self.data_type() not in TARGET_GENE_DATA_TYPES):
+            return True
+
+        tg_columns = {e for v in viewvalues(PREP_TEMPLATE_COLUMNS_TARGET_GENE)
```
squirrelo: Since we are AvoidING doing double looped list comprehensions, please break this out to make it clearer.

qiita_db/metadata_template/base_metadata_template.py
```diff
@@ -1094,6 +1095,77 @@ def categories(self):
...
+
+        # converting sql results to dataframe
+        current_map = pd.DataFrame.from_dict(current_map, orient='index')
+
+        # simple validations of sample ids and column names
+        samples_diff = set(
+            new_map.index.tolist()) - set(current_map.index.tolist())
```
wasade: I don't think the `tolist` calls are necessary

qiita_db/metadata_template/sample_template.py
```diff
@@ -189,6 +187,32 @@ def study_id(self):
...
+        ----------
+        kwargs : ignored
+            Necessary to have in parameters to support other objects.
+
+        Returns
+        -------
+        bool
+            As this is the sample template this returns always True
+        """
+        return True
```
squirrelo: HUH? This is a function that just returns true? Can you add a note why it's always true?

qiita_db/metadata_template/base_metadata_template.py
wasade: It does seem odd...

is there a situation where `'study_id'` won't be present?

qiita_pet/handlers/study_handlers/description_handlers.py
```diff
@@ -423,6 +423,63 @@ def add_prep_template(self, study, user, callback):
...
+        msg = "The prep template '%s' has been updated" % prep_template
+        msg_level = "success"
+        # Get the uploads folder
+        _, base_fp = get_mountpoint("uploads")[0]
+        # Get the path of the prep template in the uploads folder
+        fp = join(base_fp, str(study.id), prep_template)
+
+        if not exists(fp):
+            # The file does not exist, fail nicely
+            raise HTTPError(400, "This file doesn't exist: %s" % fp)
```
squirrelo: 404 error, not 400

squirrelo:
few comments.

qiita_db/metadata_template/prep_template.py
```diff
@@ -289,6 +289,54 @@ def data_type(self, ret_id=False):
...
+                      for e in v.columns.keys()}
```
wasade: call to `.keys()` is redundant

qiita_db/metadata_template/base_metadata_template.py
antgonza: This was in the original code and I think is to be "safe" as every update leaves a row without study_id ... I can change if you want to, suggestions?

qiita_db/metadata_template/base_metadata_template.py
antgonza: ok

qiita_db/metadata_template/base_metadata_template.py
josenavas: I think I should comment here.

Note that `current_map` is a dict of {sample: {metadata: value}} so in each loop we have to remove the 'study_id' entry.

Re: situation where `study_id` won't be present: in the prep template. It is always present in the sample template but not in the prep templates. By doing `current_map[k].pop('study_id', None)` we make this code applicable to both templates.

qiita_db/metadata_template/base_metadata_template.py
antgonza: ok

qiita_db/metadata_template/base_metadata_template.py
wasade: Why not just bulk remove it from the subsequently produced DataFrame? It
will be more expressive, and is likely a more optimized operation.

josenavas: Actually, taken a closer look to the code, this can be refactored so it can be highly simplified. Lines 1119 to 1131 can be changed by these ones:

```python
current_map = self._transform_to_dict(conn_handler.execute_fetchall(
    "SELECT * FROM qiita.{0}".format(self._table_name(self.id))))
current_map = pd.DataFrame.from_dict(current_map, orient='index')
```

The reason behind is that all the metadata is stored in the dynamic table, and the query in 1119 is only retrieving the `study_id` which we are then filtering in line 1128. The current code looks like leftover from the previous DB schema that I forgot to change. Sorry about that, but at least I saw it now!

qiita_db/metadata_template/prep_template.py
antgonza: k

qiita_db/metadata_template/prep_template.py
antgonza: k

qiita_db/metadata_template/sample_template.py
antgonza: Do you think we need more info that what's in the doc? Which suggestions do you have?

qiita_pet/handlers/study_handlers/description_handlers.py
antgonza: Not sure what you mean, can you be more explicit? Thanks.

qiita_db/metadata_template/sample_template.py
squirrelo: Why is it always true because sample template? Just a sentence or two explaining that in a notes section would be useful.

qiita_pet/handlers/study_handlers/description_handlers.py
squirrelo: HTTP Error for a file not found is 404, not 400 as this states.

josenavas: @squirrelo although is a file does not exist error, I don't think this actually fits the 404 error. We don't want the user to be redirected to the 404 page, we want to get him in the study description page and let him know that the file he selected for processing as prep template does not exist. My guess is that this error will be hit almost never, but is better to be safe and put a check here.

Does this makes sense?

qiita_db/metadata_template/base_metadata_template.py
antgonza: OK, that worked. Thanks.

qiita_db/metadata_template/sample_template.py
antgonza: k

antgonza:
ping @wasade, @josenavas and @squirrelo

qiita_db/metadata_template/base_metadata_template.py
```diff
@@ -1092,6 +1093,67 @@ def categories(self):
...
+            If md_template and db do not have the same column headers
+            If self.can_be_updated is not True
+        """
+        conn_handler = SQLConnectionHandler()
+
+        # Clean and validate the metadata template given
+        new_map = self._clean_validate_template(md_template, self.study_id,
+                                                self.columns_restrictions)
+        # Retrieving current metadata
+
```
josenavas: Non-blocking: extra space.

qiita_db/metadata_template/base_metadata_template.py
```diff
@@ -1092,6 +1093,67 @@ def categories(self):
...
+        current_map = self._transform_to_dict(conn_handler.execute_fetchall(
+            "SELECT * FROM qiita.{0}".format(self._table_name(self.id))))
+        current_map = pd.DataFrame.from_dict(current_map, orient='index')
+
+        # simple validations of sample ids and column names
+        samples_diff = set(new_map.index) - set(current_map.index)
```
josenavas: Non-blocking but there is a fastest way of doing this:
```
In [16]: %timeit set(l1) - set(l2)
10 loops, best of 3: 120 ms per loop

In [17]: %timeit set(l1).difference(l2)
10 loops, best of 3: 103 ms per loop
```

qiita_db/metadata_template/base_metadata_template.py
```diff
@@ -1092,6 +1093,67 @@ def categories(self):
...
+        if samples_diff:
+            raise QiitaDBError('The new template differs from what is stored '
+                               'in database by these samples names: %s'
+                               % ', '.join(samples_diff))
+        columns_diff = set(new_map.columns) - set(current_map.columns)
```
josenavas: Same as above.

qiita_pet/handlers/study_handlers/description_handlers.py
```diff
@@ -423,6 +423,63 @@ def add_prep_template(self, study, user, callback):
 
         callback((msg, msg_level, 'prep_template_tab', pt_id, None))
 
+    def update_prep_template(self, study, user, callback):
```
josenavas: @antgonza I noticed that when a new sample/prep template is added, the uploaded file is removed from the upload folder. However, when we update the values this doesn't happen. Is this desired? IMO we should remove the file from the upload folder, so we can keep the FS cleaner (otherwise for every update that needs to be done the file is still in there...).

Another advantage of this is that if the user wants to update a value, it will download our current template, instead of using his/her old template, which I think is better. Does this make sense?

josenavas:
A couple of non-blocking comments and a question. Otherwise this looks good!

qiita_pet/handlers/study_handlers/description_handlers.py
antgonza: Agree, will update.

qiita_pet/handlers/study_handlers/description_handlers.py
```diff
@@ -423,6 +423,64 @@ def add_prep_template(self, study, user, callback):
...
+        fp = join(base_fp, str(study.id), prep_template)
+
+        if not exists(fp):
+            # The file does not exist, fail nicely
+            raise HTTPError(400, "This file doesn't exist: %s" % fp)
+        try:
+            with warnings.catch_warnings(record=True) as warns:
+                pt = PrepTemplate(pt_id)
+                pt.update(load_template_to_dataframe(fp))
+                remove(fp)
```
josenavas: Thanks for adding this. However, the removal is not happening in the update_sample_template function either. Can you add it there? I know that you didn't modified it here, but I think is easy enough to fix it here too. Sorry if I was not explicit enough with my previous comment about this.

josenavas:
Just one comment about consistency. Thanks @antgonza 
:+1: once that is solved!

antgonza:
Done.

josenavas:
:+1: @squirrelo you also reviewed this PR. Can you review, or merge if there are no more comments?

antgonza:
ping @squirrelo

qiita_db/metadata_template/sample_template.py
```diff
@@ -189,6 +187,38 @@ def study_id(self):
...
+            Necessary to have in parameters to support other objects.
+
+        Returns
+        -------
+        bool
+            As this is the sample template this returns always True
+
+        Notes
+        -----
+            This always return True as the sample template can always be
```
squirrelo: This makes no grammatical sense. Also, according to this we can update the sample id column?

antgonza: OK, do you have suggestions? Not sure, have you tried?

qiita_pet/handlers/study_handlers/description_handlers.py
squirrelo: Makes sense.

squirrelo: Might want to note that in a comment though, so we don't have this issue again if someone is reviewing the code.

qiita_db/metadata_template/sample_template.py
squirrelo: No I have not tried, have you? Should probably be in the tests to make sure we can't update that column if we don't want it to be updated.

squirrelo:
Couple comments and questions. I'll test the interface shortly.

qiita_db/metadata_template/sample_template.py
squirrelo: Thinking about it more, we definitely do NOT what to have that column updated using the current code. If we do want to be able to update the sample IDs, we also need the updates to propagate to the study->samples join table and the other templates as well, which could get interesting.

antgonza: I just did and that's validated in another step, so is not possible. You will need to erase the sample template and add a new one. Will update the text and hopefully you will like it.

Please test your concerns before raising them ...

antgonza:
@squirrelo fixed the texts that you requested ...

qiita_db/metadata_template/sample_template.py
josenavas: As @antgonza said, that is tested in another step - by definition of update
you can update values of a sample, not the sample itself...
Anyways, updating a sample ID is as simple as updating the sample ID on
this table. Everything else will be updated because we have a cascade on
update. However, the only way of doing such update is going directly to the
SQL.

2015-06-17 9:10 GMT-07:00 Joshua Shorenstein <notifications@github.com>:

> In qiita_db/metadata_template/sample_template.py
> <https://github.com/biocore/qiita/pull/1242#discussion_r32643864>:
>
> > +    def can_be_updated(self, **kwargs):
> > +        """Gets if the template can be updated
> > +
> > +        Parameters
> > +        ----------
> > +        kwargs : ignored
> > +            Necessary to have in parameters to support other objects.
> > +
> > +        Returns
> > +        -------
> > +        bool
> > +            As this is the sample template this returns always True
> > +
> > +        Notes
> > +        -----
> > +            This always return True as the sample template can always be
>
> Thinking about it more, we definitely do NOT what to have that column
> updated using the current code. If we do want to be able to update the
> sample IDs, we also need the updates to propagate to the study->samples
> join table and the other templates as well, which could get interesting.
>
> —
> <https://github.com/biocore/qiita/pull/1242/files#r32643864>.
>



-- 
Jose Navas

squirrelo:
Text is much clearer, thanks.

PR #1243, opened 2015-06-05
fix #1237

### LINKED ISSUE #1237: analysis.dropped_samples should always return same type
Not sure if related to #1236 but if you select those 2 bioms and select to do beta the system errors out cause the analysis has no biom tables associated. 

Note that analysis.dropped_samples returns None if there are no biom files associated vs. returning []. Which causes a 500 error [here](https://github.com/biocore/qiita/blob/master/qiita_pet/handlers/analysis_handlers.py#L146)

antgonza:
While fixing I realized that also analysis.biom_tables will return None vs. [] so fixed.

### DIFF
--- a/qiita_db/analysis.py
+++ b/qiita_db/analysis.py
@@ -423,8 +423,7 @@ class Analysis(QiitaStatusObject):
         Returns
         -------
         dict
-            Dictonary in the form {data_type: full BIOM filepath} or None if
-            not generated
+            Dictonary in the form {data_type: full BIOM filepath}
         """
         conn_handler = SQLConnectionHandler()
         fptypeid = convert_to_id("biom", "filepath_type")
--- a/qiita_db/analysis.py
+++ b/qiita_db/analysis.py
@@ -337,7 +337,7 @@ class Analysis(QiitaStatusObject):
         """
         bioms = self.biom_tables
         if not bioms:
-            return None
+            return []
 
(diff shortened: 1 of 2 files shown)


josenavas:
Thanks @antgonza 

Do any of these changes impact the interface? i.e. there isn't any check in the interface for the result of these calls being `None`. If so, that code should be removed. Otherwise :+1:

antgonza:
AFAIK there are no checks in the interfase, which cause the problem.

josenavas:
Thanks! :+1:

qiita_db/analysis.py
```diff
@@ -337,7 +337,7 @@ def dropped_samples(self):
         """
         bioms = self.biom_tables
         if not bioms:
-            return None
+            return []
```
ElDeveloper: Can you update the docstring?

qiita_db/analysis.py
```diff
@@ -435,7 +435,7 @@ def biom_tables(self):
                "WHERE af.analysis_id = %s AND f.filepath_type_id = %s")
         tables = conn_handler.execute_fetchall(sql, (self._id, fptypeid))
         if not tables:
-            return None
+            return []
```
ElDeveloper: Can you update the docstring?

ElDeveloper:
:+1: ready for merge once tests pass.

PR #1244, opened 2015-06-05
Issue 285 - Add twitter feed sidebar

squirrelo:
This adds the base code needed to create the twitter feed storage in redis, as well as showing the feed in qiita-pet.


qiita_db/user.py
```diff
@@ -463,6 +468,45 @@ def _change_pass(self, newpass):
...
+        message : str
+            Message to add to queue
+        """
+        count = r_client.lpush("%s:messages" % self._id, '%s\n%s' %
+                               (message, datetime.now()))
+        if count > 100:
+            # only store 100 messages, so pop oldest
+            r_client.rpop()
+
+    def messages(self, count=100, as_html=False):
```
josenavas: I disagree on having any kind of HTML in qiita_db. Please, remove this. It is easy enough on the interface to do the replaces that you are doing here. (Note that this is already done in other places in the code)

qiita_pet/handlers/base_handlers.py
```diff
@@ -61,7 +61,12 @@ class MainHandler(BaseHandler):
     def get(self):
         msg = self.get_argument('message', '')
         lvl = self.get_argument('level', '')
-        self.render("index.html", message=msg, level=lvl)
+        if self.current_user:
+            msgs = self.current_user.messages(5, as_html=True)
```
josenavas:
```python
msgs = self.current_user.messages(5)
msgs = ['<p>%s</p>' % x.replace('\n', '<br />') for x in msgs]
```

qiita_pet/handlers/base_handlers.py
```diff
@@ -61,7 +61,12 @@ class MainHandler(BaseHandler):
...
+        else:
+            msgs = []
+        self.render("index.html", message=msg, level=lvl,
```
josenavas: `message` and `messages`: really easy to mess up
Can we change `messages` to `user_notifications`? It will make easier to avoid errors and it is way more descriptive.

qiita_pet/templates/user_messages.html
```diff
@@ -0,0 +1,10 @@
+{% extends sitebase.html %}
+{% autoescape None %}
+{% block content%}
+<h2 style="text-decoration: underline;">System Messages</h2>
+<ul>
```
josenavas: Fix indentation

josenavas:
Few comments. I've mixed feelings on having qiita_db using moi, but I guess if we are going to store information in redis rather in the database, it is ok to have it there as redis becomes a part of qiita_db.

qiita_db/user.py
```diff
@@ -463,6 +468,45 @@ def _change_pass(self, newpass):
...
 
+    def add_message(self, message):
+        """Add a message to the user's message queue
+
+        Parameters
+        ----------
+        message : str
+            Message to add to queue
+        """
+        count = r_client.lpush("%s:messages" % self._id, '%s\n%s' %
```
wasade: why "\n" as the delimiter? that's risky

josenavas: Good catch

squirrelo: Already done in other places. Why is it risky?

squirrelo:
I have mixed feelings on the div placement for system messages as I have it currently. My other thought is to have the dropdown from messages link in the navbar be the latest 5 messages, then click the messages link and get the rest.

qiita_pet/templates/user_messages.html
squirrelo: That is correct indentation.

qiita_db/user.py
squirrelo: Fair enough. Done.

qiita_pet/handlers/base_handlers.py
squirrelo: done

qiita_db/user.py
wasade: that's bad. it makes an incredible assumption about the content of the data.

squirrelo: fair enough.

qiita_pet/templates/user_messages.html
josenavas: No it's not. The indentation in the html files are 2 spaces. This is how it should look (accordingly to other files in qiita)
```html
<ul>
  {% for msg in messages %}
    <li>{% raw msg %}</li>
  {% end %}
</ul>
```

Also note that I think this should be `raw msg` instead of only `msg` since the msg is in html format, correct?

squirrelo: Ah, thought you meant just the line you commented on as that's all I had to go on. 

Autoescape None is set, so it will always print raw on this page. Makes sense considering what this page is.

josenavas: Uhm... that's not standard procedure across the code base... Unsure if it can generate problems in the future if we decide to add more functionality to this page. My vote goes to better be safe than sorry, so please change to follow what is more widely used in the system.

squirrelo:
Comments addressed and also made the change for the messages placement since to me it looks better and makes more sense.

qiita_pet/templates/sitebase.html
```diff
@@ -133,6 +133,14 @@
                 <li><a href="/study/list/">View Studies</a></li>
               </ul>
             </li>
+            <li class="dropdown">
+              <a href="/user/messages/">Messages</a>
+              <ul class="dropdown-menu">
+              {% for msg, timestamp in current_user.messages(5) %}
```
josenavas: Please, do not perform database accesses in the templates. There has been a fair amount of work to remove any kind of those from the templates. Why did you change this if it was already correct before?

squirrelo: because this is on sitebase, so without this doing it this way we need every single template call to pass messages, which is inefficient.

squirrelo: see line 144 of this very file.

josenavas:
This looks a bit weird when there are no messages. It should say something along the lines of `You don't have any messages`.

screen shot 2015-06-05 at 11 39 44 am
screen shot 2015-06-05 at 11 40 09 am

Also, I don't like the placement of the messages thing. I think it will make more sense to have a bootstrap envelope next to the user name (where it says `welcome foo@foo.bar`) and next to the envelope have a counter of the messages. It would be great to keep track of the messages that have been read and the ones that doesn't.
screen shot 2015-06-05 at 11 44 17 am

qiita_pet/templates/sitebase.html
josenavas: This can be easily accomplish with a UI module in which you insert that staff in the bar, so we can remove any database access in the template.

This should be applied for the entire navigation bar, but that is out of scope of the PR, but I think we should not be adding more potential issues to the code.

josenavas:
Re: keeping track of the ones that have been read and not. If it is easy to do I would like to see it implemented. If not, I'm happy with an issue and add it later.

squirrelo:
Count is going to be a pain, so I'l open an issue. Otherwise implementing now.

squirrelo:
ready for more review.

qiita_db/user.py
```diff
@@ -463,6 +469,42 @@ def _change_pass(self, newpass):
...
 
+    def add_message(self, message):
+        """Add a message to the user's message queue
+
+        Parameters
+        ----------
+        message : str
+            Message to add to queue
+        """
+        #
```
antgonza: Not needed, right?

squirrelo: Not even sure how that got there.

josenavas:
The comment with the empty messages has still not been resolved.

squirrelo:
The empty message thing is a side effect of using bootstrap. I can't think of a clean way to fix that.

qiita_pet/templates/sitebase.html
```diff
@@ -162,7 +162,15 @@
               <a href="/auth/logout/">Log Out</a>
             </li>
             <li class="dropdown">
-              <a href="/analysis/selected/" id="dflt-sel-info"><span class="glyphicon glyphicon-list-alt"></span> Selected</a>
+              <a href="/user/messages/"><span class="glyphicon glyphicon-envelope"></span></a>
+              <ul class="dropdown-menu">
+              {% for um, ts in current_user.messages(5) %}
```
josenavas:
```html
{% set user_msgs = current_user.messages(5) %}
{% if not user_msgs %}
  <li>You don't have any messages</span></li>
{% else %}
  {% for um, ts in user_msgs %}
    <li>&nbsp;{{um}}<br /><span style="font-size:60%;color:gray;">&nbsp;{{ts}}</span></li>
  {% end %}
{% end %}
```

josenavas:
I've just commented in the code how this can be done in case of the dropdown. something similar can be done in the page...

antgonza:
We are gonna be adding messages with html. These messages do not get displayed nicely in the feed ...

squirrelo:
fixed

josenavas:
I still don't see any fix for the `UserMessagesHander` when there are no messages... Please, read all the comments and perform all the changes or comment back...

squirrelo:
I have no idea what you are talking about. The UserMessagesHandler returns an empty list if there are no messages, and it displays properly.

antgonza:
Did you test your fix? It doesn't look right:
screen shot 2015-06-05 at 3 32 30 pm

josenavas:
What I'm talking about is that pages like this are not ok:
screen shot 2015-06-05 at 11 40 09 am

It is really easy to show a message like `There are no messages available`, and it highly improves the user feedback...

squirrelo:
both fixed

antgonza:
Thanks. What will be the best way to let a user know that there are unread messages? Blinking, adding a star or something similar will be nice.

squirrelo:
That would require the unread messages counting, which an issue is already open for.

josenavas:
I've been thinking a little bit about keeping track of the (un)read messages.

I think an easy way of doing it will be to keep 2 queues per user, one for unread messages and another for read messages. We can move from the unread queue to the read queue easily so this will solve the issue.

Note that we don't have a way to "path" the redis DB, so having a design close to the actual end design is desirable, as I don't know how we can change this later on.

Another issue that I found is that messages can be long, which doesn't look right. Limiting the number of characters in the dropdown is another thing that should be added. An example is here, and you can read the message itself to get more insight on why this is needed:
screen shot 2015-06-08 at 10 20 04 am

squirrelo:
The two lists thing is a bit of a pain since we need to describe what we consider "read."

For the long message thing, I can truncate at 100 or 140 characters, but that will cause issues with HTML in the strings so that will take some time to do proper parsing.

josenavas:
Doesn't matter how is the implementation, we still need to define how is "read".

The reason why I'm still discussing this is that if we need to change the structure in redis is not going to be trivial, so I don't want to add something quick if we now that it is not the ideal solution and we need something else.

For defining something as "read" I will suggest we add a button to the dropdown that says "mark as read", and the other one is when users go to the message list page, everything get's marked as read. The other solution is to always ask the user to mark something as read.

squirrelo:
I'm actually partial to doing it like twitter and facebook do, where if it's in the top bar and viewed, it's considered 'read.' And yes, implementation matters since how we define read influences how it's implemented.

josenavas:
Exactly, how we define it affects implementation, not the other way around ^^

josenavas:
and BTW, it makes sense what you say. Note that FB doesn't mark them as "read" you have to click the button to mark it as read unless you actually go to the page, which automatically marks it as read.

That being said, do you think that the 2 lists is a good solution?

squirrelo:
If it's an explicit "read" button, no. If it's marked as read the first time it's seen, yes. 

What I mean for facebook is if you click the little globe with messages in it, it marks all those read and takes the "new messages" trigger off of it.

josenavas:
Uhm... Facebook actually works in 2 steps. If you hit the globe, the counter goes away. However, the stories that you have not actively click or actively mark as read, they're still highlighted in blue. But this is something that we can decide.

In my opinion, and given how users can be, I like more the idea of actively asking the user to mark the message as read. What do others think?

antgonza:
I like mark (this is how banks work) ...

squirrelo:
My only issue with explicit mark as read button is that it can get cluttered and will need a 'mark all as read" which can cause people not to read things.

antgonza:
I do not think we actually need a button but as long as the user clicks on the message it could change color (for example, from light grey to white).

squirrelo:
Not feasible if you have many messages. We can't have it as a "click each message to dismiss" as some people may not log in for many months and have a ton of messages on next login.

antgonza:
I think that's fine: (1) if a user doesn't log for a while and wants to know what's new, that's a good way to figure it out; (2) hopefully we will not get those many messages; (3) we only store up to 100 messages. Another thing is that we can simply show up to 5 in the dropdown (but we show how many are unread) and the user can click on all of them in the full display. We can discuss more about this on Wed ...

squirrelo:
I've truncated the messages in the menu, and I think since the issue for read is open and it's obviously a larger problem than expected, this should be merge-able now.

josenavas:
How are you planning to "patch" the redis database if we need to change some of the information on the queues? That is my concern on merging this as it is...

squirrelo:
python patch

josenavas:
Uhm, I'm not sure. @biocore/qiita-dev can somebody else comment in here?

antgonza:
It sounds like #1245 will basically modify the lines that are in this PR and perhaps add a few ones. I think we should solve it here. Furthermore, it sounds like @squirrelo already knows what needs to be done and could simply add it here ...

josenavas:
Which is the status of this PR? It looks like it is missing fixing #1245

squirrelo:
Yes, that and setting up the websockets. Portals are more urgent so I'm focusing on that right now.

antgonza:
@squirrelo what's the status of this PR?

squirrelo:
Was working on things for american gut last week. Can start up on this one again now.

antgonza:
Great, thanks!

squirrelo:
Ready for review once tests pass. This now has notifications for new messages and a cleaner message interactivity.

qiita_db/user.py
```diff
@@ -509,6 +525,68 @@ def _change_pass(self, newpass):
...
+            sql_info = [self._id]
+            sql = """SELECT message_id, message, message_time, read
+                     FROM qiita.message_user
+                     JOIN qiita.message USING (message_id)
+                     WHERE email = %s ORDER BY message_time DESC"""
+            if count is not None:
+                sql += " LIMIT %s"
+                sql_info.append(count)
+            TRN.add(sql, sql_info)
+            return TRN.execute_fetchindex(-1)
```
josenavas: The `-1` is not needed, remove for consistency with other codebase and for magic numbers.

qiita_db/user.py
```diff
@@ -509,6 +525,68 @@ def _change_pass(self, newpass):
...
+            Message IDs to mark as read/unread
+        read : bool, optional
+            Marks as read if True, unread if False. Default True
+        """
+        with TRN:
+            sql = """UPDATE qiita.message_user
+                     SET read = %s
+                     WHERE message_id in %s and email = %s"""
+            TRN.add(sql, [read, tuple(messages), self._id])
+            return TRN.execute_fetchindex(-1)
```
josenavas: Same here

qiita_db/user.py
```diff
@@ -509,6 +525,68 @@ def _change_pass(self, newpass):
...
+        ----------
+        messages : list of ints
+            Message IDs to mark as read/unread
+        read : bool, optional
+            Marks as read if True, unread if False. Default True
+        """
+        with TRN:
+            sql = """UPDATE qiita.message_user
+                     SET read = %s
+                     WHERE message_id in %s and email = %s"""
```
josenavas: `in` -> `IN`
`and` -> `AND`

All sql reserved words should be capitalized.

qiita_db/user.py
```diff
@@ -509,6 +525,68 @@ def _change_pass(self, newpass):
...
+
+        Parameters
+        ----------
+        messages : list of ints
+            Message IDs to delete
+        """
+        with TRN:
+            # remove message from user
+            sql = """DELETE FROM qiita.message_user
+                     WHERE message_id in %s and email = %s"""
```
josenavas: `in` -> `IN`
`and` -> `AND`

qiita_db/user.py
```diff
@@ -441,6 +445,18 @@ def shared_analyses(self):
...
+    def unread_messages(self):
+        """Returns all unread messages for a user"""
+        with TRN:
+            sql = """SELECT message_id, message, message_time, read
+                     FROM qiita.message_user
+                     JOIN qiita.message USING (message_id)
+                     WHERE email = %s AND read = FALSE
+                     ORDER BY message_time DESC"""
+            TRN.add(sql, [self._id])
+            return TRN.execute_fetchindex(-1)
```
josenavas: The -1 is not needed, remove for consistency with other codebase and for magic numbers.

qiita_db/util.py
```diff
@@ -1098,3 +1099,25 @@ def infer_status(statuses):
...
+    message : str
+        Message to add
+    users : list of User objects
+        Users to connect the message to
+    """
+    with TRN:
+        sql = """INSERT INTO qiita.message (message) VALUES (%s)
+                 RETURNING message_id"""
+        TRN.add(sql, [message])
+        msg_id = TRN.execute_fetchflatten()[0]
```
josenavas: `TRN.execute_fetchflatten()[0]` -> `TRN.execute_fetchlast()`

qiita_db/util.py
```diff
@@ -1098,3 +1099,25 @@ def infer_status(statuses):
...
+        sql = """INSERT INTO qiita.message_user (email, message_id)
+                 VALUES (%s, %s)"""
+        for user in users:
```
josenavas: Change with:
```python
sql_args = [[user.id, msg_id] for user in users]
TRN.add(sql, sql_args, many=True)
```

josenavas:
Few comments in the code. GUI-wise, I found the following error:
josh_pr
The main issue is that there are no checks in place to see if the users has checked something or not:
```
  File "/Users/jose/.virtualenvs/qiita/lib/python2.7/site-packages/tornado/web.py", line 1346, in _execute
  File "/Users/jose/qiime_software/QiiTa/qiita_pet/handlers/user_handlers.py", line 166, in post
    self.current_user.delete_messages(messages)
  File "/Users/jose/qiime_software/QiiTa/qiita_db/user.py", line 588, in delete_messages
    TRN.execute()
  File "/Users/jose/qiime_software/QiiTa/qiita_db/sql_connection.py", line 474, in wrapper
    return func(self, *args, **kwargs)
  File "/Users/jose/qiime_software/QiiTa/qiita_db/sql_connection.py", line 798, in execute
    return self._execute()
  File "/Users/jose/qiime_software/QiiTa/qiita_db/sql_connection.py", line 745, in _execute
    self._raise_execution_error(sql, sql_args, e)
  File "/Users/jose/qiime_software/QiiTa/qiita_db/sql_connection.py", line 618, in _raise_execution_error
    % (sql, str(sql_args), str(error)))
ValueError: Error running SQL query:
Query: DELETE FROM qiita.message_user
                     WHERE message_id in %s and email = %s
Arguments: [(), 'test@foo.bar']
Error: syntax error at or near ")"
LINE 2:                      WHERE message_id in () and email = 'tes...
                                                  ^
```
Also, can you add a space between the envelope and the number? It will improve visibility:
screen shot 2015-08-03 at 1 56 30 pm

squirrelo:
Added the changes, ready to review again.

josenavas:
@squirrelo these 2 files got added to your commits, please remove and pull from upstream, since they're now included in the gitignore: `qiita_pet/static/vendor/js/moi.js`, `qiita_pet/static/vendor/js/moi_list.js`

squirrelo:
Done

josenavas:
Few more comments wrt the GUI. Now, although it doesn't fail, there is no feedback at all... I would suggest to use the bootstrap error bar on top to show that the user should at least choose one message to apply the functionality.

Another thing is that in the dropdown menu, there is no way to know which messages are read and which ones are not. Adding some stuff in there to differentiate those will be useful.
screen shot 2015-08-05 at 9 49 37 am

squirrelo:
It bolds unread in the dropdown. Guess that's not enough differentiation

squirrelo:
dropdown now greys out read messages. I don't think we need an error message for the actions not doing anything. They didn't select anything, so no action is taken, which should be intuitive.

josenavas:
Re no messages: that is not the strategy that we followed in other parts of the code, i.e. in the portals, if no study is selected, a message that at least one study should be selected is shown. This breaks consistency across the project. Please add message or either deactivate the buttons if no study is selected... "useless" buttons are always bad in GUI...

squirrelo:
Now shows/hides depending on checkbox status.

josenavas:
Thanks @squirrelo. Now, if the message is read and I mark it as read, nothing happens in the system, but since the buttons disappear and the checkboxes get reset, the user has feedback and it doesn't look like the system is blocked.

gif

Good job! :+1: @antgonza, @wasade, @ElDeveloper or @adamrp can either of you review?

antgonza:
Looks fine, a few questions/comments:
- Is there an easy way to add a message system wide?
- I'm not sure red is the color we want to use when we have unread messages, what about green?
- Is it possible to make the html tags active?
- If you use the "select all" button the "mark as read ..." menu is not shown.

squirrelo:
No way to add a message system wide other than just sending it to all users through the existing function in this pull reqeust. I thought that was what the messaging system through redis was for though?

I made it red because that's the color facebook, google, and a lot of other messaging people use. Plus, we already use green for the studies.

What do you mean "make the html tags active"?

Fixed that last one, about to push.

antgonza:
I think having both is a good idea. I see the redis messages as something we use for quick messages and this for a longer term message. For example, "The system is going to be updated in a week". If posted with redis it will be there for a week, all the time, vs. having it in this new interfase without being in the way of anything else. We can discuss during today's meeting.

Red: I see, note that the only part that is red is the number and not the full logo ...

See:
screen shot 2015-08-05 at 1 04 44 pm

squirrelo:
Number now is only red thing, and disappears if no unread messages. Also print raw message so html works.

antgonza:
Thanks. Note that you are missing the space between the envelop and the number ...

squirrelo:
Space is actually there...

antgonza:
Yes, I can see it in the code. I think it looks like is not there cause you have <span envelope></span><span messages></span> vs. <span envelope><span messages></span></span> but not sure.

BTW could you add these changes to the documentation? Some of the screenshots need to be replaced and you will need to add a section about messages.

squirrelo:
Admin messaging added, documentation forthcoming.

qiita_db/test/test_util.py
```diff
@@ -616,6 +619,32 @@ def test_check_access_to_analysis_result(self):
 
         self.assertEqual(obs, exp)
 
+    def test_add_message(self):
+        users = [User('shared@foo.bar'), User('admin@foo.bar')]
+        add_message("TEST MESSAGE", users)
+
+        obs = [[x[0], x[1]] for x in User('shared@foo.bar').messages()]
```
antgonza: If someone adds a message in the populate_test_db.sql file the 4 is gonna change, could you replace for a `count` and then `+ 1`? Same for the tests below.

josenavas:
@squirrelo ETA on adding documentation and note to the CHANGELOG.md?

squirrelo:
Tonight most likely

antgonza:
@squirrelo ping!

josenavas:
@squirrelo note that this is modifying the DB so you're potentially holding more development (there are still some issues in alpha 0.2 that require DB changes, not sure if we are going to do them for the upcoming release, but better not to hold potential development).

squirrelo:
All the code works, and it's only docs that are needed, so if this is holding stuff up you can merge and docs can be a separate pull request. It's actually probably better to do it this way, as we still don't know what will message and how this system will be used, only what it looks like. Might be better to write the documentation once we have actual use cases.

antgonza:
I'm OK with merging as long as the documentation changes soon (before deploy) mainly cause some of the current screenshots of the site do not match with the deploy. However, I can see that we can not add how it works until after we start adding functionality ...

squirrelo:
I just flipped through and none of the current images are affected by this pull request, so is fine to merge.

antgonza:
I think you are right. Thanks for checking. :+1:

josenavas:
@squirrelo can you add a comment to CHANGELOG.md?

josenavas:
Also, I think it will be useful to have a small tutorial before doing the release. Since we are targeting that release soon (ideally by the end of this week), I would suggest to have the tutorial in the following days. @squirrelo will that be possible?

squirrelo:
Tutorial for what?

josenavas:
If we have the admin message system, it will be good to have a piece of user documentation (i.e. a tutorial or something similar) that explain why admin messages cannot be removed. The general interaction with the messaging system is easy enough, but that aspect is different in our system from other messaging systems.

squirrelo:
That was not what we discussed, just that systemwide messages can be added. Currently you can delete system messages, they just get sent out to everyone but you can delete them from your queue. If we don't want users to do that, I can add that functionality.

josenavas:
Please pay attention in meetings. That was discussed on the last meeting and I specifically ask for that functionally...

squirrelo:
You asked that system messages not be deleteable so new users may get the messages, not that users couldn't delete them from their queue. Please pay attention to system architecture when making requests, so that they are clear. I will add the functionality though.

squirrelo:
Functionality added.

qiita_db/test/test_user.py
```diff
@@ -353,6 +361,53 @@ def test_change_forgot_password_bad_code(self):
...
+        obs = user.messages()
+        exp = [True, True, False]
+        self.assertEqual([x[3] for x in obs], exp)
+
+        user.mark_messages([1], read=False)
+        obs = user.messages()
+        exp = [False, True, False]
+        self.assertEqual([x[3] for x in obs], exp)
+
+    def test_delete_messages(self):
```
josenavas: Can you add a test for making sure that admin messages cannot be deleted by the user?

qiita_db/test/test_user.py
```diff
@@ -353,6 +361,53 @@ def test_change_forgot_password_bad_code(self):
...
+        user = User('test@foo.bar')
+        user.delete_messages([1, 2])
+        obs = user.messages()
+        exp_msg = [(3, 'message <a href="#">3</a>')]
+        self.assertEqual([(x[0], x[1]) for x in obs], exp_msg)
+
+        sql = ("SELECT message_id FROM qiita.message")
```
josenavas: Extra parenthesis

qiita_db/test/test_user.py
```diff
@@ -353,6 +361,53 @@ def test_change_forgot_password_bad_code(self):
...
+                'Nunc vel mi accumsan, porttitor eros ut, pharetra elit. Nulla'
+                ' ac nisi quis dui egestas malesuada vitae ut mauris. Morbi '
+                'blandit non nisl a finibus. In erat velit, congue at ipsum '
+                'sit amet, venenatis bibendum sem. Curabitur vel odio sed est '
+                'rutrum rutrum. Quisque efficitur ut purus in ultrices. '
+                'Pellentesque eu auctor justo.'),
+            (3, 'message <a href="#">3</a>')]
+        self.assertEqual([(x[0], x[1]) for x in obs], exp_msg)
+        self.assertTrue(all(x[2] < datetime.now() for x in obs))
+        self.assertFalse(all(x[3] for x in obs))
```
josenavas: You are missing the test for `x[4]` as there are 5 elements on the returned list:
```
In [1]: from qiita_db.user import User

In [2]: u = User('test@foo.bar')

In [3]: u.messages()
Out[3]: 
[[1L,
  'message 1',
  datetime.datetime(2015, 8, 10, 12, 57, 17, 499881),
  False,
  False],
 [2L,
  'Lorem ipsum dolor sit amet, consectetur adipiscing elit. Pellentesque sed auctor ex, non placerat sapien. Vestibulum vestibulum massa ut sapien condimentum, cursus consequat diam sodales. Nulla aliquam arcu ut massa auctor, et vehicula mauris tempor. In lacinia viverra ante quis pellentesque. Nunc vel mi accumsan, porttitor eros ut, pharetra elit. Nulla ac nisi quis dui egestas malesuada vitae ut mauris. Morbi blandit non nisl a finibus. In erat velit, congue at ipsum sit amet, venenatis bibendum sem. Curabitur vel odio sed est rutrum rutrum. Quisque efficitur ut purus in ultrices. Pellentesque eu auctor justo.',
  datetime.datetime(2015, 8, 10, 12, 57, 17, 499881),
  False,
  False],
 [3L,
  'message <a href="#">3</a>',
  datetime.datetime(2015, 8, 10, 12, 57, 17, 499881),
  False,
  False]]
```

qiita_db/user.py
```diff
@@ -509,6 +530,81 @@ def _change_pass(self, newpass):
...
+            [(msg_id, msg, timestamp, read, system_message), ...]
+
+        Notes
+        -----
+        system_message is a bool. When True, this is a systemwide message.
+        """
+        with TRN:
+            sql_info = [self._id]
+            sql = """SELECT message_id, message, message_time, read,
+                     (expiration IS NOT NULL) as system_message
```
josenavas: `as` -> `AS` and add an extra indent here to be compliant with the new SQL coding guidelines.

qiita_db/user.py
```diff
@@ -509,6 +530,81 @@ def _change_pass(self, newpass):
...
+        Parameters
+        ----------
+        messages : list of ints
+            Message IDs to delete
+        """
+        with TRN:
+            # remove message from user
+            sql = """DELETE FROM qiita.message_user
+                     WHERE message_id IN %s AND email = %s
+                     AND message_id NOT IN
```
josenavas: Add an extra indentation here and the next 2 lines, for SQL guidelines compliance

qiita_db/user.py
```diff
@@ -509,6 +530,81 @@ def _change_pass(self, newpass):
...
+                        (SELECT message_id FROM qiita.message
+                         WHERE expiration IS NOT NULL)"""
+            TRN.add(sql, [tuple(messages), self._id])
+            # Remove any messages that no longer are attached to a user
+            # and are not system messages
+            sql = """DELETE FROM qiita.message
+                     WHERE message_id NOT IN
+                     (SELECT DISTINCT message_id FROM qiita.message_user
```
josenavas: Indent this block.

josenavas:
Few comments. @squirrelo can you also add a CLI command to add a system message? That functionality will mainly be performed by us when we are planning a system update, so having at least a CLI will be desirable.

qiita_db/test/test_user.py
squirrelo: done

qiita_db/test/test_user.py
squirrelo: That is tested in `test_mark_messages` since 4 is if the message is read or not.

squirrelo:
I've added a `--banner` flag to the `sysmessage` maintenance command. If set, it will show as a site banner like before, otherwise it will send messages as a standard system message.

josenavas:
@squirrelo errors

squirrelo:
Should be good to go

antgonza:
Thanks, looks great! 

Some suggestions:
- What about selecting a message when clicking on it? This should make easier selecting messages
- Not sure if the user needs to see the milliseconds, what do you think? 
- Could you either make the messages be selected even if they were system wide or make them expire (read and deleted) after the time is gone?

squirrelo:
I can do the first two, but the third is what the discussion above is about. I'll wait to make any more changes on this until the meeting today, since apparently there are two differing opinions on how to take care of system messages.

squirrelo:
Added suggestions. Ready for review.

qiita_db/test/test_user.py
josenavas: @squirrelo In test_mark_messages you also have `x[3]`. `x[4]` is not present anywhere. Anyways I think you should test all the fields in this function.

josenavas: Just checked the code. And you previous comment is wrong. `x[3]` is the one that tells you if the message is read or not. `x[4]` tells you if the message is a sys message or not. Please update the tests to reflect this and add a sys message to be able to test that all the functionality is working as expected.

qiita_db/test/test_util.py
```diff
@@ -616,6 +619,47 @@ def test_check_access_to_analysis_result(self):
...
+        self.assertEqual(obs, exp)
+
+    def test_clear_system_messages(self):
+        obs = [[x[0], x[1]] for x in User('shared@foo.bar').messages()]
+        exp = [[1, 'message 1']]
+        self.assertEqual(obs, exp)
+
+        add_system_message("SYS MESSAGE", datetime(2015, 8, 5, 19, 41))
+        obs = [[x[0], x[1]] for x in User('shared@foo.bar').messages()]
+        exp = [[1, 'message 1'], [4, 'SYS MESSAGE']]
```
josenavas: Remove the magic number 4 and use count, as in previous tests.

josenavas: BTW, you're missing the `self.assertEqual(obs, exp) here and in the test below... so you're not testing anything right now.

qiita_db/util.py
```diff
@@ -1098,3 +1099,60 @@ def infer_status(statuses):
...
+        msg_id = TRN.execute_fetchlast()
+        sql = """INSERT INTO qiita.message_user (email, message_id)
+                 SELECT email, %s FROM qiita.qiita_user"""
+        TRN.add(sql, [msg_id])
+        TRN.execute()
+
+
+def clear_system_messages():
+    with TRN:
+        sql = """SELECT message_id FROM qiita.message WHERE expiration < %s"""
```
josenavas: You don't need triple quotes here as it fits in a single line. It applies to the other 2 SQL commands below.

scripts/qiita
```diff
@@ -384,13 +386,20 @@ def unlock():
               help='The amount of time to show message, in seconds')
 @click.option('--message', required=True, type=str,
               help="Message to show users")
-def sysmessage(message, time):
-    r_client.setex('sysmessage', message, time)
+@click.option('--banner', is_flag=True,
+              help="If set, show as site banner instead of a standard message")
+def sysmessage(message, time, banner):
```
josenavas: I'm gonna leave this to other devs, but the `--time` parameter makes sense for the redis system. However, for the messaging system that we are putting in place, I think the most natural input will be to provide hour, minutes, day, month and year. Either in a specified format or in multiple parameters.

I'm bringing this up because I saw today that when we create a system message we end up putting a really big number so we can clear it up afterwards... this makes the functionality effectively useless as we are not going to count the seconds that we need until the day that we are gonna do the upgrade...

josenavas:
Few more comments @squirrelo

squirrelo:
This is good to go.

scripts/qiita
squirrelo: This may be a discussion, but it is well outside the already ever-increasing scope of features in this pull request. To keep it from getting any bigger, and to make the deadline, please open an issue and we can discuss there.

antgonza: I think is something we need, however I agree with @squirrelo about opening an issue.

squirrelo:
If there's nothing else besides the time discussion, can this get merged?

scripts/qiita
josenavas: Done: https://github.com/biocore/qiita/issues/1421

qiita_db/test/test_user.py
```diff
@@ -302,6 +307,11 @@ def test_verify_code(self):
                  WHERE email = 'new@test.bar' AND dflt = true"""
         self.assertEqual(self.conn_handler.execute_fetchone(sql)[0], 2)
 
+        # Make sure new system messages are linked to user
+        sql = """SELECT message_id FROM qiita.message_user
+                 WHERE email = 'new@test.bar'"""
+        self.assertEqual(self.conn_handler.execute_fetchall(sql), [[5]])
```
josenavas: Please remove magic number and use a `count` variable.

squirrelo: done

josenavas:
@squirrelo if there are no system messages and you call clear_sysmessage, an error is raised:
```
$ qiita maintenance clear_sysmessage
/Users/jose/.virtualenvs/qiita/lib/python2.7/site-packages/IPython/parallel/client/client.py:446: RuntimeWarning: 
            Controller appears to be listening on localhost, but not on this machine.
            If this is true, you should specify Client(...,sshserver='you@137.110.95.142')
            or instruct your controller to listen on an external IP.
  RuntimeWarning)
  File "/Users/jose/qiime_software/QiiTa/scripts/qiita", line 498, in <module>
    qiita()
  File "/Users/jose/.virtualenvs/qiita/lib/python2.7/site-packages/click/core.py", line 664, in __call__
  File "/Users/jose/.virtualenvs/qiita/lib/python2.7/site-packages/click/core.py", line 644, in main
    rv = self.invoke(ctx)
  File "/Users/jose/.virtualenvs/qiita/lib/python2.7/site-packages/click/core.py", line 991, in invoke
  File "/Users/jose/.virtualenvs/qiita/lib/python2.7/site-packages/click/core.py", line 991, in invoke
  File "/Users/jose/.virtualenvs/qiita/lib/python2.7/site-packages/click/core.py", line 837, in invoke
  File "/Users/jose/.virtualenvs/qiita/lib/python2.7/site-packages/click/core.py", line 464, in invoke
  File "/Users/jose/qiime_software/QiiTa/scripts/qiita", line 402, in clear_sysmessage
    clear_system_messages()
  File "/Users/jose/qiime_software/QiiTa/qiita_db/util.py", line 1158, in clear_system_messages
    TRN.execute()
  File "/Users/jose/qiime_software/QiiTa/qiita_db/sql_connection.py", line 474, in wrapper
    return func(self, *args, **kwargs)
  File "/Users/jose/qiime_software/QiiTa/qiita_db/sql_connection.py", line 798, in execute
    return self._execute()
  File "/Users/jose/qiime_software/QiiTa/qiita_db/sql_connection.py", line 745, in _execute
    self._raise_execution_error(sql, sql_args, e)
  File "/Users/jose/qiime_software/QiiTa/qiita_db/sql_connection.py", line 618, in _raise_execution_error
    % (sql, str(sql_args), str(error)))
ValueError: Error running SQL query:
Query: DELETE FROM qiita.message_user WHERE message_id IN %s
Arguments: [()]
Error: syntax error at or near ")"
LINE 1: DELETE FROM qiita.message_user WHERE message_id IN ()
                                                            ^
```

The usage of tuple is pretty cool, but you should remember to check if there is actually some data in the tuple before calling the SQL command. In the last few PR, this is the most common issue that I found in your code, so probably it can be good to write down a note or some other technique that will help you to remember to check this.

squirrelo:
Fixed, ready to go.

qiita_db/test/test_util.py
```diff
@@ -616,6 +619,50 @@ def test_check_access_to_analysis_result(self):
...
+        obs = [[x[0], x[1]] for x in User('admin@foo.bar').messages()]
+        exp = [[count, 'SYS MESSAGE']]
+        self.assertEqual(obs, exp)
+
+        sql = "SELECT expiration from qiita.message WHERE message_id = %s"
+        obs = self.conn_handler.execute_fetchall(sql, [count])
+        exp = [[datetime(2015, 8, 5, 19, 41)]]
+        self.assertEqual(obs, exp)
+
+    def test_clear_system_messages(self):
```
josenavas: Please add a test with empty system messages, as it is always useful to have tests for bugs that we already found.

squirrelo: test added

josenavas:
:+1: @antgonza do you mind going over this one and merge if the tests pass and you don't have any further comments?

PR #1246, opened 2015-06-05
Showing the raw data of all accessible studies, not only the owned ones

josenavas:
The current code only allows to add the raw data to a prep template if the user owns the raw data. It should be if the user has access.

This is blocking people work, please a fast review is appreciated.

### DIFF
--- a/qiita_pet/uimodules/prep_template_tab.py
+++ b/qiita_pet/uimodules/prep_template_tab.py
@@ -35,16 +35,18 @@ fp_type_by_ft = defaultdict(
 
 
 def _get_accessible_raw_data(user):
-    """Retrieves a tuple of raw_data_id and the last study title for that
+    """Retrieves a tuple of raw_data_id and one study title for that
     raw_data
     """
     d = {}
     accessible_studies = user.user_studies.union(user.shared_studies)
     for sid in accessible_studies:
-        for rdid in Study(sid).raw_data():
+        study = Study(sid)
+        study_title = study.title
+        for rdid in study.raw_data():
             rdid = int(rdid)
             if rdid not in d:
-                d[rdid] = Study(RawData(rdid).studies[-1]).title
+                d[rdid] = study_title
     return d
 
 
--- a/qiita_pet/uimodules/prep_template_tab.py
+++ b/qiita_pet/uimodules/prep_template_tab.py
@@ -39,9 +39,11 @@ def _get_accessible_raw_data(user):
     raw_data
     """
     d = {}
-    for sid in user.user_studies:
+    accessible_studies = user.user_studies.union(user.shared_studies)
+    for sid in accessible_studies:
         for rdid in Study(sid).raw_data():
-            d[int(rdid)] = Study(RawData(rdid).studies[-1]).title
+            if rdid not in d:
+                d[int(rdid)] = Study(RawData(rdid).studies[-1]).title
     return d
 
 
--- a/qiita_pet/uimodules/prep_template_tab.py
+++ b/qiita_pet/uimodules/prep_template_tab.py
@@ -44,9 +44,7 @@ def _get_accessible_raw_data(user):
         study = Study(sid)
         study_title = study.title
         for rdid in study.raw_data():
-            rdid = int(rdid)
-            if rdid not in d:
-                d[rdid] = study_title
+            d[int(rdid)] = study_title
     return d
 
 
--- a/qiita_pet/uimodules/prep_template_tab.py
+++ b/qiita_pet/uimodules/prep_template_tab.py
@@ -42,8 +42,9 @@ def _get_accessible_raw_data(user):
     accessible_studies = user.user_studies.union(user.shared_studies)
     for sid in accessible_studies:
         for rdid in Study(sid).raw_data():
+            rdid = int(rdid)
             if rdid not in d:
-                d[int(rdid)] = Study(RawData(rdid).studies[-1]).title
+                d[rdid] = Study(RawData(rdid).studies[-1]).title
     return d
 
 


qiita_pet/uimodules/prep_template_tab.py
```diff
@@ -39,9 +39,11 @@ def _get_accessible_raw_data(user):
     raw_data
     """
     d = {}
```
antgonza: d = defaultdict(list), right?

wasade: the present pattern wouldn't benefit from a `defaultdict` from what i can see

qiita_pet/uimodules/prep_template_tab.py
```diff
@@ -39,9 +39,11 @@ def _get_accessible_raw_data(user):
...
-    for sid in user.user_studies:
+    accessible_studies = user.user_studies.union(user.shared_studies)
+    for sid in accessible_studies:
         for rdid in Study(sid).raw_data():
-            d[int(rdid)] = Study(RawData(rdid).studies[-1]).title
+            if rdid not in d:
```
wasade: this test is not accurate since the key is cast prior to insertion, and type does matter as `str(1) != int(1)`

josenavas: Good catch, changed

qiita_pet/uimodules/prep_template_tab.py
josenavas: No, note that I'm not creating a list. The only reason why I'm asking if the rdid exists in the dictionary or not is to avoid all the DB accesses performed by this call:
```python
d[int(rdid)] = Study(RawData(rdid).studies[-1]).title
```
which are at least 4.

josenavas:
comments addresed

qiita_pet/uimodules/prep_template_tab.py
```diff
@@ -39,9 +39,12 @@ def _get_accessible_raw_data(user):
...
     """
     d = {}
-    for sid in user.user_studies:
+    accessible_studies = user.user_studies.union(user.shared_studies)
+    for sid in accessible_studies:
         for rdid in Study(sid).raw_data():
-            d[int(rdid)] = Study(RawData(rdid).studies[-1]).title
+            rdid = int(rdid)
+            if rdid not in d:
+                d[rdid] = Study(RawData(rdid).studies[-1]).title
```
squirrelo: This seems overly complicated and wasteful. Why can't we just use the study ID and study object instantiated above in Study(sid).raw_data(), get the title from that as well, then use that title for the raw data?

josenavas: @antgonza this line of code was extracted from what you had before. Do you have any reason why this was done in this way, instead of using the already available study?

squirrelo: Also doing it the way I proposed allows use of defaultdict.

josenavas: Again, no need of a defaultdict...

antgonza: Not sure but note that you are getting all raw datas of which a Study has access and then adding the title of where it was used lastly. Just to be clear, not sure what that gives us.

josenavas: Ok, since there is no reason to do this, I'm going to change this to go the more efficient way. I noticed this when I was reworking this part of the code, but I preferred to leave it as it was rather than add more and more changes to an already gigantic PR. Thanks for the discussion.

qiita_pet/uimodules/prep_template_tab.py
```diff
@@ -35,13 +35,18 @@
...
-    for sid in user.user_studies:
-        for rdid in Study(sid).raw_data():
-            d[int(rdid)] = Study(RawData(rdid).studies[-1]).title
+    accessible_studies = user.user_studies.union(user.shared_studies)
+    for sid in accessible_studies:
+        study = Study(sid)
+        study_title = study.title
+        for rdid in study.raw_data():
+            rdid = int(rdid)
+            if rdid not in d:
```
squirrelo: You can actually defaultdict this now, since it is technically part of whatever study you have it in. If the raw data ID gets over-ridden it is still part of the study that overrides the existing one. Also removes the if statement.

josenavas: I can remove the if statement if it bothers you. However, still no need for the defaultdict...

squirrelo: it's not needed, so better to remove it.

josenavas:
Ready for review/merge.

antgonza:
:+1:

qiita_pet/uimodules/prep_template_tab.py
```diff
@@ -35,13 +35,16 @@
...
     """
     d = {}
-    for sid in user.user_studies:
-        for rdid in Study(sid).raw_data():
-            d[int(rdid)] = Study(RawData(rdid).studies[-1]).title
+    accessible_studies = user.user_studies.union(user.shared_studies)
+    for sid in accessible_studies:
+        study = Study(sid)
+        study_title = study.title
+        for rdid in study.raw_data():
```
wasade: syntactic, non-blocking, and purely for note: this could be done as `d.update({int(rdid): study_title for rdid in study.raw_data()})`

...out of curiosity, I benched these, and the insert is faster.

```python
19:48:26 (mcdonadt@Daniels-MBP-3):/Users/mcdonadt
> def insert():
    d = {}
    for i in xrange(100):
        for j in xrange(100):
            d[(i, j)] = i
   ...:

19:49:22 (mcdonadt@Daniels-MBP-3):/Users/mcdonadt
> def merge():
   ...:     d = {}
   ...:     for i in xrange(100):
   ...:         d.update({(i, j): i for j in xrange(100)})
   ...:

19:49:51 (mcdonadt@Daniels-MBP-3):/Users/mcdonadt
> %timeit insert()
1000 loops, best of 3: 1.49 ms per loop

19:50:02 (mcdonadt@Daniels-MBP-3):/Users/mcdonadt
> %timeit merge()
1000 loops, best of 3: 1.97 ms per loop
```

josenavas: Thanks @wasade 

I tried creating a generator that yields (rdid, study_title) and just using
````python
d = {rdid: title for rdid, title in generator}
```

but with a real database, the accesses to the DB are so much expensive that the time differences due to this are minimal.

ackermag: Is this deployed? And, if not when?

On Sat, Jun 6, 2015 at 7:33 PM, josenavas <notifications@github.com> wrote:

> In qiita_pet/uimodules/prep_template_tab.py
> <https://github.com/biocore/qiita/pull/1246#discussion_r31873261>:
>
> >      raw_data
> >      """
> >      d = {}
> > -    for sid in user.user_studies:
> > -        for rdid in Study(sid).raw_data():
> > -            d[int(rdid)] = Study(RawData(rdid).studies[-1]).title
> > +    accessible_studies = user.user_studies.union(user.shared_studies)
> > +    for sid in accessible_studies:
> > +        study = Study(sid)
> > +        study_title = study.title
> > +        for rdid in study.raw_data():
>
> Thanks @wasade <https://github.com/wasade>
>
> I tried creating a generator that yields (rdid, study_title) and just using
>
> d = {rdid: title for rdid, title in generator}
>
> but with a real database, the accesses to the DB are so much expensive
> that the time differences due to this are minimal.
>
> —
> <https://github.com/biocore/qiita/pull/1246/files#r31873261>.
>



-- 
Gail Ackermann
Knight Lab
UCSD

josenavas: @antgonza ?

antgonza: Done

PR #1249, opened 2015-06-08
DO NOT MERGE: Adding database layout for Single Strain WGS

josenavas:
@RNAer can you check if I missed something?

@biocore/qiita-dev This is the base db layout for the Single Strain WGS. As you'll seem, it is not affecting any part of the code so the tests should pass.

With this in master, @RNAer, me and whoever wants to work on this section will be able to add all the needed bits to add this pipeline to Qiita. Once this is merged, I will create a branch on upstream and a dummy PR with a list of all the things that need to be added for this pipeline to be in place.

As you know, DB changes conflicts are hard to resolve, and there is a fair amount of work on re-doing this PR (not hard, but really time consuming). so I will appreciate fast reviews on this.

Thanks!

### DIFF
--- a/qiita_db/support_files/patches/26.sql
+++ b/qiita_db/support_files/patches/26.sql
@@ -94,9 +94,9 @@ CREATE TABLE qiita.annotation_filepath (
 	CONSTRAINT idx_annotation_filepath PRIMARY KEY ( annotation_id, filepath_id )
  ) ;
 
-CREATE INDEX idx_annotation_filepath ON qiita.annotation_filepath ( annotation_id ) ;
+CREATE INDEX idx_annotation_filepath_0 ON qiita.annotation_filepath ( annotation_id ) ;
 
-CREATE INDEX idx_annotation_filepath ON qiita.annotation_filepath ( filepath_id ) ;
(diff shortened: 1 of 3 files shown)


RNAer:
:+1:

josenavas:
This is ready for review/merge.

antgonza:
We should discuss this in the next meeting, some points to consider:
- Should we create a new schema for these?
- Should we create a new DB for these?
- What about making these tables/code a plug in for qiita?

josenavas:
Open to discussion. Just want to highlight that the tables are not fully independent of the rest of the system, they are using some tables that are already present on the system (e.g. sample, prep template, raw_data, and PreprocessedData).

This means that new schema or new DB are not possible because it will not allow the ability of FK to these tables.

Re: plug in. If this becomes a plugin, I would suggest that any pipeline should be a plugin. This starts smelling to QIIME 2, so we should be careful with this.

antgonza:
Cool, let's talk about it ... the idea of start working in QIIME 2 "smells" great!

mortonjt:
Noticed that this hasn't been touched for a while.  What is needed for this?

josenavas:
This no longer applies - it can be closed - This needs to be completely reworked and done as a plugin. The main blocker for this project is micronota.

PR #1252, opened 2015-06-09
Merging code from glowing dangerzone - BLOCKS DEVELOPMENT - please review

josenavas:
This PR is merging the code from glowing dangerzone.

The code in SQLConnectionHandler was a bit hard to follow in some cases because there was no logical structure on the code. I've moved some functions around and github make a fair good job here, it looks like it only had a problem with the `execute_queue` function.

The functions `fetchone_with_types` and `fetchall_with_types` have been removed. They're not used anywhere and nobody knows why they're there.

I've rewritten some of the tests and added a BUNCH of new tests, that is one of the reasons why it shows so many changes.

Once this is merged, I will start implementing all the actual improvements to the SQLConnectionHandler that will require changes in the code base - but that will be done in a different branch.

As a reminder, all this changes are needed to get the portal types in place.


josenavas:
Ready for review. This is blocking development.

qiita_db/sql_connection.py
```diff
@@ -100,6 +101,18 @@ def flatten(listOfLists):
 
 
 class SQLConnectionHandler(object):
+    """Encapsulates the DB connection with the Postgres DB
```
ElDeveloper: Maybe -> "Postgres DB connection object"?

qiita_db/sql_connection.py
```diff
@@ -100,6 +101,18 @@ def flatten(listOfLists):
...
+
+    Parameters
+    ----------
+    admin : {'no_admin', 'no_database', 'database'}, optional
```
ElDeveloper: These values don't match with the options described in the docstring.

qiita_db/sql_connection.py
```diff
@@ -100,6 +101,18 @@ def flatten(listOfLists):
...
+        Whether or not to connect as the admin user. Options other than
+        `no_admin` depend on admin credentials in the qiita configuration. If
```
ElDeveloper: Minor but you are using a mixture of ` and of ' to refer to the options.

qiita_db/sql_connection.py
```diff
@@ -255,33 +256,30 @@ def get_postgres_cursor(self):
         except PostgresError as e:
             raise QiitaDBConnectionError("Cannot get postgres cursor! %s" % e)
 
-    def set_autocommit(self, on_or_off):
-        """Sets the isolation level to autocommit or default (read committed)
```
ElDeveloper: Would be good to move this docstring to the autocommit property. Otherwise this information is being lost.

qiita_db/sql_connection.py
```diff
@@ -255,33 +256,30 @@ def get_postgres_cursor(self):
...
-        else:
-            level = ISOLATION_LEVEL_READ_COMMITTED
+    @property
+    def autocommit(self):
+        return self._connection.isolation_level == ISOLATION_LEVEL_AUTOCOMMIT
 
+    @autocommit.setter
+    def autocommit(self, value):
+        if not isinstance(value, bool):
+            raise TypeError("The value for autocommit should be a boolean")
```
ElDeveloper: This exception should be documented in the autocommit docstring.

qiita_db/sql_connection.py
```diff
@@ -320,158 +318,55 @@ def _sql_executor(self, sql, sql_args=None, many=False):
...
-            Default False
+            The SQL query
+        sql_args : tuple, list or dict, optional
+            The arguments for the SQL query
 
         Raises
         ------
-        KeyError
-            queue does not exist
+        QiitaDBExecutionError
```
ElDeveloper: I'm OK with this as per our discussions before, but I want to remark (again) that with this we explicitly say that people shouldn't use `SQLConnectionHandler` outside of Qiita.

qiita_db/sql_connection.py
```diff
@@ -320,158 +318,55 @@ def _sql_executor(self, sql, sql_args=None, many=False):
...
+            if there is some error executing the SQL query
 
         Notes
         -----
-        Queues are executed in FIFO order
+        From psycopg2 documentation, only variable values should be bound
```
ElDeveloper: Would be nice to include a `References` section with that link.

qiita_db/sql_connection.py
```diff
@@ -320,158 +318,55 @@ def _sql_executor(self, sql, sql_args=None, many=False):
...
-        else:
-            self._check_sql_args(sql_args)
-            self.queues[queue].append((sql, sql_args))
+        with self._sql_executor(sql, sql_args):
+            pass
 
-    def execute_fetchall(self, sql, sql_args=None):
-        """ Executes a fetchall SQL query
+    def executemany(self, sql, sql_args_list):
+        """ Executes an executemany SQL query with no results
```
ElDeveloper: There's a leading space at the beginning of this docstring.

qiita_db/sql_connection.py
```diff
@@ -479,15 +374,13 @@ def execute_fetchall(self, sql, sql_args=None):
 
         Notes
         -----
-        from psycopg2 documentation, only variable values should be bound
+        From psycopg2 documentation, only variable values should be bound
```
ElDeveloper: Same as above.

qiita_db/sql_connection.py
```diff
@@ -521,22 +414,20 @@ def execute_fetchone(self, sql, sql_args=None):
 
         return result
 
-    def fetchall_with_types(self, sql, sql_args=None):
-        """Executes a fetchall SQL query with column information
+    def execute_fetchall(self, sql, sql_args=None):
+        """ Executes a fetchall SQL query
```
ElDeveloper: There's a leading space in this docstring.

qiita_db/sql_connection.py
```diff
@@ -545,99 +436,156 @@ def fetchall_with_types(self, sql, sql_args=None):
...
-                     for desc in pgcursor.description}
 
-        return result, types
+        return result
 
-    def fetchone_with_types(self, sql, sql_args=None):
-        """Executes a fetchone SQL query with column information
+    def _check_queue_exists(self, queue_name):
+        if queue_name not in self.queues:
+            raise KeyError("Queue %s does not exist" % queue_name)
```
ElDeveloper: For the sake of clarity with whitespaces, etc you may want to do `"Queue '%s' does not exist"`.

qiita_db/sql_connection.py
```diff
@@ -545,99 +436,156 @@ def fetchall_with_types(self, sql, sql_args=None):
...
+        self.queues[queue_name] = []
 
-    def execute(self, sql, sql_args=None):
-        """ Executes an SQL query with no results
+    def list_queues(self):
+        """Returns list of all queue names currently in handler
+
+        Returns
+        -------
+        list
```
ElDeveloper: list of str ... correct?

qiita_db/sql_connection.py
```diff
@@ -545,99 +436,156 @@ def fetchall_with_types(self, sql, sql_args=None):
...
+        for args in sql_args:
+            self._check_sql_args(args)
+            self.queues[queue].append((sql, args))
+
+    def _rollback_raise_error(self, queue, sql, sql_args, e):
+        self._connection.rollback()
+        # wipe out queue since it has an error in it
+        del self.queues[queue]
+        raise QiitaDBExecutionError(
+            ("\nError running SQL query in queue %s: %s\nARGS: %s\nError: %s"
```
ElDeveloper: Why do we need the extra parenthesis here and the leading `'\n'`?

qiita_db/sql_connection.py
```diff
@@ -545,99 +436,156 @@ def fetchall_with_types(self, sql, sql_args=None):
...
-            via sql_args, it shouldn't be used to set table or field names. For
-            those elements, ordinary string formatting should be used before
-            running execute.
+        Queues are executed in FIFO order
         """
-        with self._sql_executor(sql, sql_args_list, True):
-            pass
+        self._check_queue_exists(queue)
+
+        with self.get_postgres_cursor() as cur:
```
ElDeveloper: This is a rather complex block, are we sure that it cannot be decomposed/restructured in any other way?

ElDeveloper:
A few comments, overall looks good and the tests are passing!

qiita_db/sql_connection.py
squirrelo: agree, I can't quite tell whats going on here.

qiita_db/sql_connection.py

qiita_db/sql_connection.py
josenavas: Good catch. Done

qiita_db/sql_connection.py

qiita_db/sql_connection.py
josenavas: Yeah, it was not documented on GD and I forgot to move the docs. Thanks!

qiita_db/sql_connection.py

qiita_db/sql_connection.py
josenavas: What do you mean by this? Note that the SQLConnectionHandler is part of Qiita, so if they are using this code they are using Qiita. If at some point we actually need this code outside of Qiita, we can then create a separate project. The problem right now is that the SQLConnectionHandler still needs a fair amount of development, and it is hard to maintain the SQLConnectionHandler and Qiita up to date if they're 2 different projects. Once all this code is more or less stable, we can think if this can be move to another project if we need it.

qiita_db/sql_connection.py

qiita_db/sql_connection.py

qiita_db/sql_connection.py

qiita_db/sql_connection.py
josenavas: Grr sorry about that! Done

qiita_db/sql_connection.py
squirrelo: Similar to MOI, where we try and keep a logical separation even if it is being developed inside qiita.

qiita_db/sql_connection.py
josenavas: Done - good catch

qiita_db/sql_connection.py
josenavas: yup!

qiita_db/sql_connection.py
josenavas: Good catch! I just move this function around, good to know that reviewing code that has been already reviewed improves it!

qiita_db/sql_connection.py
ElDeveloper: @josenavas, yes exactly what you mention is what I was referring to!

josenavas: @squirrelo MOI is a separate project, so there is a difference.
@ElDeveloper I can remove any kind of QiitaDB errors and use standard python errors so if we actually want to move it to its own project will be easier. What do you think?

qiita_db/sql_connection.py
josenavas: Yeah... I will try to decompose this. When even its own original developer doesn't know what is going on in the code (@squirrelo) means that the code is really convoluted. Note that I just moved this function around :smile:

qiita_db/sql_connection.py
ElDeveloper: I think that this should be a concern, if we think this will be moving
outside of Qiita at some point.

josenavas: Ok, removing any QiitaDB error and using standard python ones.

qiita_db/sql_connection.py
josenavas: Thinking a little bit more about this, I need to re-work this block of code in a future PR (which I will start working on it as soon as this one gets merged). Can we leave it as it is acknowledging that is going to disappear soon anyways?

josenavas:
Ready for review again!

adamrp:
@josenavas Looks like there are legitimate failures (something to do with `group` being an SQL reserved term, I think?)

josenavas:
thanks @adamrp 
the issue is that we change some of the errors that are raised in the connection handler, so we use only standard python errors. Other parts of the code rely on those errors to work correctly, so I change those parts of the code to catch the new error types.

I've also checked the interface and I make sure that everywhere that `QiitaDBExecutionError` was catch, `ValueError` is also catch.

josenavas:
BTW, there was a bug in one of the tests, but it was masked by the assertRaises so when we changed the type of the errors from the sql handler I found it and fix it.

qiita_db/sql_connection.py
ElDeveloper: OK

ElDeveloper:
Thanks @josenavas :+1: 

One thing though, with the exceptions that you found that needed to be updated, are there any others where this update needs to happen as well? These places would be untested and thus not showing up as errors when the test suite is run.

josenavas:
I did a grep and search for all of them in the code. I think I covered all of them :-)

ElDeveloper:
fantastic, :+1: :fireworks:

josenavas:
@adamrp, can you review?

qiita_db/sql_connection.py
```diff
@@ -504,140 +410,212 @@ def execute_fetchone(self, sql, sql_args=None):
...
 
         Raises
         ------
-        QiitaDBExecutionError
-            if there is some error executing the SQL query
+        KeyError
+            If the queue `queue_name` does not exist in the handler
+        """
+        if queue_name not in self.queues:
+            raise KeyError("Queue '%s' does not exist" % queue_name)
```
adamrp: This function seems a little weird to me. It checks whether a queue exists, which I would think returns a boolean. Instead, if it exists, None is returned, and if it doesn't exist, an exception is raised.

qiita_db/sql_connection.py
```diff
@@ -504,140 +410,212 @@ def execute_fetchone(self, sql, sql_args=None):
...
+        ------
+        KeyError
+            Queue name already exists
         """
-        with self._sql_executor(sql, sql_args) as pgcursor:
-            result = pgcursor.fetchone()
-            types = {desc[0]: self.TYPE_CODES[desc[1]]
-                     for desc in pgcursor.description}
+        if queue_name in self.queues:
+            raise KeyError("Queue %s already exists" % queue_name)
```
adamrp: might as well use the `_check_queue_exists` function here, right?

qiita_db/sql_connection.py
```diff
@@ -504,140 +410,212 @@ def execute_fetchone(self, sql, sql_args=None):
...
+        KeyError
+            queue does not exist
 
-        Note: from psycopg2 documentation, only variable values should be bound
-            via sql_args, it shouldn't be used to set table or field names. For
-            those elements, ordinary string formatting should be used before
-            running execute.
+        Notes
+        -----
+        Queues are executed in FIFO order
```
adamrp: This note seems unnecessary, since queues are by definition FIFO, but I guess it wouldn't hurt to keep it here

qiita_db/sql_connection.py
```diff
@@ -504,140 +410,212 @@ def execute_fetchone(self, sql, sql_args=None):
...
+            self._check_sql_args(args)
+            self.queues[queue].append((sql, args))
+
+    def _rollback_raise_error(self, queue, sql, sql_args, e):
+        self._connection.rollback()
+        # wipe out queue since it has an error in it
+        del self.queues[queue]
+        raise ValueError(
+            "Error running SQL query in queue %s: %s\nARGS: %s\nError: %s"
+            % (queue, sql, str(sql_args), e))
```
adamrp: For these exceptions, it might be useful to use [`mogrify`](http://initd.org/psycopg/docs/cursor.html#cursor.mogrify) to get the exact query that it was trying to execute. Not blocking though

qiita_db/sql_connection.py
```diff
@@ -504,140 +410,212 @@ def execute_fetchone(self, sql, sql_args=None):
...
-            The arguments for the SQL query
+        queue : str
+            Name of queue to execute
+
+        Notes
+        -----
+        Does not support executemany command. Instead, enter the multiple
+        SQL commands as multiple entries in the queue.
+
+        Queues are executed in FIFO order
```
adamrp: same comment about FIFO note here

qiita_db/sql_connection.py
```diff
@@ -504,140 +410,212 @@ def execute_fetchone(self, sql, sql_args=None):
...
         ------
-        QiitaDBExecutionError
-            If there is some error executing the SQL query
-
-        Note: from psycopg2 documentation, only variable values should be bound
-            via sql_args, it shouldn't be used to set table or field names. For
-            those elements, ordinary string formatting should be used before
-            running execute.
+        IndexError
+            If there a placeholder cannot be replaced with a real value
```
adamrp: I don't understand this explanation; maybe a word was skipped?

adamrp:
I added a few comments, but I don't think any of them should be considered blocking.

qiita_db/sql_connection.py
josenavas: I think you're right, semantically your proposal makes more sense that the code that I have. I will change that.

qiita_db/sql_connection.py

qiita_db/sql_connection.py
josenavas: Removed - easy to maintain

qiita_db/sql_connection.py
josenavas: We had mogrify previously in the code and we end up removing it because it was masking some errors, i.e. mogrify was also raising an error so we where not able to check the actual error. I think it is safer to leave as it is, but if you feel strong about it, I can change.

qiita_db/sql_connection.py
josenavas: Removed

qiita_db/sql_connection.py
adamrp: No, this is fine -- did not remember running into those issues, but I can how that would be problematic. I guess you could wrap mogrify in a try/except (and on exception print what is currently being printed), but it's not a big deal

qiita_db/sql_connection.py
josenavas: Done.

josenavas:
Thanks @adamrp 
I think I've addressed all your comments!

adamrp:
:+1: when tests pass. Thanks @josenavas !

josenavas:
Tests passed!!

PR #1253, opened 2015-06-10
fix #1240

### LINKED ISSUE #1240: When study completes pre-processing there is no longer an option to reprocess
If I preprocess a study and it is successful but the sequence counts are very low, I cannot reprocess and select reverse complement barcodes.  I have to reload the prep-template re-link the sequence files and start all over (This happened in study 10178

antgonza:


josenavas:
I think this is ok given that we need to find a fast solution for this and it is blocking people work. However, I think this is showing that we need to pay more attention to that page and probably do some kind of refactoring so it simplifies the template logic (or even moves it to the handler).

I think this falls in the broader group of the qiita_pet refactoring, so :+1: for this PR.

ElDeveloper:
:+1:

PR #1254, opened 2015-06-11
URGENT: Allowing preprocessing single-lane raw data that do not have run prefix

josenavas:
@ackermag found today that she cannot preprocess a study that had a RawData with a single lane because the PrepTemplate was missing the run_prefix.

There was a bug in the way the constraints was checked, as it was assuming that only the forward reads and the barcodes where added. If the reverse reads were added, the test fail and asked for the run_prefix.

This PR fix that problem. This is blocking the processing of current studies so having this merged and deployed is really urgent. @antgonza @ElDeveloper @squirrelo can I have a fast review of this?

### DIFF
--- a/qiita_pet/uimodules/prep_template_tab.py
+++ b/qiita_pet/uimodules/prep_template_tab.py
@@ -191,7 +191,9 @@ class PrepTemplateInfoTab(BaseUIModule):
                     "linked with the Raw Data")
             else:
                 if prep_template.data_type() in TARGET_GENE_DATA_TYPES:
-                    key = ('demultiplex_multiple' if len(raw_data_files) > 2
+                    raw_forward_fps = [fp for _, fp, ftype in raw_data_files
+                                       if ftype == 'raw_forward_seqs']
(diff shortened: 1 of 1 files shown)


antgonza:
:+1:

ElDeveloper:
:+1: the code looks good to me, but I don't know how to reproduce the bug and check that the problem is gone.

josenavas:
Add a prep template without the run_prefix column, then add a RawData with 3 files (forward, reverse and barcodes). In the old version, it will not allow you to preprocess it and it will say that you need the run_prefix column, which is not true because you only have one Illumina lane.

ElDeveloper:
I tried doing that, but I still see the same error. Maybe it will be
easier if we look at it for 5 mins and try to figure out what's going
on.

josenavas:
Sure, let me know when do you have some time and we go over.

ElDeveloper:
Any time after 2:00 or before 1:00 should work.

ElDeveloper:
Thanks a lot!

PR #1256, opened 2015-06-11
Portals db changes

squirrelo:


josenavas:
@squirrelo failing tests

squirrelo:
Ready for review.

josenavas:
@squirrelo what is missing here to get portal types?
Also, can you provide descriptions of what is done here? In a complex issue like this one is better if we have short descriptions of what is done, what is expected, what needs to be done, etc, specially if you expect other devs to help you with this issue...

squirrelo:
This is the first pull request to start the branch. I can explain whats going in this specifically:
1) Adding a `portal-type_id` with fk to `portal_type` table to the analyses, so they are also tied to a portal. 
2) updating the analysis objects and objects dependent on the analysis tables to reflect this change.

josenavas:
Thanks @squirrelo 
That makes the review process easier. Once this PR is merged, create a PR from the `biocore:portals` branch to master and list there everything that is missing to get portal types up and ready. Thanks for getting this started!

qiita_core/support_files/config_test.txt
```diff
@@ -36,6 +36,9 @@ BASE_DATA_DIR =
 # Valid upload extension, comma separated. Empty for no uploads
 VALID_UPLOAD_EXTENSION = fastq,fastq.gz,txt,tsv,sff,fna,qual
 
+#Portal the site is working under
+PORTAL = QIIME_EMP
```
josenavas: The default should be `QIITA`

squirrelo: That does not currently exist in the DB. I can add it though.

qiita_db/analysis.py
```diff
@@ -107,7 +108,8 @@ def get_by_status(cls, status):
         return {x[0] for x in conn_handler.execute_fetchall(sql, (status,))}
 
     @classmethod
-    def create(cls, owner, name, description, parent=None, from_default=False):
+    def create(cls, owner, name, description, portal, parent=None,
```
josenavas: I don't think you need to pass the portal. The portal should be automatically retrieved from the config file.

qiita_db/analysis.py
```diff
@@ -266,6 +275,21 @@ def name(self):
         return conn_handler.execute_fetchone(sql, (self._id, ))[0]
 
     @property
+    def portal(self):
```
josenavas: Do we actually want to retrieve the portal as a property? In my opinion, the portal is something that happens in a transparent way in the system.

squirrelo: Can be useful for debugging, and is a small thing that is easy to maintain. I'd rather have access to it through the object if for nothing else than an API measure.

qiita_pet/handlers/analysis_handlers.py
```diff
@@ -75,7 +76,7 @@ def post(self):
         name = self.get_argument('name')
         desc = self.get_argument('description')
         analysis = Analysis.create(self.current_user, name, desc,
-                                   from_default=True)
+                                   qiita_config.portal, from_default=True)
```
josenavas: If you are going to populate it anyways from the DB, better to have it already embedded on qiita db....

qiita_db/analysis.py
squirrelo: yeah, makes sense. I'll fix this in all the tests too then.

qiita_pet/handlers/analysis_handlers.py
squirrelo: can make it part of the config object. That way I can also make sure the portal passed actually exists before starting the web server.

qiita_core/support_files/config_test.txt
josenavas: We currently have this values:
```
 portal_type_id |  portal   |  portal_description   
----------------+-----------+-----------------------
              1 | QIIME     | QIIME portal
              2 | EMP       | EMP portal
              3 | QIIME_EMP | QIIME and EMP portals
```

I would suggest changing these to:
```
 portal_type_id |  portal   |  portal_description   
----------------+-----------+-----------------------
              1 | QIITA     | QIITA portal
              2 | EMP       | EMP portal
```

Note that we don't need a `QIITA_EMP` since QIITA has access to everything. The first one `QIITA` is used to represent does studies that do not belong to any other portal.

qiita_db/analysis.py
josenavas: That's a good reason. Thanks.

qiita_pet/handlers/analysis_handlers.py
josenavas: It is already part of the config object. But I agree, it will be extremely useful to fail at the moment that Qiita starts if the portal type does not exist. That can be the last step of the config reading. This way, even if the configuration of the DB is wrong, it will not allow the system to start.

squirrelo:
Ready for another review.

qiita_db/support_files/patches/26.sql
```diff
@@ -0,0 +1,23 @@
...
+ADD CONSTRAINT fk_analysis_portal
+FOREIGN KEY (portal_type_id)
+REFERENCES qiita.portal_type(portal_type_id);
+
+-- Drop defaut value constraint so it is now required to give portal type
+ALTER TABLE qiita.analysis
+ALTER portal_type_id DROP DEFAULT;
+
+-- Remove existing portals and replace with more relevant ones
+DELETE FROM qiita.portal_type WHERE portal = 'QIIME_EMP';
```
josenavas: Since we are going to start using this actively, we need to make sure that the studies already present in the database has the correct portal type. Are we sure that removing this will not leave the DB in an inconsistent state?

josenavas:
One concern about the live database. Otherwise :+1:

qiita_db/support_files/patches/26.sql
squirrelo: It will not remove it if there are FK to it, so we should be safe. The patch will error though.

squirrelo: I've added an update before the delete that will take care of the issue. Now should be 100% OK for live system.

squirrelo:
Took care of that one comment, so needs one more reviewer.

qiita_db/support_files/patches/26.sql
```diff
@@ -0,0 +1,25 @@
...
+ADD CONSTRAINT fk_analysis_portal
+FOREIGN KEY (portal_type_id)
+REFERENCES qiita.portal_type(portal_type_id);
+
+-- Drop defaut value constraint so it is now required to give portal type
+ALTER TABLE qiita.analysis
+ALTER portal_type_id DROP DEFAULT;
+
+-- Remove existing portals and replace with more relevant ones
+UPDATE qiita.study SET portal_type_id = 1 WHERE portal_type_id = 3;
```
antgonza: Should we only replace `portal_type_id = 3` or all?

antgonza: Also, this will mean that a study can only belong to one portal, which I'm not sure is what we need/want. I can see having 4 portals: qiita (all studies), emp, sloan, hmp2 and having studies that belong to the 3 of the 4 (qiita, sloan, hmp2).

antgonza:
A few of comments.

qiita_db/support_files/patches/26.sql
rob-knight: It’s critical that the same study can be seen in multiple portals.

josenavas: The qiita portal is implicit, all studies belong to the study portal because all the studies can be seen from Qiita. However, the other case (sloan and hmp2) it is not currently supported. I was unaware that a single study can belong to more than one portal (w/o taking into account Qiita) so this needs more changes in the DB... Good that we figure this out before it got merged!

squirrelo: More join tables! Fixed.

squirrelo:
Portals have now been broken out to their own table for studies. Also added accessory functions to the study object for getting studies by portal and manipulating what portals a study is in.

qiita_db/study.py
```diff
@@ -133,6 +133,8 @@ class Study(QiitaObject):
     preprocessed_data
     processed_data
     add_pmid
+    add_portal
+    remove_portal
```
antgonza: you are missing get_by_portal, right?

qiita_db/support_files/patches/26.sql
antgonza: The update is not needed, right? What about the next delete?

qiita_db/study.py
```diff
@@ -340,6 +360,17 @@ def create(cls, owner, title, efo, info, investigation=None):
                "efo_id) VALUES (%s, %s)".format(cls._table))
         conn_handler.executemany(sql, [(study_id, e) for e in efo])
 
+        # insert portal into database
+        portal_id = convert_to_id(qiita_config.portal, 'portal_type', 'portal')
+        portal_sql = ""
+        if qiita_config.portal != "QIITA":
```
antgonza: Do we really need to add studies to qiita? I think we do not need to add portal qiita as we should be able to select every study, right?

qiita_db/study.py
squirrelo: class level, so not in this list

antgonza: OK, thanks.

qiita_db/study.py
squirrelo: It allows for very quick grabbing of all studies in the portal, and allows us to have a universal way of getting studies by just asking for studies by portal. Since it's a one-to-many anyway, this ease of use is worth it. Every study belongs to QIITA as well as whatever other portal.

qiita_db/support_files/patches/26.sql
squirrelo: Update makes it so any study with invalid portal type is fixed, then all portal types moved to the new table. It is needed.

qiita_db/study.py
```diff
@@ -206,6 +207,26 @@ def get_by_status(cls, status):
 
         return studies
 
+    @staticmethod
+    def get_by_portal(portal):
```
josenavas: I'm nervous about adding this kind of functions. This will make the portal types explicit instead of implicit, which is not what we want with portal types.

The way that I'm thinking this is when we invoke an iterator over all the studies, we get the ones only for the portal that is accessing. As this is getting right now, is going to require a lot of changes in the code base and make the entire system aware of the portal.

I'm strongly against this kind of functions...

josenavas:
Given the latest changes on this PR, I remove my :+1: and this PR gets now a :-1: 
I think this implementation is highly overcomplicating the issue and there are better ways of doing it (as we already discussed in the meetings...).

qiita_db/study.py
squirrelo: This is here because it will allow much fewer changes to the code base. For example, we can just add a single set intersection check instead of modifying every single SQL query interacting with the studies.

josenavas: I don't think that is the correct way of doing it. Also, performance wise, is going to kill us. That is a big no-no-no from my end.

squirrelo: I can remove the function then, but once removed the pull request is good to go.

josenavas: Is not only this function. I disagree with other parts. With the add/remove portal functions each portal can give itself access to all the studies; which I disagree. Only QIITA should be able to change the portal - this way Qiita becomes also de admin part of the database.

The EMP portal should not know that there are other portals. It only sees part of the data and it has the illusion that everything that it can see is the entire DB.

qiita_db/analysis.py
josenavas: Changed my mind about this. We don't want to use this across the code base and it is only intended for debugging. So this is a "DO NOT USE" in the codebase, I suggest changing it's name to use a double leading underscore to identify that is super-private.

josenavas:
Given the actual disagreements on this PR, we should discuss this in tomorrow's PR.

squirrelo:
Remember though, this is at the DB level. We need an API level way to add and remove studies from a portal, get by portal, etc if for nothing else than the admin side of things. At the -ware and -pet levels, yes, it should be transparent, but we still need a way on -db to say "Yes, this study belongs to these 5 portals" or "What studies belong to this portal?"

josenavas:
Agree with being able to ask those questions, however disagree on exposing this as part of the object API - at the moment that you expose the API it becomes non-transparent for qiita-ware and qiita-pet...

qiita_db/analysis.py
wasade: Please do not use double underscores as they act unexpectedly. A single underscore is fine. If this is a debug only method, it could include a check that raises if the environment is not in debug mode.

```python
> class foo(object):
   ....:     def __bar(self):
   ....:         print self
   ....:

16:04:29 (mcdonadt@Daniels-MacBook-Pro-3):/Users/mcdonadt/ResearchWork/greengenes_release/gg_13_8_otus/trees
> x = foo()

16:04:31 (mcdonadt@Daniels-MacBook-Pro-3):/Users/mcdonadt/ResearchWork/greengenes_release/gg_13_8_otus/trees
> x.__bar()
AttributeError                            Traceback (most recent call last)
<ipython-input-14-883553e2ea7d> in <module>()
----> 1 x.__bar()

AttributeError: 'foo' object has no attribute '__bar'
```

josenavas: That is not unexpected. Accordingly to the [python documentation](https://docs.python.org/2/tutorial/classes.html#private-variables-and-class-local-references):
> Any identifier of the form __spam (at least two leading underscores, at most one trailing underscore) is textually replaced with _classname__spam


```python
In [21]: class foo(object):
   ....:     def __bar(self):
   ....:         print self
   ....:         

In [22]: x = foo()

In [23]: x.__bar()
AttributeError                            Traceback (most recent call last)
<ipython-input-23-883553e2ea7d> in <module>()
----> 1 x.__bar()

AttributeError: 'foo' object has no attribute '__bar'

In [24]: x._foo__bar()
<__main__.foo object at 0x103f9d1d0>
```
which I think is what we want. Adding a check is not ok as it will not allow us to use it to debug the live system.

josenavas:
@squirrelo and I discussed this offline. A summary:
The main issue is that once a user gets into the system, it shouldn't be anything in the system that allow him/her to know that there are other portals sharing the same database. Thus, exposing anything related to portals through the API will break this behavior.

In an ideal world, we would have a site that will be the portal administrator site, so we can change things from portal to portal, add portals, etc. In our specific case, however, this site is Qiita, so the solution will be to have everything portal related in its own file and the functions will be defined/imported if and only if the portal is Qiita (which will then become also our portal administrator site).

squirrelo:
Changes have been made, tests should pass, so good for review.

qiita_db/analysis.py
wasade: I know how it works, but it is unexpected behavior as the instance is
inconsistent with the definition and the name change is implicit. Since the
purpose of this method is debug, flag it as such in the doc string or even
the function name if you want to be explicit (eg `_DEBUG_foo`. Using a
double underscore will only confuse a developer, which is the last thing
you want to do when you need to urgently use debug methods.

squirrelo:
@biocore/qiita-admin Anyone have time to review this?

josenavas:
I will review shortly.

squirrelo:
Ok, assuming tests pass this now has a table for analysis portals as well, so should be good for another review.

qiita_db/analysis.py
```diff
@@ -160,6 +162,22 @@ def create(cls, owner, name, description, parent=None, from_default=False):
             conn_handler.add_to_queue(
                 queue, sql, (owner.id, name, description, status_id))
 
+        # make changes to SQL needed to add to both QIITA and given portal
+        insert_portal = ""
```
josenavas: This entire block of code can be replaced by:
```python
sql = """INSERT INTO qiita.analysis_portal (analysis_id, portal_type_id)
         VALUES (%s, %s)"""
args = [('{0}', portal_id)]
if qiita_config.portal != 'QIITA':
    qp_id = convert_to_id('QIITA', 'portal_type', 'portal')
    args.append(('{0}', qp_id))
conn_handler.add_to_queue(queue, sql, args, many=True)
```
which is easy to understand and maintain.

qiita_db/analysis.py
```diff
@@ -160,6 +162,22 @@ def create(cls, owner, name, description, parent=None, from_default=False):
...
+        portal_info = []
+        if qiita_config.portal != 'QIITA':
+            qp = convert_to_id('QIITA', 'portal_type', 'portal')
+            insert_portal = '(%s, {}), '.format(qp)
+            portal_info.append("{0}")
+        portal_info.extend(["{0}", portal_id, "{0}"])
+
+        # insert into qiita_portal table
```
josenavas: Also, this comment is wrong

qiita_db/analysis.py
josenavas: Note that I removed the 'returning' clause, I don't think it is needed...

qiita_db/portal.py
```diff
@@ -0,0 +1,53 @@
+from .sql_connection import SQLConnectionHandler
```
josenavas: This file is missing the copyright.

qiita_db/portal.py
```diff
@@ -0,0 +1,53 @@
...
+from .util import convert_to_id
+from qiita_core.qiita_settings import qiita_config
+
+# Only make functions available for main portal
+if qiita_config.portal == "QIITA":
+    def get_studies_by_portal(portal):
+            """Returns study id for all Studies belonging to a portal
```
josenavas: All the functions in this file are double-indented, please fix.

qiita_db/portal.py
```diff
@@ -0,0 +1,53 @@
+from .sql_connection import SQLConnectionHandler
+from .util import convert_to_id
+from qiita_core.qiita_settings import qiita_config
+
+# Only make functions available for main portal
+if qiita_config.portal == "QIITA":
+    def get_studies_by_portal(portal):
```
josenavas: Given the nature of the functions here, it looks like it makes sense to create a `Portal` object. This will allow us to create portals from the "Admin" Qiita interface by calling `Portal.create(name)`. The functions here will then become:
 - `get_studies_by_portal` -> property `studies`
 - `add_study_to_portal` -> `add_study`
 - `remove_study_from_portal` -> `remove_study`

qiita_db/study.py
```diff
@@ -340,6 +337,17 @@ def create(cls, owner, title, efo, info, investigation=None):
                "efo_id) VALUES (%s, %s)".format(cls._table))
         conn_handler.executemany(sql, [(study_id, e) for e in efo])
 
+        # insert portal into database
+        portal_id = convert_to_id(qiita_config.portal, 'portal_type', 'portal')
```
josenavas: Please, replace this block of code with a variation of the code that I wrote for the analysis object, as it is conceptually doing the same thing. It will also remove magic numbers and will be consistent.

Also the comment above is incorrect, as it is not adding any portal.

qiita_db/support_files/patches/26.sql
```diff
@@ -0,0 +1,52 @@
+-- June 11,, 2015
+-- Adds portal type to analyses
```
josenavas: Please, improve description of the patch. It's doing a lot more things that just adding portal type to the analysis object.

qiita_db/support_files/patches/26.sql
```diff
@@ -0,0 +1,52 @@
...
+
+-- Remove existing portals and replace with more relevant ones
+UPDATE qiita.study SET portal_type_id = 1 WHERE portal_type_id = 3;
```
josenavas: This is actually incorrect. If a study previously belonged to the `QIIME_EMP` portal, means that now it belongs to the EMP portal (because all of them belong to the QIITA portal by default):
```sql
UPDATE qiita.study SET portal_type_id = 2 WHERE portal_type_id = 3;
```

Also, the comment is misleading, as this line is not replacing the portal with relevant ones.

qiita_db/support_files/patches/26.sql
```diff
@@ -0,0 +1,52 @@
...
+
+CREATE INDEX idx_study_portal_0 ON qiita.study_portal ( portal_type_id );
+
+COMMENT ON TABLE qiita.study_portal IS 'Controls what studies are visible on what portals';
+
+ALTER TABLE qiita.study_portal ADD CONSTRAINT fk_study_portal FOREIGN KEY ( study_id ) REFERENCES qiita.study( study_id );
+
+ALTER TABLE qiita.study_portal ADD CONSTRAINT fk_study_portal_0 FOREIGN KEY ( portal_type_id ) REFERENCES qiita.portal_type( portal_type_id );
+
+INSERT INTO qiita.study_portal (study_id, portal_type_id) SELECT s.study_id, s.portal_type_id FROM qiita.study s;
```
josenavas: This is not correct. If the study was attached to the `EMP` portal, after the patch it will not be attached to the QIITA portal. Please modify.

qiita_db/support_files/populate_test_db.sql
```diff
@@ -30,15 +30,18 @@ INSERT INTO qiita.study_person (name, email, affiliation, address, phone) VALUES
...
     study_abstract, vamps_id) VALUES
     ('test@foo.bar', 2, '2014-05-19 16:10', NULL, 1, 1, TRUE, TRUE,
-    '2014-05-19 16:11', 27, 27, 2, 3, FALSE, FALSE,
+    '2014-05-19 16:11', 27, 27, 3, FALSE, FALSE,
     'Identification of the Microbiomes for Cannabis Soils', 'Cannabis Soils', 'Analysis of the Cannabis Plant Microbiome',
     'This is a preliminary study to examine the microbiota associated with the Cannabis plant. Soils samples from the bulk soil, soil associated with the roots, and the rhizosphere were extracted and the DNA sequenced. Roots from three independent plants of different strains were examined. These roots were obtained November 11, 2011 from plants that had been harvested in the summer. Future studies will attempt to analyze the soils and rhizospheres from the same location at different time points in the plant lifecycle.',
     NULL);
 
+-- Add portal to the study
+INSERT INTO qiita.study_portal (study_id, portal_type_id) VALUES (1, 2);
```
josenavas: Shouldn't the study also be added to portal 1 (The qiita portal)? The idea of having a test database is to be as realistic as possible to the real data.

qiita_db/test/test_portal.py
```diff
@@ -0,0 +1,33 @@
+from unittest import TestCase, main
+
+from qiita_core.util import qiita_test_checker
+from qiita_db.portal import (
+    add_study_to_portal, remove_study_from_portal, get_studies_by_portal)
+from qiita_db.study import Study
+
+
+@qiita_test_checker()
```
josenavas: All the tests in here will fail if in the config file the portal is not defined as Qiita. Please, enclose the entire file in an if statement.

qiita_db/user.py
```diff
@@ -232,10 +233,15 @@ def create(cls, email, password, info=None):
         # create user default sample holder
         sql = ("INSERT INTO qiita.analysis "
                "(email, name, description, dflt, analysis_status_id) "
-               "VALUES (%s, %s, %s, %s, 1)")
+               "VALUES (%s, %s, %s, %s, 1) RETURNING analysis_id")
         conn_handler.add_to_queue(queue, sql,
                                   (email, '%s-dflt' % email, 'dflt', True))
 
+        portal_id = convert_to_id('QIITA', 'portal_type', 'portal')
```
josenavas: This is not correct, it is adding the dflt analysis to the Qiita portal. However, all the other portals will not have a default analysis, which will make this to fail.

josenavas:
@squirrelo more comments.

qiita_db/user.py
squirrelo: That is something we need to discuss. The default to me is supposed to reside on qiita only, but do we then need defaults for each portal in order to track things correctly?

josenavas: Is the interface going to work if the portal is EMP and there is no default analysis available? Users will not be able to select the samples from the study listing because there is no analysis attached.

squirrelo: So then yes, every portal will need a default analysis attached to work properly.

josenavas: :smile:

josenavas: Also, if the user has selected some samples in the EMP portal, and another ones in the Qiita portal, we want those to be independent.

qiita_db/analysis.py
squirrelo: Returning is needed for analysis_chain to insert correctly.

josenavas: Not in this one. This is linking the analysis to the portal, so here you already have the id. At least, it didn't look like you're using the returned value for anything.

squirrelo: It's how you need to do it for queues until your pull request goes in.

josenavas: Ok, in such case add the returning to the above `INSERT`, but this code needs to be refactored anyways.

qiita_db/portal.py
squirrelo: I don't like this idea. Since portals are rarely going to be added, it's a command line admin piece that can be done using straight SQL.

josenavas: @ElDeveloper @antgonza @wasade @adamrp 
Any comments in here? I think it will make the design much cleaner and it will make our lives easier in the long run.

squirrelo: As we previously discussed, these are essentially helper functions for the `qiita-env` command line commands so to me it makes no sense to put them into an object.

ElDeveloper: I think it's fine either way, though we certainly have been using the
design that @josenavas suggests all over the place.

squirrelo:
This is ready for another review

josenavas:
@squirrelo tests are failing.

squirrelo:
One test counting rows that has no bearing on anything right now in the
pull request. Ready for review.

josenavas:
Please fix, I'll review, but the tests are expected to pass in this PR.

qiita_db/environment_manager.py
```diff
@@ -233,9 +233,10 @@ def make_environment(load_ontologies, download_reference, add_demo_user):
              'Demo', 'Qitta Dev', '1345 Colorado Avenue', '303-492-1984')""")
         conn.execute("""
             INSERT INTO qiita.analysis (email, name, description, dflt,
-                                        analysis_status_id)
+                                        analysis_status_id, portal_type_id)
```
josenavas: This is not correct. AFAIK, the qiita.analysis table doesn't have a porta_type_id column.

qiita_db/support_files/patches/26.sql
```diff
@@ -0,0 +1,76 @@
...
+		SELECT portal_type_id from qiita.portal_type WHERE portal_type_id != 1
+	LOOP
+		INSERT INTO qiita.analysis (email, name, description, dflt, analysis_status_id) VALUES (eml, eml || '-dflt', 'dflt', true, 1) RETURNING analysis_id INTO aid;
+		INSERT INTO qiita.analysis_workflow (analysis_id, step) VALUES (aid, 2);
+		INSERT INTO qiita.analysis_portal (analysis_id, portal_type_id) VALUES (aid, portal);
+	END LOOP;
+END LOOP;
+END $do$;
+
+-- Remove study portal info to it's own table, as one study can be in multiple portals
```
josenavas: `Remove` -> `Move`

qiita_db/user.py
josenavas: @squirrelo ping on this comment, it doesn't look like it has been addressed.

josenavas:
Done with another pass.

qiita_db/user.py
squirrelo: It has, adds a user to every portal now in the patch. Will do the same for user creation, but that is another pull request.

squirrelo:
Ready for another review

qiita_db/user.py
josenavas: It should be only 2 lines of code, but since it's going to it's own branch is fine... I guess today is Monday... 
I've added a note on the WIP so we don't forget about this...

josenavas: Just a general comment. Note that you are doing partial changes here. I would be more happy if you actually do all the changes rather than just half of it. This way of working is really dangerous as there is a high probability that this slips through... but that's why I made the note on the WIP PR.

Also, it is better if you comment that this is going to be solved in another PR rather than just ignoring the comment...

squirrelo: Plus tests, etc. And more than 2 lines of code since it's in a queue.

qiita_db/environment_manager.py
```diff
@@ -231,11 +231,18 @@ def make_environment(load_ontologies, download_reference, add_demo_user):
...
-        conn.execute("""
+        analysis_id = conn.execute_fetchone("""
             INSERT INTO qiita.analysis (email, name, description, dflt,
                                         analysis_status_id)
             VALUES
-            ('demo@microbio.me', 'demo@microbio.me-dflt', 'dflt', 't', 1)""")
+            ('demo@microbio.me', 'demo@microbio.me-dflt', 'dflt', 't', 1)
+            RETURNING analysis_id
+        """)[0]
+        portal_id = convert_to_id(qiita_config.portal, 'portal_type', 'portal')
```
josenavas: What if `qiita_config.portal` is not `QIITA`? This code is incomplete. The default analysis has to be added to all the available portals...

@squirrelo please pay attention at the code that you're writing...

qiita_db/user.py
josenavas: Ok, maybe 4 plus tests.

The important point is that you're modifying this in order to fix something, and you're letting the fix not completed and saying that this will go in another PR.

the way of working here is not that we are going to review half of your commits. We are going to review full functionality. You already had a lot of examples in how atomicity is accomplished in this type of complex issues. First, solving the main part, and then keep improving each of the files independently with complete changes. What you're doing here is not a good way of programming.

Either you remove all the changes you've done in this function and you do it in the next PR, or you implement all the changes in here.

squirrelo: I can't make that change until your transaction code is added. Please get that merged before you ask for more refactors of queues.

josenavas: Remove any modification in this file then.

josenavas: And, BTW, you didn't had any problem to do this kind of modifications in other places of the code.

squirrelo:
comments addressed, ready for another review.

josenavas:
:+1: Thanks @squirrelo 
@antgonza can you review?

antgonza:
Looks fine but the patch should be 27.sql, should we leave as this and change before merging the branch to master?

squirrelo:
yeah, it won't patch on the branch otherwise I think. I can do cleanup like that right before merge into master. That way if it ends up being patch 28 or 29 it still is fine.

PR #1258, opened 2015-06-12
issue 1230

antgonza:
<del>This is the initial commit but it is still a WIP. However, it will be nice if people can take a look to make sure they agree with these changes. At this point you can upload, link and process (unsuccessfully) already demultiplexed samples without primers. Next steps are: (1) successfully process these kind of samples, and (2) allow to remove primers.</del>

### DIFF
--- a/qiita_db/support_files/patches/26.sql
+++ b/qiita_db/support_files/patches/26.sql
@@ -2,14 +2,14 @@
 
 -- Updating FASTA-Sanger -> FASTA_Sanger, needed so we can put restrictions on
 -- what kind of files the user can select in the GUI
-
 UPDATE qiita.filetype SET type='FASTA_Sanger' WHERE type = 'FASTA-Sanger';
 
--- Adding new filetype and new illumina processing params if they do not exists
--- taken from: http://stackoverflow.com/a/13902402
-
+-- Adding new filetype
 INSERT INTO qiita.filetype (type) VALUES ('per_sample_FASTQ');
 
+
+-- Adding new illumina processing params if they do not exists
+-- adapted from: http://stackoverflow.com/a/13902402
 INSERT INTO qiita.preprocessed_sequence_illumina_params (param_set_name, barcode_type)
   SELECT DISTINCT 'per sample FASTQ defaults', 'not-barcoded' FROM qiita.preprocessed_sequence_illumina_params
   WHERE NOT EXISTS (SELECT 1 FROM qiita.preprocessed_sequence_illumina_params WHERE barcode_type = 'not-barcoded');
--- a/qiita_db/support_files/patches/26.sql
+++ b/qiita_db/support_files/patches/26.sql
@@ -11,5 +11,5 @@ UPDATE qiita.filetype SET type='FASTA_Sanger' WHERE type = 'FASTA-Sanger';
 INSERT INTO qiita.filetype (type) VALUES ('per_sample_FASTQ');
 
 INSERT INTO qiita.preprocessed_sequence_illumina_params (param_set_name, barcode_type)
-  SELECT 'per sample FASTQ defaults', 'not-barcoded' FROM qiita.preprocessed_sequence_illumina_params
+  SELECT DISTINCT 'per sample FASTQ defaults', 'not-barcoded' FROM qiita.preprocessed_sequence_illumina_params
   WHERE NOT EXISTS (SELECT 1 FROM qiita.preprocessed_sequence_illumina_params WHERE barcode_type = 'not-barcoded');
--- a/qiita_db/support_files/patches/26.sql
+++ b/qiita_db/support_files/patches/26.sql
@@ -1,0 +1,15 @@
+-- Jun 11, 2015
+
+-- Updating FASTA-Sanger -> FASTA_Sanger, needed so we can put restrictions on
+-- what kind of files the user can select in the GUI
+
+UPDATE qiita.filetype SET type='FASTA_Sanger' WHERE type = 'FASTA-Sanger';
+
+-- Adding new filetype and new illumina processing params if they do not exists
+-- taken from: http://stackoverflow.com/a/13902402
+
+INSERT INTO qiita.filetype (type) VALUES ('per_sample_FASTQ');
+
+INSERT INTO qiita.preprocessed_sequence_illumina_params (param_set_name, barcode_type)
+  SELECT 'per sample FASTQ defaults', 'not-barcoded' FROM qiita.preprocessed_sequence_illumina_params
+  WHERE NOT EXISTS (SELECT 1 FROM qiita.preprocessed_sequence_illumina_params WHERE barcode_type = 'not-barcoded');
--- a/qiita_pet/support_files/doc/source/tutorials/no-raw-sequences.rst
+++ b/qiita_pet/support_files/doc/source/tutorials/no-raw-sequences.rst
@@ -10,8 +10,11 @@ able to use Qiita to process your data files.
    `QIIME <http://www.qiime.org>`__ installation and basic knowledge on the
    usage of the command line.
 
-Per sample FASTQ files
-======================
+
+Per sample FASTQ files with barcode information
+===============================================
+
+Due to the nature of Qiita this is the prefered way to add per sample files.
 
 In this example we assume we are working with uncompressed per-sample
 FASTQ files. Thus before we begin, we have to make sure that we separate
@@ -55,3 +58,29 @@ compression program to use is ``gzip``:
     # new files, named barcodes.fastq.gz, forward.fastq.gz and revers.fastq.gz
     gzip *.fastq
 
+
+Per sample FASTQ files without barcode and primer information
+=============================================================
+
+In this example we assume we are working with uncompressed per-sample
+FASTQ files and that they do not have any barcode or primer information. This
+file type normally is what you can download from BaseSpace.
+
+The current system allows to upload this kind of files. The only requirement is
+that the prep template should have the uploaded sequence file(s) name in the
+run_prefix field. This should be an exact match without extension (fastq or
+fastq.gz). For example, if your uploaded file is named sample1_L001_R1.fastq.gz
+you will need to have sample1_L001_R1 as the run_prefix.
+
+
+Per sample FASTQ files without barcode but with primer information
+==================================================================
+
+The current way to process this files is to remove the primer section of the
+reads and follow the `Per sample FASTQ files without barcode and primer information`_
+instructions after they have been removed.
+
+To remove the primer information we will use extract_barcodes.py and pass the size
+of the primer as the barcode and simply discard the barcode files created during
+this step. For this you could follow the `Per sample FASTQ files with barcode information`_
+tutorial.
--- a/qiita_pet/support_files/doc/source/tutorials/no-raw-sequences.rst
+++ b/qiita_pet/support_files/doc/source/tutorials/no-raw-sequences.rst
@@ -59,6 +59,8 @@ compression program to use is ``gzip``:
     gzip *.fastq
 
 
+.. _per_sample_fastq_files_without_barcode_or_primer_information:
+
 Per sample FASTQ files without barcode and primer information
 =============================================================
 
--- a/qiita_pet/support_files/doc/source/tutorials/no-raw-sequences.rst
+++ b/qiita_pet/support_files/doc/source/tutorials/no-raw-sequences.rst
@@ -64,7 +64,7 @@ Per sample FASTQ files without barcode and primer information
 
 In this example we assume we are working with uncompressed per-sample
 FASTQ files and that they do not have any barcode or primer information. This
-file type normally is what you can download from BaseSpace.
+file type normally is what you can download from `Illumina's BaseSpace`<https://basespace.illumina.com/home/index>_.
 
 The current system allows to upload this kind of files. The only requirement is
 that the prep template should have the uploaded sequence file(s) name in the
@@ -80,7 +80,7 @@ The current way to process this files is to remove the primer section of the
 reads and follow the `Per sample FASTQ files without barcode and primer information`_
 instructions after they have been removed.
 
-To remove the primer information we will use extract_barcodes.py and pass the size
-of the primer as the barcode and simply discard the barcode files created during
-this step. For this you could follow the `Per sample FASTQ files with barcode information`_
-tutorial.
+To remove the primer information we will use `extract_barcodes.py`<http://qiime.org/scripts/extract_barcodes.html>_
+and pass the size of the primer as the barcode and simply discard the barcode
+files created during this step. For this you could follow the
+`Per sample FASTQ files with barcode information`_ tutorial.
--- a/qiita_pet/support_files/doc/source/tutorials/no-raw-sequences.rst
+++ b/qiita_pet/support_files/doc/source/tutorials/no-raw-sequences.rst
@@ -62,9 +62,10 @@ compression program to use is ``gzip``:
 Per sample FASTQ files without barcode and primer information
 =============================================================
 
-In this example we assume we are working with uncompressed per-sample
-FASTQ files and that they do not have any barcode or primer information. This
-file type normally is what you can download from `Illumina's BaseSpace`<https://basespace.illumina.com/home/index>_.
+In this example we assume we are working with uncompressed per-sample FASTQ
+files and that they do not have any barcode or primer information. This file type
+normally is what you can download from `Illumina's BaseSpace <https://basespace.illumina.com/home/index>`__.
+
 
 The current system allows to upload this kind of files. The only requirement is
 that the prep template should have the uploaded sequence file(s) name in the
@@ -80,7 +81,7 @@ The current way to process this files is to remove the primer section of the
 reads and follow the `Per sample FASTQ files without barcode and primer information`_
 instructions after they have been removed.
 
-To remove the primer information we will use `extract_barcodes.py`<http://qiime.org/scripts/extract_barcodes.html>_
+To remove the primer information we will use `extract_barcodes.py <http://qiime.org/scripts/extract_barcodes.html>`__
 and pass the size of the primer as the barcode and simply discard the barcode
 files created during this step. For this you could follow the
 `Per sample FASTQ files with barcode information`_ tutorial.
--- a/qiita_pet/templates/study_description.html
+++ b/qiita_pet/templates/study_description.html
@@ -148,11 +148,10 @@ function get_and_check_raw_data_files(prep_id, filetype){
         break;
     }
   }
-
   if (sff.length === 0 && barcodes.length === 0 && forward.length === 0 && fasta.length === 0) {
     bootstrapAlert("You need to select at least: one barcode and one forward file, one sff file or one fasta file.");
     return [false];
-  } else if (sff.length === 0 && barcodes.length != forward.length) {
+  } else if (sff.length === 0 && barcodes.length != forward.length && filetype !== 5 && filetype != "per_sample_FASTQ") {
     bootstrapAlert("You must select the same number of barcode and forward seq files.");
     return [false];
   } else if (sff.length === 0 && (reverse.length !== 0 && reverse.length != forward.length)) {
--- a/qiita_pet/uimodules/prep_template_tab.py
+++ b/qiita_pet/uimodules/prep_template_tab.py
@@ -31,7 +31,9 @@ filepath_types = [k.split('_', 1)[1].replace('_', ' ')
                   if k.startswith('raw_')]
 fp_type_by_ft = defaultdict(
     lambda: filepath_types, SFF=['sff'], FASTA=['fasta', 'qual'],
-    FASTQ=['barcodes', 'forward seqs', 'reverse seqs'])
+    FASTQ=['barcodes', 'forward seqs', 'reverse seqs'],
+    FASTA_Sanger=['fasta'],
+    per_sample_FASTQ=['forward seqs', 'reverse seqs'])
 
 
 def _get_accessible_raw_data(user):
@@ -161,12 +163,17 @@ class PrepTemplateInfoTab(BaseUIModule):
         if raw_data_id:
             rd = RawData(raw_data_id)
             rd_ft = rd.filetype
+
             # If the prep template has a raw data associated, it can be
             # preprocessed. Retrieve the pre-processing parameters
             if rd_ft in ('SFF', 'FASTA'):
                 param_iter = Preprocessed454Params.iter()
             elif rd_ft == 'FASTQ':
-                param_iter = PreprocessedIlluminaParams.iter()
+                param_iter = [pip for pip in PreprocessedIlluminaParams.iter()
+                              if pip.name != 'per sample FASTQ defaults']
+            elif rd_ft == 'per_sample_FASTQ':
+                param_iter = [pip for pip in PreprocessedIlluminaParams.iter()
+                              if pip.name == 'per sample FASTQ defaults']
             else:
                 raise NotImplementedError(
                     "Pre-processing of %s files currently not supported."
@@ -197,13 +204,17 @@ class PrepTemplateInfoTab(BaseUIModule):
                            else 'demultiplex')
                     missing_cols = prep_template.check_restrictions(
                         [PREP_TEMPLATE_COLUMNS_TARGET_GENE[key]])
-                    show_preprocess_btn = len(missing_cols) == 0
+
+                    if rd_ft == 'per_sample_FASTQ':
+                        show_preprocess_btn = 'run_prefix' not in missing_cols
+                    else:
+                        show_preprocess_btn = len(missing_cols) == 0
+
+                    no_preprocess_msg = None
                     if not show_preprocess_btn:
                         no_preprocess_msg = (
                             "Preprocessing disabled due to missing columns in "
                             "the prep template: %s" % ', '.join(missing_cols))
-                    else:
-                        no_preprocess_msg = None
 
         preprocessing_status = prep_template.preprocessing_status
 
--- a/qiita_pet/uimodules/prep_template_tab.py
+++ b/qiita_pet/uimodules/prep_template_tab.py
@@ -170,10 +170,10 @@ class PrepTemplateInfoTab(BaseUIModule):
                 param_iter = Preprocessed454Params.iter()
             elif rd_ft == 'FASTQ':
                 param_iter = [pip for pip in PreprocessedIlluminaParams.iter()
-                              if pip.name != 'per sample FASTQ defaults']
+                              if pip.values['barcode_type'] != 'not-barcoded']
             elif rd_ft == 'per_sample_FASTQ':
                 param_iter = [pip for pip in PreprocessedIlluminaParams.iter()
-                              if pip.name == 'per sample FASTQ defaults']
+                              if pip.values['barcode_type'] == 'not-barcoded']
             else:
                 raise NotImplementedError(
                     "Pre-processing of %s files currently not supported."
--- a/qiita_ware/processing_pipeline.py
+++ b/qiita_ware/processing_pipeline.py
@@ -90,11 +90,12 @@ def _get_sample_names_by_run_prefix(prep_template):
     ----------
     prep_template : PrepTemplate
         The prep template from which we need to generate the minimal mapping
+        file
 
     Returns
     -------
     dict
-        The paths to the qiime minimal mapping files
+        Dict mapping run_prefix to path of the minimal qiime mapping file
 
     Raises
     ------
@@ -103,7 +104,6 @@ def _get_sample_names_by_run_prefix(prep_template):
     """
     from qiita_db.metadata_template import load_template_to_dataframe
 
-    # The prep templates has a QIIME mapping file, get it
     qiime_map = load_template_to_dataframe(prep_template.qiime_map_fp,
                                            index='#SampleID')
 
@@ -112,8 +112,8 @@ def _get_sample_names_by_run_prefix(prep_template):
     for prefix, df in qiime_map.groupby('run_prefix'):
         len_df = len(df)
         if len_df != 1:
-            errors.append('%s (%d): %s' % (prefix, len_df,
-                                           ', '.join(df.index.values)))
+            errors.append('%s has %d samples (%s)' % (prefix, len_df,
+                                                      ', '.join(df.index)))
         else:
             samples[prefix] = df.index.values[0]
 
@@ -149,11 +149,11 @@ def _get_preprocess_fastq_cmd(raw_data, prep_template, params):
     ValueError
         If the raw data object does not have any sequence file associated
         If the raw data filetype is not per_sample_FASTQ and the number of raw
-            forward sequences an raw barcode files are not the same
-        If the raw data filetype is per_sample_FASTQ and has barcode files
+        forward sequences and raw barcode files are not the same
+        If the raw data filetype is per_sample_FASTQ and it has barcode files
         If the raw data filetype is per_sample_FASTQ and the run_prefix values
-            don't match the file names without extensions and without the
-            raw data id prefix
+        don't match the file names without extensions and without the raw data
+        id prefix
     """
     from tempfile import mkdtemp
     from os.path import basename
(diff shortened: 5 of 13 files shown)


qiita_db/support_files/patches/26.sql
```diff
@@ -0,0 +1,15 @@
+-- Jun 11, 2015
+
+-- Updating FASTA-Sanger -> FASTA_Sanger, needed so we can put restrictions on
+-- what kind of files the user can select in the GUI
+
+UPDATE qiita.filetype SET type='FASTA_Sanger' WHERE type = 'FASTA-Sanger';
+
+-- Adding new filetype and new illumina processing params if they do not exists
+-- taken from: http://stackoverflow.com/a/13902402
```
josenavas: Minor, but can you move this comment to the line prior to the insert in line 13?
And `taken` -> `adapted`

qiita_db/support_files/patches/26.sql
```diff
@@ -0,0 +1,15 @@
+-- Jun 11, 2015
+
+-- Updating FASTA-Sanger -> FASTA_Sanger, needed so we can put restrictions on
+-- what kind of files the user can select in the GUI
+
+UPDATE qiita.filetype SET type='FASTA_Sanger' WHERE type = 'FASTA-Sanger';
```
josenavas: I don't fully understand why we need to change from `FASTA-Sanger` to `FASTA_Sanger`. Can you elaborate?

qiita_pet/templates/study_description.html
```diff
@@ -148,11 +148,10 @@
         break;
     }
   }
-
   if (sff.length === 0 && barcodes.length === 0 && forward.length === 0 && fasta.length === 0) {
     bootstrapAlert("You need to select at least: one barcode and one forward file, one sff file or one fasta file.");
     return [false];
-  } else if (sff.length === 0 && barcodes.length != forward.length) {
+  } else if (sff.length === 0 && barcodes.length != forward.length && filetype !== 5 && filetype != "per_sample_FASTQ") {
```
josenavas: Is there any specific test that needs to be done in a `per_sample_FASTQ`? i.e. the same number of forward and reverse, in case that the user chose any reverse read?

qiita_db/support_files/patches/26.sql
antgonza: In line 35 of qiita_pet/uimodules/prep_template_tab.py we define which filetypes have access to which raw data. However, to do this we need the names to match and FASTA-Sanger is not a valid identifier.

qiita_pet/uimodules/prep_template_tab.py
```diff
@@ -161,12 +163,17 @@ def render(self, study, prep_template, full_access, ena_terms,
...
             rd = RawData(raw_data_id)
             rd_ft = rd.filetype
+
             # If the prep template has a raw data associated, it can be
             # preprocessed. Retrieve the pre-processing parameters
             if rd_ft in ('SFF', 'FASTA'):
                 param_iter = Preprocessed454Params.iter()
             elif rd_ft == 'FASTQ':
-                param_iter = PreprocessedIlluminaParams.iter()
+                param_iter = [pip for pip in PreprocessedIlluminaParams.iter()
```
josenavas: This test shouldn't be against `pip.barcode_type != 'not-barcoded'`? The reason is that if we then add more non-barcode options here (for multiple parameters sweeping) we will need to list them by name in here.

qiita_pet/uimodules/prep_template_tab.py
```diff
@@ -161,12 +163,17 @@ def render(self, study, prep_template, full_access, ena_terms,
...
+                              if pip.name != 'per sample FASTQ defaults']
+            elif rd_ft == 'per_sample_FASTQ':
+                param_iter = [pip for pip in PreprocessedIlluminaParams.iter()
+                              if pip.name == 'per sample FASTQ defaults']
```
josenavas: Similarly here, but in this case testing for equality.

qiita_pet/uimodules/prep_template_tab.py
```diff
@@ -197,13 +204,17 @@ def render(self, study, prep_template, full_access, ena_terms,
...
+                        show_preprocess_btn = 'run_prefix' not in missing_cols
+                    else:
+                        show_preprocess_btn = len(missing_cols) == 0
+
+                    no_preprocess_msg = None
                     if not show_preprocess_btn:
                         no_preprocess_msg = (
                             "Preprocessing disabled due to missing columns in "
                             "the prep template: %s" % ', '.join(missing_cols))
-                    else:
```
josenavas: Why did you removed this `else`?

qiita_pet/templates/study_description.html
antgonza: Just that there are the same number of fwd and reverse files, and that test is in place.

josenavas:
@antgonza few comments. I think the approach taken here is correct, besides the couple of comments that I made here.

qiita_pet/uimodules/prep_template_tab.py
antgonza: Please read the block again, with the new structure is not needed.

josenavas: Oh, I noticed now. Sorry!

qiita_pet/uimodules/prep_template_tab.py
antgonza: Good point.

qiita_ware/processing_pipeline.py
```diff
@@ -83,6 +83,51 @@ def _get_qiime_minimal_mapping(prep_template, out_dir):
...
+    Raises
+    ------
+    ValueError
+        If there is more than 1 sample per run_prefix
+    """
+    from collections import defaultdict
+    import pandas as pd
+
+    # The prep templates has a QIIME mapping file, get it
+    qiime_map = pd.read_csv(prep_template.qiime_map_fp, sep='\t',
```
josenavas: We have identified a potential issue parsing the QIIME mapping files in this way. Can you use `qiime_map = load_template_to_dataframe(fp, index='#SampleID')`? It will solve that problem.

I know this is inconsistent with the other functions in this file, but those instances also need to be changed, so I prefer to not add more potential sources of errors.

qiita_ware/processing_pipeline.py
```diff
@@ -83,6 +83,51 @@ def _get_qiime_minimal_mapping(prep_template, out_dir):
...
+                            keep_default_na=False, na_values=['unknown'],
+                            index_col=False,
+                            converters=defaultdict(lambda: str))
+    qiime_map.set_index('#SampleID', inplace=True, drop=True)
```
josenavas: With the call to `load_template_to_dataframe` you will not need this line.

qiita_ware/processing_pipeline.py
```diff
@@ -106,13 +151,20 @@ def _get_preprocess_fastq_cmd(raw_data, prep_template, params):
     NotImplementedError
         If any of the raw data input filepath type is not supported
     ValueError
-        If the number of raw sequences an raw barcode files are not the same
         If the raw data object does not have any sequence file associated
+        If the raw data filetype is not per_sample_FASTQ number of raw
```
josenavas: This comment reads weird - probably a missing word?
```
If the raw data filetype is not per_sample_FASTQ and the number of raw forward sequences an raw barcode files are not the same
```

qiita_ware/processing_pipeline.py
```diff
@@ -106,13 +151,20 @@ def _get_preprocess_fastq_cmd(raw_data, prep_template, params):
...
+            sequences an raw barcode files are not the same
+        If the raw data filetype is per_sample_FASTQ and has barcode files
+        If the raw data filetype is per_sample_FASTQ and the run_prefix values
+            don't match the file names without extentions
```
josenavas: `...without extensions and without the raw data id prefix`

qiita_ware/processing_pipeline.py
```diff
@@ -158,13 +187,68 @@ def _get_preprocess_fastq_cmd(raw_data, prep_template, params):
...
     barcode_fps = sorted(barcode_fps)
-    mapping_fps = sorted(mapping_fps)
 
-    # Create the split_libraries_fastq.py command
-    cmd = str("split_libraries_fastq.py --store_demultiplexed_fastq -i %s -b "
-              "%s -m %s -o %s %s"
-              % (','.join(forward_seqs), ','.join(barcode_fps),
-                 ','.join(mapping_fps), output_dir, params_str))
+    # Create a temporary directory to store the split libraries output
+    output_dir = mkdtemp(dir=qiita_config.working_dir, prefix='slq_out')
```
josenavas: I noticed that it is really hard to check if a given instance of split libraries is running by checking the file system because the prefix is not informative enough. Can you change to `prefix='slq_out_%d_%d_' % (prep_template.id, raw_data.id)` ?

qiita_ware/processing_pipeline.py
```diff
@@ -158,13 +187,68 @@ def _get_preprocess_fastq_cmd(raw_data, prep_template, params):
...
+            raise ValueError("per_sample_FASTQ can not have barcodes:"
+                             % (', '.join(barcode_fps)))
+
+        sn_by_rp = _get_sample_names_by_run_prefix(prep_template)
+        samples = []
+        for f in forward_seqs:
+            # getting just the main filename
+            f = basename(f).split('_', 1)[1]
+            # removing extentions: fastq or fastq.gz
+            if f.endswith('fastq.gz'):
```
josenavas: This entire if block can be replaced by `f = splitext(splitext(f)[0])[0]` (using the `splitext` function from `os.path`):
``` python
In [18]: s1 = "foo.fastq"

In [19]: f = splitext(splitext(s2)[0])[0]

In [20]: f
Out[20]: 'foo'

In [21]: s2 = "foo.fastq.gz"

In [22]: f = splitext(splitext(s2)[0])[0]

In [23]: f
Out[23]: 'foo'
```

antgonza: Do not agree. Note that I tried that but realized that:
In [5]: s = 'filename.1.we_should_keep_this.fastq'
In [6]: f = splitext(splitext(s)[0])[0]
In [7]: f
Out[7]: 'filename.1'

josenavas: Uhm, don't use that code, is actually unsafe in some cases: `foo.bar.fastq` will fail. I'm trying to find a better way of doing this.

josenavas: Found a better way of doing this, using regular expressions `f = re.sub('(.fastq.gz|.fastq)', '', f)`:
```python
In [51]: s1 = "foo.bar.fastq.gz"

In [52]: s2 = "foo.bar.fastq"

In [53]: re.sub('(.fastq.gz|.fastq)', '', s1)
Out[53]: 'foo.bar'

In [54]: re.sub('(.fastq.gz|.fastq)', '', s2)
Out[54]: 'foo.bar'
```

antgonza: Is it better? How? Faster?

qiita_ware/processing_pipeline.py
```diff
@@ -158,13 +187,68 @@ def _get_preprocess_fastq_cmd(raw_data, prep_template, params):
...
+                f = f[:-9]
+            else:
+                f = f[:-6]
+            try:
+                samples.append(sn_by_rp[f])
+                del sn_by_rp[f]
+            except KeyError:
```
josenavas: Can you add a comment that by passing here doesn't mean that we don't care about this error, is just that it is checked in line 222?

Also, it may be useful to actually put the failed prefix in a list and show that to the user, so he can actually see what he has in the run prefix and see if it is a typo or something else.

qiita_ware/processing_pipeline.py
josenavas: It removes magic numbers, reduces the lines of code, and it is more readable.

antgonza: That makes sense but I can replace the "magic" numbers (which I believe are clear in this case) for, for example, `len('fastq.gz')` and could make the `if` a ternary operation to remove lines of code. Will that work? Just wondering if the use of sub is too much (it causes overhead per file).

qiita_ware/test/test_processing_pipeline.py
```diff
@@ -271,6 +325,54 @@ def test_get_preprocess_fastq_cmd(self):
...
+        obs_cmd, obs_output_dir = _get_preprocess_fastq_cmd(raw_data,
+                                                            prep_template,
+                                                            params)
+
+        raw_fps = ','.join([fp for _, fp, _ in
+                            sorted(raw_data.get_filepaths())])
+        exp_cmd = (
+            "split_libraries_fastq.py --store_demultiplexed_fastq -i "
+            "{} --sample_ids 1.SKB8.640193,1.SKD8.640184 -o {} --barcode_type "
+            "golay_12 --max_bad_run_length 3 --max_barcode_errors 1.5 "
```
josenavas: It looks like it should be something wrong in here. The `--barcode_type` should be `not-barcoded`

antgonza: This is because I'm using `PreprocessedIlluminaParams(1)`, I could use the actual value but will need to query the db to get the actual value, thus complicating this test. Note that the test is not actually testing that but that the command is valid ...

qiita_ware/processing_pipeline.py
josenavas: I will let others weight in here, I would like to get the magic numbers removed, as they're always a potential source of problems.

Also, there is another issue, you're just checking against `fastq.gz` but you don't check against `fastq`, so you will be removing 6 chars independently if it matches `fastq` or not. I don't think it is safe to make this assumption.

antgonza: OK but note that the only way to get to this point is because the user have linked fastq or fastq.gz files (or forced via the DB) -- no other way!

qiita_ware/test/test_processing_pipeline.py
josenavas: Yeah, the problem is that such command is not a valid QIIME command, it will raise this error:
```
split_libraries_fastq.py -i filename1.fastq.gz,filename2.fastq.gz -o test --sample_ids='f1,f2' --barcode_type golay_12
...
split_libraries_fastq.py: error: Must provide --barcode_read_fps if --barcode_type is not 'not-barcoded'

```

qiita_ware/processing_pipeline.py
josenavas: That will be true if the only way of using QIITA is the GUI. QIITA is also tailored to be used through the command-line, although we have not being putting that many efforts...

qiita_ware/test/test_processing_pipeline.py
antgonza: Note that this is not a test for qiime ...

qiita_ware/processing_pipeline.py
antgonza: Yup, but that means that at this point we should always have fastq/fastq.gz, right? Either way, I can add an if that checks that and if not raises an error. Note that this will mean that your sub solution will not actually work or at least will not work as it is now.

qiita_ware/test/test_processing_pipeline.py
josenavas: I know, but the test is to make sure that Qiita generates valid QIIME commands. I think that what is missing here is a check in the actual function that if the type is `per_sample_FASTQ` the parameter set must be one of the parameters that have `not_barcoded` as barcode type. This will make the system more robust.

antgonza: OK

qiita_ware/processing_pipeline.py
wasade: `if 'fastq' in f.rsplit('.', 2):`

...then for trimming, just use `rindex`:

`f = f[:f.rindex('fastq')]`

removes all magic numbers, is pretty readable, and doesn't care about gz

qiita_ware/processing_pipeline.py
wasade: since the error is ignored, why not just do an explicit check? If the `KeyError` is common, this will be very expensive.

```python
samples.append(sn_by_rp.pop(f)) if f in sn_by_rp else None
```

qiita_ware/processing_pipeline.py
antgonza: I like that.

qiita_ware/processing_pipeline.py
antgonza: That makes sense but it will be a compromise as basically the will be testing every single filename vs. not checking and just waiting for the error. Note that in general we should never hit that KeyError cause it means that there is something wrong and we are `pass`ing it cause we want to capture all the errors before exiting (next if).

qiita_ware/processing_pipeline.py
josenavas: I like that too.

qiita_ware/processing_pipeline.py
josenavas: Agree that such error should be almost never hit, that's why I just suggested adding a comment saying that is checked in the next if. Is not clear if you're just reading the code and it took me a while to realize what is going on.

antgonza: OK adding some comments.

qiita_ware/processing_pipeline.py
```diff
@@ -83,6 +83,47 @@ def _get_qiime_minimal_mapping(prep_template, out_dir):
     return output_fps
 
 
+def _get_sample_names_by_run_prefix(prep_template):
+    """Generates a dictionary of run_prefix and sample names
+
+    Parameters
+    ----------
+    prep_template : PrepTemplate
+        The prep template from which we need to generate the minimal mapping
```
ElDeveloper: I *think* the word `file` is missing at the end of this sentence.

qiita_ware/processing_pipeline.py
```diff
@@ -83,6 +83,47 @@ def _get_qiime_minimal_mapping(prep_template, out_dir):
...
+
+    Returns
+    -------
+    dict
+        The paths to the qiime minimal mapping files
```
ElDeveloper: Shouldn't this be something like `Dict mapping run_prefix to path of the qiime minimal mapping file.`?

qiita_ware/processing_pipeline.py
```diff
@@ -83,6 +83,47 @@ def _get_qiime_minimal_mapping(prep_template, out_dir):
...
+
+    Raises
+    ------
+    ValueError
+        If there is more than 1 sample per run_prefix
+    """
+    from qiita_db.metadata_template import load_template_to_dataframe
+
+    # The prep templates has a QIIME mapping file, get it
```
ElDeveloper: has -> have ... and maybe the `get it` is not needed?

qiita_ware/processing_pipeline.py
```diff
@@ -83,6 +83,47 @@ def _get_qiime_minimal_mapping(prep_template, out_dir):
...
+    qiime_map = load_template_to_dataframe(prep_template.qiime_map_fp,
+                                           index='#SampleID')
+
+    samples = {}
+    errors = []
+    for prefix, df in qiime_map.groupby('run_prefix'):
+        len_df = len(df)
+        if len_df != 1:
+            errors.append('%s (%d): %s' % (prefix, len_df,
```
ElDeveloper: This is very useful, but perhaps the way to go here is to phrase this as a sentence i.e. `%s has %d samples (%s) % (prefix, len_df, ', '.join(df.index))`.

Also, note that you don't need to do `.values`, `Index` objects are already iterable and can be passed to `.join` seamlessly:

```python
In [3]: pd.DataFrame([[1, 2, 3], [4, 5, 6]], ['A', 'B'], ['r1', 'r2', 'r3'])
Out[3]: 
   r1  r2  r3
A   1   2   3
B   4   5   6

In [4]: _3.index
Out[4]: Index([u'A', u'B'], dtype='object')

In [5]: ', '.join(_3.index)
Out[5]: 'A, B'
```

qiita_ware/processing_pipeline.py
```diff
@@ -106,13 +147,21 @@ def _get_preprocess_fastq_cmd(raw_data, prep_template, params):
     NotImplementedError
         If any of the raw data input filepath type is not supported
     ValueError
-        If the number of raw sequences an raw barcode files are not the same
         If the raw data object does not have any sequence file associated
+        If the raw data filetype is not per_sample_FASTQ and the number of raw
+            forward sequences an raw barcode files are not the same
```
ElDeveloper: an -> and

I think this line (and others in this section) are over-indented.

qiita_ware/processing_pipeline.py
```diff
@@ -106,13 +147,21 @@ def _get_preprocess_fastq_cmd(raw_data, prep_template, params):
...
+        If the raw data filetype is per_sample_FASTQ and has barcode files
```
ElDeveloper: Not sure about this one, but "and has" -> "and it has"?

qiita_ware/processing_pipeline.py
```diff
@@ -158,13 +184,70 @@ def _get_preprocess_fastq_cmd(raw_data, prep_template, params):
...
+    # Add any other parameter needed to split libraries fastq
+    params_str = params.to_str()
+
+    if len(forward_seqs) == 0:
+        raise ValueError("Forward reads file not found on raw data %s"
+                         % raw_data.id)
+
+    if filetype == "per_sample_FASTQ":
+        if barcode_fps:
+            raise ValueError("per_sample_FASTQ can not have barcodes:"
```
ElDeveloper: This string is missing a `%s` ... and I assume there's a test missing?

qiita_ware/processing_pipeline.py
```diff
@@ -158,13 +184,70 @@ def _get_preprocess_fastq_cmd(raw_data, prep_template, params):
...
+    # Create a temporary directory to store the split libraries output
+    output_dir = mkdtemp(
+        dir=qiita_config.working_dir,
+        prefix='slq_out_%d_%d_' % (prep_template.id, raw_data.id))
+
+    # Add any other parameter needed to split libraries fastq
+    params_str = params.to_str()
+
+    if len(forward_seqs) == 0:
+        raise ValueError("Forward reads file not found on raw data %s"
```
ElDeveloper: I'm not sure but I think on -> in

qiita_ware/processing_pipeline.py
```diff
@@ -158,13 +184,70 @@ def _get_preprocess_fastq_cmd(raw_data, prep_template, params):
...
+                         % raw_data.id)
+
+    if filetype == "per_sample_FASTQ":
+        if barcode_fps:
+            raise ValueError("per_sample_FASTQ can not have barcodes:"
+                             % (', '.join(barcode_fps)))
+
+        sn_by_rp = _get_sample_names_by_run_prefix(prep_template)
+        samples = []
+        for f in forward_seqs:
```
ElDeveloper: Question about this: can we also process reverse_seqs? My understanding is that `split_libraries_fastq.py` should be capable of dealing with fwd or reverse sequences interchangeably. Can this be added then?

antgonza: That's possible but you will need to upload the reverse seqs as forward.

ElDeveloper: After talking with @antgonza offline, it seems like this is OK as is because of the way split_librarie_fastq.py works.

qiita_ware/processing_pipeline.py
```diff
@@ -158,13 +184,70 @@ def _get_preprocess_fastq_cmd(raw_data, prep_template, params):
...
+            # and display them in the next if block
+            try:
+                samples.append(sn_by_rp[f])
+                del sn_by_rp[f]
+            except KeyError:
+                pass
+
+        if sn_by_rp:
+            raise ValueError(
+                'Some run_prefix values do not match your sample names: %s'
```
ElDeveloper: The code above is not really looking for a "prefix", it's looking for an exact match of the file name once the extensions have been removed. I think the code should be changed to search by prefix and not by exact match. Because a file like `sample1_L001_R1.fastq.gz` would not be found above even though the run_prefix was set as `sample1`.

An alternative, is that we may be using the term `run_prefix` incorrectly here as we are not exactly looking for a prefix.

qiita_ware/test/test_processing_pipeline.py
```diff
@@ -271,6 +326,55 @@ def test_get_preprocess_fastq_cmd(self):
...
+                            'experiment_design_description': 'A'},
+            'SKD8.640184': {'run_prefix': "sample2", 'primer': 'A',
+                            'barcode': 'A', 'center_name': 'ANL',
+                            'platform': 'ILLUMINA',
+                            'library_construction_protocol': 'A',
+                            'experiment_design_description': 'A'}}
+        md_template = pd.DataFrame.from_dict(metadata_dict, orient='index')
+        prep_template = PrepTemplate.create(md_template, Study(1), '16S')
+
+        tmp_dir = mkdtemp()
```
ElDeveloper: I don't think you need this line nor the next two lines, this is already being done in the `setUp` method.

ElDeveloper:
Just a few comments on the code. Aside from that, It would be good if you can add a mention in the ChangeLog about this new and *exciting* and new feature. 

When selecting what type of files are being linked to the raw data, we probably want to add a button or something to select all the available files as forward seqs, otherwise in a study with a couple hundred samples this becomes a very tedious task.

BTW, If you use the example files from the mailing list, the pipeline will fail with the following message:

screen shot 2015-06-17 at 4 42 53 pm

Solution, rename sample3.fastq.gz and sample4.fastq.gz to filename3.fastq.gz and filename4.fastq.gz

qiita_ware/processing_pipeline.py
antgonza: k

qiita_ware/processing_pipeline.py
antgonza: k

qiita_ware/processing_pipeline.py
antgonza: Removing cause is not needed. I guess is a leftover from a previous iteration.

qiita_ware/processing_pipeline.py
antgonza: k, thanks.

qiita_ware/processing_pipeline.py
antgonza: k

qiita_ware/processing_pipeline.py
antgonza: k

qiita_ware/processing_pipeline.py
antgonza: k, added a test for this.

qiita_ware/processing_pipeline.py
antgonza: Yes, we are looking for an exact match and we are using `run_prefix` as the field where we are storing this info. IMOO this is safer. Note that I think this is one of those examples where the same PR should have a new (or expand) documentation entry, I will add that.

qiita_ware/test/test_processing_pipeline.py
antgonza: Two different uses, the setup creates _sff_ files while we need fastq. I think we have 2 options: (1) leave as is, (2) move this block to each of the methods that test sff, as is done in the fastq. Let me know what you want to do.

ElDeveloper: What I mean is that the setup of a partial function that builds temporary paths (`self.path_builder`) is all being done in `setUp` ([here](https://github.com/antgonza/qiita/blob/issue-1230/qiita_ware/test/test_processing_pipeline.py#L74-75) and [here](https://github.com/antgonza/qiita/blob/issue-1230/qiita_ware/test/test_processing_pipeline.py#L107)), thus you don't need to run this again. From looking at the code, it doesn't seem like there are parts that are sensitive to the contents of the directory, so it would be safe to use as is.

qiita_ware/processing_pipeline.py
ElDeveloper: A documentation entry is a great idea! The name of the field however, is
still misleading :sob: , but there's not really much we can do about
this, right?

antgonza: The only thing will be to change the name of the header to something else, then make that required, etc ... which will make things way too complicated. My suggestion is to leave it as is.

ElDeveloper: Sounds good, the suggestion to add a documentation piece on this, will
be essential to reduce further confusion.

qiita_pet/support_files/doc/source/tutorials/no-raw-sequences.rst
```diff
@@ -55,3 +58,29 @@ compression program to use is ``gzip``:
     # new files, named barcodes.fastq.gz, forward.fastq.gz and revers.fastq.gz
     gzip *.fastq
 
+
+Per sample FASTQ files without barcode and primer information
+=============================================================
+
+In this example we assume we are working with uncompressed per-sample
+FASTQ files and that they do not have any barcode or primer information. This
+file type normally is what you can download from BaseSpace.
```
ElDeveloper: BaseSpace ->

```
 `Illumina's BaseSpace`<https://basespace.illumina.com/home/index>__
```

qiita_pet/support_files/doc/source/tutorials/no-raw-sequences.rst
```diff
@@ -55,3 +58,29 @@ compression program to use is ``gzip``:
...
+
+
+Per sample FASTQ files without barcode but with primer information
+==================================================================
+
+The current way to process this files is to remove the primer section of the
+reads and follow the `Per sample FASTQ files without barcode and primer information`_
+instructions after they have been removed.
+
+To remove the primer information we will use extract_barcodes.py and pass the size
```
ElDeveloper: Can you add a link to http://qiime.org/scripts/extract_barcodes.html

qiita_ware/processing_pipeline.py
```diff
@@ -158,13 +184,70 @@ def _get_preprocess_fastq_cmd(raw_data, prep_template, params):
...
+            raise ValueError("per_sample_FASTQ can not have barcodes: %s"
+                             % (', '.join([basename(b) for b in barcode_fps])))
+
+        sn_by_rp = _get_sample_names_by_run_prefix(prep_template)
+        samples = []
+        for f in forward_seqs:
+            # getting just the main filename
+            f = basename(f).split('_', 1)[1]
+            # removing extentions: fastq or fastq.gz
+            if 'fastq' in f.rsplit('.', 2):
```
ElDeveloper: Can you make a consideration for files where the extension is `FASTQ` or a mixture of upper/lower case?

qiita_ware/processing_pipeline.py
ElDeveloper: Just saw the documentation, looks good!

ElDeveloper:
Ping on this comment:

> When selecting what type of files are being linked to the raw data, we probably want to add a button or something to select all the available files as forward seqs, otherwise in a study with a couple hundred samples this becomes a very tedious task.

ElDeveloper:
:+1:

qiita_ware/processing_pipeline.py
antgonza: k

qiita_pet/support_files/doc/source/tutorials/no-raw-sequences.rst
```diff
@@ -55,3 +58,29 @@ compression program to use is ``gzip``:
     # new files, named barcodes.fastq.gz, forward.fastq.gz and revers.fastq.gz
     gzip *.fastq
 
+
+Per sample FASTQ files without barcode and primer information
+=============================================================
+
+In this example we assume we are working with uncompressed per-sample
+FASTQ files and that they do not have any barcode or primer information. This
+file type normally is what you can download from `Illumina's BaseSpace`<https://basespace.illumina.com/home/index>_.
```
ElDeveloper: This link is missing an underscore, same below in `extract_barcodes.py`

ElDeveloper:
Looks really good thanks for working on this @antgonza, :+1: once tests pass!

qiita_pet/support_files/doc/source/tutorials/no-raw-sequences.rst
```diff
@@ -55,3 +58,30 @@ compression program to use is ``gzip``:
...
+=============================================================
+
+In this example we assume we are working with uncompressed per-sample FASTQ
+files and that they do not have any barcode or primer information. This file type
+normally is what you can download from `Illumina's BaseSpace <https://basespace.illumina.com/home/index>`__.
+
+
+The current system allows to upload this kind of files. The only requirement is
+that the prep template should have the uploaded sequence file(s) name in the
+run_prefix field. This should be an exact match without extension (fastq or
```
josenavas: I think you should mention this in the documentation about preparing templates, under the `Required fields for pre-processing target gene data` section.

josenavas:
Code looks great! Thanks @antgonza 
Just one comment about documentation, after that :+1: (I keep forgetting that we have documentation in the repo now and we have to modify it as new functionality is added!)

PR #1260, opened 2015-06-15
DOC: Add missing tutorial file

ElDeveloper:


antgonza:
:+1:

josenavas:
@antgonza this is still applies taking into account your work in #1258?

antgonza:
Yup, cause this is for files from which we can extract barcodes and #1258 is for those that we can't.

josenavas:
Oh, ok. :+1: the once tests pass.

josenavas:
That was a really bad wording: :+1: once the tests pass
hahah

PR #1262, opened 2015-06-16
WIP: Improve SQL queues

josenavas:
The current SQL queues are somehow limited, as they do not allow us to make all DB interactions a single transaction.

The main limitation is the placeholder system. Currently, if you want to use a result from a previous query, you provide an index in this form: `{#}`, and once the result is consumed, the results are wiped out. This makes impossible to re-use the result from one query more than once.

The proposed change is to use a notation like this: `{#:#}` where the first number is the index of the sql query and the second number is the index of the result within that query that needs to be used. Then, all the results will be stored and ew can re-use them as many times as we want.

In order to accomplish this, these are the steps that are needed:
 - [x] Modify the queues system in the `SQLConnectionHandler` - creation of the new Transaction object (@josenavas)~~(https://github.com/biocore/qiita/pull/1276)~~
 - [ ] Fix python patches (@josenavas)
 - [x] Fix `__init__.py`
 - [ ] Fix `analysis.py`
 - [ ] Fix `base.py` (@josenavas) (https://github.com/biocore/qiita/pull/1298)
 - [ ] Fix `commands.py`
 - [ ] Fix `data.py`
 - [ ] Fix `environment_manager.py` (@josenavas)
 - [x] Fix `exceptions.py`
 - [x] Fix `investigation.py`
 - [ ] Fix `job.py`
 - [ ] Fix `logger.py`
 - [ ] Fix `meta_util.py`
 - [ ] Fix `metadata_template/*` (@josenavas)
 - [ ] Fix `ontology.py` (@josenavas) (https://github.com/biocore/qiita/pull/1307)
 - [ ] Fix `parameters.py`(@josenavas)
 - [ ] Fix `reference.py` (@josenavas) (https://github.com/biocore/qiita/pull/1306)
 - [ ] Fix `search.py`
 - [ ] Fix `sql_connection.py` - remove unused code
 - [ ] Fix `study.py` (@josenavas) (https://github.com/biocore/qiita/pull/1301)
 - [x] Fix `user.py`(@josenavas) ~~(https://github.com/biocore/qiita/pull/1294)~~
 - [ ] Fix `util.py`
 - [ ] Add mention to the changelog
 - [ ] Modify `sql_connection.py` documentation (or delete it...)
 - [ ] Adapt `qiita_ware` to the new API
 - [ ] Adapt `qiita_pet` to the new API

Does this sounds like a reasonable way of doing this? I'm going to start modifying the queue system right now, so if the design is not good I would appreciate a fast comment.

Note that while we are not just fixing the files to make use of the new queue system. We are fixing them also to make sure that everywhere that needs to be a single transaction is a single transaction!


josenavas:
I just want to notify everybody the strategy that I'm going to follow in here. I'm going to start with a file that will produce the minimal impact to the rest of the code. I will fix that file and any other function from other objects that I need to get the tests passing on such object. Then I'm going to move to the next file that impacts a bit more of code and do the same.

This will allow me to do the most atomic changes possible. This also means that if I didn't change a property or a function in one class, it might be changed later. If at the end is not changed, that means that such property/method is not used within any other transaction and such there is no need on overcomplicating the API.

Does this seems like a good strategy?

josenavas:
I will close the PR once I've gather the comments from all of them. Thanks!

PR #1266, opened 2015-06-16
WIP: Add portals to qiita - needs 3 👍

squirrelo:
This is a lot of work, and will require many steps. The general goal is to have portals be a seamless backend thing that users of the system will never see. This means that all queries that require portal types will seamlessly add a "where portal_type_id = ?????" and act as if everything returned is what is available in the database, along with other considerations. The full list of changes is as follows:

- [x] Add portal types to analyses (PR #1256)
- [x] Move portal types out of study table and into join table, since one study can be in multiple portals (PR #1256)
- [x] Create portal manipulation scripts for qiita_db (PR #1256)
- [x] Create portal manipulation scripts for qiita scripts (PR #1316)
- [x] Create portal manipulation interface for qiita_pet (PR #1310)
- [ ] ****Make all functions of objects work in queues**** (PR #1262)
- [x] Update analysis object to be portal aware (PR #1289)
- [x] Update study object to be portal aware (PR #1287)
- [x] Update search object to be portal aware (PR #1286)
- [x] Update user object to be portal aware (PR #1285)
- [x] Add branding ability to qiita_pet (PR #1275)
- [x] Add a note to the changelog (PR #1312)
- [x] Add documentation on how to create and maintain a new portal. (PR #1314)
- [x] Add the default analysis on user creation (PR #1285)
- [x] Change the way we are using `convert_to_id` in the this PR. (PR #1328)


rob-knight:
I don’t understand this comment: the whole point is that users will see different branding, and perhaps have access to different functionality, depending on the portal?

squirrelo:
Correct. In discussing how this will work, we decided that it would be best if, to the users, the qiita_db objects acted as if they had access to the entire database regardless of what they actually have access to, essentially walling off access at below the qiita_db object level. This means calling `Study.get_by_status()` or `User.available_studies` will automatically limit what is available based on portal, without the user explicitly doing anything regarding portals.

This is going to require retooling a lot of the objects, as well as the interface. After it is implemented, though, changing portals should be as easy as changing the portal name in the setting file and everything else will be automatic and walled off safely.

rob-knight:
Ok thanks for the explanation.

On Jun 16, 2015, at 7:03 PM, Joshua Shorenstein <notifications@github.com>
wrote:

Correct. In discussing how this will work, we decided that it would be best
if, to the users, the qiita_db objects acted as if they had access to the
entire database regardless of what they actually have access to,
essentially walling off access at below the qiita_db object level. This
means calling Study.get_by_status() or User.available_studies will
automatically limit what is available based on portal, without the user
explicitly doing anything regarding portals.

This is going to require retooling a lot of the objects, as well as the
interface. After it is implemented, though, changing portals should be as
easy as changing the portal name in the setting file and everything else
will be automatic and walled off safely.

—
<https://github.com/biocore/qiita/pull/1266#issuecomment-112622917>.

josenavas:
@squirrelo Please do not check the items in the list until the code is actually merged on this branch. I unchecked it and added the PR # next to them. If those numbers are not correct, please correct. If you want to mark that you're working on them, please just add your name next to the item.

ElDeveloper:
I just added another bullet point 'Add documentation on how to create and maintain a new portal.'.

squirrelo:
How is  `Add the default analysis on user creation` not part of `Update user object to be portal aware`?

josenavas:
It is part, but taking into account that you're doing partial changes I want to make sure that everything is changed... Again, the way you're proposing doing the changes is a way of adding a lot of bugs. What you should have done is not modify any of the other files until you don't want to completely change them.

josenavas:
@squirrelo There is a merge conflict in here, you probably want to solve it before continuing working on this. Since this is a special case, I would suggest that in order to resolve the merge conflicts, you should do a PR against this branch, so we can check that the merge conflicts have been correctly fixed. Does this make sense?

josenavas:
Just wanted to add a comment in here. Note that there is not a problem on merging this into master once is ready even #1262 is not merged. We can't deploy multiple portals if #1262 is not finished, but that shouldn't block this PR as right now we only have one portal.

squirrelo:
All items in the list of things to do now have pull requests in, so please review quickly so we can get this merged into master and start working on Transactions for these object changes.

ElDeveloper:
We now need 3 devs on this PR to review and merge as soon as possible so the transition into transactions can go as smooth as possible, can we get 3 volunteers? I'll have a look later today.

ElDeveloper:
:+1: just tried this out and it looks very good!

antgonza:
Agree, tested several possibilities and looks great! :+1:

josenavas:
I tried multiple things in here and everything worked!

squirrelo:
Fireworks

PR #1275, opened 2015-06-18
Portals styling

squirrelo:
This adds styling ability to the portals. You can change the colors of the navbar and all items in it, as well as the site title and index text. It also allows changing of the example search on the study listing page.


squirrelo:
Ready for review. Small PR so should be fast to review.

qiita_pet/support_files/portal_media.py
```diff
@@ -0,0 +1,53 @@
+_SITEBASE = {
```
josenavas: This file is missing the copyright.

josenavas:
I don't think the approach taken here is correct. The goal is to be able to change the qiita_config and then the entire website becomes a different portal. However, with the approach taken here, when you change the portal you need to change `qiita_pet/support_files/portal_media.py` in order to put the correct values.

I think the best approach is as follows:
 - All data that is stored in `qiita_pet/support_files/portal_media.py` should be stored in a configuration file.
 - Each portal stores in the DB the path to its configuration file
 - When Qiita is started, the configuration file for the given portal is parsed and stored in the `portal_styling` variable
 - All the other modifications done shouldn't be affected by this change.

Also, it is really weird that we are importing from `support_files`. Just create a new module in `qiita_pet, `style.py` where the parsing of the configuration file is implemented and `portal_styling` is defined at the end. The configuration files that we distribute with the system (qiita, emp, etc...) can live then in `support_files/style/`.

ElDeveloper:
I think this looks good! The only thing that I think is not right, is that styling is being mixed with content i.e. `portal_media.py`. The way content is currently being modified seems fine, however style should probably be modified through a [CSS file](http://sphinx-doc.org/theming.html), see for example what Sphinx does, you only have to define a CSS file and it gets applied to the whole site, which I think would be nice to have in this case.

In addition there's no way to define which style/content belongs to which portal.

josenavas:
Good point @ElDeveloper! I think we should merge my solution with yours and make sure that each portal define it's own CSS file.

squirrelo:
I can't do the CSS file and have the variables inserted as far as I can tell when it comes to tornado. 

As for @josenavas suggestion, I do not agree at all. We said that a portal is it's own running webserver, so modifying the config file allows for DB setup. Then modifying the media info changes the look of the site. These are two completely separate actions, especially if we want to keep logical separation of qiita_db and qiita_pet. The way I have done it allows for fast and easy changes to the CSS without having to have someone go in and write the actual CSS themselves, which is what we wanted in the first place.

josenavas:
@squirrelo your suggestion will not work if you want to start 2 servers in the same machine and using a single installation of Qiita...

Completely against this solution.

squirrelo:
Neither will our current setup for config files, since it relies on a system variable. Are you committing to fixing that issue?

josenavas:
that is as easy as doing `bash test.sh` with the following contents in test.sh"
```bash
source /path/to/bash_profile
export QIITA_CONFIG_FP=path/to/config/file
qiita pet webserver start
```
:smile:

josenavas:
While your solution will require 2 different installations of the qiita code

squirrelo:
That is not intuitive for most users, and bash scripting is a hack work around. It needs to be a parameter you pass to web server start.

josenavas:
Well even easier, with a parameter is solved, if you want, we add it to the list and I will have it in 5 minutes.
But, of course, modifying the code is waay more intuitive...

squirrelo:
OK, now works as config file.

qiita_pet/portal.py
```diff
@@ -0,0 +1,94 @@
...
+        Welcome text header for the website
+    index_text : str
+        Welcome text for the website
+    example_search : str
+        Example search to be shown on the study listing page
+    """
+    def __init__(self, conf_fp=None):
+        # If conf_fp is None, we default to the test configuration file
+        try:
+            conf_fp = environ['QIITA_PORTAL_FP']
```
josenavas: I don't agree with this one. Just add another key to the current configuration, like you added portal.

qiita_pet/support_files/portal.txt
```diff
@@ -0,0 +1,45 @@
+[sitebase]
```
josenavas: Rename file to `qiita_portal.txt` and add a comment on top as it is done for `qiita_core/support_files/config_test.txt`.

josenavas:
Thanks @squirrelo!!!
This really looks way better! Can you add tests to the `qiita_pet/portal.py` module?

Also, can you elaborate why you can't use CSS for the style?

squirrelo:
I don't believe there is a way for tornado to replace values in a css file, unless we directly write it out to file on every startup of the system. That to me seems messier than the current configuration.

squirrelo:
Also this is using CSS for style, but is inline on the page, not a separate file.

ElDeveloper:
Hmmm, I believe all you would have to do is load the contents of the
specified CSS path and put it in sitebase.html right?

josenavas:
You don't need to replace values in an css, you can have the css defined in its own file and then just inject it in the page using [`embedded_css`](http://www.tornadoweb.org/en/branch2.2/web.html#tornado.web.UIModule.embedded_css).

squirrelo:
embedded_css only works on UIModules, not on straight WebHandler pages. Just tested and it will not work. Also, I am trying really hard not to have users edit the CSS directly, but instead just specify colors. 

As for the portals being its own settings file, I think this is needed. It does not fit the standard setup we have for the settings file, as it is directly for colors and text in the qiita_pet site, not settings related to the site. Furthermore, since the ability to turn on and off features by portal has been brought up, having that in a separate settings file is also useful for condensing things.

josenavas:
@squirrelo I agree on having portals in its own settings file. Probably I was not explicit enough on my comment: instead of defining a new system variable for this file, store the path of that file in the QIITA config file. In your PR #1256 you added PORTAL, I'm asking for this:
```
#Portal configuration the site is working under
PORTAL = QIITA
PORTAL_SETTING_FP = /path/to/settings.txt
```

BTW, I discussed with @ElDeveloper about the CSS, and he has a really good point. On top of the sitebase.html you can include the CSS using the path directly (which means that it should live under static). The major advantage about this solution is that the different portals can completely changes the CSS of the page, instead of only the few values that you're allowing here.

Note that portals wills not be created by normal users, they will be admins (as otherwise there is no point of having portals). So asking them modifying CSS is not a big deal (anyways they will need to know more or less CSS in order to know what they're changing in the configuration file)...

@ElDeveloper can you add/extend the advantages of using CSS directly?

ElDeveloper:
Yes, thanks. I agree this is not an average user feature, this is an
admin-level feature. As for the custom.css, I think the idea would be to
read the contents and paste them into sitebase.html itself, not to add a
path that lives under /static/. The reason you want to do this, is so
that multiple profiles can be added independently of the location of the
source.

squirrelo:
My worry is if you allow full customization of CSS, the help documentation goes right out the window for any portal that is not qiita. The current way allows for consistency in how you interact with the site and what things look like, just different colors. Allowing full re-skinning destroys that consistency.

squirrelo:
Another thing to consider is plug-ins .  Having stable CSS in the base code makes plug-ins much easier and safer.  Allowing custom CSS has the potential to cause display issues for a plug-in just because of CSS conflicts.  I say leave it a it is now and if, after plug-ins go in, we find a way to overcome this and the above problems, it can be implemented as full custom skinning. Until then the current implementation will work fine.

josenavas:
I don't understand how this will affect plug-ins, as plug-ins should follow the same CSS as the rest of the site (i.e. the user don't even know that he is using a plugin).
Also, I don't understand neither how this will affect the help documentation, as the processes will be the same although the page has a different look. Note that although the CSS can change a lot, the contents of the page will not change.

josenavas:
However, in the interest of time and how critical is to have this functionality in, I'd suggest that we can change this later. Do you agree @ElDeveloper ?

ElDeveloper:
> My worry is if you allow full customization of CSS, the help documentation goes right out the window for any portal that is not qiita.

The documentation is something independent from the system, if the styles don't match that's fine. 

> The current way allows for consistency in how you interact with the site and what things look like, just different colors. Allowing full re-skinning destroys that consistency.

With either solution, you could find a way to "destroy consistency" in one or another way.

> Another thing to consider is plug-ins . Having stable CSS in the base code makes plug-ins much easier and safer.

How is CSS less safe than python code modifications. Note that these are features you want to control as the maintainer of the system, so it is your responsibility when you deploy new plugins, styles, etc.

> I say leave it a it is now and if, after plug-ins go in, we find a way to overcome this and the above problems, it can be implemented as full custom skinning.

:-1:.

----

Maybe having a quick call will help us reach an agreement.

squirrelo:
If the styles dont match the documentation, we are going to have a lot of very confused end users. You can theoretically remove bits,  hide bits, move bits, etc just through CSS. Having a consistent interface to the help documentation, especially for our target users (biologists), is very important. 

Also, assume you have a plugin that expects default behavior for the HTML or CSS. Now the portal overrides that functionality. You have the possibility of having a gigantic mess at that point, with the interface not rendering correctly. By hard coding all but the colors, we have a consistent interface to build off of for plugins.

squirrelo:
Also @ElDeveloper: "With either solution, you could find a way to "destroy consistency" in one or another way."
How exactly would you do it short of straight up editing the template in this case? This keeps it so all you do is add different colors, no css knowledge needed.

ElDeveloper:
I see what you mean with "consistency". However I don't see this as a real concern. I have expressed my opinion and view (:-1:), if two developers want to :+1: this, I'd be fine though.

josenavas:
I agree with @ElDeveloper view, but I was tired of arguing with @squirrelo 

My :-1: for this PR to.

ElDeveloper:
Also, if the ultimate decision is to use the configuration file in `qiita_pet/support_files/portal.txt`, it would be good to include a new section in the documentation that describes the file and the parameters.

ElDeveloper:
@adamrp, @wasade & @antgonza can someone *qiime* in?

rob-knight:
Please be nice, although I am seeing a lot of arguing that doesn’t always seem warranted. I appreciate everyone’s efforts to reach consensus rapidly.

squirrelo:
Using custom CSS file now. Good to go.

qiita_pet/portal.py
```diff
@@ -0,0 +1,87 @@
...
+    example_search : str
+        Example search to be shown on the study listing page
+    custom_css : str
+        custom css for the portal
+    conf_fp : str
+        The filepath to the portal styling config file
+    css_fp : str
+        The filepath to the portal styling custom CSS
+    """
+    def __init__(self, conf_fp=None):
```
josenavas: Looks like the `conf_fp` parameter is ignored (probably leftover code). Can you remove?

squirrelo: woops, yup. Done.

qiita_pet/portal.py
```diff
@@ -0,0 +1,87 @@
...
+        if qiita_config.portal_fp:
+            self.conf_fp = qiita_config.portal_fp
+        else:
+            self.conf_fp = join(dirname(abspath(__file__)),
+                                'support_files/config_portal.txt')
+        self.css_fp = qiita_config.css_fp
```
josenavas: Shouldn't this be included in the above if statement? In the same way that we have a default config_portal.txt we should have the default css.

squirrelo: Default CSS is the bootstrap styling in this case. Giving custom CSS changes how this looks, so by default we just use standard with no modification.

josenavas: Oh, ok

josenavas:
:+1: @ElDeveloper can you re-view?

@squirrelo I'm wondering if this PR will have a merge conflict with #1256? Both are adding stuff to the config file.

squirrelo:
Shouldn't. This one is adding text to the end, that one is adding text to the middle.

qiita_pet/support_files/config_portal.txt
```diff
@@ -0,0 +1,41 @@
+# ---------- Base information for the website ----------
```
ElDeveloper: Can you change the extension to `.cfg`? That way the format will be recognized both by github and text editors, for example like [this](https://gist.github.com/ElDeveloper/c0ea363168b6f86cf63f).

squirrelo: Can do. Should I change the base config file to .cfg instead of .txt as well?

qiita_pet/support_files/config_portal.txt
```diff
@@ -0,0 +1,41 @@
...
+[sitebase]
+LOGO = /static/img/logo-clear.png
```
ElDeveloper: Might be useful to note here (or in the docs in a later PR) what the size of this logo should be.

qiita_pet/portal.py
```diff
@@ -0,0 +1,87 @@
...
+    title : str
+        Site title
+    index header : str
+        Welcome text header for the website
+    index_text : str
+        Welcome text for the website
+    example_search : str
+        Example search to be shown on the study listing page
+    custom_css : str
+        custom css for the portal
```
ElDeveloper: css -> CSS

qiita_pet/support_files/config_portal.txt
ElDeveloper: That would be good, but is not necessary as part of this PR ... though
cheers if you do do that!

qiita_core/support_files/config_test.txt
```diff
@@ -122,13 +122,23 @@ EBI_CENTER_NAME = CCME-COLORADO
...
 USER = user
 PASSWORD = password
 URL = https://vamps.mbl.edu/mobe_workshop/getfile.php
+
+# ----------------------------- Portal settings -----------------------------
+[portal]
+# Full path to portal styling config file
+PORTAL_FP =
+# Full path to portal custom CSS styling file
+CSS_FP =
```
ElDeveloper: Shouldn't this FP be living inisde the portal config file?

ElDeveloper:
This works pretty well, I just deployed this on my local system, downloaded [this bootswatch theme](https://bootswatch.com/cerulean/) and modified my config:

web-surfin

Thanks for making the changes @squirrelo :+1:

qiita_core/support_files/config_test.txt
squirrelo: I'm keeping all the various filepath things in the main config file just to centralize things.

ElDeveloper: The semantic separation between portal specific attributes makes more sense to me than value types. If not, we might as well just have everything in a single file. This is not a blocking comment though.

squirrelo: Eh. Anyone want to be tie breaker?

squirrelo: Thinking more yeah, that makes sense. Change made.

ElDeveloper: :+1:

squirrelo:
I don't know why this is failing. The files were renamed and work on my computer, and the commits seem good for the name changes.

qiita_pet/support_files/config_portal.cfg
```diff
@@ -0,0 +1,44 @@
...
+  Earth Microbiome Project</a> and the
+  <a target=_blank href="https://ibdmdb.org/">
+  Inflammatory Bowel Disease Multi'omics Database</a>.</p>
+  <br />
+  <p align="left">
+    Qiita is built on top of the following open-source technologies:
+    <ul>
+      <li><a target=_blank href="http://qiime.org">QIIME</a></li>
+      <li><a target=_blank href="http://scikit-bio.org">scikit-bio</a></li>
+      <li><a target=_blank href="http://emperor.colorado.edu">EMPeror</a></li>
```
ElDeveloper: Just noticed this, can you change colorado.edu -> microbio.me

ElDeveloper:
I think the problem is in the travis.yml file, see [here](https://github.com/biocore/qiita/blob/master/.travis.yml#L26).

squirrelo:
Awesome, that's probably it. Thanks.

ElDeveloper:
:+1:, thanks @squirrelo!

PR #1276, opened 2015-06-19
Add `Transaction` object

josenavas:
This PR is the first of a series of PR to be able to modify Qiita DB to make use of transactions everywhere.

As discussed in the last meeting, instead of modifying the current queue system it will be better to add a `Transaction` object that encapsulates the transaction.

A few changes from the original design:
 - The placeholders, instead of being of the form `{#:#} -> {query_idx:result_idx}` are of the form `{#:#:#} -> {query_idx:row_id:value_idx}`. This allow for more flexibility and it is easier to reference an actual result. It also removes the need of flattening the results.
 - The `execute` method has a boolean parameter `commit`, optional and default `True`. This allows to execute all the current queries in the transaction without committing it, which is useful if you need an intermediate result of the transaction to execute some python code (e.g. instantiate an object to retrieve some other information needed for adding more queries to the DB)
 - The `execute` methods returns always all the results. You can retrieve then a specific result using the index, or using `[-1]` to retrieve the result of the last query. This last part was impossible before unless you knew how many results where returned by the last query.

The tests are expected to fail, since the rest of the code has not been modified. Those will be further PR.


qiita_db/sql_connection.py
```diff
@@ -460,176 +450,199 @@ def execute_fetchall(self, sql, sql_args=None):
...
-        self.queues[queue_name] = []
+    def _replace_placeholders(self, sql, sql_args):
+        """Replaces the placeholder in `sql_args` with the actual value
 
-    def list_queues(self):
-        """Returns list of all queue names currently in handler
+        Parameters
+        ----------
+        sql : str
+            The sql query
```
ElDeveloper: minor sql -> SQL

qiita_db/sql_connection.py
```diff
@@ -460,176 +450,199 @@ def execute_fetchall(self, sql, sql_args=None):
...
+        sql_args : list or None
+            The arguments of the SQL query
 
         Returns
         -------
-        list of str
-            names of queues in handler
+        tuple of (str, list of objects)
+            The input sql query (unmodified) and the sql arguments with the
```
ElDeveloper: sql -> SQL ... :sheep:

qiita_db/sql_connection.py
```diff
@@ -460,176 +450,199 @@ def execute_fetchall(self, sql, sql_args=None):
...
+                                "any previous result"
+                                % (q_idx, r_idx, v_idx))
+                        except TypeError:
+                            # The query that the placeholder is pointing to
+                            # is not expected to retrieve any value
+                            # (e.g. an INSERT w/o RETURNING clause)
+                            self._raise_execution_error(
+                                sql, sql_args,
+                                "The placeholder {%d:%d:%d} is referring to "
+                                "an sql query that do not retrieve data"
```
ElDeveloper: do -> does ... and sql -> SQL

qiita_db/sql_connection.py
```diff
@@ -460,176 +450,199 @@ def execute_fetchall(self, sql, sql_args=None):
...
+        return sql, sql_args
+
+    def add(self, sql, sql_args=None, many=False):
+        """Add an sql query to the transaction
+
+        If the current query needs a result of a previous query in the
+        transaction, a placeholder of the form '{#:#:#}' can be used. The first
+        number is the index of the previous SQL query in the transaction, the
+        second number is the row from that query result and the third number is
+        the index of the value withing the query result row.
```
ElDeveloper: withing -> within

qiita_db/sql_connection.py
```diff
@@ -460,176 +450,199 @@ def execute_fetchall(self, sql, sql_args=None):
...
-            self.queues[queue].append((sql, args))
-
-    def _rollback_raise_error(self, queue, sql, sql_args, e):
-        self._connection.rollback()
-        # wipe out queue since it has an error in it
-        del self.queues[queue]
-        raise ValueError(
-            "Error running SQL query in queue %s: %s\nARGS: %s\nError: %s"
-            % (queue, sql, str(sql_args), e))
+            if args and type(args) != list:
```
ElDeveloper: can you use `isinstance`?

qiita_db/sql_connection.py
```diff
@@ -460,176 +450,199 @@ def execute_fetchall(self, sql, sql_args=None):
...
         ----------
-        queue : str
-            Name of queue to execute
-
-        Notes
-        -----
-        Does not support executemany command. Instead, enter the multiple
-        SQL commands as multiple entries in the queue.
+        commit : bool, optional
+            Whether if the transaction should be committed or not. Defaults
```
ElDeveloper: `Whether if` -> `Whether`

qiita_db/sql_connection.py
```diff
@@ -460,176 +450,199 @@ def execute_fetchall(self, sql, sql_args=None):
...
-            Queue name already exists
+        ValueError
         """
-        if self._check_queue_exists(queue_name):
-            raise KeyError("Queue %s already exists" % queue_name)
+        self.rollback()
+        raise ValueError(
+            "Error running SQL query in transaction %s:\n"
+            "Query: %s\nArguments: %s\nError: %s\n"
+            % (self._name, sql, str(sql_args), error))
```
ElDeveloper: Might be a good idea to do `str(error)` since the places where this is being used pass an exception object.

ElDeveloper: Or alternatively pass the exception objects as strings.

qiita_db/sql_connection.py
```diff
@@ -460,176 +450,199 @@ def execute_fetchall(self, sql, sql_args=None):
...
         return result
 
-    def _check_queue_exists(self, queue_name):
-        """Checks if queue `queue_name` exists in the handler
 
-        Parameters
-        ----------
-        queue_name : str
-            The name of the queue
+class Transaction(object):
```
ElDeveloper: Can you update the module docstring with any appropriate changes?

josenavas: I've just added `Transaction` to the classes list and removed the examples. I will add the examples back once I'm 100% sure that I don't have to do any further change in the Transaction class (maybe adding some functionality that is needed and I didn't foresee). In order to not forget about this, I've added an item in #1262

qiita_db/sql_connection.py
josenavas: Added `str(error)` since it minimizes the code changes and the fact of remembering that we need to cast to str.

qiita_db/sql_connection.py

qiita_db/sql_connection.py

qiita_db/sql_connection.py

qiita_db/sql_connection.py
josenavas: Hahaha... my fingers are faster than my brain...

qiita_db/sql_connection.py
josenavas: Sure! Done

qiita_db/sql_connection.py

josenavas:
Thanks @ElDeveloper ! Comments addressed

ElDeveloper:
Awesome, thanks! :+1:

qiita_db/sql_connection.py
```diff
@@ -460,176 +397,199 @@ def execute_fetchall(self, sql, sql_args=None):
...
-        if self._check_queue_exists(queue_name):
-            raise KeyError("Queue %s already exists" % queue_name)
+        self.rollback()
+        raise ValueError(
+            "Error running SQL query in transaction %s:\n"
+            "Query: %s\nArguments: %s\nError: %s\n"
+            % (self._name, sql, str(sql_args), str(error)))
 
-        self.queues[queue_name] = []
+    def _replace_placeholders(self, sql, sql_args):
```
squirrelo: Please add an example of the placeholder to the notes section, and an explanation of each of the three bits.

qiita_db/test/test_sql_connection.py
```diff
@@ -193,173 +195,302 @@ def test_execute_fetchall_with_sql_args(self):
...
+class TestTransaction(TestBase):
+    def test_init(self):
+        obs = Transaction("test_init")
+        self.assertEqual(obs._name, "test_init")
+        self.assertEqual(obs._queries, [])
+        self.assertEqual(obs._results, [])
+        self.assertEqual(obs._index, 0)
+        self.assertTrue(isinstance(obs._conn_handler, SQLConnectionHandler))
+
+    def test_replace_placeholders(self):
```
squirrelo: can you add a test here for when there are a few inserts or updates followed by another select, so for example res5a would be next in the results list, as query 5.

squirrelo: Basically we need a test for when you have a select, followed by inserts that use the select, then anther select, then inserts that use the second select.

josenavas: Done

josenavas:
@squirrelo I can't find one of your comments, I assume that you deleted it? If so, please, instead of deleting the comment, comment back saying why no longer applies...

qiita_db/sql_connection.py
```diff
@@ -460,176 +397,199 @@ def execute_fetchall(self, sql, sql_args=None):
...
-        Parameters
-        ----------
-        queue_name : str
-            Name of the new queue
+    def __init__(self, name):
+        # The name is useful for debugging, since we can identify the
+        # failed queue in errors
+        self._name = name
+        self._queries = []
+        self._results = []
```
squirrelo: What does memory use look like for this list on large transaction blocks? Since this is a dense representation of all SQL queries' results, this is a concern.

qiita_db/sql_connection.py
```diff
@@ -460,176 +397,199 @@ def execute_fetchall(self, sql, sql_args=None):
...
+                                "any previous result"
+                                % (q_idx, r_idx, v_idx))
+                        except TypeError:
+                            # The query that the placeholder is pointing to
+                            # is not expected to retrieve any value
+                            # (e.g. an INSERT w/o RETURNING clause)
+                            self._raise_execution_error(
+                                sql, sql_args,
+                                "The placeholder {%d:%d:%d} is referring to "
+                                "an SQL query that does not retrieve data"
```
squirrelo: a SQL query...

qiita_db/sql_connection.py
```diff
@@ -460,176 +397,199 @@ def execute_fetchall(self, sql, sql_args=None):
...
-            queue does not exist
-        """
-        if not self._check_queue_exists(queue):
-            raise KeyError("Queue '%s' does not exist" % queue)
+        TypeError
+            If `sql_args` is provided and is not a list
 
+        Notes
+        -----
+        If `many` is true, `sql_args` should be a list of lists, in which each
```
squirrelo: This is confusingly worded. Something along the lines of "...each list in the list contains the parameters for one SQL query of the many."

qiita_db/sql_connection.py
josenavas: Not different as what you had before.

josenavas: Note that this depends only on what you're returning. If you are not returning anything the memory footprint is only the representation of a None. On the other hand, if you're returning some values, you probably want them in memory anyways right?

qiita_db/sql_connection.py

qiita_db/sql_connection.py

qiita_db/sql_connection.py
squirrelo: What is the overhead of a dictcursor though? Remember that it was originally storing just the last query, and wiping that out when used, so it was never going to have a large list. This stores all returned things as a dictcursor.

josenavas:
Ready for another pass!

qiita_db/sql_connection.py
josenavas: Nope, originally was not storing the results of the last query. Originally was flattening the results of all queries and storing them in the list until something was used. If anything was used, the list was wiped out, but if not, it was stored everything in memory.

Again, if you are returning something from the database, you're going to have that memory footprint in the system anyways. If you think that the list is containing something that you don't need, probably you should re-visit your SQL queries.

You'll have the DictCursor anyway when you're returning a value from the database (i.e. executing a fetchall). Note that the only real overhead that this is adding is the 'None' values when the query is not returning anything. This value can be avoided using a dict instead of a list, however you then have an overhead on solving the indirections of the dictionary, so probably better as it is.

josenavas:
@squirrelo since this is "blocking" your work (or at least that what you stated) you should finish reviewing this PR...

squirrelo:
Something I've run into and want to make sure can be done: can you take a list from a query and loop over the results, e.g.

Select all portal ids

loop over ids and add analysis to them

all in a queue.

josenavas:
with `execute(commit=False)` there are some examples in my other PRs...

josenavas:
Just for completeness, if you want to do that automatically in the queue, makes the placeholders mostly break, since that is technically an executemany, but is done internally and the user do not know that. This makes any future placeholder to be shifted wrt what the user expects, which then makes the placeholders useless. I've already try to find an elegant solution, and the best one is to allow the partial execution of the transaction w/o committing, do some python work, and finish the transaction. This still encapsulates the nice Transaction object, while it still allows you to do whatever you want.

In short, it allows to do any execute_fetch{all/one} without committing and still having the transaction open.

squirrelo:
Your example doesn't work as it executes before the queue is filled. I need to create a User and analyses all in one queue, and do what is stated, then edit the user more.  Your example looks like it will leave that user and analysis  in a limbo state

josenavas:
Forget about queues, we are talking about transactions. It is the same transaction as it is not committed... You add some queries to the transaction, you execute them w/o committing, adding some more, and execute them committing. There is an example [here](https://github.com/biocore/qiita/pull/1276/files#diff-fd8fa1ae44937ec711b82368c103fe10R423).

Also, note that this is allowing you to do what is done internally in the `execute_queue`, execute a bunch of queries while you're executing some intermediate python and committing at the end.

wasade:
Does commit=False lock tables?

josenavas:
The actual documentation is in [here](http://initd.org/psycopg/docs/connection.html#connection.autocommit) although a bit misleading given the title.

psycopg2 locks the table until a rollback or a commit is done. Since we are not doing any commit/rollback, those tables are locked until the last query is executed, which is when we actually commit.

josenavas:
So, @wasade yes tables are locked as long as you're executing the transaction with the `commit=False` parameters. However, note that if you forgot to commit and/or rollback, those tables can be locked forever. But this is common in SQL programming, and we didn't hit this issue before because we're auto-committing always.

wasade:
Yes. Should this be allowed? The chance of a bug impacting the whole
infrastructure is high. If the infrastructure halts, we risk unanticipated
consistency issues.

@squirrelo, can your usecase be expressed as an actual transaction block?

josenavas:
There is no way we can avoid this. Even in the execute_fetchall function, if the infrastructure halts just before the call to commit is executed, we have the same problem. Or if we completely remove the SQL connection handler and we use plain psycopg2, we have the exact same issue. Example:
```python
         with connect(user=qiita_config.user, password=qiita_config.password,
                      host=qiita_config.host, port=qiita_config.port,
                      database=qiita_config.database) as con:
             with con.cursor() as cur:
                 cur.executemany(sql, sql_args)
             con.commit()
```
If the system halts between the cur.executemany and the con.commit, will be the exact same situation.
However, I think that this is something that the postgres/psycopg2 developers have already faced and have something in place. In any case, I'm not adding any other issue that is not already in there, and that we can tackle.

Re @squirrelo use case. Yes:
```python
         with connect(user=qiita_config.user, password=qiita_config.password,
                      host=qiita_config.host, port=qiita_config.port,
                      database=qiita_config.database) as con:
             with con.cursor() as cur:
                 cur.executemany(sql, sql_args)
             # Do something in python
             # Do something in SQL
             con.commit()
```
The overall idea is this. The Transaction object, is simplifying this entire chunk of code by doing:
```python
t = Transaction()
t.add(sql, sql_args, many=True)
t.execute(commit=False)
# Do something in python
t.add("#Do something in SQL")
t.execute()
```

wasade:
Yes, this is one reason why execute calls are wrapped in a try/except. This
is a different situation that is not analogous as:

t = Transaction()
t.add(stuff)
t.execute(commit=False)
return None

...will rely on gc if gc picks it up (as resource dependent objects tend to
behave unexpectedly). This is a toy example obviously, but assuming the
developer is always going to write perfect code is risky.

The other scenario I can think of that is concerning is:

t = Transaction()
t.add(stuff)
t.execute(commit=False)
raise ValueError

Will the transaction block rollback in the destructor, or will an exception
take out the whole infrastructure? If it rolls back in the destructor, are
we forced to wait on gc? Are there assurances that gc will pick it up?

qiita_db/sql_connection.py
```diff
@@ -460,176 +397,202 @@ def execute_fetchall(self, sql, sql_args=None):
...
+    def _replace_placeholders(self, sql, sql_args):
+        """Replaces the placeholder in `sql_args` with the actual value
 
-    def list_queues(self):
-        """Returns list of all queue names currently in handler
+        Parameters
+        ----------
+        sql : str
+            The SQL query
+        sql_args : list or None
```
wasade: Why not use `[]` instead of `None`? Less type confusion, and it reduces a level of nesting

josenavas:
@wasade Reading the psycopg2 documentation [here](http://initd.org/psycopg/docs/connection.html) and [here](http://initd.org/psycopg/docs/usage.html#thread-safety) we can see that the transaction rollbacks automatically on deletion. Of course, we will depend on gc but I don't think there is a lot that we can do in here (doesn't matter what we do, if we use the transaction object or not, I can see examples that will have the same situation).

However, one thing that popped up while reading such documentation is that it looks like it is not safe to be using a single connection unless we are ensuring that any given call executed within a transaction is not committed, as it will partially commit the transaction.

An alternative will be to use a new connection for each transaction (can be encapsulated in the same transaction object).

Independently of the solution that we reach here, reading all this documentation means that we need to mostly modify all the code in Qiita DB to actually accept a transaction (whatever representation we want of it) as a parameter so we can ensure that all the changes in a given operation are performed in a transactional way.

What do you think?

qiita_db/sql_connection.py
josenavas: Uhm, good point. At this point, the parameter `sql_args` comes from the user, but I think one of the first checks that I can do when that parameters gets added to the transaction is to check that if it is None, I can put an empty list. psycopg2 does not file passing an empty list or None, so it will simplify code. Good suggestion! Thanks,

wasade:
What I'm generally wondering about is whether we actually need `commit=False`, or if we can phrase those problems differently? If we need it, then we deal with it. But if there are other ways to go about the problems we're looking at, then we may save ourselves a huge future headache. 

It is possible the gc won't trigger as this assumes that psycopg2 doesn't hold references. It may here due to the uncommitted nature.

I don't understand the single connection concern, can you elaborate? The only thing I can think of is:

```python
t = Transaction()
t.add(stuff)
t.execute(commit=False)
# do some other db query
t.add(stuff)
t.execute()
```

If we

josenavas:
Thinking a little bit more about this, it was a bad idea to use only a single connection. In psycopg2, a connection defines a transaction. At the very first SQL query executed in the connection, psycopg2 defines a transaction and it does not start a new one until it is committed.

What this means is that almost all our qiita_db functions/classes should have a transaction parameter (or connection - if we don't want to use the transaction object), optional, that allow to add the given operations to that transaction if is provided. If the transaction is provided, the function does not commit the changes. If it is not provided, it actually does.

What do people thing about this? Note that this requires a lot of changes that I'm willing to do, but we should reach a consensus soon. Does everybody see the problem that we are currently facing?

cc @antgonza @wasade @adamrp @ElDeveloper @squirrelo

josenavas:
Do people think a call might be useful in here?

wasade:
I don't understand the concern. Remember, we have a single threaded system,
and our execute calls are committed (right?)

josenavas:
Sorry @wasade I wrote my previous comment before reading yours. An example of the single connection problem:
```
t = Transaction()
t.add(stuff)
t.execute(commit=False)
value = Study(1).title # !!!
t.add(stuff)
t.execute()
```

The python code marked with `!!!` is internally executing a SQL query that at the end commits. Since we have a single connection (shared by the Transaction object and the Study object), the commit inside the `Study(1).title` will commit half of the transaction. Does this make sense?

wasade:
That is only a problem if we allow `commit=False`...

wasade:
And further highlights the issues with supporting it

josenavas:
But if we don't allow commit=False, we cannot represent all the possible transactions in the code base. Those transactions that you need to loop over the results of the first query to add the subsequent queries are not representable using the Transaction object. What this means is that in those situations we will need to do:
```
with conn_handler.get_postgres_cursor() as cur:
   #(1) Do the first query
   #(2) Loop over the results to generate the following queries
   #(3) Execute the next X queries (dependent of the first query)
conn_handler._connection.commit()
```
However, I think this solution have the same problems that using the transaction object. Example:
```
with conn_handler.get_postgres_cursor() as cur:
   #(1) Do the first query
   #(2) Loop over the results to generate the following queries
   #(3) Python call with an internal SQL query that commits
   #(4) Execute the next X queries (dependent of the first query)
conn_handler._connection.commit()
```
In this case (3) will commit our current transaction...

wasade:
The minimal change here would be for `Transaction` to instantiate its own
connection. I do not think we should revert `SQLConnection` back given the
huge prior performance overhead and, in general, a lack of needing multiple
connections.

In the first example, shouldn't that just be done in pure SQL? In other
words, is all this so that Python can be used as a convenience? We are not
the first people to encounter the need to have extended transactions, where
the subsequent database state changes are contingent on the present
database state. How are these handled in, say, SQLAlchemy?

josenavas:
I'm cc'ing the entire @biocore/qiita-dev since I've been discussing with @ElDeveloper offline and it turns out that the issue is quite big. I'll try to do my best to summarize the different issues that we have, the potential solutions and the solution that I think is the best one to move forward.

All the problems that we are encountering here is that our modifications to the database are not atomic, i.e. a function that interacts with the database executes multiple commits instead of a single one at the end of the function.

@wasade this is related with your question. The way SQLAlchemy or sqlite handles this situation is by always doing an explicit commit. The problem that we have here occurs due to the implicit commits that the SQLConnectionHandler is doing. As a remainder, those implicit commits were done to avoid ~5 lines of code for each call to the database, which I think is legitimate.

However, when we talk about transactions, the story changes. What should happen is that all the steps executed inside a function are committed at once, or none of them is committed.

Few examples of the current implementation that will leave the system in an inconsistent state:
  - `SampleTemplate.create` The DB changes are encapsulated in a transaction. Once the transaction is executed (and committed) the files are generated. However, if there is a problem during the creation of the files, the DB will contain the information although the files are not present in the FS, which is one of our assumptions. This is an inconsistency.
  - Creating SampleTemplate and PrepTemplate from a QIIME mapping file. If the PrepTemplate creation fails, the SampleTemplate still exists. This is inconsistent. Either both are created or None are created.

The above two examples are from the metadata, but there are other examples in the code. I've used those examples because they allow me to expose the following issue. A function is a transaction itself or not depending on its caller. If we just want to create a SampleTemplate, the function `SampleTemplate.create` it's a single transaction. However, if we want to create the SampleTemplate and the PrepTemplate from the QIIME mapping file, `SampleTemplate.create` becomes part of a bigger transaction, so the changes that it performs shouldn't be committed unless the entire, bigger transaction is successfully executed.

This example is an extreme, but this is what is continuously happening in our code with small functions (i.e. calls to `check_table`, `chek_required_columns`, etc; in general anything under `util.py`, but calls like `study.title` also fall in this category).

Thus, a function should commit or not depending on the context of its caller.

Possible solutions:
 1. This might be possible to solve (not sure 100%) if inside the objects we only use pure SQL queries, and we never use any other python call to the util functions. Drawbacks of this solution: not 100% sure that will work, duplicates a lot of code, small changes in the DB schema might required the rewrite of half of the queries in qiita_db. Advantages: we might gain performance (just a guess... not actual data).
 2. We use the transaction object implemented in this PR and we modify all the API in the qiita_db code base that interacts with the database to accept an optional parameter `transaction=None`. If the transaction **IS NOT** provided, the function then creates it's own, and at the end of the function **commits** the changes. If the transaction **IS** provided, it adds the queries to the transaction, it potentially executes it with `commit=False` to return the needed values, and **do not commit** the changes. Drawbacks of this solution: it changes a lot of the current API. Advantages: i'm 100% that works, allows to represent any kind of transaction, and reduces the code duplication. Since only the function that creates the transactions will commit it, we make sure that it actually works as a transaction. If there is a problem with the creation of the files (first example with the creation of the SampleTemplate) the function will not be committed, so no changes in the database and everything gets clean up.

My vote goes for 2 and I volunteer to change it. I think I can have it done by the end of the week if I start working on this tomorrow morning (or tonight if we reach a consensus earlier).

Sorry for the long comment, but there are a lot of potential issues in our qiita_db code base, and I think it is really important that we solve them ASAP. I will not start working on this until I know that other developers agree. There is a lot of work in here, and I don't want to spend time in something that we're not going to use.

josenavas:
@wasade @squirrelo @ElDeveloper 

I've updated the Transaction object to be a context manager. Few things to note:
 - The Transaction does not try to automatically execute in the `__exit__`. I've tried it and it generated some unexpected behavior (try to execute things have previously errored). Also, I prefer to have things explicit rather than implicit. Idea: Do not modify the DB if the programmer do not ask for it.
 - I did not remove the old queue system code. The reason behind this is we can then execute the tests and tests **now should pass**. I've added a comment in #1262 so we don't forget.
 - The 'execute' code has been moved to an internal `_execute` function and the exposed `execute` wraps the `_execute` one in a try/except, so we can catch and rollback any changes in there.

ElDeveloper:
I'm having a look at this right now.

qiita_db/sql_connection.py
```diff
@@ -633,3 +581,239 @@ def get_temp_queue(self):
...
+
+        The error message contains the name of the transaction, the failed
+        query, the arguments of the failed query and the error generated.
+
+        Raises
+        ------
+        ValueError
+        """
+        self.rollback()
+        raise ValueError(
```
squirrelo: Can this be a PostgresError? ValueError doesn't fit to me, since it can be a malformed query or other non-value based error. Maybe RuntimeError if Postgres is not doable?

qiita_db/sql_connection.py
```diff
@@ -633,3 +581,239 @@ def get_temp_queue(self):
...
+        list of the list contains the parameters for one SQL query of the many.
+        Each element on the list is all the parameters for a single one of the
+        many queries added. The amount of SQL queries added to the list is
+        len(sql_args).
+        """
+        if not many:
+            sql_args = [sql_args]
+
+        for args in sql_args:
+            if args:
```
squirrelo: can you combine this into one if:

`if args and not isinstance(args, list):`
Python short circuits so should be fine.

josenavas: No, I can't. Note that the else only applies to the first part of the if, not to the second.

squirrelo: Also, this won't fail if the passed thing is a None, e.g. sql_args = [None]. Not sure if this is an issue in the later code.

squirrelo: Ah, missed that else at the end, which takes care of both concerns. Cool.

qiita_db/sql_connection.py
```diff
@@ -633,3 +581,239 @@ def get_temp_queue(self):
...
+                except PostgresError as e:
+                    # Some other error happened during the execution of the
+                    # query, so we need to rollback
+                    self._raise_execution_error(sql, sql_args, e)
+
+                # Store the results of the current query
+                self._results.append(res)
+
+        # wipe out the already executed queries
+        self._index += len(self._queries)
```
squirrelo: Would it make more sense to just increment index during add? That way you can have index as a property instead of a defined decorator property.

squirrelo:
Couple of comments.

ElDeveloper:
This looks pretty good, as for your comment:

> Also, I prefer to have things explicit rather than implicit. Idea: Do not modify the DB if the programmer do not ask for it.

When I tested this out on an interactive session, it kinda doesn't make sense to have it as a context manager then, see for example here:

```python-traceback
In [24]: with Transaction('foo') as t:
   ....:     t.add('select 42+a')
   ....:     

In [25]: t.execute()
ValueError                                Traceback (most recent call last)
<ipython-input-25-904e3650ed03> in <module>()
----> 1 t.execute()

/Users/yoshikivazquezbaeza/git_sw/qiita/qiita_db/sql_connection.py in execute(self, commit)
    801         """
    802         try:
--> 803             return self._execute(commit=commit)
    804         except Exception:
    805             self.rollback()

/Users/yoshikivazquezbaeza/git_sw/qiita/qiita_db/sql_connection.py in _execute(self, commit)
    752                     # We catch any exception as we want to make sure that we
    753                     # rollback every time that something went wrong
--> 754                     self._raise_execution_error(sql, sql_args, e)
    755 
    756                 try:

/Users/yoshikivazquezbaeza/git_sw/qiita/qiita_db/sql_connection.py in _raise_execution_error(self, sql, sql_args, error)
    631             "Error running SQL query in transaction %s:\n"
    632             "Query: %s\nArguments: %s\nError: %s\n"
--> 633             % (self._name, sql, str(sql_args), str(error)))
    634 
    635     def _replace_placeholders(self, sql, sql_args):

ValueError: Error running SQL query in transaction foo:
Query: select 42+a
Arguments: []
Error: column "a" does not exist
LINE 1: select 42+a
```

The context of that transaction had seemingly "exited", but since the execute is never called, then nothing quite happens until you call `execute`. Therefore either (1) it doesn't make sense to have this as a context manager or (2) we should add an extra argument that controls whether or not changes will be executed once the context is exited. Thoughts?

qiita_db/sql_connection.py
josenavas: Uhm... The beauty of the ValueError is that it is already cached in everywhere else in the code where it is needed. The main issue in here is that this is a single Error point but the actual error can be for multiple reasons: malformed query, index error in the placeholder, a PostgresError, or any general exception that can be raised in the execute method.

I can change it to RuntimeError, however this means that the SQLConnectionHandler also should change the type of errors (for consistency) to RuntimeError. If you feel strong about it, I can change, but in such case I would prefer to open an issue and solve it later, as it minimizes the number of errors added by this PR. Makes sense?

squirrelo:
I'd agree to something in init, like moving `commit` variable to init instead of on execute. In theory a whole queue should be filled before the execute call, so having the call automatic in `__exit__` makes sense, but with this it would execute at the end regardless, and the commit would be controlled by the initial init setting of the queue.

qiita_db/sql_connection.py
squirrelo: Ah, did not realize that was so embedded in the system. Worrying, but not worth the change right now.

josenavas:
Uhm... Thanks @ElDeveloper. The original idea is that it gets executed inside the context manager, not outside.... I can try to play with it a little bit more and see if I can make it work by automatically executing once it exits the context AND preventing any other call once you're out of the context manager.

That is, if you try to execute this:
```python
with Transaction("t") as t:
    t.add("stuff")
t.execute()
```
The call to `t.execute()` will generate an error (it's outside of the context manger, so it's not safe to do it there). Does it make sense?

squirrelo:
Are transactions able to execute more than once? If not, a simple "executed" flag should be enough. If they can be run more than once, I guess check for SQL in the queue before executing?

josenavas:
You cannot execute the full transaction more than once, but you can partially execute a transaction (with commit=False).

ElDeveloper:
@josenavas, Yes I think that's the solution that makes the most sense!

qiita_db/sql_connection.py
josenavas: Good call! Done.

josenavas:
@ElDeveloper, @squirrelo,  there is still one case that I'm still not sure what to do. The case is when there are no more queries to execute, but the latest changes in the DB hasn't been committed. I can see arguments in one way and in another if I should commit them or not...

One reason to commit it automatically is that the developer forgot to commit, so doing it at the exit time makes sense.

However, in a situation like this one:
```python
with Transaction('t') as t:
   t.add("Do changes in the DB")
   t.add("Retrieve some value from the DB")
   value = t.execute(commit=False)[-1][0][0]
   if value == 'something that the developer doesn't want to commit':
       return False
   # Do some other stuff
```
The return False will make the context manager to actually commit the changes.... We can make clear in the documentation that it will commit automatically, which means that the developer should do this:
```python
with Transaction('t') as t:
   t.add("Do changes in the DB")
   t.add("Retrieve some value from the DB")
   value = t.execute(commit=False)[-1][0][0]
   if value == 'something that the developer doesn't want to commit':
       t.rollback()
       return False
   # Do some other stuff
```

Do you think this is a good solution?

squirrelo:
I'm still not liking the ability to execute without commit in there. You can use if statements in the SQL itself, and there is nothing in the transaction stopping you. I'd say if this is the case, it should be checked at the SQL level, not the python level, that way only things that are supposed to change actually change during the execution.

squirrelo:
Actually, the transaction objects are a good example of how we need to take better advantage of SQL's full capabilities, not just simple inserts and updates.

josenavas:
@squirrelo I agree with you, but if we do this, we end up with the other solution, in that everything is executed in pure SQL, doing everything that we need. This means that if something changes in the DB, we may end up needing to update +40 queries across the qiita_db code base.

josenavas:
And we have a system that consistency should be ensured between DB and FS, which means that once the DB is done, doesn't mean that the actual transaction si complete. Today @antgonza had an example: He added a new template to the system, the DB didn't had a problem inserting it, but there was a problem writing the file. This lead the system in an inconsistent state, because we assume that those files exist but they didn't. This end up breaking the interface...

squirrelo:
We can still do a hybrid of both. For things like inserts and whatnot, we can break it down by query. But for more complicated things like adding the default analysis for portals, we can take advantage of SQL for loops instead of doing the loop in python and a bunch of SQL inserts lined up. It would still be broken up calls, but more of a melding between SQL and python than the almost totally strict python we are doing right now.

As for the files, if we do the files work inside the transaction `with` statement, the errors on file stuff will cause the entire thing to roll back, files issues and SQL alike.

josenavas:
@squirrelo, exactly, but sometimes to generate those files you need the data in the DB. Transactions allow you to execute some queries, try to generate the files, and then commit them. There is no other way of doing this consistency that executing one (or many) SQL queries w/o committing and then commit after everything is in place...

squirrelo:
Is there a way of querying inside the transaction cursor, as opposed to the database itself? That way you are querying as if the items are already committed, but they actually aren't. I know sqlalchemy can do this.

Alternately, if you need the query information, you can make the query info list available through a function, e.g. `get_result(index)` so you can use the SQL results without having to commit.

josenavas:
@squirrelo What you're proposing is just executing the query and not commit it. You're saying this:
```python
with connect(params) as con:
    with con.get_cursor() as cur:
        try:
            cur.execute("DO STUFF")
        except:
            con.rollback()
     if checks_pass:
         con.commit()
     else:
         con.rollback()
```

This the exact same thing that I'm proposing with the `commit=False` parameter, but the transaction object allows you to do this like this:
```python
with Transaction('t') as t:
    t.add("DO STUFF")
    t.execute(commit=False)
    if not check_pass:
        t.rollback()
```

Don't let the terminology confuse you, you don't do anything "inside" a cursor, you do it directly to the DB, the difference is that you commit it or not.

squirrelo:
Actually what I'm proposing is more like:
```python
with Transaction('t') as t:
    t.add("DO STUFF")
    t.add("QUERY NEEDED STUFF")
    needed_info = t.get_results(t.index-1)
    # use needed info to make files
```

josenavas:
@squirrelo 

And how can I return the result if I don't execute the query w/o committing?
Your proposal, will be doing this internally:
```python
def get_results(index):
    res = self.execute(commit=False)
    return res[-1]
```

squirrelo:
I've found a thing that may fix all our problems: [Tornado sessions object](https://queries.readthedocs.org/en/latest/tornado_session.html) This locks a transaction thread until explicitly freed. If this works the way I think it does, we really should be using this as our backend instead of a straight connection like we do now.

josenavas:
You're proposing 2 things:
 1. Removing psycopg2 as a dependency
 2. Coupling qiita_db with tornado.

I don't agree with this, we have already seen that is a bad idea coupling qiita_pet and qiita_db...

josenavas:
Also, I'm not seeing anything about commit/rollback on that documentation... I don't even know how this would work...

squirrelo:
That's fair. ISOLATION_LEVEL_SERIALIZABLE might be the best in this case then, which if I'm reading correctly would make the cursor essentially stall the database while running the execution. This is the only way I can see not having the issue of running execute(commit=False) and not worrying about the database changing before the actual database execute with commit.

josenavas:
No, that would not solve the issue.

Let me resume the issue that we have in one sentence:
**We are doing commits that we shouldn't do**

That's the only problem that we're having, I don't know how else I can explain this.

squirrelo:
I get that, but my worry is something else, like another user, starts a transaction and adds something we needed between the commit=false and the actual commit. If this is not a worry and I'm just being paranoid, then OK, but just want to make sure.

josenavas:
The question is, how another user is going to add something that we need? If we need it, we shouldn't be relying in another user to add it.

Also, with commit = False, we are locking the tables, so no other user can modify those tables, which is the idea of SQL transaction.

squirrelo:
OK, cool. I'm paranoid then. Carry on.

ElDeveloper:
> We can make clear in the documentation that it will commit automatically, ...

Seems like a fine solution to me!

josenavas:
@ElDeveloper @squirrelo This is ready for another review.

qiita_db/sql_connection.py
```diff
@@ -633,3 +581,276 @@ def get_temp_queue(self):
...
+        try:
+            status = self._conn_handler._connection.get_transaction_status()
+            if exc_type is not None:
+                # An exception occurred during the execution of the transaction
+                # Make sure that we leave the DB w/o any modification
+                self.rollback()
+            elif self._queries:
+                # There are still queries to be executed, execute them
+                # It is safe to use the execute method here, as internally is
+                # wrapped in a tr/except and rollbacks in case of failure
```
ElDeveloper: tr -> try

qiita_db/sql_connection.py
```diff
@@ -633,3 +581,276 @@ def get_temp_queue(self):
...
+    Parameters
+    ----------
+    name : str
+        Name of the transaction.
+
+    Notes
+    -----
+    When the execution leaves the context manager, any remaining queries in
+    the transaction will be executed and committed.
+    The Transaction methods can only be executed inside a context
```
ElDeveloper: Maybe it would be appropriate to note here or in each method (up to you) that a RuntimeError is raised in this case.

qiita_db/sql_connection.py
```diff
@@ -633,3 +581,276 @@ def get_temp_queue(self):
...
+                # It is safe to use the execute method here, as internally is
+                # wrapped in a tr/except and rollbacks in case of failure
+                self.execute()
+            elif status != TRANSACTION_STATUS_IDLE:
+                # There are no queries to be executed, however, the transaction
+                # is still not committed. Commit it so the changes are not lost
+                self.commit()
+        finally:
+            self._is_inside_context = False
+
```
wasade: on exit, can the connection be explicitly closed? or is there a benefit in it possibly staying open? just concerned as the destructors with psycopg2 haven't always worked implicitly as expected

wasade: ...that, and the leaky nature of python really bites you here as references will be maintained:

```python
> with open('/tmp/foo.txt', 'w') as fp:
   ...:     pass
   ...:

08:51:11 (mcdonadt@Daniels-MBP-3):/Users/mcdonadt/Dropbox/manuscripts/2015/ag
> fp
Out[2]: <closed file '/tmp/foo.txt', mode 'w' at 0x10ecc8270>
```

...so unless there is a good reason to possibly leave the connection open, I think it should be explicitly closed.

ElDeveloper:
BTW, :+1:.

qiita_db/sql_connection.py
```diff
@@ -633,3 +581,276 @@ def get_temp_queue(self):
...
+        self._is_inside_context = False
+
+    def __enter__(self):
+        self._is_inside_context = True
+        return self
+
+    def __exit__(self, exc_type, exc_value, traceback):
+        # We need to wrap the entire function in a try/finally because
+        # at the end of the function we need to set _is_inside_context to false
+        try:
```
squirrelo: This try/finally can be removed by just moving `self._is_inside_context = False` tothe first line of this function. Regardless of what happens after, you are always going to be outside the context when exit is called.

qiita_db/sql_connection.py
josenavas: I will note it here and in each method, for completeness.

qiita_db/sql_connection.py
josenavas: Nope, I can't. If I move the `self._is_inside_context = False` to the beginning of the function, any call to the Transaction methods (`rollback`, `execute` and `commit`) done in this method will raise a `RuntimeError`

qiita_db/sql_connection.py

qiita_db/sql_connection.py
josenavas: @wasade Given the improvements to the Transaction object, and the fact that now we are going to pass the transaction object around (so we can ensure the same DB view) the Transaction object uses the same connection as the rest of the system, so there is only one connection in the entire system. Does this makes sense?

wasade: Ah, okay, sorry looked at it wrong. Thanks!!

josenavas:
Ready for another review!

qiita_db/sql_connection.py
josenavas: No problem!

PR #1277, opened 2015-06-19
Fix environment_manager.py

josenavas:
Built on top of https://github.com/biocore/qiita/pull/1276.

This fixes `environment_manager.py` to make use of the transaction object instead of the deprecated queue system.

There are some parts of this code that was not using queues and it could potentially use the transaction object. However, since any of the code in this module is intended to be executed while the system is running, I think is better to not change it to transactions because it will make the "verbose" option of the environment creation impossible: the prints will be executed when it is added to the queue, not when they're actually executed. I think this will make our debug ability harder in any of these steps, and it will completely remove the idea of the verbose option which is providing an update as steps are being successfully executed.

Tests are expected to fail


josenavas:
Closing as the latest changes in the transaction are note reflected here, and starting on this file is no longer the best approach.

PR #1278, opened 2015-06-19
Fix python patches

josenavas:
Built on top of https://github.com/biocore/qiita/pull/1277

This PR modifies the python patches that were using the queue system to use now the Transaction object.

Tests expected to fail.


josenavas:
Closing as the latest changes in the transaction are note reflected here, and starting on this file is no longer the best approach.

PR #1281, opened 2015-06-22
Metadata templates now use transactions

josenavas:
Built on top of https://github.com/biocore/qiita/pull/1278

This PR modifies the metadata template modules to make use of the transaction objects.
Check carefully in case I missed a case in which the transactions are needed and I did not use them.

During the development of this PR I found a couple of issues that I solved in here:
 - The `SampleTemplate.delete` function had a small bug. It checked if the study had any RawData attached, and if it did, it did not allow the deletion. With the latest changes on the DB structure, the correct thing to do is to check if there is a PrepTemplate attached to the study, not a RawData. I fixed this.
 - The `_get_categories` function in the BaseSample object accepted a `conn_handler` parameters that was not used. I removed it and fixed the calls.


josenavas:
Closing as the latest changes in the transaction are note reflected here.

PR #1282, opened 2015-06-22
Fixing type lookup

josenavas:
There was an error in the way the error was checked. This prevents us to actually see the source of the error. It will be great if we can have this merged fast, as it is blocking work on the live system.

### DIFF
--- a/qiita_db/metadata_template/util.py
+++ b/qiita_db/metadata_template/util.py
@@ -25,6 +25,20 @@ else:
     from string import letters, digits
 
 
+def type_lookup(dtype):
+    """Lookup function """
+    if dtype in [np.int8, np.int16, np.int32, np.int64]:
+        return 'integer'
+    elif dtype in [np.float16, np.float32, np.float64]:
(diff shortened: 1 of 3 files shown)


antgonza:
:+1:

qiita_db/metadata_template/util.py
```diff
@@ -25,6 +25,20 @@
     from string import letters, digits
 
 
+def type_lookup(dtype):
+    """Lookup function """
```
ElDeveloper: Can you make this numpydoc compliant?

ElDeveloper:
Looks good, 1 comment though.

qiita_db/metadata_template/util.py
josenavas: hahaha sure! I think I haven't finished the docstring >.< too involved on the other issue...

josenavas:
Thanks @ElDeveloper 
I addressed the comment!

ElDeveloper:
:+1:

PR #1285, opened 2015-06-23
Portals user obj

squirrelo:
This adds portal checks to the SQL in the user object. It also moves the creation of default analyses to after the user is verified, so we don't clutter up the DB with potentially unused, unverified users.

### DIFF
--- a/qiita_db/support_files/populate_test_db.sql
+++ b/qiita_db/support_files/populate_test_db.sql
@@ -383,7 +383,7 @@ INSERT INTO qiita.job_results_filepath (job_id, filepath_id) VALUES (1, 10), (2,
 
 -- Insert Analysis
 INSERT INTO qiita.analysis (email, name, description, analysis_status_id, pmid) VALUES ('test@foo.bar', 'SomeAnalysis', 'A test analysis', 1, '121112'), ('test@foo.bar', 'SomeSecondAnalysis', 'Another test analysis', 1, '22221112');
-
+INSERT INTO qiita.analysis_portal (analysis_id, portal_type_id) VALUES (1, 1), (2, 1);
 -- Insert Analysis Workflow
 INSERT INTO qiita.analysis_workflow (analysis_id, step) VALUES (1, 3), (2, 3);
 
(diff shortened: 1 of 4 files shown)


qiita_db/test/test_user.py
```diff
@@ -224,38 +219,85 @@ def test_set_info_bad_info(self):
...
+
+            qiita_config.portal = "EMP"
+            self.assertEqual(user.private_analyses, set())
+        finally:
+            qiita_config.portal = self.portal
 
     def test_get_shared_analyses(self):
-        self.assertEqual(self.user.shared_analyses, set([]))
+        user = User('shared@foo.bar')
+        try:
```
wasade: If the exception is deterministic, can it be caught explicitly? If it is stochastic, then can the test be designed to be deterministic? Same thing with the other try/except blocks

qiita_db/user.py
```diff
@@ -349,55 +356,60 @@ def info(self, info):
 
     @property
     def default_analysis(self):
-        sql = ("SELECT analysis_id FROM qiita.analysis WHERE email = %s AND "
-               "dflt = true")
-        conn_handler = SQLConnectionHandler()
-        return conn_handler.execute_fetchone(sql, [self._id])[0]
-
-    @property
-    def sandbox_studies(self):
```
josenavas: Why did you remove the `sandbox_studies` property?

squirrelo: It was used nowhere in the code, didn't even have a test in the user obj tests, and still references qiita.study_status table.

qiita_db/test/test_user.py
squirrelo: These try/finally are more for making absolutely sure that the portal stays what it should. That way if as you're testing the test fails, it will still reset back to the correct portal no matter what. Safety precaution really.

qiita_db/user.py
josenavas: Ok, thanks

qiita_db/test/test_user.py
josenavas: Given that there are multiple functions where this needs to be done, can you add it to the tearDown function? That is the whole idea of that tearDown function. Also, it doesn't hurt if it is executed always. and make the tests more readable.

squirrelo: Ok.

qiita_db/support_files/populate_test_db.sql
```diff
@@ -440,7 +440,8 @@ INSERT INTO qiita.collection_job (collection_id, job_id) VALUES (1, 1);
...
+INSERT INTO qiita.analysis (email, name, description, dflt, analysis_status_id) VALUES ('test@foo.bar', 'test@foo.bar-dflt-1', 'dflt', true, 1), ('admin@foo.bar', 'admin@foo.bar-dflt-1', 'dflt', true, 1), ('shared@foo.bar', 'shared@foo.bar-dflt-1', 'dflt', true, 1), ('demo@microbio.me', 'demo@microbio.me-dflt-1', 'dflt', true, 1), ('test@foo.bar', 'test@foo.bar-dflt-2', 'dflt', true, 1), ('admin@foo.bar', 'admin@foo.bar-dflt-2', 'dflt', true, 2), ('shared@foo.bar', 'shared@foo.bar-dflt-2', 'dflt', true, 2), ('demo@microbio.me', 'demo@microbio.me-dflt-2', 'dflt', true, 2);
+INSERT INTO qiita.analysis_portal (analysis_id, portal_type_id) VALUES (3, 1), (4, 1), (5, 1), (6, 1), (7, 2), (8, 2), (9, 2), (10, 2);
```
josenavas: If I understand correctly, what you're doing here is making the default analysis only visible on their portal, right? If so, I think this is correct.

squirrelo: yup

qiita_db/user.py
```diff
@@ -266,9 +263,9 @@ def verify_code(cls, email, code, code_type):
         elif code_type == 'reset':
             column = 'pass_reset_code'
         else:
-            raise IncompetentQiitaDeveloperError("code_type must be 'create'"
-                                                 " or 'reset' Uknown type "
-                                                 "%s" % code_type)
+            raise IncompetentQiitaDeveloperError(
+                "code_type must be 'create' or 'reset' Uknown type %s" %
```
josenavas: This is not for this PR, and we probably want to open an issue, but, if there are only 2 options `create` and  `reset`, can we represent this as a bool? If `True` means reset, if false means create...

qiita_db/user.py
```diff
@@ -280,15 +277,34 @@ def verify_code(cls, email, code, code_type):
...
+                     WHERE email = %s""".format(cls._table)
             conn_handler.execute(sql, (level, email))
+            # create user default sample holders once verified
+            # create one per portal
+            analysis_sql = """INSERT INTO qiita.analysis
+                (email, name, description, dflt, analysis_status_id)
+                VALUES (%s, %s, %s, %s, 1)"""
+            analysis_args = []
+            sql = "SELECT portal_type_id from qiita.portal_type"
+            for portal in conn_handler.execute_fetchall(sql):
```
josenavas: Can this be done in a list comprehension? They're faster than doing appends to the list.

qiita_db/user.py
```diff
@@ -280,15 +277,34 @@ def verify_code(cls, email, code, code_type):
...
+                analysis_args.append(
+                    (email, '%s-dflt-%d' % (email, portal[0]), 'dflt', True))
+            conn_handler.executemany(analysis_sql, analysis_args)
+
+        if db_code == code:
```
josenavas: This comparison is done up to 3 times in the code (previous if, this one and return statement). Can be computed once and stored in a variable?

josenavas:
Thanks @squirrelo 
I'm done with the review. You've some test failures but the code looks pretty good!

I really like how this is shaping up, with the modifications in the SQL queries. It is a bit tedious, but the end product is going to be awesome!

qiita_db/user.py
squirrelo: There could potentially be other codes, for example API keys, so it was left open ended.

qiita_db/user.py
squirrelo: done

qiita_db/user.py
squirrelo: done

squirrelo:

josenavas:
:+1:

squirrelo:
This is ready to be merged, and should probably be soon so @josenavas can start working on transactions in #1294

josenavas:
Don't worry @squirrelo, I will solve the merge conflicts later on, otherwise is going to be 2 branches really dirty and one dependent on the other.

qiita_db/user.py
```diff
@@ -280,16 +277,36 @@ def verify_code(cls, email, code, code_type):
             return False
 
         db_code = db_code[0]
+        if not db_code:
+            raise QiitaDBError("No %s code for user %s" %
```
ElDeveloper: Can you add this exception to the `Raises` section of the docstring? Also ... just noticed that `code_type` does not have a description, would you mind adding that as well?

ElDeveloper:
:+1: once the documentation comment is addressed.

PR #1286, opened 2015-06-24
add portal restrictions to search object

squirrelo:
Adds portal restrictions to searches.


josenavas:
:+1: I've to say that I'm really surprise that more changes were not needed in the search object!

squirrelo:
I know, thought it would be bad but apparently my two query system works to our advantage here.

ElDeveloper:
:+1:

PR #1287, opened 2015-06-24
Portals study object

squirrelo:
This finishes off the portal restrictions for the study object.

Now depends on #1288

### DIFF
--- a/qiita_db/study.py
+++ b/qiita_db/study.py
@@ -155,6 +155,18 @@ class Study(QiitaObject):
         if self.status != 'sandbox':
             raise QiitaDBStatusError("Illegal operation on non-sandbox study!")
 
+    def _check_portal_access(study_ids):
+        """Raises QiitaDBError if non-portal-accesable studies given"""
+        conn_handler = SQLConnectionHandler()
+        # Make sure portal has access to all studies asked for
+        sql = """SELECT COUNT(study_id) FROM qiita.study_portal
+                 JOIN qiita.portal_type USING portal_type_id
+                 WHERE study_id IN ({}) and portal = %s""".format(
+            ','.join(str(x) for x in study_ids))
+        accesable = conn_handler.execute_fetchall(sql, [qiita_config.portal])
+        if accesable != len(study_ids):
+            raise QiitaDBError('Non-portal-accessable studies asked for!')
+
     @property
     def status(self):
         r"""The status is inferred by the status of its processed data"""
@@ -226,11 +238,6 @@ class Study(QiitaObject):
         list of DictCursor
             Table-like structure of metadata, one study per row. Can be
             accessed as a list of dictionaries, keyed on column name.
-
-        Raises
-        ------
-        QiitaDBError
-            Portal-restricted studies are asked for
         """
         if info_cols is None:
             info_cols = cls._info_cols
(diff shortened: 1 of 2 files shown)


qiita_db/study.py
```diff
@@ -155,6 +155,20 @@ def _lock_non_sandbox(self, conn_handler):
         if self.status != 'sandbox':
             raise QiitaDBStatusError("Illegal operation on non-sandbox study!")
 
+    @staticmethod
+    def _check_portal_access(study_ids):
+        """Raises QiitaDBError if non-portal-accesable studies given"""
```
josenavas: accesable -> accessible

qiita_db/study.py
```diff
@@ -155,6 +155,20 @@ def _lock_non_sandbox(self, conn_handler):
...
+        conn_handler = SQLConnectionHandler()
+        # Make sure portal has access to all studies asked for
+        sql = """SELECT COUNT(study_id) FROM qiita.study_portal
+                 JOIN qiita.portal_type USING (portal_type_id)
+                 WHERE study_id IN ({}) and portal = %s""".format(
+            ','.join(str(x) for x in study_ids))
+        accesable = conn_handler.execute_fetchone(
```
josenavas: accesable -> accessible

qiita_db/study.py
```diff
@@ -234,11 +254,15 @@ def get_info(cls, study_ids=None, info_cols=None):
...
             pmid FROM qiita.study_pmid GROUP BY study_id) sp USING (study_id)
-            )""".format(search_cols)
+            JOIN qiita.study_portal USING (study_id)
+            JOIN qiita.portal_type USING (portal_type_id)
+            ) WHERE portal = '{1}'""".format(search_cols, qiita_config.portal)
         if study_ids is not None:
-            sql = "{0} WHERE study_id in ({1})".format(
-                sql, ','.join(str(s) for s in study_ids))
+            # Make sure portal has access to all studies asked for
+            cls._check_portal_access(study_ids)
```
josenavas: `_check_portal_access` is only used here. I think it is actually not needed and we can avoid a query in the database:
you can do this at the end, it is cheaper (saves a SQL query):
```python
res = conn_handler.execute_fetchall(sql)
if study_ids is not None and len(res) != len(study_ids):
    raise QiitaDBError('Non-portal-accessible studies asked for!')
```

josenavas:
Few comments

qiita_db/study.py
squirrelo: I actua;ly need to also put it in `has_access` just forgot in this. Will do in next round of push.

josenavas: Uhm... I'm not sure it is actually needed in `has_access`. `has_access` uses other functions that are already aware of the portal, so that doesn't require any change...

squirrelo: The admin `return True` will return true regardless of portal access.

josenavas: Also note that has_access is an instance method, which means that in order to instantiate it, the portal has been previously checked. i.e. a study shouldn't be instantiated if the portal doesn't have access to it.

squirrelo: That is currently not implemented, and is going to take a lot more work. The checks are in the base class, so thats going to take some re-tooling.

josenavas: In that case means that there is something wrong. Checks should be put in the __init__ method, not in each instance method. Allowing instantiation of a study at any portal gives you full access to all the information.

Please, limit the __init__ method and the classmethods. With this is enough to prohibit access to a study if the portal is not correct.

squirrelo: Yup, another pull request doing that coming shortly.

josenavas: Can we have the other PR first? it makes me nervous to have incomplete things.

squirrelo: Yeah, It will go in shortly like I said. Once merged, I'll update these other two (analysis and study)

josenavas: Thanks

squirrelo:

josenavas:
@squirrelo shouldn't the `delete` method have a check that the study that is trying to delete actually belongs to the portal?

squirrelo:
Good call.

squirrelo:
Actually, with the new base class, the line in delete:
```python
# checking that the id_ exists
        cls(id_)
```
will also check for portal access. So this check is already in place.

josenavas:
Thanks @squirrelo 

We really need to change the delete methods to be instance methods rather than classmethods... we are instantiating the object anyways inside...

:+1:

squirrelo:
Agree, but that is for another round of refactors :)

PR #1288, opened 2015-06-24
Add portals checking to base class

squirrelo:


qiita_db/base.py
```diff
@@ -151,6 +177,10 @@ def __init__(self, id_):
         if not self._check_id(id_):
             raise QiitaDBUnknownIDError(id_, self._table)
 
+        if not self._check_portal(id_):
+            raise QiitaDBError("Inaccessible in current portal: %s" %
```
josenavas: `"%s with id %d inaccessible in current portal: %s" % (self.__class__.__name__, id_, qiita_config.portal)`
It can be useful while looking at logs.

qiita_db/base.py
```diff
@@ -136,6 +139,29 @@ def _check_id(self, id_):
...
 
+    def _check_portal(self, id_):
+        """Checks that object is accessible in current portal
+
+        Parameters
+        ----------
+        id_ : object
+            The ID to test
+        """
+        self._check_subclass()
```
josenavas: Can you remove this call from here, and from `_check_id` and add it as a first call in the `__init__` function. Right now it is executed twice and since this is an instance method, if it is in the init it will be tested first.

josenavas:
A couple of comments.

qiita_db/base.py
squirrelo: Done

qiita_db/base.py
squirrelo: Done

squirrelo:

josenavas:
:+1:

qiita_db/base.py
```diff
@@ -128,14 +131,34 @@ def _check_id(self, id_):
...
 
+    def _check_portal(self, id_):
+        """Checks that object is accessible in current portal
+
+        Parameters
+        ----------
+        id_ : object
+            The ID to test
+        """
+        if self._portal_table is None:
```
antgonza: Is this possible? My concern is that _protal_table can be different types ...

antgonza:
One question.

qiita_db/base.py
squirrelo: If it's None that means it wasn't set by the actual subclass.

squirrelo: Same check that the _check_subclass function is doing actually. It is None by default in the base class, then changed to an actual value in the subclass.

antgonza: Then should it be an error here? Either way can you add a test?

squirrelo: No. It works exactly like the _check_id function right above this one. Also, there is a test already in this pull request. Please read the full pull request before commenting.

antgonza: Thanks for the explanation. However, I can't see the test for the None example, I can see the test for two different portals, can you explain?

squirrelo: Ah, that was not made clear. I can add that test.

antgonza: Sorry for not being clear, I though it was cause we were talking about that line. Next time I will try to be ...

squirrelo: Test added.

PR #1289, opened 2015-06-24
Portals analysis object

squirrelo:
Depends on #1288

### DIFF
--- a/qiita_db/analysis.py
+++ b/qiita_db/analysis.py
@@ -73,6 +73,7 @@ class Analysis(QiitaStatusObject):
     """
 
     _table = "analysis"
+    _portal_table = "analysis_portal"
     _analysis_id_column = 'analysis_id'
 
     def _lock_check(self, conn_handler):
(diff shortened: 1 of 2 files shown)


squirrelo:
This is ready for review

josenavas:
@squirrelo It looks like you're missing a check in `delete` and the `Collection` object at the end of the file.

squirrelo:
Delete calls exists, which does have the check in it. I can update the Collection object.

squirrelo:
Actually, the Collection object is not tied to a portal. Should it be indirectly tied to a portal through the analyses in it, or should we just leave it as is?

josenavas:
@squirrelo What I meant is that you forgot to check in the delete if the analysis belong to that portal. An analysis that belongs only to Qiita cannot be removed from the EMP portal, so there should be a check for that.

Re:collection object. All the analysis in a collection object should belong to the same portal. Not sure which is the best solution: connect the collection to the portals or ensure that all the analysis of a collection belong to the same portal. Probably the former, as it is more explicit.

The main problem is that AFAIK we are not using the collection anywhere, so it's kind of hard to understand which is the actual flow.

squirrelo:
What I mean is that `delete` does not need explicit check of portal, because if it is not in the portal `exists` will return false, and the analysis will not be deleted.

As for the collection object, I think we should just open an issue that it needs to be portaled and we can discuss it when we actually implement it in places and know more about its function. I'm leaning towards the indirect approach of making sure all analyses are in the same portal, since it is one less table to work with and we can check analysis portal. Analyses can only exist in at most 2 portals, so we should be OK in that regard.

josenavas:
Oh, I see now. Sorry. I though the other function you modified was `create` not `exists`... This is what I meant by submitting half of the solution... it complicates the review process cause I didn't remember that what was already done. Please, next time include a description of what is done, what is missing, and what was already previously done, so we can avoid this type of confusions...

Re: collection. Agree, can you open an issue?

:+1:

squirrelo:
Issue #1297 added for collections

PR #1291, opened 2015-06-25
Transaction context improvement

josenavas:
While I started working transforming the code base to use transactions, I noticed that it was really hard to maintain a nice code. With the improvement in this PR to the Transaction context management, the code highly improves. It allows to enter multiple times in the context and only do something if we are leaving the last context.
Another change is the removal of the parameter `commit=False`. The execute function **will never commit**. Since the committing is automatically performed when exiting the context, you never want to explicitly commit, since through the code you don't know if there is something else in the transaction that shouldn't be committed...

What the changes in here allows you to do is:
```python
def test(trans=None):
    trans = trans if trans is not None else Transaction('test')
    with trans:
       trans.add("do something")
       return trans.execute() # This does not commit

trans = Transaction('foo')
with trans:
   trans.add("do_staff")
   val = test(trans) # This call will not commit the changes inside `test`
   res = trans.execute() # This does not commit

val = test() # This call will commit the changes inside `test`
```

### DIFF
--- a/qiita_db/test/test_sql_connection.py
+++ b/qiita_db/test/test_sql_connection.py
@@ -556,7 +556,7 @@ class TestTransaction(TestBase):
                 args = [['insert1', 1], ['insert2', 2], ['insert3', 3]]
                 trans.add(sql, args, many=True)
 
-            # We exited the second context, anything should have been executed
+            # We exited the second context, nothing should have been executed
             self.assertEqual(trans._contexts_entered, 1)
             self.assertEqual(
                 trans._conn_handler._connection.get_transaction_status(),
--- a/qiita_db/test/test_sql_connection.py
+++ b/qiita_db/test/test_sql_connection.py
@@ -312,7 +312,9 @@ class TestTransaction(TestBase):
             trans.add(sql, [20, False, "test_insert"])
             obs = trans.execute()
             self.assertEqual(obs, [None, None])
-            self._assert_sql_equal([("test_insert", False, 20)])
+            self._assert_sql_equal([])
+
+        self._assert_sql_equal([("test_insert", False, 20)])
 
     def test_execute_many(self):
         with Transaction("test_execute_many") as trans:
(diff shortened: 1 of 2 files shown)


qiita_db/test/test_sql_connection.py
```diff
@@ -526,16 +532,47 @@ def test_context_manager_no_commit(self):
...
+
+            trans.add("SELECT 42")
+            with trans:
+                self.assertEqual(trans._contexts_entered, 2)
+                sql = """INSERT INTO qiita.test_table (str_column, int_column)
+                         VALUES (%s, %s) RETURNING str_column, int_column"""
+                args = [['insert1', 1], ['insert2', 2], ['insert3', 3]]
+                trans.add(sql, args, many=True)
+
+            # We exited the second context, anything should have been executed
```
squirrelo: anything -> nothing

qiita_db/test/test_sql_connection.py
```diff
@@ -526,16 +532,47 @@ def test_context_manager_no_commit(self):
...
+            self.assertEqual(trans._contexts_entered, 1)
+            self.assertEqual(
+                trans._conn_handler._connection.get_transaction_status(),
+                TRANSACTION_STATUS_IDLE)
             self._assert_sql_equal([])
 
+        # We have exited the first context, everything should have been
+        # executed and committed
+        self.assertEqual(trans._contexts_entered, 0)
```
squirrelo: These tests should go after the end of the final `with` as well, to make sure it actually executed and committed when exiting the final with.

squirrelo:
Couple comments

qiita_db/test/test_sql_connection.py
josenavas: @squirrelo these tests are actually at the end of the final `with` note that after the first with there is a test: `self._assert_sql_equal([])`. The reason why that test is empty data is because `self._assert_sql_equal` uses a different connection, so it doesn't see the changes done in the transaction; which is testing that the changes are actually not committed to the DB :-)

qiita_db/test/test_sql_connection.py
josenavas: Done

josenavas:
Ready for another round!

qiita_db/test/test_sql_connection.py
squirrelo: Cool.

squirrelo:
:+1: if tests pass.

PR #1298, opened 2015-06-26
Fix base object

josenavas:
Tests are expected to fail.

There are some case where we need to instantiate an object in a transaction, in order to do that, we need to modify the base object.

Also sometimes we check the status inside a transaction. This means that the `status` property is no longer a property and it becomes a method, so we can pass in the transaction in case that it is needed.

The changes in here affected the `User` object, so I also fix it since it was already checked on the list.

### DIFF
--- a/qiita_db/base.py
+++ b/qiita_db/base.py
@@ -27,7 +27,7 @@ Classes
 
 from __future__ import division
 from qiita_core.exceptions import IncompetentQiitaDeveloperError
-from .sql_connection import SQLConnectionHandler
+from .sql_connection import SQLConnectionHandler, Transaction
 from .exceptions import QiitaDBNotImplementedError, QiitaDBUnknownIDError
 
 
@@ -192,47 +192,60 @@ class QiitaStatusObject(QiitaObject):
     _status_setter_checks
     """
 
-    @property
-    def status(self):
-        r"""String with the current status of the analysis"""
-        # Check that self._table is actually defined
-        self._check_subclass()
+    def status(self, trans=None):
+        r"""String with the current status of the object
 
+        Parameters
+        ----------
+        trans: Transaction, optional
+            Transaction in which this method should be executed
+        """
         # Get the DB status of the object
-        conn_handler = SQLConnectionHandler()
-        return conn_handler.execute_fetchone(
-            "SELECT status FROM qiita.{0}_status WHERE {0}_status_id = "
-            "(SELECT {0}_status_id FROM qiita.{0} WHERE "
-            "{0}_id = %s)".format(self._table),
-            (self._id, ))[0]
+        trans = trans if trans is not None else Transaction(
+            "status_%s_%s" % (self.__class__.__name__, self._id))
+        with trans:
+            sql = """SELECT status FROM qiita.{0}_status
+                     WHERE {0}_status_id = (
+                        SELECT {0}_status_id FROM qiita.{0}
+                        WHERE {0}_id = %s)""".format(self._table)
+            trans.add(sql, [self._id])
+            return trans.execute()[-1][0][0]
 
-    def _status_setter_checks(self, conn_handler):
+    def _status_setter_checks(self, trans):
         r"""Perform any extra checks that needed to be done before setting the
         object status on the database. Should be overwritten by the subclasses
+
+        Parameters
+        ----------
+        trans: Transaction
+            Transaction in which this method should be executed
         """
         raise QiitaDBNotImplementedError()
 
-    @status.setter
-    def status(self, status):
-        r"""Change the status of the analysis
+    def set_status(self, status, trans=None):
+        r"""Change the status of the object
 
         Parameters
         ----------
         status: str
             The new object status
+        trans: Transaction, optional
+            Transaction in which this method should be executed
         """
-        # Check that self._table is actually defined
-        self._check_subclass()
+        trans = trans if trans is not None else Transaction(
+            "set_status_%s_%s" % (self.__class__.__name__, self._id))
 
-        # Perform any extra checks needed before we update the status in the DB
-        conn_handler = SQLConnectionHandler()
-        self._status_setter_checks(conn_handler)
-
-        # Update the status of the object
-        conn_handler.execute(
-            "UPDATE qiita.{0} SET {0}_status_id = "
-            "(SELECT {0}_status_id FROM qiita.{0}_status WHERE status = %s) "
-            "WHERE {0}_id = %s".format(self._table), (status, self._id))
+        with trans:
+            # Perform any extra checks needed before we update the
+            # status in the DB
+            self._status_setter_checks(trans)
+            # Update the status of the object
+            sql = """UPDATE qiita.{0} SET {0}_status_id = (
+                        SELECT {0}_status_id
+                        FROM qiita.{0}_status WHERE status = %s)
+                    WHERE {0}_id = %s""".format(self._table)
+            self.add(sql, [status, self._id])
+            self.execute()
 
     def check_status(self, status, exclude=False):
         r"""Checks status of object.
@@ -264,21 +277,20 @@ class QiitaStatusObject(QiitaObject):
         Table setup:
         foo: foo_status_id  ----> foo_status: foo_status_id, status
         """
-        # Check that self._table is actually defined
-        self._check_subclass()
+        trans = Transaction("check_status_%s_%s"
+                            % (self.__class__.__name__, self._id))
+        with trans:
+            sql = "SELECT DISTINCT status FROM qiita.{0}_status".format(
+                self._table)
+            trans.add(sql, [self._id])
+            statuses = [x[0] for x in trans.execute()[-1]]
 
-        # Get all available statuses
-        conn_handler = SQLConnectionHandler()
+            # Check that all the provided statuses are valid statuses
+            if set(status).difference(statuses):
+                raise ValueError("%s are not valid status values"
+                                 % set(status).difference(statuses))
 
-        statuses = [x[0] for x in conn_handler.execute_fetchall(
-            "SELECT DISTINCT status FROM qiita.{0}_status".format(self._table),
-            (self._id, ))]
+            # Get the DB status of the object
+            dbstatus = self.status(trans)
 
-        # Check that all the provided statuses are valid statuses
-        if set(status).difference(statuses):
-            raise ValueError("%s are not valid status values"
-                             % set(status).difference(statuses))
-
-        # Get the DB status of the object
-        dbstatus = self.status
         return dbstatus not in status if exclude else dbstatus in status
--- a/qiita_db/base.py
+++ b/qiita_db/base.py
@@ -27,7 +27,7 @@ Classes
 
 from __future__ import division
 from qiita_core.exceptions import IncompetentQiitaDeveloperError
-from .sql_connection import SQLConnectionHandler, Transaction
+from .sql_connection import Transaction
 from .exceptions import QiitaDBNotImplementedError, QiitaDBUnknownIDError
 
 
--- a/qiita_db/base.py
+++ b/qiita_db/base.py
@@ -293,7 +293,7 @@ class QiitaStatusObject(QiitaObject):
             # Check that all the provided statuses are valid statuses
             if set(status).difference(statuses):
                 raise ValueError("%s are not valid status values"
-                                 % set(status).difference(statuses))
+                                 % ', '.join(set(status).difference(statuses)))
 
             # Get the DB status of the object
             dbstatus = self.status(trans)
--- a/qiita_db/base.py
+++ b/qiita_db/base.py
@@ -27,7 +27,7 @@ Classes
 
 from __future__ import division
 from qiita_core.exceptions import IncompetentQiitaDeveloperError
-from .sql_connection import Transaction
+from .sql_connection import get_transaction
 from .exceptions import QiitaDBNotImplementedError, QiitaDBUnknownIDError
 
 
@@ -121,7 +121,7 @@ class QiitaObject(object):
         id_ : object
             The ID to test
         trans: Transaction
-            Transaction in which this method should be executed
+            Transaction in which this method is executed
 
         Notes
         -----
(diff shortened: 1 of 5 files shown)


qiita_db/base.py
```diff
@@ -128,28 +130,34 @@ def _check_id(self, id_):
...
+        return trans.execute()[-1][0][0]
 
-    def __init__(self, id_):
+    def __init__(self, id_, trans=None):
         r"""Initializes the object
 
         Parameters
         ----------
-        id_: the object identifier
+        id_: object
```
ElDeveloper: Isn't this a `str` type?

qiita_db/base.py
```diff
@@ -128,28 +130,34 @@ def _check_id(self, id_):
...
+            The object identifier
+        trans: Transaction
```
ElDeveloper: This is missing the 'optional' part.

qiita_db/base.py
```diff
@@ -184,47 +192,60 @@ class QiitaStatusObject(QiitaObject):
...
-        # Check that self._table is actually defined
-        self._check_subclass()
+    def status(self, trans=None):
+        r"""String with the current status of the object
 
+        Parameters
+        ----------
+        trans: Transaction, optional
+            Transaction in which this method should be executed
+        """
```
ElDeveloper: Can you describe what this method returns?

qiita_db/base.py
```diff
@@ -256,21 +277,20 @@ def check_status(self, status, exclude=False):
...
+                            % (self.__class__.__name__, self._id))
+        with trans:
+            sql = "SELECT DISTINCT status FROM qiita.{0}_status".format(
+                self._table)
+            trans.add(sql, [self._id])
+            statuses = [x[0] for x in trans.execute()[-1]]
+
+            # Check that all the provided statuses are valid statuses
+            if set(status).difference(statuses):
+                raise ValueError("%s are not valid status values"
```
ElDeveloper: I know you didn't really change this code, but can you add this exception to the raises section and format the set using `', '.join(...)`?

ElDeveloper:
:+1:, @squirrelo can you review this?

qiita_db/base.py
josenavas: Nope, each subclass define which type is it. Usually it is an integer, but there is a couple of cases where it actually is a string; so that's why I have to go back to the common class, which is object.

qiita_db/base.py
josenavas: Done! Forgot to add after transforming the property into a method.

qiita_db/base.py
josenavas: Done

qiita_db/base.py
josenavas: done

qiita_db/base.py
ElDeveloper: Got it, thanks!

qiita_db/base.py
```diff
@@ -128,28 +130,34 @@ def _check_id(self, id_):
         the other classes. However, still defining here as there is only one
         subclass that doesn't follow this convention and it can override this.
         """
-        self._check_subclass()
-
-        conn_handler = SQLConnectionHandler()
+        sql = "SELECT EXISTS(SELECT * FROM qiita.{0} WHERE {0}_id=%s)".format(
+            self._table)
+        trans.add(sql, [id_])
+        return trans.execute()[-1][0][0]
```
wasade: magic number voodoo

qiita_db/base.py
```diff
@@ -113,13 +113,15 @@ def _check_subclass(cls):
...
-    def _check_id(self, id_):
+    def _check_id(self, id_, trans):
         r"""Check that the provided ID actually exists on the database
 
         Parameters
         ----------
         id_ : object
             The ID to test
+        trans: Transaction
+            Transaction in which this method should be executed
```
wasade: The Transaction in which this method is executed

qiita_db/base.py
```diff
@@ -128,28 +130,34 @@ def _check_id(self, id_):
...
         Raises
         ------
         QiitaDBUnknownIDError
             If `id_` does not correspond to any object
         """
-        if not self._check_id(id_):
-            raise QiitaDBUnknownIDError(id_, self._table)
+        self._check_subclass()
+
+        trans = trans if trans is not None else Transaction(
```
wasade: decompose this

qiita_db/base.py
```diff
@@ -184,47 +192,65 @@ class QiitaStatusObject(QiitaObject):
...
         # Get the DB status of the object
-        conn_handler = SQLConnectionHandler()
-        return conn_handler.execute_fetchone(
-            "SELECT status FROM qiita.{0}_status WHERE {0}_status_id = "
-            "(SELECT {0}_status_id FROM qiita.{0} WHERE "
-            "{0}_id = %s)".format(self._table),
-            (self._id, ))[0]
-
-    def _status_setter_checks(self, conn_handler):
+        trans = trans if trans is not None else Transaction(
```
wasade: decompose, see above, will not comment on this further

qiita_db/base.py
```diff
@@ -184,47 +192,65 @@ class QiitaStatusObject(QiitaObject):
...
+            "status_%s_%s" % (self.__class__.__name__, self._id))
+        with trans:
+            sql = """SELECT status FROM qiita.{0}_status
+                     WHERE {0}_status_id = (
+                        SELECT {0}_status_id FROM qiita.{0}
+                        WHERE {0}_id = %s)""".format(self._table)
+            trans.add(sql, [self._id])
+            return trans.execute()[-1][0][0]
```
wasade: magic number voodoo, will not comment on this further

qiita_db/base.py
```diff
@@ -256,21 +282,20 @@ def check_status(self, status, exclude=False):
...
-
-        # Get all available statuses
-        conn_handler = SQLConnectionHandler()
+        trans = Transaction("check_status_%s_%s"
+                            % (self.__class__.__name__, self._id))
+        with trans:
+            sql = "SELECT DISTINCT status FROM qiita.{0}_status".format(
+                self._table)
+            trans.add(sql, [self._id])
+            statuses = [x[0] for x in trans.execute()[-1]]
```
wasade: statusi?

wasade:
Since tests are expected to fail, unsure how else to proceed except a static analysis

josenavas:
Thanks @wasade 

Static analysis is ok since this PR is the first one of a series of PR that is doing a lot of changes and going to its own branch. The reason this is done in this way is to avoid a 50 files change PR impossible to review.

wasade:
I understand the motivation, but I do not fully understand the utility of reviewing code that doesn't work

josenavas:
The reason is that if I make the entire sitebase to pass tests is going to be a 50 files change... we agreed that this is OK as long as it goes to its own branch.

The code in here works, but other parts in the code that uses this have not been adapted.

josenavas:
Thanks @wasade comments addressed

qiita_db/test/test_sql_connection.py
```diff
@@ -611,5 +612,13 @@ def test_index(self):
             trans.add(sql, args, many=True)
             self.assertEqual(trans.index, 7)
 
+    def test_get_transaction(self):
+        trans = Transaction("test_get_transaction")
+        obs = get_transaction(trans, "test")
+        self.assertEqual(obs, trans)
+
+        obs = get_transaction(None, "test")
+        self.assertEqual(obs._name, "test")
```
squirrelo: can you also add a check on `type` for obs, to make sure it's a Transaction.

qiita_db/base.py
```diff
@@ -128,28 +130,36 @@ def _check_id(self, id_):
...
-            "SELECT EXISTS(SELECT * FROM qiita.{0} WHERE "
-            "{0}_id=%s)".format(self._table), (id_, ))[0]
-
-    def __init__(self, id_):
+        sql = "SELECT EXISTS(SELECT * FROM qiita.{0} WHERE {0}_id=%s)".format(
+            self._table)
+        trans.add(sql, [id_])
+        # The value that we want is the result of the last SQL query,
+        # and it is stored in the first value of the first row
+        return trans.execute()[-1][0][0]
```
squirrelo: Magic numbers of doom.

qiita_db/base.py
```diff
@@ -184,49 +194,69 @@ class QiitaStatusObject(QiitaObject):
...
+            trans, "status_%s_%s" % (self.__class__.__name__, self._id))
+        with trans:
+            sql = """SELECT status FROM qiita.{0}_status
+                     WHERE {0}_status_id = (
+                        SELECT {0}_status_id FROM qiita.{0}
+                        WHERE {0}_id = %s)""".format(self._table)
+            trans.add(sql, [self._id])
+            # The value that we want is the result of the last SQL query,
+            # and it is stored in the first value of the first row
+            return trans.execute()[-1][0][0]
```
squirrelo: Magic numbers of doom. Last comment, but applies to all.

qiita_db/test/test_base.py
```diff
@@ -83,7 +85,7 @@ def setUp(self):
 
     def test_status(self):
         """Correctly returns the status of the object"""
-        self.assertEqual(self.tester.status, "in_construction")
+        self.assertEqual(self.tester.status(), "in_construction")
```
squirrelo: This also needs inside transaction checks for `_status_setter_checks`, `check_status`, and all other functions that are transactified.

qiita_db/user.py
```diff
@@ -66,26 +66,26 @@ class User(QiitaObject):
     # The following columns are considered not part of the user info
     _non_info = {"email", "user_level_id", "password"}
 
-    def _check_id(self, id_):
+    def _check_id(self, id_, trans):
```
squirrelo: This function needs a transaction test.

squirrelo:
Few comments

qiita_db/base.py
josenavas: Please read the comment. This is exactly what @wasade commented before and I added a comment.

I have to access a 3 nested list in some way, and those indexes are not provided by anybody, they just happen to be those:
 - -1 is the SQL query (last one)
 - 0 which row on that query I'm searching the result
 - 0 which value on that row is the result that I want

If you don't like this solution, I'm open to ideas. Applies to all.

qiita_db/test/test_sql_connection.py
josenavas: If it's not a transaction, `obs._name` will fail.

qiita_db/base.py
josenavas: BTW, not going to continue development until this is resolved. I've already spent 2 days on code and this is spread all over the new code. So until we reach a consensus I'm not going to waste more time.

This is way it is important to review PR as they come in. Specially in this type of issues.

qiita_db/test/test_sql_connection.py
squirrelo: Now, but another class may have that as a class attribute in the future. Better to be safe and check explicitly.

qiita_db/base.py
squirrelo: Comment is not clear on this. Please add some indication that the indices are needed and not magic numbers. That should be enough.

josenavas: Proposal?

squirrelo: Thinking, if `trans.execute()[-1][0][0]` really shows up as much as you say it does, it's probably a good idea to have something like `trans.execute(return_last=True)` that will automatically take care of the indexing. This will make it much cleaner throughout the codebase.

josenavas: That will change to this: `trans.execute(return_last=True)[0][0]`. Is that an improvement? Also, I have a use case in which I need the last X results, so I can do `for res in trans.execute()[x:][0][0]`, which will not be possible with your solution.

squirrelo: No. As stated, it would do full indexing of `[-1][0][0]` or `[-1][0]` if that makes it more widely applicable. Looking through the pull requests, this would cover a large fraction of the use cases. This does not affect other cases where `trans.execute()[-1]` is still applicable, but cleans up the longer, more convoluted indexing. Plus, two reviewers now have found the indexing to be an issue, so it is best to centralize it into the execute function with a detailed explanation.

josenavas: You can't centralize the indexing of [-1][0][0], because from the queue system you don't know if it returns a single value or multiple values. Unless with that parameter we document it really well and we say that it just does like [-1][0][0]... is that what you're proposing?

josenavas: However, it makes the return type of execute dependent of the parameters, which has been already criticized in other parts of the code, and I agree that it is not a good design...

squirrelo: Yes, that is what I'm proposing. `trans.execute(return_one=True)` or whatever the parameter is called replaces  `trans.execute()[-1][0][0]` for readability and understandability throughout the code.

josenavas: @wasade @antgonza @ElDeveloper what do you think about this solution?
`trans.execute()` will return a triple nested list while
`trans.execute(return_one=True)` will return a single value (the same as `trans.execute()[-1][0][0]`)

Since the return type of the function is different and this has been already criticized in another parts of the code, I would like to hear form others before doing this change.

ElDeveloper: I would opt for having a `execute_return_one`, not really a fan of different return types.

wasade: Agree with @ElDeveloper. The other option is an adapter:

```python
return Transaction.get_one(trans.execute())
```

... or something

wasade: `execute` is described as returning a `list` of `DictCursor`, so how does this indexing work out, or is the docstring inaccurate?

squirrelo: Dictcursor acts as both a list and a dictionary. You can index in if you know the index, or alternately use it as a dictionary if you know just the column name.

squirrelo: To be consistent with previous naming scheme, `execute_fetchlast` might be a good name for a second function.

josenavas: Thanks for all the comments, fetchlast is doable

PR #1302, opened 2015-06-28
Portals - fix conflicts with master

squirrelo:
Only issue ws patch 26 needing to be renamed to 27.sql


qiita_db/support_files/patches/26.sql
```diff
@@ -1,76 +1,16 @@
...
-INSERT INTO qiita.study_portal (study_id, portal_type_id) SELECT s.study_id, 1 FROM qiita.study s WHERE s.portal_type_id != 1;
-
-ALTER TABLE qiita.study DROP COLUMN portal_type_id
\ No newline at end of file
+-- Adding new illumina processing params if they do not exists
+-- adapted from: http://stackoverflow.com/a/13902402
+INSERT INTO qiita.preprocessed_sequence_illumina_params (param_set_name, barcode_type)
+  SELECT DISTINCT 'per sample FASTQ defaults', 'not-barcoded' FROM qiita.preprocessed_sequence_illumina_params
+  WHERE NOT EXISTS (SELECT 1 FROM qiita.preprocessed_sequence_illumina_params WHERE barcode_type = 'not-barcoded');
+>>>>>>> 1322f888a3cd1e91b90b7e4b62315b6a14e4b69f
```
josenavas: Merge conflicts have not been correctly fixed!!!

josenavas:
I though this will be easier to actually take a look U.U 
@squirrelo every time that something is merged in master, try to do a pull upstream master in the portals branch, so if another merge conflict occurs, we can see easier which are the differences...

I'm assuming tests are failing due to the problem on the patch.

Tests should pass, right?

squirrelo:
Tests should pass, yeah. Missed a marker, but fixed now.

squirrelo:
Also, I was just going to take care of all the conflicts in one go in the final pull request into master, but if it is wanted incrementally instead I can do that way too.

josenavas:
It can be pretty hard to identify if you introduce some bug at the end if you try to solve of the merge conflicts at once because that would force us to review the entire PR again, which will be almost impossible...

PR #1303, opened 2015-06-28
Portals scripts

squirrelo:
This adds a portal group to db, and the following commands to portal:
add-analyses
remove-analyses
add-studies
remove-studies

 They all take a portal (as text) and a comma separated list of study or analysis ids. For example, if you wanted to add studies 5 and 6 to portal EMP, you would do the following:
`qiita db portal add-studies --portal EMP --studies 5,6`

### DIFF
--- a/qiita_db/portal.py
+++ b/qiita_db/portal.py
@@ -32,11 +32,13 @@ if qiita_config.portal == "QIITA":
         return {x[0] for x in
                 conn_handler.execute_fetchall(sql, [portal_id])}
 
-    def add_study_to_portal(study, portal):
-        """Adds study to given portal
+    def add_studies_to_portal(portal, studies):
+        """Adds studies to given portal
 
         Parameters
         ----------
+        studies : list of int
+            Study ids to attach to portal
         portal : str
             Portal to associate with.
         """
@@ -44,13 +46,15 @@ if qiita_config.portal == "QIITA":
         sql = """INSERT INTO qiita.study_portal (study_id, portal_type_id)
                  VALUES (%s, %s)"""
         conn_handler = SQLConnectionHandler()
-        conn_handler.execute(sql, [study.id, portal_id])
+        conn_handler.executemany(sql, [(s, portal_id) for s in studies])
 
-    def remove_study_from_portal(study, portal):
-        """Adds study to given portal
+    def remove_studies_from_portal(portal, studies):
+        """Removes studies to given portal
 
         Parameters
         ----------
+        studies : list of int
+            Study ids to remove from portal
         portal : str
             Portal to associate with.
         """
@@ -60,4 +64,59 @@ if qiita_config.portal == "QIITA":
         sql = """DELETE FROM qiita.study_portal WHERE study_id = %s
                  AND portal_type_id = %s"""
         conn_handler = SQLConnectionHandler()
-        conn_handler.execute(sql, [study.id, portal_id])
+        conn_handler.executemany(sql, [(s, portal_id) for s in studies])
+
+    def get_analyses_by_portal(portal):
+        """Returns analysis id for all Analyses belonging to a portal
+
+        Parameters
+        ----------
+        portal : str
+           Portal to check analyses belong to
+
+        Returns
+        -------
+        set of int
+            All analysis ids in the database that match the given portal
+        """
+        conn_handler = SQLConnectionHandler()
+        portal_id = convert_to_id(portal, 'portal_type', 'portal')
+        sql = """SELECT analysis_id FROM qiita.analysis_portal
+                 WHERE portal_type_id = %s"""
+        return {x[0] for x in
+                conn_handler.execute_fetchall(sql, [portal_id])}
+
+    def add_analyses_to_portal(portal, analyses):
+        """Adds analyses to given portal
+
+        Parameters
+        ----------
+        analyses : list of int
+            Analysis ids to attach to portal
+        portal : str
+            Portal to associate with.
+        """
+        portal_id = convert_to_id(portal, 'portal_type', 'portal')
+        sql = """INSERT INTO qiita.analysis_portal
+                 (analysis_id, portal_type_id)
+                 VALUES (%s, %s)"""
+        conn_handler = SQLConnectionHandler()
+        conn_handler.executemany(sql, [(a, portal_id) for a in analyses])
+
+    def remove_analyses_from_portal(portal, analyses):
+        """Removes  analyses from given portal
+
+        Parameters
+        ----------
+        analyses : list of int
+            Analysis ids to remove from portal
+        portal : str
+            Portal to associate with.
+        """
+        if portal == "QIITA":
+            raise ValueError('Can not remove from main QIITA portal!')
+        portal_id = convert_to_id(portal, 'portal_type', 'portal')
+        sql = """DELETE FROM qiita.analysis_portal WHERE analysis_id = %s
+                 AND portal_type_id = %s"""
+        conn_handler = SQLConnectionHandler()
+        conn_handler.executemany(sql, [(a, portal_id) for a in analyses])
--- a/qiita_db/portal.py
+++ b/qiita_db/portal.py
@@ -9,11 +9,62 @@ import warnings
 
 from .sql_connection import SQLConnectionHandler
 from .util import convert_to_id
-from .exceptions import QiitaDBError
+from .exceptions import QiitaDBError, QiitaDBDuplicateError
 from qiita_core.qiita_settings import qiita_config
+from qiita_core.exceptions import IncompetentQiitaDeveloperError
 
 # Only make functions available for main portal
 if qiita_config.portal == "QIITA":
+    def create_portal(portal, desc):
+        """Creates a new portal on the system
+
+        Parameters
+        ----------
+        portal : str
+            The name of the portal to add
+        desc : str
+            Description of the portal
+
+        Raises
+        ------
+        QiitaDBDuplicateError
+            Portal already exists
+        """
+        try:
+            convert_to_id(portal, 'portal_type', 'portal')
+        except IncompetentQiitaDeveloperError:
+            # Portal does not exist, so add it
+            sql = """INSERT INTO qiita.portal_type (portal, portal_description)
+                     VALUES (%s, %s)"""
+            conn_handler = SQLConnectionHandler()
+            conn_handler.execute(sql, [portal, desc])
+        else:
+            # Portal already exists, so raise error
+            raise QiitaDBDuplicateError("Portal", portal)
+
+    def remove_portal(portal):
+        """Removes a portal from the system
+
+        Parameters
+        ----------
+        portal : str
+            The name of the portal to add
+        desc : str
+            Description of the portal
+
+        Raises
+        ------
+        QiitaDBError
+            Portal has analyses or studies attached to it
+        """
+        sql = "DELETE FROM qiita.portal_type WHERE portal = %s"
+        conn_handler = SQLConnectionHandler()
+        try:
+            conn_handler.execute(sql, [portal])
+        except ValueError:
+            # Portal attached to things, so can't delete
+            raise QiitaDBError("Portal %s in use, can not delete!" % portal)
+
     def get_studies_by_portal(portal):
         """Returns study id for all Studies belonging to a portal
 
--- a/qiita_db/portal.py
+++ b/qiita_db/portal.py
@@ -5,6 +5,7 @@
 #
 # The full license is in the file LICENSE, distributed with this software.
 # -----------------------------------------------------------------------------
+import warnings
 
 from .sql_connection import SQLConnectionHandler
 from .util import convert_to_id
@@ -43,10 +44,23 @@ if qiita_config.portal == "QIITA":
             Portal to associate with.
         """
         portal_id = convert_to_id(portal, 'portal_type', 'portal')
+        conn_handler = SQLConnectionHandler()
+        sql = """SELECT study_id from qiita.study_portal
+                 WHERE portal_type_id != %s AND study_id IN %s"""
+        clean_studies = [x[0] for x in conn_handler.execute_fetchall(
+                         sql, [portal_id, tuple(studies)])]
+
+        if len(clean_studies) != len(studies):
+            rem = map(str, set(studies).difference(clean_studies))
+            warnings.warn("The following studies will not be added: %s" %
+                          ', '.join(rem))
+
+        # Add cleaned list to the portal
+        portal_id = convert_to_id(portal, 'portal_type', 'portal')
+        conn_handler = SQLConnectionHandler()
         sql = """INSERT INTO qiita.study_portal (study_id, portal_type_id)
                  VALUES (%s, %s)"""
-        conn_handler = SQLConnectionHandler()
-        conn_handler.executemany(sql, [(s, portal_id) for s in studies])
+        conn_handler.executemany(sql, [(s, portal_id) for s in clean_studies])
 
     def remove_studies_from_portal(portal, studies):
         """Removes studies to given portal
@@ -57,14 +71,32 @@ if qiita_config.portal == "QIITA":
             Study ids to remove from portal
         portal : str
             Portal to associate with.
+
+        Raises
+        ------
+        ValueError
+            Try and delete from QIITA portal
         """
         if portal == "QIITA":
             raise ValueError('Can not remove from main QIITA portal!')
+
         portal_id = convert_to_id(portal, 'portal_type', 'portal')
-        sql = """DELETE FROM qiita.study_portal WHERE study_id = %s
-                 AND portal_type_id = %s"""
         conn_handler = SQLConnectionHandler()
-        conn_handler.executemany(sql, [(s, portal_id) for s in studies])
+        sql = """SELECT study_id from qiita.study_portal
+                 WHERE portal_type_id = %s AND study_id IN %s"""
+        clean_studies = [x[0] for x in conn_handler.execute_fetchall(
+                         sql, [portal_id, tuple(studies)])]
+
+        if len(clean_studies) != len(studies):
+            rem = map(str, set(studies).difference(clean_studies))
+            warnings.warn("The following studies will not be removed: %s" %
+                          ', '.join(rem))
+
+        sql = """DELETE FROM qiita.study_portal
+                 WHERE study_id IN %s AND portal_type_id = %s"""
+        if len(clean_studies) != 0:
+            conn_handler = SQLConnectionHandler()
+            conn_handler.execute(sql, [tuple(studies), portal_id])
 
     def get_analyses_by_portal(portal):
         """Returns analysis id for all Analyses belonging to a portal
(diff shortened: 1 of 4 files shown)


squirrelo:
Ready for review

qiita_db/portal.py
```diff
@@ -60,4 +64,59 @@ def remove_study_from_portal(study, portal):
...
+
+        Parameters
+        ----------
+        analyses : list of int
+            Analysis ids to attach to portal
+        portal : str
+            Portal to associate with.
+        """
+        portal_id = convert_to_id(portal, 'portal_type', 'portal')
+        sql = """INSERT INTO qiita.analysis_portal
```
josenavas: Should this function first check that the analyses actually exist? Note that the error is not going to be that helpful from the DB directly, so we probably want to check that the analysis exists, then that they've not been already linked with the portal (it will otherwise also raise another useless error) and then add the analyses to the portal.

We should probably do the same for adding the studies to the portal.

qiita_db/portal.py
```diff
@@ -60,4 +64,59 @@ def remove_study_from_portal(study, portal):
...
+        ----------
+        analyses : list of int
+            Analysis ids to remove from portal
+        portal : str
+            Portal to associate with.
+        """
+        if portal == "QIITA":
+            raise ValueError('Can not remove from main QIITA portal!')
+        portal_id = convert_to_id(portal, 'portal_type', 'portal')
+        sql = """DELETE FROM qiita.analysis_portal WHERE analysis_id = %s
```
josenavas: I know that SQL will not fail, but shouldn't we check here that the analyses is actually linked to the portal? (Same for studies)

Another think to check is that it is not trying to remove one of the default analyses, as those are needed for the correct behavior of the interface.

josenavas:
A couple of comments.

qiita_db/portal.py
squirrelo: Do we care if we try and unlink an analysis that is not part of the portal though? The end result is exactly the same: The analysis is not part of the portal at the end of the function. 

I can add a check for default analysis in there.

josenavas: I have mixed feelings about it... but I'm thinking in be safe later... Imagine that for some reason we have to remove an analysis (or a study) from one portal to whatever reason (e.g.a study is actually not a EMP). We do a typo, we put the wrong study and we don't know that we did that... at least in this way is another check...

squirrelo:
Comments addressed, ready to go

qiita_db/portal.py
```diff
@@ -32,32 +33,156 @@ def get_studies_by_portal(portal):
...
         Parameters
         ----------
+        studies : list of int
+            Study ids to attach to portal
         portal : str
             Portal to associate with.
         """
         portal_id = convert_to_id(portal, 'portal_type', 'portal')
+        conn_handler = SQLConnectionHandler()
+        sql = """SELECT study_id from qiita.study_portal
```
josenavas: I don't fully understand this query. If I'm reading correctly, it is checking for the studies that the portal id is different from the current portal that you want to add and the study id is passed in the current list. I think the thought process here is pretty complex... Shouldn't you just check if adding duplicates? For checking duplicates its easier to do:
```python
sql = "SELECT study_id FROM qiita.study_portal WHERE portal_type_id=%s AND study_id IN %s"
duplicates = [x[0] for x in conn_handler.execute_fetchall(sql, [portal_id, tuple(studies)])]

if duplicates:
   warnings.warn("...")
```

And I'd say that we want to do an error here rather than a warning. These operation will be executed by sysadmins so it's better to not allow to execute something if there is an error.

What is your rationale to do a warning rather than an error?

@biocore/qiita-dev can somebody else weight here if this should be an error or a warning?

josenavas:
It would be good to have a more useful error/warning message. In the case of the analyses, for example, it would be good to differentiate why an analysis cannot be removed (because it is a default, or because it is not associated with such portal).

josenavas:
Same situation for the study. It is possible that the study ID does not exist, that the study is already associated with the portal, etc... differentiating between all these case is pretty useful at troubleshooting time.

qiita_db/portal.py
squirrelo: The query is doing exactly the same thing as your query above, just giving a "clean" list of studies instead of a "dirty" list of duplicates. Hence why a warning is OK in my mind, since we are only operating on the cleaned list of studies.

squirrelo:
If  I am going to split this up by type of issue with the study, I can raise an error for studies/analyses that don't exist, but a warning should suffice for studies that already belong to/don't belong to the portal. As stated earlier, for those add/remove functions it's the same thing at the end of the function: The studies act as if they are added/removed from the portal.

josenavas:
Probably good as a warning if they're already in the portal (or not) but an error for those studies/analyses that do not exist, and for the the default analyses I would suggest also an error, rather than a warning.

qiita_db/portal.py
josenavas: Ok

squirrelo:
Comments addressed, errors now raised as appropriate. Ready for review.

qiita_db/portal.py
```diff
@@ -32,32 +34,198 @@ def get_studies_by_portal(portal):
...
+            bad = map(str, (set(studies).difference(existing)))
+            raise QiitaDBError("The following studies do not exist: %s" %
+                               ", ".join(bad))
+
+    def add_studies_to_portal(portal, studies):
+        """Adds studies to given portal
 
         Parameters
         ----------
+        studies : list of int
```
josenavas: The parameters list in here should be in the same order as in the function signature.

Also missing the raises section, in case that some of the provided studies do not exist.

qiita_db/portal.py
```diff
@@ -32,32 +34,198 @@ def get_studies_by_portal(portal):
...
+                         sql, [portal_id, tuple(studies)])]
+
+        if len(clean_studies) != len(studies):
+            rem = map(str, set(studies).difference(clean_studies))
+            warnings.warn("The following studies area already part of %s: %s" %
+                          (portal, ', '.join(rem)))
+
+        # Add cleaned list to the portal
         portal_id = convert_to_id(portal, 'portal_type', 'portal')
+        conn_handler = SQLConnectionHandler()
```
josenavas: You already have a conn_handler (line 61).

qiita_db/portal.py
```diff
@@ -32,32 +34,198 @@ def get_studies_by_portal(portal):
...
+        conn_handler.executemany(sql, [(s, portal_id) for s in clean_studies])
 
-    def remove_study_from_portal(study, portal):
-        """Adds study to given portal
+    def remove_studies_from_portal(portal, studies):
+        """Removes studies to given portal
 
         Parameters
         ----------
+        studies : list of int
```
josenavas: Same as above re parameters order

qiita_db/portal.py
```diff
@@ -32,32 +34,198 @@ def get_studies_by_portal(portal):
...
+            Study ids to remove from portal
         portal : str
             Portal to associate with.
+
+        Raises
+        ------
+        ValueError
```
josenavas: Missing QiitaDBError if the study doesn't exist.

qiita_db/portal.py
```diff
@@ -32,32 +34,198 @@ def get_studies_by_portal(portal):
...
+                               ", ".join(bad))
+
+        # Check if any analyses given are default
+        sql = ("SELECT analysis_id from qiita.analysis WHERE analysis_id IN %s"
+               " AND dflt = True")
+        default = [x[0] for x in conn_handler.execute_fetchall(
+            sql, [tuple(analyses)])]
+        if len(default) > 0:
+            bad = map(str, set(analyses).difference(default))
+            raise QiitaDBError("The following analyses are default: %s" %
```
josenavas: `The following analyses are default and can't be removed: %s`

qiita_db/portal.py
```diff
@@ -32,32 +34,198 @@ def get_studies_by_portal(portal):
...
+                               ", ".join(bad))
+
+    def add_analyses_to_portal(portal, analyses):
+        """Adds analyses to given portal
+
+        Parameters
+        ----------
+        analyses : list of int
```
josenavas: Parameter order.

qiita_db/portal.py
```diff
@@ -32,32 +34,198 @@ def get_studies_by_portal(portal):
...
+            warnings.warn("The following analyses are already part of %s: %s" %
+                          (portal, ', '.join(rem)))
+
+        sql = """INSERT INTO qiita.analysis_portal
+                 (analysis_id, portal_type_id)
+                 VALUES (%s, %s)"""
         portal_id = convert_to_id(portal, 'portal_type', 'portal')
-        sql = """DELETE FROM qiita.study_portal WHERE study_id = %s
-                 AND portal_type_id = %s"""
         conn_handler = SQLConnectionHandler()
```
josenavas: You already have a conn_handler above, remove.

qiita_db/portal.py
```diff
@@ -32,32 +34,198 @@ def get_studies_by_portal(portal):
...
+    def add_analyses_to_portal(portal, analyses):
+        """Adds analyses to given portal
+
+        Parameters
+        ----------
+        analyses : list of int
+            Analysis ids to attach to portal
+        portal : str
+            Portal to associate with.
+        """
```
josenavas: Missing raises section with the QiitaDBError

qiita_db/portal.py
```diff
@@ -32,32 +34,198 @@ def get_studies_by_portal(portal):
...
         conn_handler = SQLConnectionHandler()
-        conn_handler.execute(sql, [study.id, portal_id])
+        conn_handler.executemany(sql, [(a, portal_id) for a in clean_analyses])
+
+    def remove_analyses_from_portal(portal, analyses):
+        """Removes  analyses from given portal
+
+        Parameters
+        ----------
+        analyses : list of int
```
josenavas: Same here. Parameters order and raises section with QiitaDBError

qiita_db/portal.py
```diff
@@ -32,32 +34,198 @@ def get_studies_by_portal(portal):
...
+
+        if len(clean_analyses) != len(analyses):
+            rem = map(str, set(analyses).difference(clean_analyses))
+            warnings.warn("The following analyses are not part of %s: %s" %
+                          (portal, ', '.join(rem)))
+
+        sql = """DELETE FROM qiita.analysis_portal
+                 WHERE analysis_id IN %s AND portal_type_id = %s"""
+        if len(clean_analyses) != 0:
+            conn_handler = SQLConnectionHandler()
```
josenavas: You already have a connection handler (line 212)

qiita_db/portal.py
josenavas: Actually, this is better:
`The following analyses are default and can't be deleted or assigned to another portal: %s`

josenavas:
Few more comments @squirrelo

squirrelo:
Comments addressed, ready for another review.

qiita_db/portal.py
```diff
@@ -32,32 +34,209 @@ def get_studies_by_portal(portal):
...
         sql = """INSERT INTO qiita.study_portal (study_id, portal_type_id)
                  VALUES (%s, %s)"""
-        conn_handler = SQLConnectionHandler()
-        conn_handler.execute(sql, [study.id, portal_id])
+        conn_handler.executemany(sql, [(s, portal_id) for s in clean_studies])
 
-    def remove_study_from_portal(study, portal):
-        """Adds study to given portal
+    def remove_studies_from_portal(portal, studies):
+        """Removes studies to given portal
```
josenavas: `to -> from`

qiita_db/portal.py
```diff
@@ -32,32 +34,209 @@ def get_studies_by_portal(portal):
...
+                          (portal, ', '.join(rem)))
+
+        sql = """INSERT INTO qiita.analysis_portal
+                 (analysis_id, portal_type_id)
+                 VALUES (%s, %s)"""
+        portal_id = convert_to_id(portal, 'portal_type', 'portal')
+        conn_handler.executemany(sql, [(a, portal_id) for a in clean_analyses])
+
+    def remove_analyses_from_portal(portal, analyses):
+        """Removes  analyses from given portal
```
josenavas: There is an extra space between Removes and analyses

josenavas:
2 small comments on documentation.

:+1: once they're solved!

squirrelo:
Comments addressed, ready for another reviewer.

qiita_db/portal.py
```diff
@@ -5,13 +5,66 @@
...
 
 from .sql_connection import SQLConnectionHandler
 from .util import convert_to_id
+from .exceptions import QiitaDBError, QiitaDBDuplicateError
 from qiita_core.qiita_settings import qiita_config
+from qiita_core.exceptions import IncompetentQiitaDeveloperError
 
 # Only make functions available for main portal
 if qiita_config.portal == "QIITA":
+    def create_portal(portal, desc):
```
josenavas: This totally looks now like it should be an object... It has create, delete, properties...
At the end you implemented everything that an object requires. I don't understand why you didn't want at the beginning to implement the object... This type of (bad) decisions is what makes Qiita use up a lot of developer time. I can see in the future doing a refactoring of this code to make it an object as this is what it fits, some time that we could avoid if this was an object since the very beginning...

qiita_db/portal.py
```diff
@@ -5,13 +5,66 @@
...
+
+        Raises
+        ------
+        QiitaDBError
+            Portal has analyses or studies attached to it
+        """
+        sql = "DELETE FROM qiita.portal_type WHERE portal = %s"
+        conn_handler = SQLConnectionHandler()
+        try:
+            conn_handler.execute(sql, [portal])
```
josenavas: We never did this across the code base: i.e. leveraging on the DB to check if something can be deleted or not. We directly check it. Note that if there is any other error here, you'll mask it and say that the portal is in use. Please specifically check for the things that you want. It will be also easy to point out what is in use, either studies or analyses and which ones.

scripts/qiita
```diff
@@ -30,6 +30,10 @@ from qiita_db.commands import (load_sample_template_from_cmd,
                                load_prep_template_from_cmd,
                                load_parameters_from_cmd, SUPPORTED_PARAMS,
                                update_preprocessed_data_from_cmd)
+from qiita_db.portal import (add_studies_to_portal,
```
josenavas: If the configuration doesn't specify the QIITA portal, this import will fail. Also, the commands shouldn't be defined if the portal is not QIITA.

scripts/qiita-env
```diff
@@ -155,5 +159,19 @@ def job_details(job):
             click.echo("%s : %s" % (k, value))
 
 
+@env.command(name="create-portal")
+@click.option('--portal', required=True, type=str, help='The portal name')
+@click.option('--desc', required=True, type=str, help='The portal description')
+def add_portal(portal, desc):
+    """Creates a new portal on the database"""
+    create_portal(portal, desc)
```
josenavas: This will fail if portal in the config file is not QIITA

qiita_db/portal.py
squirrelo: I don't think it needs to be an object. It doesn't make sense in my mind to do something like Portal('EMP').add_studies([...]) since this is primarily used really in the command line functions. Instantiating an object each time doesn't seem necessary.

josenavas:
Besides the object comment (which we can leave it as it is right now in the interest of time) the other comments need to be fixed.

qiita_db/portal.py
wasade: I agree with both of you. The encapsulation of these methods under a namespace makes sense, however there doesn't appear to be much common state so they'd likely end up being `staticmethod`s. If the object is entirely composed of `staticmethod`s, then it probably shouldn't be an object. 

The bigger issue I see is that these methods are optionally present in the namespace. The current approach will require that any caller of, say, `create_portal` will first have to either interrogate the `globals` to see if the method exists or check against `qiita_config.portal`. Even then, this is fragile as imports are not assured to work:

```python
try:
    from qiita_db.portal import create_portal
except ImportError:
    create_portal = lambda *args: None
``` 

Why does `qiita_db` care what portal is specified in the config when doing things like `create_portal`? Are the portals nested? If so, then I apologize for missing that design goal. If they aren't, then really I think the consideration should be whether an administrator is performing the action.

josenavas: The main reason behind doing conditional imports is that Qiita is tailored to be used both through the GUI and through the command-line. Right now we don't have users at the command-line level, which is something that we will need to fix (I think), but the idea is that these functions should be defined if and only if the "portal" that is accessing the DB is qiita.

wasade: Why?

josenavas: Re staticmethods. Only create and delete should be static methods.

Note that a similar argument can be done for the rest of objects in the system, which means that we probably shouldn't have created objects and just use specialized functions to access/modify the DB...

josenavas: Because the "qiita" portal is also the "portal manager", but none of the other portals are. In theory, the portals manager should be its own portal, but we make it to be qiita for convenience.

squirrelo: Thinking more, this definitely should not be an object. Unlike the other objects, this is admin-only, potentially dangerous functionality. Therefore, if you instantiate something like `Portal('EMP')` you would get access to all the functionality, which shouldn't be done for something admin-only. To me, it's safer to explicitly import the functionality needed (as in the qiita pet pull request, where only the add and remove studies functions are imported).

qiita_db/portal.py
squirrelo: I would argue we aren't leveraging the full functionality of the database if we are doing explicit checks that can instead be handled by foreign keys. This removes two, and potentially more, SQL calls along with the needed python logic. Also, the ValueError raised is only for foreign key concerns or when the portal text is formatted incorrectly, both of which are concerns here. Any other error is raised as a psycopg2 Error or no-ValueError, which will not be caught in the try except and raised as usual.

josenavas: I thought everything raised from the SQLConnectionHandler was ValueError. Probably it is only on the transactions.

qiita_db/portal.py
wasade: The other methods in here don't appear to have common state, so there is no instance state to manipulate, which would make them `staticmethod`s. A `property` can still be considered a `staticmethod` if it doesn't modify or access instance state. These could be coerced into instance methods if `portal` was stored but it doesn't really buy you much except for the namespace.

These methods could probably be coerced into a subclass of a `QiitaObject`, which I support simply for consistency. 

I disagree that the argument extends to the rest of the objects in the system as there are notions of common state, (sane) operator overloading and the encapsulation is pleasant. Keep in mind though, it was not necessary to use objects, and without seeing alternative implementations, it's difficult to evaluate if that was the best decision (I am in no way suggesting we revisit this). 

I still don't understand why these functions should be defined IFF the configuration file says 'qiita'. It sounds like the "portal manager" is really just a means of access control. If that is the case, then this approach is flawed.

wasade: @squirrelo, if the concern is access, then we _really_ need to do this correctly. Limiting what functions are in the namespace is not a reliable security measure given that a user could just change their qiita config.

qiita_db/portal.py
squirrelo: Ah, so it is, but we should really rethink that. You can get some nicely fine-grained stuff for errors in psycopg2 Details [here](http://initd.org/psycopg/docs/module.html#psycopg2.InterfaceError) and [here](http://initd.org/psycopg/docs/errorcodes.html). This will also allow us to leverage the errors for foreign key checks, example [here](http://stackoverflow.com/a/12495497).

qiita_db/portal.py
josenavas: @wasade I don't understand the fact that the other objects have notions of common state... I don't see anything in the objects that will point out that notion.

Re alternative implementations. I also disagree on re-visiting this, but we the changes introduced by the transaction object this "alternative" implementation will play a key role, as a lot of the logics that we are doing in qiita_pet is going to move to qiita_db in order to make sure that everything is executed in a single transaction.

josenavas: However, I think this is another discussion and we should move it outside of this PR if we want to continue talking about this.

wasade: You lost me with `Transaction` and qiita_pet.

Regarding state, this is a very coarse check, but there are quite a few references to instance state:

```bash
$ grep -c self\. *.py
__init__.py:0
analysis.py:147
base.py:31
commands.py:0
data.py:129
environment_manager.py:0
exceptions.py:12
investigation.py:6
job.py:53
logger.py:19
meta_util.py:0
ontology.py:14
parameters.py:23
reference.py:10
search.py:38
sql_connection.py:62
study.py:98
user.py:44
util.py:0
```

Bringing this back on track, @squirrelo, my vote here is to make this a subclass of `QiitaObject` for consistency with the rest of the database interaction. I also vote against conditional imports. 

I'm not going to be responding to this thread further unless there are specific questions.

squirrelo: I guess I can make this a `QiitaObject` then with portal as the ID.

josenavas: :+1:

squirrelo:
in light of the massive changes needed for this, I'm closing this pull request and opening a new one for the changes.

PR #1307, opened 2015-06-29
Fix the Ontology object

josenavas:
Built on top of https://github.com/biocore/qiita/pull/1306

This PR fixes the Ontology object to use Transactions

Tests expected to fail.

### DIFF
--- a/qiita_db/base.py
+++ b/qiita_db/base.py
@@ -27,7 +27,7 @@ Classes
 
 from __future__ import division
 from qiita_core.exceptions import IncompetentQiitaDeveloperError
-from .sql_connection import SQLConnectionHandler
+from .sql_connection import Transaction
 from .exceptions import QiitaDBNotImplementedError, QiitaDBUnknownIDError
 
 
@@ -113,13 +113,15 @@ class QiitaObject(object):
             raise IncompetentQiitaDeveloperError(
                 "Could not instantiate an object of the base class")
 
-    def _check_id(self, id_):
+    def _check_id(self, id_, trans):
         r"""Check that the provided ID actually exists on the database
 
         Parameters
         ----------
         id_ : object
             The ID to test
+        trans: Transaction
+            Transaction in which this method should be executed
 
         Notes
         -----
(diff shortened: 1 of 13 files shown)


qiita_db/base.py
```diff
@@ -113,13 +113,15 @@ def _check_subclass(cls):
...
-    def _check_id(self, id_):
+    def _check_id(self, id_, trans):
         r"""Check that the provided ID actually exists on the database
 
         Parameters
         ----------
         id_ : object
             The ID to test
+        trans: Transaction
+            Transaction in which this method should be executed
```
wasade: `The Transaction this method will use`

qiita_db/base.py
```diff
@@ -128,28 +130,34 @@ def _check_id(self, id_):
...
-
-        conn_handler = SQLConnectionHandler()
-
-        return conn_handler.execute_fetchone(
-            "SELECT EXISTS(SELECT * FROM qiita.{0} WHERE "
-            "{0}_id=%s)".format(self._table), (id_, ))[0]
+        sql = "SELECT EXISTS(SELECT * FROM qiita.{0} WHERE {0}_id=%s)".format(
+            self._table)
+        trans.add(sql, [id_])
+        return trans.execute()[-1][0][0]
```
wasade: magic number voodoo

josenavas:
@wasade note that this PR depend on others, and it is noted in the PR comment. Please review those other ones first, so the amount of code per PR is smaller.

qiita_db/base.py
```diff
@@ -128,28 +130,34 @@ def _check_id(self, id_):
...
         Raises
         ------
         QiitaDBUnknownIDError
             If `id_` does not correspond to any object
         """
-        if not self._check_id(id_):
-            raise QiitaDBUnknownIDError(id_, self._table)
+        self._check_subclass()
+
+        trans = trans if trans is not None else Transaction(
```
wasade: this should be a method. what about the following signature:

```python
trans = Transaction.create_if_none(trans, self, id_)
```

The code could look something like:

```python
@classmethod
def create_if_none(cls, trans, obj, id_):
    if trans is None:
        return trans
    else:
        return cls("init_%s_%s" % (obj.__class__.__name__, id_))
```

qiita_db/base.py
```diff
@@ -184,47 +192,60 @@ class QiitaStatusObject(QiitaObject):
...
         # Get the DB status of the object
-        conn_handler = SQLConnectionHandler()
-        return conn_handler.execute_fetchone(
-            "SELECT status FROM qiita.{0}_status WHERE {0}_status_id = "
-            "(SELECT {0}_status_id FROM qiita.{0} WHERE "
-            "{0}_id = %s)".format(self._table),
-            (self._id, ))[0]
-
-    def _status_setter_checks(self, conn_handler):
+        trans = trans if trans is not None else Transaction(
```
wasade: see above, won't comment again on this pattern

wasade: ...though the comment above can be generalized slightly further to allow passing in a name or something

qiita_db/base.py
```diff
@@ -184,47 +192,60 @@ class QiitaStatusObject(QiitaObject):
...
+            "status_%s_%s" % (self.__class__.__name__, self._id))
+        with trans:
+            sql = """SELECT status FROM qiita.{0}_status
+                     WHERE {0}_status_id = (
+                        SELECT {0}_status_id FROM qiita.{0}
+                        WHERE {0}_id = %s)""".format(self._table)
+            trans.add(sql, [self._id])
+            return trans.execute()[-1][0][0]
```
wasade: magic number voodoo

wasade:
The comments are valid nonetheless

PR #1310, opened 2015-06-29
Portals pet - Jose & Yoshiki

squirrelo:
~~Depends on #1316~~

This adds the qiita-pet interface for adding studies to another portal. Some features include:
* Only shows up for admin on the `QIITA` portal.
* Sortable by checkbox, so you can see what studies you selected.
* The template html is extendable for any type of information, so the same template can theoretically be used for adding studies, analyses, collections, etc.

### DIFF
--- a/qiita_db/portal.py
+++ b/qiita_db/portal.py
@@ -32,11 +32,13 @@ if qiita_config.portal == "QIITA":
         return {x[0] for x in
                 conn_handler.execute_fetchall(sql, [portal_id])}
 
-    def add_study_to_portal(study, portal):
-        """Adds study to given portal
+    def add_studies_to_portal(portal, studies):
+        """Adds studies to given portal
 
         Parameters
         ----------
+        studies : list of int
+            Study ids to attach to portal
         portal : str
             Portal to associate with.
         """
@@ -44,13 +46,15 @@ if qiita_config.portal == "QIITA":
         sql = """INSERT INTO qiita.study_portal (study_id, portal_type_id)
                  VALUES (%s, %s)"""
         conn_handler = SQLConnectionHandler()
-        conn_handler.execute(sql, [study.id, portal_id])
+        conn_handler.executemany(sql, [(s, portal_id) for s in studies])
 
-    def remove_study_from_portal(study, portal):
-        """Adds study to given portal
+    def remove_studies_from_portal(portal, studies):
+        """Removes studies to given portal
 
         Parameters
         ----------
+        studies : list of int
+            Study ids to remove from portal
         portal : str
             Portal to associate with.
         """
@@ -60,4 +64,59 @@ if qiita_config.portal == "QIITA":
         sql = """DELETE FROM qiita.study_portal WHERE study_id = %s
                  AND portal_type_id = %s"""
         conn_handler = SQLConnectionHandler()
-        conn_handler.execute(sql, [study.id, portal_id])
+        conn_handler.executemany(sql, [(s, portal_id) for s in studies])
+
+    def get_analyses_by_portal(portal):
+        """Returns analysis id for all Analyses belonging to a portal
+
+        Parameters
+        ----------
+        portal : str
+           Portal to check analyses belong to
+
+        Returns
+        -------
+        set of int
+            All analysis ids in the database that match the given portal
+        """
+        conn_handler = SQLConnectionHandler()
+        portal_id = convert_to_id(portal, 'portal_type', 'portal')
+        sql = """SELECT analysis_id FROM qiita.analysis_portal
+                 WHERE portal_type_id = %s"""
+        return {x[0] for x in
+                conn_handler.execute_fetchall(sql, [portal_id])}
+
+    def add_analyses_to_portal(portal, analyses):
+        """Adds analyses to given portal
+
+        Parameters
+        ----------
+        analyses : list of int
+            Analysis ids to attach to portal
+        portal : str
+            Portal to associate with.
+        """
+        portal_id = convert_to_id(portal, 'portal_type', 'portal')
+        sql = """INSERT INTO qiita.analysis_portal
+                 (analysis_id, portal_type_id)
+                 VALUES (%s, %s)"""
+        conn_handler = SQLConnectionHandler()
+        conn_handler.executemany(sql, [(a, portal_id) for a in analyses])
+
+    def remove_analyses_from_portal(portal, analyses):
+        """Removes  analyses from given portal
+
+        Parameters
+        ----------
+        analyses : list of int
+            Analysis ids to remove from portal
+        portal : str
+            Portal to associate with.
+        """
+        if portal == "QIITA":
+            raise ValueError('Can not remove from main QIITA portal!')
+        portal_id = convert_to_id(portal, 'portal_type', 'portal')
+        sql = """DELETE FROM qiita.analysis_portal WHERE analysis_id = %s
+                 AND portal_type_id = %s"""
+        conn_handler = SQLConnectionHandler()
+        conn_handler.executemany(sql, [(a, portal_id) for a in analyses])
--- a/qiita_db/portal.py
+++ b/qiita_db/portal.py
@@ -14,6 +14,12 @@ from qiita_core.qiita_settings import qiita_config
 
 # Only make functions available for main portal
 if qiita_config.portal == "QIITA":
+    def list_portals():
+        """Returns list of portals available in system"""
+        sql = "SELECT portal FROM qiita.portal_type"
+        conn_handler = SQLConnectionHandler()
+        return [x[0] for x in conn_handler.execute_fetchall(sql)]
+
     def get_studies_by_portal(portal):
         """Returns study id for all Studies belonging to a portal
 
--- a/qiita_db/portal.py
+++ b/qiita_db/portal.py
@@ -10,11 +10,26 @@ import warnings
 from .sql_connection import SQLConnectionHandler
 from .util import convert_to_id
 from .base import QiitaObject
-from .exceptions import QiitaDBError, QiitaDBDuplicateError
+from .exceptions import QiitaDBError, QiitaDBDuplicateError, QiitaDBWarning
 from qiita_core.exceptions import IncompetentQiitaDeveloperError
 
 
 class Portal(QiitaObject):
+    r"""Portal object to create and maintain portals in the system
+
+    Attributes
+    ----------
+    portal
+
+    Methods
+    -------
+    get_studies
+    add_studies
+    remove_studies
+    get_analyses
+    add_analyses
+    remove_analyses
+    """
     _table = 'portal_type'
 
     def __init__(self, portal):
@@ -43,7 +58,7 @@ class Portal(QiitaObject):
 
     @classmethod
     def create(cls, portal, desc):
-        """Creates a new portal on the system
+        """Creates a new portal and its default analyses on the system
 
         Parameters
         ----------
@@ -55,7 +70,7 @@ class Portal(QiitaObject):
         Raises
         ------
         QiitaDBDuplicateError
-            Portal already exists
+            Portal name already exists
         """
         if cls.exists(portal):
             raise QiitaDBDuplicateError("Portal", portal)
@@ -93,7 +108,7 @@ class Portal(QiitaObject):
 
     @staticmethod
     def delete(portal):
-        """Removes a portal from the system
+        """Removes a portal and its default analyses from the system
 
         Parameters
         ----------
@@ -113,7 +128,7 @@ class Portal(QiitaObject):
         studies = conn_handler.execute_fetchall(sql, [portal_id])
         if studies:
             raise QiitaDBError(
-                "Studies still attached to portal %s: %s" %
+                " Cannot delete portal '%s', studies still attached: %s" %
                 (portal, ', '.join([str(s[0]) for s in studies])))
 
         # Check if attached to any analyses
@@ -123,7 +138,7 @@ class Portal(QiitaObject):
         analyses = conn_handler.execute_fetchall(sql, [portal_id])
         if studies:
             raise QiitaDBError(
-                "Analyses still attached to portal %s: %s" %
+                " Cannot delete portal '%s', analyses still attached: %s" %
                 (portal, ', '.join([str(a[0]) for a in analyses])))
 
         # Remove portal and default analyses for all users
@@ -153,6 +168,18 @@ class Portal(QiitaObject):
 
     @staticmethod
     def exists(portal):
+        """Returns whether the portal name already exists on the system
+
+        Parameters
+        ----------
+        portal : str
+            Name of portal to check
+
+        Returns
+        -------
+        bool
+            Whether the portal exists or not
+        """
         try:
             convert_to_id(portal, 'portal_type', 'portal')
         except IncompetentQiitaDeveloperError:
@@ -181,7 +208,7 @@ class Portal(QiitaObject):
         existing = [x[0] for x in conn_handler.execute_fetchall(
             sql, [tuple(studies)])]
         if len(existing) != len(studies):
-            bad = map(str, (set(studies).difference(existing)))
+            bad = map(str, set(studies).difference(existing))
             raise QiitaDBError("The following studies do not exist: %s" %
                                ", ".join(bad))
 
@@ -190,29 +217,32 @@ class Portal(QiitaObject):
 
         Parameters
         ----------
-        studies : list of int
+        studies : iterable of int
             Study ids to attach to portal
 
         Raises
         ------
         QiitaDBError
-            Some studies given do not exist
+            Some studies given do not exist in the system
+        QiitaDBWarning
+            Some studies already exist in the given portal
         """
         self._check_studies(studies)
 
         conn_handler = SQLConnectionHandler()
         # Clean list of studies down to ones not associated with portal already
         sql = """SELECT study_id from qiita.study_portal
-                 WHERE portal_type_id != %s AND study_id IN %s"""
-        clean_studies = [x[0] for x in conn_handler.execute_fetchall(
-                         sql, [self._id, tuple(studies)])]
+                 WHERE portal_type_id = %s AND study_id IN %s"""
+        duplicates = [x[0] for x in conn_handler.execute_fetchall(
+                      sql, [self._id, tuple(studies)])]
 
-        if len(clean_studies) != len(studies):
-            rem = map(str, set(studies).difference(clean_studies))
+        if len(duplicates) > 0:
             warnings.warn("The following studies area already part of %s: %s" %
-                          (self.portal, ', '.join(rem)))
+                          (self.portal, ', '.join(map(str, duplicates))),
+                          QiitaDBWarning)
 
         # Add cleaned list to the portal
+        clean_studies = set(studies).difference(duplicates)
         sql = """INSERT INTO qiita.study_portal (study_id, portal_type_id)
                  VALUES (%s, %s)"""
         conn_handler.executemany(sql, [(s, self._id) for s in clean_studies])
@@ -222,7 +252,7 @@ class Portal(QiitaObject):
 
         Parameters
         ----------
-        studies : list of int
+        studies : iterable of int
             Study ids to remove from portal
 
         Raises
@@ -230,7 +260,9 @@ class Portal(QiitaObject):
         ValueError
             Trying to delete from QIITA portal
         QiitaDBError
-            Some studies given do not exist
+            Some studies given do not exist in the system
+        QiitaDBWarning
+            Some studies already do not exist in the given portal
         """
         if self.portal == "QIITA":
             raise ValueError('Can not remove from main QIITA portal!')
@@ -246,7 +278,7 @@ class Portal(QiitaObject):
         if len(clean_studies) != len(studies):
             rem = map(str, set(studies).difference(clean_studies))
             warnings.warn("The following studies are not part of %s: %s" %
-                          (self.portal, ', '.join(rem)))
+                          (self.portal, ', '.join(rem)), QiitaDBWarning)
 
         sql = """DELETE FROM qiita.study_portal
                  WHERE study_id IN %s AND portal_type_id = %s"""
@@ -294,13 +326,16 @@ class Portal(QiitaObject):
 
         Parameters
         ----------
-        analyses : list of int
+        analyses : iterable of int
             Analysis ids to attach to portal
 
         Raises
         ------
         QiitaDBError
-            Some given analyses do not exist, or are default analyses
+            Some given analyses do not exist in the system,
+            or are default analyses
+        QiitaDBWarning
+            Some analyses already exist in the given portal
         """
         self._check_analyses(analyses)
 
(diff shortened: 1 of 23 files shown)


josenavas:
@squirrelo I just noticed one thing: in my configuration file I do not have the portal entry, shouldn't this fail?

qiita_pet/handlers/portal.py
```diff
@@ -0,0 +1,41 @@
+from tornado.web import authenticated, HTTPError
```
ElDeveloper: Can you add the Qiita copyright header?

josenavas:
Even if add PORTAL = QIITA to my config file I can't see the portals admin page...

qiita_pet/templates/portals_edit.html
```diff
@@ -0,0 +1,96 @@
+{% extends sitebase.html%}
+
+{%block head%}
+<link rel="stylesheet" href="/static/vendor/css/jquery.dataTables.css" type="text/css">
+<script src="/static/vendor/js/jquery.dataTables.min.js"></script>
+<script type="text/javascript">
+function check_submit(action) {
+  error = false;
```
ElDeveloper: can you do `var error` instead?

qiita_pet/handlers/portal.py
```diff
@@ -0,0 +1,41 @@
...
+
+
+class StudyPortalHandler(BaseHandler):
+    def check_admin(self):
+        if self.current_user.level != "admin":
+            raise HTTPError(403, "%s does not have access to portal editing!" %
+                            self.current_user.id)
+
+    def render_page(self):
+        # COLS MUST HAVE AN ID COLUMN!
```
josenavas: I don't understand what are you trying to say (I also don't understand why you're yelling either hehehe)

qiita_pet/handlers/portal.py
```diff
@@ -0,0 +1,41 @@
...
+        studies = self.get_arguments('selected')
+        action = self.get_argument('action')
+
+        portal = Portal(portal)
+        if action == "Add":
+            portal.add_studies(studies)
+        elif action == "Remove":
+            portal.remove_studies(studies)
+        else:
+            raise ValueError("Unknown action: %s" % action)
```
josenavas: Shouldn't this be a Tornado error?

josenavas: e.g. 403 Forbidden?

qiita_pet/templates/portals_edit.html
```diff
@@ -0,0 +1,96 @@
...
+  var data = oTable.$('input').serialize();
```
josenavas: Not sure how it works (and I can't see the interface so far) but is it possible to add an event listener to the dataTable object so you can catch when the user choses something and then you make the button clickable? This way you can avoid this check here...

qiita_pet/webserver.py
```diff
@@ -93,9 +95,16 @@ def __init__(self):
...
             (r"/vamps/(.*)", VAMPSHandler),
-            # 404 PAGE MUST BE LAST IN THIS LIST!
-            (r".*", NoPageHandler)
         ]
+        if qiita_config.portal == "QIITA":
+            # Add portals editing pages only on main portal
+            portals = [
+                (r"/admin/portals/studies/", StudyPortalHandler)
+            ]
+            handlers.extend(portals)
```
josenavas: Why are you using extend? You can use a simple append:
handlers.append((r"/admin/portals/studies/", StudyPortalHandler))

josenavas:
few comments, however I couldn't check this locally...

josenavas:
@squirrelo it was actually my fault, I was in the wrong branch...

I don't really like the layout of the current page, there is a lot of wasted space. Having an entire column on the left to just put the portal selection on top is useless. Can we just have the dropdown menu on top?

Also, for me is not showing the QIITA portal, it just shows a blank space and the EMP portal.

If I change the selected portal is not updating the study list.

I think this interface needs a bit more thought and a lot of testing, I couldn't get anything to work...

josenavas:
It also needs to be similar to the other pages when showing the errors, using the bootstrap error system on top...

qiita_pet/webserver.py
```diff
@@ -93,9 +95,16 @@ def __init__(self):
             (r"/stats/", StatsHandler),
             (r"/download/(.*)", DownloadHandler),
             (r"/vamps/(.*)", VAMPSHandler),
-            # 404 PAGE MUST BE LAST IN THIS LIST!
-            (r".*", NoPageHandler)
         ]
+        if qiita_config.portal == "QIITA":
+            # Add portals editing pages only on main portal
+            portals = [
+                (r"/admin/portals/studies/", StudyPortalHandler)
```
ElDeveloper: I guess you foresee adding new handlers specific to Qiita and that's why you are making this section a list?

squirrelo: Yes, correct. I'd like to have it ready for extension now, instead of adding it later.

qiita_pet/webserver.py
squirrelo: See above answer to @ElDeveloper: there will most likely be more portal-specific maintenance pages added, and I'd like this extensible now instead of later. Mostly so the 404 works properly.

qiita_pet/templates/portals_edit.html
squirrelo: This is absolutely needed no matter what, because of the way datatables interacts with checkboxes. When you get access to the interface, you'll see the error messages added.

qiita_pet/templates/portals_edit.html
squirrelo: done

squirrelo:
@josenavas What errors are you referring to? The submission validation is done in the standard way on every other page.

qiita_pet/handlers/portal.py
squirrelo: 400 error if anything. Can add.

josenavas:
In the other pages, the error is shown just below the navigation bar, here are shown under the portals stuff.

squirrelo:
Can I get a screenshot? I'm not quite sure what you mean. The straight errors are shown in bootstrap errors system, but form validation issues are never shown that way.

ElDeveloper:
@squirrelo, I'm getting this error when I try to create a new environment using this branch:

```python-traceback
SKIPPING download of ontologies: File already exists at /Users/yoshikivazquezbaeza/git_sw/qiita/qiita_db/support_files/test_data/reference/ontologies.sql.gz. To download the file again, delete the existing file first.
  File "/Users/yoshikivazquezbaeza/git_sw/qiita/scripts/qiita-env", line 183, in <module>
    env()
  File "/Users/yoshikivazquezbaeza/.virtualenvs/qiita/lib/python2.7/site-packages/click/core.py", line 664, in __call__
  File "/Users/yoshikivazquezbaeza/.virtualenvs/qiita/lib/python2.7/site-packages/click/core.py", line 644, in main
    rv = self.invoke(ctx)
  File "/Users/yoshikivazquezbaeza/.virtualenvs/qiita/lib/python2.7/site-packages/click/core.py", line 991, in invoke
  File "/Users/yoshikivazquezbaeza/.virtualenvs/qiita/lib/python2.7/site-packages/click/core.py", line 837, in invoke
  File "/Users/yoshikivazquezbaeza/.virtualenvs/qiita/lib/python2.7/site-packages/click/core.py", line 464, in invoke
  File "/Users/yoshikivazquezbaeza/git_sw/qiita/scripts/qiita-env", line 75, in make
    make_environment(load_ontologies, download_reference, add_demo_user)
  File "/Users/yoshikivazquezbaeza/git_sw/qiita/qiita_db/environment_manager.py", line 251, in make_environment
    conn.execute_many(sql, args)
AttributeError: 'SQLConnectionHandler' object has no attribute 'execute_many'
```

josenavas:
The study in the test database is not showing up in any portal.
What I mean is that this type of errors:
screen shot 2015-07-01 at 5 50 30 pm
Usually are shown directly where you have to chose something. Since here it's saying choose a row, it's kind of confusing because is not in the table itself. I think using bootstrap in a similar way than when you're selecting samples will be way more useful here.

Also, I would suggest to always have a portal selected.

I think this page needs a bit of re-design: it needs a dropdown on top to chose the portals so the page is updated with the studies of that portal. Then, you should have an action part which is pretty similar to what you currently have, but I think this workflow is better: He has 2 dropdowns: first one to chose the option (add/remove). Second one, to choose the portal (add to/ remove from). Finally a submit button to perform the operation.

In the current interface there is no way to know to which portals each study belongs to...

squirrelo:
I am very, very against having a portal selected automatically, as that can lead to Bad Things (tm) if someone selects studies then click add accidentally without changing to the portal they want. Also, QIITA is not in that portal list because you should never be adding or removing things from that portal. It's a safety thing.

I am redesigning the rest of the interface slightly as we speak, stay tuned for push of the changes.

josenavas:
Please read carefully my comment

You actually have 2 dropdowns with the portal. The one on top is to show which studies you want to show: i.e. if you choose EMP, you will see listed only the studies of the EMP, if you select Qiita (which I will suggest having it as the default) you see all of them. Then you have another set of dropdowns which let you chose which option to perform (add/remove) and to/from portal to do so. And that last one the portal is not  selected by default so you don't have the issue that you're explaining.

Again, note that the main drawback of the current interface is that there is no feedback on which studies are in which portal...

Note that if some developer the first time that wants to use the interface is a bit lost and doesn't feel comfortable, other users can potentially have the same issues. Interfaces should be intuitive, that do not need anybody to explain how they work

squirrelo:
That seems like a lot of unnecessary dropdowns. I'll finish my recode and you can take another look.

squirrelo:
Interface redesign pushed. The select on the top right will dynamically update the table, similar to doing a search in the list studies page, but did not want to code in that functionality until we have an interface.

ElDeveloper:
@squirrelo, the interface seems to be working fine, though one thing that's extremely confusing is that you get no feedback once you associate a study with a portal:

broken-1

If you click the button twice (like I did when I saw nothing change), you get the following error on the console:

```python-trackeback
File "/Users/yoshikivazquezbaeza/.virtualenvs/qiita/lib/python2.7/site-packages/tornado/web.py", line 1141, in _when_complete
    callback()
  File "/Users/yoshikivazquezbaeza/.virtualenvs/qiita/lib/python2.7/site-packages/tornado/web.py", line 1162, in _execute_method
  File "/Users/yoshikivazquezbaeza/git_sw/qiita/qiita_pet/handlers/portal.py", line 38, in post
    portal.add_studies(studies)
  File "/Users/yoshikivazquezbaeza/git_sw/qiita/qiita_db/portal.py", line 248, in add_studies
    conn_handler.executemany(sql, [(s, self._id) for s in clean_studies])
  File "/Users/yoshikivazquezbaeza/git_sw/qiita/qiita_db/sql_connection.py", line 395, in executemany
    with self._sql_executor(sql, sql_args_list, True):
  File "/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/contextlib.py", line 17, in __enter__
    return self.gen.next()
  File "/Users/yoshikivazquezbaeza/git_sw/qiita/qiita_db/sql_connection.py", line 346, in _sql_executor
    "Error: %s" % (sql, str(sql_args), e))
ValueError: Error running SQL query: INSERT INTO qiita.study_portal (study_id, portal_type_id)
                 VALUES (%s, %s)
ARGS: [(u'1024', 2L), (u'1222', 2L), (u'1001', 2L), (u'1198', 2L)]
Error: duplicate key value violates unique constraint "pk_study_portal"
DETAIL:  Key (study_id, portal_type_id)=(1222, 2) already exists.
```

When you resize the window and the navigation bar grows in size, the menu is hidden:

screen shot 2015-07-01 at 8 39 42 pm

------

I think a column in this list should be the portal to which the study belongs to and perhaps that would get rid of the dropdown on the right.

josenavas:
Not tested locally, but taking a look to @ElDeveloper giff, I can see that you ignored my comment.

We need a dropdown on top that when you select the portal, it changes the list of studies to show the ones available on that portal...

josenavas:
And then, in another place, you have whatever mechanism you think it is needed to change the studies across portals

squirrelo:
@josenavas `The select on the top right will dynamically update the table, similar to doing a search in the list studies page, but did not want to code in that functionality until we have an interface.` No, I only ignored your dropdown for action, as that seems way less intuitive than two buttons. Please check locally before commenting. 

@ElDeveloper Hmm, I thought I disabled the submit buttons while it was working but I guess not. Will do a pull request with that in it. Also not sure why the resize covers the bar, as I can't get it to do that on my computer. I'll investigate further.

ElDeveloper:
Thanks!!

squirrelo:
OK, all @ElDeveloper issues should be fixed. Ready for another look. Once we have a layout, I will add the code for changing the studies list by selected portal.

squirrelo:
Also, for adding the portals as another column: that is computationally expensive and one study can be attached to multiple portals, so not sure how to take that into account.

qiita_db/portal.py
```diff
@@ -40,10 +40,12 @@ def __init__(self, portal):
...
         """Returns list of portals available in system
+
         Returns
         -------
         list of str
             List of portal names for the system
+
         Notes
         -----
         This does not return the QIITA portal in the list, as it is a required
```
wasade: Doesn't this imply that the short docstring is inaccurate? Is there a benefit to this method not including "qiita"? The argument to include it is that it is a valid portal.

I realize this is outside of the present PR, so can divert to an issue if desired.

qiita_pet/handlers/portal.py
```diff
@@ -0,0 +1,51 @@
...
+        portal = Portal(portal)
+        with warnings.catch_warnings(record=True) as warns:
+            if action == "Add":
+                portal.add_studies(studies)
+            elif action == "Remove":
+                portal.remove_studies(studies)
+            else:
+                raise HTTPError(400, "Unknown action: %s" % action)
+
+            if warns:
```
wasade: Is this ever not an `Iterable`? If, when no warnings are thrown, this variable is an empty Iterable, then there is no need to perform the conditional and no need to stage `msg` on line 38. As an example:

```python
08:28:23 (mcdonadt@Daniels-MacBook-Pro-3):/Users/mcdonadt
> '; '.join([])
Out[1]: ''
```

squirrelo: This `warns` catch is to get the warning out of adding or removing studies that are already part of or not part of the portal, respectively. This allows for showing that information to the user.

wasade: Right, but all it does is format a string and store it in `msg`. I believe the conditional, and the setting of `msg` to the empty string on line 38 are both redundant.

qiita_db/portal.py
squirrelo: I did this because the QIITA portal is a default portal that you can't edit. Would it make it clearrer to change the docstring to `...list of non-default portals...`?

qiita_pet/templates/portals_edit.html
```diff
@@ -0,0 +1,134 @@
...
+  if(data.length === 0) {
+    //No checked rows, so add error message if needed and then don't submit
+    if(!$('#checkbox-error').length) { $("#errors").append("<li id='checkbox-error'>Please select at least one row</li>"); }
+    error = true;
+  }
+  if(error) {
+    $("#add-button").prop('disabled', false);
+    $("#remove-button").prop('disabled', false);
+    return false;
```
wasade: `check_submit` returns `false` on `error` but never appears to return `true`, should it at the end of the method?

qiita_pet/templates/portals_edit.html
```diff
@@ -0,0 +1,134 @@
...
+          return $('input', td).prop('checked') ? '1' : '0';
+      } );
+  };
+
+  oTable = $('#info-table').dataTable({
+    "deferRender": true,
+    "iDisplayLength": 50,
+    "aoColumns": [
+            { "sSortDataType": "dom-checkbox" },
+            {% for i in headers %}
```
wasade: This forloop doesn't appear to do anything. Is it necessary?

qiita_pet/handlers/portal.py
squirrelo: Ah, did not understand what you meant in the first comment. Apparently no, it's just an empty list if no warnings. Fixed.

qiita_pet/handlers/portal.py
squirrelo: done

qiita_db/portal.py
wasade: eh, sure :) thanks

qiita_pet/handlers/portal.py
wasade: Thanks!

qiita_pet/handlers/portal.py
squirrelo: Also, lookng at this, most of the handlers do not have the copyright header. Issue opened: #1321

wasade:
Just a few minor (I believe) comments. Only able to do static analysis, but :+1: on that

qiita_pet/templates/portals_edit.html
squirrelo: `return false;` stops the submission in its tracks. There is no need to return true since it will continue on and submit at the end. Yay javascript conventions.

qiita_pet/templates/portals_edit.html
squirrelo: Yes, you need to have the `null,` foreach column other than the first in this list. This is what allows sorting by checkbox status.

wasade: ohhhhhh... okay

qiita_pet/templates/portals_edit.html
wasade: so `null === true`? What I'm wondering about is:

```javascript
if(check_status()) {
    // do stuff
}
```

squirrelo: check status also does the form submission, so I would advise against that.

squirrelo:
Also, if we are good on the interface now I will add the code for changing portal views.

ElDeveloper:
Thanks @squirrelo, the interface looks good! However when the message is too long, it looks a bit clunky:

screen shot 2015-07-02 at 8 50 18 am

I also noticed you fixed the problems with the navigation bar, thanks for looking into that!!! :+1:!

The only thing that's still not quite there yet is the fact that the feedback after adding a study to a portal is very subtle. See this example where I (1) add a study to the EMP portal, (2) reload the page, (3) try to add it again (forgot to select a portal :P ) and (4) I select a portal and add a study to it (which completes successfully also saying that it was already there); I think this will lead people to think the studies were never added to a portal. Thoughts?

anticipate

squirrelo:
I agree it may be a bit subtle, but I can't think of a better way to do it. Any ideas?

ElDeveloper:
That's why I suggested having a column where you could see what portals each study was associated with. Maybe something like dropbox when it shows you the initial of who you've shared a file with:

screen shot 2015-07-02 at 9 21 13 am

josenavas:
I really like that idea @ElDeveloper !
Also, portals shouldn't be long (QIITA, EMP, SLOAN, HMP, etc...) so if we can keep them short it will be awesome!

ElDeveloper:
Either the name or the initial + the full name on hover?

squirrelo:
Changes added. Take another look and check functionality.

ElDeveloper:
Thanks @squirrelo, this is looks great! The addition of the select all/ invert/ none buttons is very slick! I hereby declare this :+1:. @josenavas, would you mind reviewing and merging?

josenavas:
@squirrelo I have a single study in the DB and I assigned it to the EMP portal, when I removed it from the EMP portal I received this error:
```
  File "/Users/jose/.virtualenvs/qiita/lib/python2.7/site-packages/tornado/web.py", line 1346, in _execute
  File "/Users/jose/.virtualenvs/qiita/lib/python2.7/site-packages/tornado/web.py", line 2626, in wrapper
  File "/Users/jose/qiime_software/QiiTa/qiita_pet/handlers/portal.py", line 77, in get
    info = self.get_info(portal=portal)
  File "/Users/jose/qiime_software/QiiTa/qiita_pet/handlers/portal.py", line 30, in get_info
    info_cols=self.study_cols)
  File "/Users/jose/qiime_software/QiiTa/qiita_db/study.py", line 253, in get_info
    res = conn_handler.execute_fetchall(sql)
  File "/Users/jose/qiime_software/QiiTa/qiita_db/sql_connection.py", line 455, in execute_fetchall
    with self._sql_executor(sql, sql_args) as pgcursor:
  File "/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/contextlib.py", line 17, in __enter__
    return self.gen.next()
  File "/Users/jose/qiime_software/QiiTa/qiita_db/sql_connection.py", line 346, in _sql_executor
    "Error: %s" % (sql, str(sql_args), e))
ValueError: Error running SQL query: SELECT study_alias,study_id,study_title FROM (
            qiita.study
            JOIN qiita.timeseries_type  USING (timeseries_type_id)
            LEFT JOIN (SELECT study_id, array_agg(pmid ORDER BY pmid) as
            pmid FROM qiita.study_pmid GROUP BY study_id) sp USING (study_id)
            JOIN qiita.study_portal USING (study_id)
            JOIN qiita.portal_type USING (portal_type_id)
            ) WHERE portal = 'QIITA' AND study_id in ()
ARGS: None
Error: syntax error at or near ")"
LINE 8:             ) WHERE portal = 'QIITA' AND study_id in ()
```
In some place an empty list to SQL is passed.
Also there is a typo in here: https://github.com/biocore/qiita/pull/1266/files#diff-3fcea9f71411fb6fdd4a0edd9a184106R240 (area -> are)

squirrelo:
Both addressed. Ready to go.

ElDeveloper:
@josenavas, can you have a look?

josenavas:
Thanks for the heads up @ElDeveloper going over it right now.

josenavas:
Still seeing this error @squirrelo 
```
$ qiita pet webserver start
Building documentation ...
Documentation successfully built
Qiita started on port 21174
ERROR:tornado.application:Uncaught exception GET /admin/portals/studiesAJAX/?sEcho=274&view-portal=EMP&_=1435981162480 (::1)
HTTPServerRequest(protocol='http', host='localhost:21174', method='GET', uri='/admin/portals/studiesAJAX/?sEcho=274&view-portal=EMP&_=1435981162480', version='HTTP/1.1', remote_ip='::1', headers={'Accept-Language': 'es,en-US;q=0.8,en;q=0.6,ca;q=0.4', 'Accept-Encoding': 'gzip, deflate, sdch', 'X-Requested-With': 'XMLHttpRequest', 'Host': 'localhost:21174', 'Accept': 'application/json, text/javascript, */*; q=0.01', 'User-Agent': 'Mozilla/5.0 (Macintosh; Intel Mac OS X 10_10_3) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/43.0.2357.124 Safari/537.36', 'Connection': 'keep-alive', 'Referer': 'http://localhost:21174/admin/portals/studies/', 'Cookie': 'galaxysession=c6ca0ddb55be603a2ee83535219dee0032ba4c80762b1465e15e9db14df212b1f42cf3c0900a845d; user="2|1:0|10:1435981160|4:user|20:ImFkbWluQGZvby5iYXIi|b4e833086b900d778fb71b12d8b4faaf99f0767517da0dacede3df395d95011a"'})
  File "/Users/jose/.virtualenvs/qiita/lib/python2.7/site-packages/tornado/web.py", line 1346, in _execute
  File "/Users/jose/.virtualenvs/qiita/lib/python2.7/site-packages/tornado/web.py", line 2626, in wrapper
  File "/Users/jose/qiime_software/QiiTa/qiita_pet/handlers/portal.py", line 77, in get
    info = self.get_info(portal=portal)
  File "/Users/jose/qiime_software/QiiTa/qiita_pet/handlers/portal.py", line 30, in get_info
    info_cols=self.study_cols)
  File "/Users/jose/qiime_software/QiiTa/qiita_db/study.py", line 253, in get_info
    res = conn_handler.execute_fetchall(sql)
  File "/Users/jose/qiime_software/QiiTa/qiita_db/sql_connection.py", line 455, in execute_fetchall
    with self._sql_executor(sql, sql_args) as pgcursor:
  File "/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/contextlib.py", line 17, in __enter__
    return self.gen.next()
  File "/Users/jose/qiime_software/QiiTa/qiita_db/sql_connection.py", line 346, in _sql_executor
    "Error: %s" % (sql, str(sql_args), e))
ValueError: Error running SQL query: SELECT study_alias,study_id,study_title FROM (
            qiita.study
            JOIN qiita.timeseries_type  USING (timeseries_type_id)
            LEFT JOIN (SELECT study_id, array_agg(pmid ORDER BY pmid) as
            pmid FROM qiita.study_pmid GROUP BY study_id) sp USING (study_id)
            JOIN qiita.study_portal USING (study_id)
            JOIN qiita.portal_type USING (portal_type_id)
            ) WHERE portal = 'QIITA' AND study_id in ()
ARGS: None
Error: syntax error at or near ")"
LINE 8:             ) WHERE portal = 'QIITA' AND study_id in ()

```

In the test DB, add the study to the EMP portal, then change `View from portal` to the EMP and delete that study from the EMP.

squirrelo:
fixed and ready for another look.

josenavas:
I still see something weird in the interface and I think it can be improved. However, I have not a clear idea on the best path to improve it and I will need to do some research.

This is functional so I will not block it.

Thanks @squirrelo

PR #1316, opened 2015-06-30
Portals scripts

squirrelo:
Continuation from #1303 

Portals now is an object. The scripts also interact with the portals through this object.

Commands are now in the form `qiita db portal add-studies [PORTALNAME] [STUDIES]`, so adding studies 5,6, and 7 to EMP would look like `qiita db portal add-studies EMP 5 6 7`


squirrelo:
Let the review talks commence!

qiita_db/portal.py
```diff
@@ -5,20 +5,145 @@
...
+    @staticmethod
+    def delete(portal):
+        """Removes a portal from the system
 
         Parameters
         ----------
         portal : str
-           Portal to check studies belong to
+            The name of the portal to add
+        desc : str
```
josenavas: desc is not a parameter here

qiita_db/portal.py
```diff
@@ -5,20 +5,145 @@
...
+        sql = """SELECT study_id from qiita.study_portal
+                 WHERE portal_type_id = %s"""
+        studies = conn_handler.execute_fetchall(sql, [portal_id])
+        if studies:
+            raise QiitaDBError(
+                "Studies still attached to portal %s: %s" %
+                (portal, ', '.join([str(s[0]) for s in studies])))
+
+        # Check if attached to any analyses
+        sql = """SELECT analysis_id from qiita.analysis_portal
```
josenavas: This will always returns some analyses given that the default analyses are attached to the portal right? You will need to modify this line to avoid counting those analyses.

squirrelo: Good catch, fixed.

qiita_db/portal.py
```diff
@@ -26,38 +151,193 @@ def get_studies_by_portal(portal):
...
         sql = """INSERT INTO qiita.study_portal (study_id, portal_type_id)
                  VALUES (%s, %s)"""
+        conn_handler.executemany(sql, [(s, self._id) for s in clean_studies])
+
+    def remove_studies(self, studies):
+        """Removes studies from given portal
+
+        Parameters
+        ----------
+        portal : str
```
josenavas: portal is not a parameter

qiita_db/portal.py
```diff
@@ -26,38 +151,193 @@ def get_studies_by_portal(portal):
...
+                "assigned to another portal: %s" % ", ".join(bad))
 
-    def remove_study_from_portal(study, portal):
-        """Adds study to given portal
+    def add_analyses(self, analyses):
+        """Adds analyses to given portal
 
         Parameters
         ----------
         portal : str
```
josenavas: portal is not a parameter

qiita_db/portal.py
```diff
@@ -26,38 +151,193 @@ def get_studies_by_portal(portal):
...
+
+        sql = """INSERT INTO qiita.analysis_portal
+                 (analysis_id, portal_type_id)
+                 VALUES (%s, %s)"""
+        conn_handler.executemany(sql, [(a, self._id) for a in clean_analyses])
+
+    def remove_analyses(self, analyses):
+        """Removes analyses from given portal
+
+        Parameters
```
josenavas: Documentation is not accurate.

josenavas:
@squirrelo did you change the permissions on the scripts file? Github shows some change as if they're not executable now.

squirrelo:
Odd, let me check. Nothing should have happened because I just edited them, but I'll do another push if it is the case.

squirrelo:
Weird, but fixed.

qiita_db/portal.py
squirrelo: fixed

squirrelo:
Ready for more review.

qiita_db/portal.py
```diff
@@ -5,20 +5,144 @@
...
+                    WHERE analysis_id = aid;
+                END LOOP;
+
+                DELETE FROM qiita.portal_type WHERE portal_type_id = %s;
+            END $do$;"""
+        conn_handler = SQLConnectionHandler()
+        conn_handler.execute(sql, [portal_id, portal_id])
+
+    @staticmethod
+    def exists(portal):
```
josenavas: This method is missing the documentation.

qiita_db/portal.py
```diff
@@ -26,38 +150,191 @@ def get_studies_by_portal(portal):
...
+        conn_handler = SQLConnectionHandler()
+        # Clean list of studies down to ones not associated with portal already
+        sql = """SELECT study_id from qiita.study_portal
+                 WHERE portal_type_id != %s AND study_id IN %s"""
+        clean_studies = [x[0] for x in conn_handler.execute_fetchall(
+                         sql, [self._id, tuple(studies)])]
+
+        if len(clean_studies) != len(studies):
+            rem = map(str, set(studies).difference(clean_studies))
+            warnings.warn("The following studies area already part of %s: %s" %
```
josenavas: Can you use the QiitaDBWarnings?

@ElDeveloper do the warnings need to be in the numpy doc?

qiita_db/portal.py
```diff
@@ -26,38 +150,191 @@ def get_studies_by_portal(portal):
...
         conn_handler = SQLConnectionHandler()
-        conn_handler.execute(sql, [study.id, portal_id])
+        sql = """SELECT study_id from qiita.study_portal
+                 WHERE portal_type_id = %s AND study_id IN %s"""
+        clean_studies = [x[0] for x in conn_handler.execute_fetchall(
+                         sql, [self._id, tuple(studies)])]
+
+        if len(clean_studies) != len(studies):
+            rem = map(str, set(studies).difference(clean_studies))
+            warnings.warn("The following studies are not part of %s: %s" %
```
josenavas: Same as above

josenavas:
A couple more comments @squirrelo

qiita_db/portal.py
squirrelo: added

qiita_db/portal.py
squirrelo: Done on both.

josenavas:
:+1:

qiita_db/portal.py
ElDeveloper: There's nothing specific about them in the documentation, but I would
suggest noting them and the reason why they will appear (we do the same
in load_template_to_dataframe ...). For reference here's the
documentation:
https://github.com/numpy/numpy/blob/master/doc/HOWTO_DOCUMENT.rst.txt

josenavas: Thanks @ElDeveloper !

ackermag: Do any/all of you know which studies are assigned to which portal?  do i
need to make a list by portal?

On Tue, Jun 30, 2015 at 11:18 PM, josenavas <notifications@github.com>
wrote:

> In qiita_db/portal.py
> <https://github.com/biocore/qiita/pull/1316#discussion_r33651272>:
>
> >          """
> > -        portal_id = convert_to_id(portal, 'portal_type', 'portal')
> > +        self._check_studies(studies)
> > +
> > +        # Clean list of studies down to ones not associated with portal already
> > +        sql = """SELECT study_id from qiita.study_portal
> > +                 WHERE portal_type_id != %s AND study_id IN %s"""
> > +        clean_studies = [x[0] for x in conn_handler.execute_fetchall(
> > +                         sql, [self._id, tuple(studies)])]
> > +
> > +        if len(clean_studies) != len(studies):
> > +            rem = map(str, set(studies).difference(clean_studies))
> > +            warnings.warn("The following studies area already part of %s: %s" %
>
> Thanks @ElDeveloper <https://github.com/ElDeveloper> !
>
> —
> <https://github.com/biocore/qiita/pull/1316/files#r33651272>.
>



-- 
Gail Ackermann
Knight Lab
UCSD

antgonza: @ackermag please make a list of which studies should be in another portal. We know that all go to qiita but we need a list of which go to SloanBuiltEnvironment and EMP (other?).

qiita_db/portal.py
```diff
@@ -5,20 +5,156 @@
...
+from .base import QiitaObject
+from .exceptions import QiitaDBError, QiitaDBDuplicateError, QiitaDBWarning
+from qiita_core.exceptions import IncompetentQiitaDeveloperError
 
-# Only make functions available for main portal
-if qiita_config.portal == "QIITA":
-    def get_studies_by_portal(portal):
-        """Returns study id for all Studies belonging to a portal
+
+class Portal(QiitaObject):
```
ElDeveloper: Can you add a docstring describing the class?

qiita_db/portal.py
```diff
@@ -5,20 +5,156 @@
...
+        ----------
+        portal : str
+            The name of the portal to add
+        desc : str
+            Description of the portal
+
+        Raises
+        ------
+        QiitaDBDuplicateError
+            Portal already exists
```
ElDeveloper: Should probably note that it checks based on the portal name.

qiita_db/portal.py
```diff
@@ -5,20 +5,156 @@
...
+        """
+        conn_handler = SQLConnectionHandler()
+        # Check if attached to any studies
+        portal_id = convert_to_id(portal, 'portal_type', 'portal')
+        sql = """SELECT study_id from qiita.study_portal
+                 WHERE portal_type_id = %s"""
+        studies = conn_handler.execute_fetchall(sql, [portal_id])
+        if studies:
+            raise QiitaDBError(
+                "Studies still attached to portal %s: %s" %
```
ElDeveloper: Are IDs the most useful thing to display here? Would it be better to display the study aliases?

Also can you rephrase the error to something like: `Cannot delete portal '%s', the following studies are associated with it, please delete them or change their portal: %s`. Or something like that ...

qiita_db/portal.py
```diff
@@ -5,20 +5,156 @@
...
+                (portal, ', '.join([str(s[0]) for s in studies])))
+
+        # Check if attached to any analyses
+        sql = """SELECT analysis_id from qiita.analysis_portal
+                 JOIN qiita.analysis USING (analysis_id)
+                 WHERE portal_type_id = %s and dflt = FALSE"""
+        analyses = conn_handler.execute_fetchall(sql, [portal_id])
+        if studies:
+            raise QiitaDBError(
+                "Analyses still attached to portal %s: %s" %
```
ElDeveloper: Asame as above.

qiita_db/portal.py
```diff
@@ -5,20 +5,156 @@
...
+                END LOOP;
+            END $do$;"""
+        conn_handler = SQLConnectionHandler()
+        conn_handler.execute(sql, [portal, desc])
+
+        return cls(portal)
+
+    @staticmethod
+    def delete(portal):
+        """Removes a portal from the system
```
ElDeveloper: Removes a portal from the system -> Removes a portal and default analyses from the system.

qiita_db/portal.py
```diff
@@ -5,20 +5,156 @@
...
+                END LOOP;
+
+                DELETE FROM qiita.portal_type WHERE portal_type_id = %s;
+            END $do$;"""
+        conn_handler = SQLConnectionHandler()
+        conn_handler.execute(sql, [portal_id, portal_id])
+
+    @staticmethod
+    def exists(portal):
+        """Returns whether the portal already exists
```
ElDeveloper: This sentence is a bit weird, can you revise it?

qiita_db/portal.py
```diff
@@ -26,38 +162,199 @@ def get_studies_by_portal(portal):
...
+                conn_handler.execute_fetchall(sql, [self._id])}
+
+    def _check_studies(self, studies):
+        conn_handler = SQLConnectionHandler()
+        # Check if any study IDs given do not exist.
+        sql = "SELECT study_id from qiita.study WHERE study_id IN %s"
+        existing = [x[0] for x in conn_handler.execute_fetchall(
+            sql, [tuple(studies)])]
+        if len(existing) != len(studies):
+            bad = map(str, (set(studies).difference(existing)))
```
ElDeveloper: There's an extra parenthesis (I think) around set.difference.

qiita_db/portal.py
```diff
@@ -26,38 +162,199 @@ def get_studies_by_portal(portal):
...
-    def add_study_to_portal(study, portal):
-        """Adds study to given portal
+    def add_studies(self, studies):
+        """Adds studies to given portal
 
         Parameters
         ----------
-        portal : str
-            Portal to associate with.
+        studies : list of int
```
ElDeveloper: I think you really want a tuple of int, not a list, that way you can skip the `tuple` call in 203.

qiita_db/portal.py
```diff
@@ -26,38 +162,199 @@ def get_studies_by_portal(portal):
...
+            Study ids to attach to portal
+
+        Raises
+        ------
+        QiitaDBError
+            Some studies given do not exist
```
ElDeveloper: I would clarify and say `do not exist **in the system**`.

ElDeveloper: ignore this.

qiita_db/portal.py
```diff
@@ -26,38 +162,199 @@ def get_studies_by_portal(portal):
...
+        ----------
+        studies : list of int
+            Study ids to remove from portal
+
+        Raises
+        ------
+        ValueError
+            Trying to delete from QIITA portal
+        QiitaDBError
+            Some studies given do not exist
```
ElDeveloper: I would clarify and say do not exist **in the system**.

qiita_db/portal.py
squirrelo: That goes to close the map call.

scripts/qiita
```diff
@@ -289,6 +294,49 @@ def update_preprocessed_data(sl_out_dir, study,  preprocessed_data):
                                             preprocessed_data)
     click.echo("Preprocessed data %s successfully updated" % ppd.id)
 
+# #############################################################################
+# PORTAL COMMANDS
+# #############################################################################
+
+
+@portal.command(name="add-analyses")
+@click.option('--portal', required=True, type=str,
```
ElDeveloper: For the semantics of the command I think it would be better to call this option `name` instead of portal:

```bash
$ qiita portal --portal lolercoaster --analyses 1,2,3

$ qiita portal --name lolercoaster --analyses 1,2,3
```

Right?

scripts/qiita
```diff
@@ -289,6 +294,49 @@ def update_preprocessed_data(sl_out_dir, study,  preprocessed_data):
...
+              help='Portal to add the analyses to')
+@click.option('--analyses', required=True, type=str,
```
ElDeveloper: Instead of using comma-separated lists can you use "multiple options", see this [link](http://click.pocoo.org/4/options/#multiple-options). Same comment for the rest of the interfaces.

scripts/qiita
```diff
@@ -289,6 +294,49 @@ def update_preprocessed_data(sl_out_dir, study,  preprocessed_data):
...
+              help='Comma seperated list of analyses to add')
+def add_analysis(portal, analyses):
+    ids = map(int, analyses.split(','))
+    Portal(portal).add_analyses(ids)
```
ElDeveloper: For these command line interfaces, you probably want to catch the possible exceptions and raise one of Click's [exceptions](http://click.pocoo.org/4/exceptions/#which-exceptions-exist). That way the traceback is avoided and a message is just printed out :smiley:

ElDeveloper:
@squirrelo, looks good so far, I added a few comments!

qiita_db/portal.py
squirrelo: Aliases can quickly get really long and messy, especially if a long list of studies are already attached to the portal (Think trying to delete EMP or QIITA currently). This error will only be seen by admins on the command line or in the system error logging, so having IDs is probably fine. Will reword the error message though.

qiita_db/portal.py
ElDeveloper: I mean:

map(str, (set(studies).difference(existing))) -> map(str, set(studies).difference(existing))

qiita_db/portal.py
ElDeveloper: Right on, if it's not a concern to anyone else, it should be good to
leave as IDs :+1:.

qiita_db/portal.py
squirrelo: oh, yeah. Removed.

qiita_db/portal.py
squirrelo: people have taken `list of int` to mean any iterator of ints in the past. I wanted to be explicit with the tuple because then the function will work with any iterator, since psycopg2 is the one requiring a tuple.

scripts/qiita
squirrelo: The full command is `qiita db portal add-analyses --portal lolercoaster --analyses 1,2,3` So not sure if name would clarify anything in this case. @biocore/qiita-dev thoughts?

scripts/qiita
squirrelo: I can, but that seems messy to me. If we have 30 analyses/studies to add, one comma list seems better to me than "-a 1 -a 2 -a 3 -a 4 -a 5 -a 6 -a 7 -a 8 ....."

scripts/qiita
squirrelo: I didn't know click did that. That's pretty cool!

qiita_db/portal.py
ElDeveloper: Hmmm, that's wrong then. List of int means something very specific, not
an iterable. If you decide to not take a tuple, I would at least change
this piece of documentation to ask for an iterable of int.

squirrelo: Can do.

scripts/qiita
ElDeveloper: Unless someone else thinks that's atrocious, I would do it. Plus you can
set the type as Click.INT (or maybe even Click.Range, to avoid negative
numbers)and all the identifiers will be validated (which is also nice).

squirrelo: I think a good compromise is using `nargs=-1`. This should allow `-a 1 2 3 4 5 6 7 8 9 10 ....` so you don't need the -a after each one.

ElDeveloper: Seems good!

squirrelo: ok, so with `nargs=-1` the command looks like: 

`qiita db portal add-analyses -p lolercoaster 1 2 3 4 5 6 7 8 9 ...

squirrelo: ok, so with `nargs=-1` the command looks like: 

`qiita db portal add-analyses -p lolercoaster 1 2 3 4 5 6 7 8 9 ...`

ElDeveloper: No `-a`?? Yargh, if that's the case then it would make more sense to have that as an argument and not an option ...

squirrelo: nargs only works for arguments, not options.

ElDeveloper: Got it, did not know that. Maybe both options should become arguments?
...

squirrelo: Doable. and with the portal editing already implied in the command it makes sense.

squirrelo:
This is ready for another round of review.

qiita_db/portal.py
ElDeveloper: ping ...

qiita_db/test/test_portal.py
```diff
@@ -1,38 +1,117 @@
...
-            with self.assertRaises(ValueError):
-                remove_study_from_portal(self.study, 'QIITA')
-
-            add_study_to_portal(self.study, 'EMP')
-            remove_study_from_portal(self.study, 'EMP')
-            obs = self.study._portals
-            self.assertEqual(obs, ['QIITA'])
+
+@qiita_test_checker()
+class TestPortal(TestCase):
```
ElDeveloper: I missed this before, but can you add tests for the methods that raise warnings?

scripts/qiita
```diff
@@ -289,6 +294,49 @@ def update_preprocessed_data(sl_out_dir, study,  preprocessed_data):
                                             preprocessed_data)
     click.echo("Preprocessed data %s successfully updated" % ppd.id)
 
+# #############################################################################
+# PORTAL COMMANDS
+# #############################################################################
+
+
+@portal.command(name="add-analyses")
+@click.argument('portal', required=True, type=str)
```
ElDeveloper: Just realized ... this argument should be a Choice type, because we know the name of all the portals and it should really only be one of those available.

squirrelo: Ah, good call.

qiita_db/portal.py
squirrelo: Missed that. Fixed now.

qiita_db/portal.py
ackermag: I have the EMP list and will update the Sloan list

On Wed, Jul 1, 2015 at 5:01 AM, Antonio Gonzalez <notifications@github.com>
wrote:

> In qiita_db/portal.py
> <https://github.com/biocore/qiita/pull/1316#discussion_r33671473>:
>
> >          """
> > -        portal_id = convert_to_id(portal, 'portal_type', 'portal')
> > +        self._check_studies(studies)
> > +
> > +        # Clean list of studies down to ones not associated with portal already
> > +        sql = """SELECT study_id from qiita.study_portal
> > +                 WHERE portal_type_id != %s AND study_id IN %s"""
> > +        clean_studies = [x[0] for x in conn_handler.execute_fetchall(
> > +                         sql, [self._id, tuple(studies)])]
> > +
> > +        if len(clean_studies) != len(studies):
> > +            rem = map(str, set(studies).difference(clean_studies))
> > +            warnings.warn("The following studies area already part of %s: %s" %
>
> @ackermag <https://github.com/ackermag> please make a list of which
> studies should be in another portal. We know that all go to qiita but we
> need a list of which go to SloanBuiltEnvironment and EMP (other?).
>
> —
> <https://github.com/biocore/qiita/pull/1316/files#r33671473>.
>



-- 
Gail Ackermann
Knight Lab
UCSD

qiita_db/test/test_portal.py
squirrelo: Added

squirrelo:
Comments addressed, ready for another go.

scripts/qiita-env
```diff
@@ -155,5 +157,28 @@ def job_details(job):
             click.echo("%s : %s" % (k, value))
 
 
+@env.command(name="create-portal")
+@click.option('--portal', '-p', required=True, type=str,
```
ElDeveloper: For consistency, I think these should also become arguments.

scripts/qiita-env
```diff
@@ -155,5 +157,28 @@ def job_details(job):
...
+def add_portal(portal, desc):
+    """Creates a new portal on the database"""
+    try:
+        Portal.create(portal, desc)
+    except QiitaDBDuplicateError:
+        raise click.BadParameter("Portal name already exists!")
+
+
+@env.command(name="remove-portal")
+@click.option('--portal', '-p', required=True, type=str,
```
ElDeveloper: Can this also be a choice type?

squirrelo: Tried, but since this is qiita-env, you can possibly have a broken database when calling the commands, so having `Portal.list_portals()` can throw an error if database is not set up correctly/test database is not in a consistent state. Found this out when I called `qiita-env clean_test`

ElDeveloper: ahhh true, forgot about that!

PR #1320, opened 2015-07-02
URGENT: Transaction object

josenavas:
This PR adds the Transaction object.

The Transaction object now creates it's own connection. It also includes the execute_fetchlast operation. Test for that new function have been added and a couple of changes in the commit/rollback functions have been applied. Otherwise, the code is pretty much the same with the previous version of the object.
It sets up the system with the singleton pattern for the transaction.

It will be great if I can have a fast review as this object is the central change for any further changes.

### DIFF
--- a/qiita_db/environment_manager.py
+++ b/qiita_db/environment_manager.py
@@ -19,7 +19,7 @@ from future.utils import viewitems
 
 from qiita_core.exceptions import QiitaEnvironmentError
 from qiita_core.qiita_settings import qiita_config
-from .sql_connection import SQLConnectionHandler, transaction
+from .sql_connection import SQLConnectionHandler, TRN
 from .reference import Reference
 from natsort import natsorted
 
@@ -251,7 +251,7 @@ def drop_environment(ask_for_confirmation):
     """
     # The transaction has an open connection to the database, so we need
     # to close it in order to drop the environment
-    transaction.close()
+    TRN.close()
     # Connect to the postgres server
     conn = SQLConnectionHandler()
     settings_sql = "SELECT test FROM settings"
--- a/qiita_db/environment_manager.py
+++ b/qiita_db/environment_manager.py
@@ -19,7 +19,7 @@ from future.utils import viewitems
 
 from qiita_core.exceptions import QiitaEnvironmentError
 from qiita_core.qiita_settings import qiita_config
-from .sql_connection import SQLConnectionHandler
+from .sql_connection import SQLConnectionHandler, transaction
 from .reference import Reference
 from natsort import natsorted
 
@@ -249,6 +249,9 @@ def make_environment(load_ontologies, download_reference, add_demo_user):
 def drop_environment(ask_for_confirmation):
     """Drops the database specified in the configuration
     """
+    # The transaction has an open connection to the database, so we need
+    # to close it in order to drop the environment
+    transaction.close()
     # Connect to the postgres server
     conn = SQLConnectionHandler()
     settings_sql = "SELECT test FROM settings"
--- a/qiita_db/sql_connection.py
+++ b/qiita_db/sql_connection.py
@@ -14,6 +14,7 @@ Classes
    :toctree: generated/
 
    SQLConnectionHandler
+   Transaction
 
 Examples
 --------
@@ -79,7 +80,7 @@ conn_handler.execute_fetchall(
 from __future__ import division
 from contextlib import contextmanager
 from itertools import chain
-from functools import partial
+from functools import partial, wraps
 from tempfile import mktemp
 from datetime import date, time, datetime
 import re
@@ -88,7 +89,8 @@ from psycopg2 import (connect, ProgrammingError, Error as PostgresError,
                       OperationalError)
 from psycopg2.extras import DictCursor
 from psycopg2.extensions import (
-    ISOLATION_LEVEL_AUTOCOMMIT, ISOLATION_LEVEL_READ_COMMITTED)
+    ISOLATION_LEVEL_AUTOCOMMIT, ISOLATION_LEVEL_READ_COMMITTED,
+    TRANSACTION_STATUS_IDLE)
 
 from qiita_core.qiita_settings import qiita_config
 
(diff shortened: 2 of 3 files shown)


qiita_db/environment_manager.py
```diff
@@ -249,6 +249,9 @@ def make_environment(load_ontologies, download_reference, add_demo_user):
 def drop_environment(ask_for_confirmation):
     """Drops the database specified in the configuration
     """
+    # The transaction has an open connection to the database, so we need
+    # to close it in order to drop the environment
+    transaction.close()
```
ElDeveloper: Since this is altering the system, it may be worth adding a note in the docstring.

qiita_db/sql_connection.py
```diff
@@ -630,3 +632,381 @@ def get_temp_queue(self):
...
+
+    A transaction is defined by a series of consecutive queries that need to
+    be applied to the database as a single block.
+
+    Notes
+    -----
+    When the execution leaves the context manager, any remaining queries in
+    the transaction will be executed and committed.
+    The Transaction methods can only be executed inside a context, if they are
+    invoked outside a context, a RuntimeError is raised.
```
ElDeveloper: I'm on the fence on whether this should be part of the Notes section or of the Raises section.

qiita_db/sql_connection.py
```diff
@@ -630,3 +632,381 @@ def get_temp_queue(self):
         self.create_queue(temp_queue_name)
 
         return temp_queue_name
+
+
+def _checker(func):
```
ElDeveloper: :ok_hand:

qiita_db/sql_connection.py
```diff
@@ -630,3 +632,381 @@ def get_temp_queue(self):
...
+
+        Raises
+        ------
+        RuntimeError
+            If invoked outside a context
+
+        Notes
+        -----
+        If any exception occurs during the execution transaction, a rollback
+        is executed an no changes are reflected in the database.
```
ElDeveloper: an -> and

qiita_db/sql_connection.py
```diff
@@ -630,3 +632,381 @@ def get_temp_queue(self):
...
+            If invoked outside a context
+        """
+        self._connection.rollback()
+        # Reset the queries, the results and the index
+        self._queries = []
+        self._results = []
+        self.index = 0
+
+# Singleton pattern, create the transaction for the entire system
+transaction = Transaction()
```
ElDeveloper: A mention of this should be made at the module level docstring.

qiita_db/sql_connection.py
```diff
@@ -630,3 +632,381 @@ def get_temp_queue(self):
...
+        execute_fetchlast
+        """
+        try:
+            return self._execute()
+        except Exception:
+            self.rollback()
+            raise
+
+    @_checker
+    def execute_fetchlast(self):
```
ElDeveloper: I like this a lot!

Later or maybe right now, we should add:

```python
def execute_fetchcolumn(self, idx=0):
    '''Returns the column in `idx`'''
    # code similar to what we do when we "flatten"
    pass

```

qiita_db/environment_manager.py
ElDeveloper: Also since this changes the way the module behaves, I think we should probably re-open the transaction after an environment is created (line 246). What I'm imagining is a scenario where you do:

drop_environment()

# now let's bring back everything up ....
make_environment()

ElDeveloper:
Overall it makes sense and I think it will play nicely. One thing that I don't particularly like is that `transaction` looks like any other variable even though it's a rather important one, some alternatives I was considering were:

- `T`.
- `TR`.
- `Trx`.

Aside from that, I think most of the code looks fine! :+1:

wasade:
what about `_o0OtrxO0o_`?

qiita_db/sql_connection.py
```diff
@@ -630,3 +632,381 @@ def get_temp_queue(self):
...
+        ------
+        RuntimeError
+            if the cursor cannot be created
+        """
+        self._open_connection()
+
+        try:
+            with self._connection.cursor(cursor_factory=DictCursor) as cur:
+                yield cur
+        except PostgresError as e:
```
wasade: Will `PostgresError` catch the other error types that may be thrown (e.g., a malformed query being executed)?

qiita_db/sql_connection.py
```diff
@@ -630,3 +632,381 @@ def get_temp_queue(self):
...
+
+    def __exit__(self, exc_type, exc_value, traceback):
+        # We only need to perform some action if this is the last context
+        # that we are entering
+        if self._contexts_entered == 1:
+            # We need to wrap the entire function in a try/finally because
+            # at the end we need to decrement _contexts_entered
+            try:
+                self._clean_up(exc_type)
+            finally:
```
wasade: This could be redone slightly and reduce a slight bit of code replication:

```python
if self._contexts_entered == 1:
    try:
        self._clean_up(exc_type)
    except:
        pass

self._contexts_entered -= 1
```

qiita_db/sql_connection.py
```diff
@@ -630,3 +632,381 @@ def get_temp_queue(self):
...
+            # It is safe to use the execute method here, as internally is
+            # wrapped in a try/except and rollbacks in case of failure
+            self.execute()
+            self.commit()
+        elif self._connection.get_transaction_status() != \
+                TRANSACTION_STATUS_IDLE:
+            # There are no queries to be executed, however, the transaction
+            # is still not committed. Commit it so the changes are not lost
+            self.commit()
+
```
wasade: If an `else` block existed, would it ever get triggered?

qiita_db/sql_connection.py
```diff
@@ -630,3 +632,381 @@ def get_temp_queue(self):
...
+        See Also
+        --------
+        execute
+        """
+        return self.execute()[-1][0][0]
+
+    @_checker
+    def commit(self):
+        """Commits the transaction and reset the queries
+        Raises
```
wasade: Space please, same with `rollback`

qiita_db/sql_connection.py
```diff
@@ -630,3 +632,381 @@ def get_temp_queue(self):
...
+
+        for args in sql_args:
+            if args:
+                if not isinstance(args, list):
+                    raise TypeError("sql_args should be a list. Found %s"
+                                    % type(args))
+            else:
+                args = []
+            self._queries.append((sql, args))
+            self.index += 1
```
wasade: I can't see where `index` is used other than being set to `0` in the constructor and reset in `commit` and `rollback`. Is this public variable intended for the caller when forming placeholders?

wasade: ...isn't `self.index == len(self._queries)` at all times? If so, what do you think about making `index` a property that returns `len(self._queries)`?

qiita_db/sql_connection.py
```diff
@@ -630,3 +632,381 @@ def get_temp_queue(self):
...
+                except PostgresError as e:
+                    # Some other error happened during the execution of the
+                    # query, so we need to rollback
+                    self._raise_execution_error(sql, sql_args, e)
+
+                # Store the results of the current query
+                self._results.append(res)
+
+        # wipe out the already executed queries
+        self._queries = []
```
wasade: Should `index` be reset here?

qiita_db/sql_connection.py
```diff
@@ -630,3 +632,381 @@ def get_temp_queue(self):
...
+
+    @_checker
+    def commit(self):
+        """Commits the transaction and reset the queries
+        Raises
+        ------
+        RuntimeError
+            If invoked outside a context
+        """
+        self._connection.commit()
```
wasade: it is possible that this line will except (same with 1005). Given this potential, and that calls to `Transaction.commit` are wrapped in a try/except caught, and that the singleton `transaction` is to be reused through the lifetime of the process, should the member variables (`_queries`, `_results`, `index`) be reset prior to these possibly exceptional(?) calls?

wasade:
Few comments, otherwise this looks awesome

ElDeveloper:
Better yet, what about: ':train2::saxophone:' (trainsax) :P

wasade:
because python isn't a hipster like julia

ElDeveloper:
We should rewrite Qiita in Julia!

josenavas:
ehm.. don't count me in for that rewrite...

After a bit of googling it looks like `TRN` is a good abbreviation. Also I will use all caps since it is a global variable.

qiita_db/environment_manager.py
josenavas: @ElDeveloper I don't think it is needed. The Transaction object re-opens the connection if it has been closed once you enter into the context (see [here](https://github.com/biocore/qiita/pull/1320/files#diff-29a37d36652014d33c798df45a644b46R738))

qiita_db/sql_connection.py
josenavas: You're probably right. Changing.

qiita_db/environment_manager.py
ElDeveloper: That's true, cool!

qiita_db/sql_connection.py
josenavas: You're right, but every time that we use `cur.execute` it is wrapped in its own `try/except` so that type of errors will never reach here. Through the code, the only postgres erros that can be caught here are the ones generated by `self._connection.cursor(cursor_factory=DictCursor)`. Note that this function is private and it is only used in `_execute`.

wasade: If `self._connection.cursor(cursor_factory=DictCursor)` is capable of throwing any other type of exception (or at least that we expect may happen), then an uncaught exception will be propagated through the context on line 899. 

One possibility, though I think it is extremely unlikely given the design of `SQLConnection` is an `InterfaceError` which can result if `cursor` is called on a closed connection. In this case, `PostgresError` is a superclass, so it would still be caught.

I don't know if a finer grain of exception handling is necessary, just inquiring. I think this is all good. I'm trying to really enumerate different potential issues here as this object has a central role, and because of the singleton pattern, it is expected to operate correctly and tolerate error through out the life of the process.

josenavas: Although the line 899 can propagate any other type of exception, the internal function `_execute` (where the line 899 is) is only called from the `execute` function in line 960, wrapping it in a try/except with a general Exception so it will catch everything, rollback the transaction and propagate the exception.

qiita_db/sql_connection.py
josenavas: The only situation in which the else situation will be triggered is if ` self._connection.get_transaction_status() == TRANSACTION_STATUS_IDLE`, in such case, we don't want to do anything (the transaction is already idle so anything is blocked in the DB).

qiita_db/sql_connection.py
josenavas: Your solution will mask the exception raised in `self._clean_up` while the current solution will propagate the exception but it will update `self._contexts_entered`. Does it make sense?

qiita_db/sql_connection.py
wasade: Thanks!

qiita_db/sql_connection.py
josenavas: You're correct, the index is intended for the caller to form the placeholders. However, it is not always `len(self._queries)`. The reason is transactions can be partially executed but not committed. When that happens, we reset `self._queries` so we don't re-execute them, but `self._results` is kept, and self.index is the one who keeps track of how many queries have been inserted (which is not the same as executed).

The only reason why I think it will make sense to have it as  property is to avoid the consumer of this to modify it. What do you think?

qiita_db/sql_connection.py
wasade: Yes, I see that now, thanks.

I've been staring at the code trying to think of every way (short of a developer being "creative") to get the `contexts_entered` out of sync. I can't identify one -- awesome :)

josenavas: Awesome! Those are the best reviews!!

qiita_db/sql_connection.py
josenavas: Nope, since the transaction has not been committed, so the previous results are still available.

qiita_db/sql_connection.py
josenavas: Done!

qiita_db/sql_connection.py
wasade: That makes sense, thanks. It sounds like `index == len(self._queries) +
len(self._results)` at any given time then?

Not tooo concerned about someone setting `index` but agree that would be
one way to mitigate

qiita_db/sql_connection.py
josenavas: Uhm... right now we are not "flattening" results in here, that is why I have an extra index...
I will also be surprised of a use case for that. If you're interested in a single column value, why didn't you format your SQL to return only that value?

qiita_db/sql_connection.py
josenavas: Done!

qiita_db/sql_connection.py
josenavas: You're probably right, although if those calls fail, the transaction will not be left IDLE, so it is blocking the DB... If these calls fail we have a bigger problem that I don't know how to solve, since the way of resetting the transaction is doing a rollback or a commit. A potential solution will be forcing to close the connection as it internally will do a rollback, but not sure if in that such-a-hard-case-to-test this will work. Worst case scenario, these exceptions are propagated and I don't think they will be caught anywhere, which I think is a desirable behavior if this case arises.

qiita_db/sql_connection.py
josenavas: Done!

qiita_db/sql_connection.py
wasade: Should this then be wrapped in a try/except? From some light googling, it does appear there are cases where a commit can not go through, but I confess I'm not familiar enough here to know if the probability here is nonzero

qiita_db/sql_connection.py
josenavas: You're right! Changed.

qiita_db/sql_connection.py
josenavas: Yup, I think a try/except is need. I'm closing the connection in the the `except` clause and re-raising the error, as I think we don't want to mask these errors...

josenavas:
Thanks @ElDeveloper and @wasade 
I think I've addressed/answered to all the comments!

PR #1322, opened 2015-07-02
Portals object removal fix

squirrelo:
There was some checks missing in the tests and therefore an issue was not caught. If an analysis is created in the portal, it will be unlinked. Also, selected samples for a default analysis will now be deleted properly.

### DIFF
--- a/qiita_db/portal.py
+++ b/qiita_db/portal.py
@@ -139,7 +139,7 @@ class Portal(QiitaObject):
                 " Cannot delete portal '%s', analyses still attached: %s" %
                 (portal, ', '.join([str(a[0]) for a in analyses])))
 
-        # Remove portal and default analyses for all users
+        # Remove portal and default analyses for all users, unlink others
         sql = """DO $do$
             DECLARE
                 pid bigint;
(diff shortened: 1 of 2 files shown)


josenavas:
Uhm... I'm not sure if I understand the issue.
The issue is that there are analyses attached to the portal? Shouldn't this be an error rather than unlink them blindly? We are not doing this in any other removal.

squirrelo:
I can make it an error.

josenavas:
Yeah it will be better for consistency. And safer, since this is a "delete"...

qiita_db/portal.py
```diff
@@ -139,7 +139,7 @@ def delete(portal):
                 " Cannot delete portal '%s', analyses still attached: %s" %
                 (portal, ', '.join([str(a[0]) for a in analyses])))
 
-        # Remove portal and default analyses for all users
+        # Remove portal and default analyses for all users, unlink others
```
ElDeveloper: Minor comment, but I just noticed that `pid` is being used anywhere in the SQL below.

ElDeveloper:
I think this looks good, though I don't really understand the problem.

squirrelo:
OK, fixed the tests and caught another bug. Should be good to go now.

josenavas:
:+1:

PR #1323, opened 2015-07-02
Fixing base and user objects

josenavas:
This fixes the base and user object to use the transaction.

Since this is the first time the transaction object is used in the code base, I will appreciate if people can take a quick look to see that everything works as expected.

Thanks!

### DIFF
--- a/qiita_db/base.py
+++ b/qiita_db/base.py
@@ -27,7 +27,7 @@ Classes
 
 from __future__ import division
 from qiita_core.exceptions import IncompetentQiitaDeveloperError
-from .sql_connection import SQLConnectionHandler
+from .sql_connection import transaction
 from .exceptions import QiitaDBNotImplementedError, QiitaDBUnknownIDError
 
 
@@ -130,11 +130,12 @@ class QiitaObject(object):
         """
         self._check_subclass()
 
-        conn_handler = SQLConnectionHandler()
-
-        return conn_handler.execute_fetchone(
-            "SELECT EXISTS(SELECT * FROM qiita.{0} WHERE "
-            "{0}_id=%s)".format(self._table), (id_, ))[0]
+        with transaction:
+            sql = """SELECT EXISTS(
+                        SELECT * FROM qiita.{0}
+                        WHERE {0}_id=%s)""".format(self._table)
+            transaction.add(sql, [id_])
+            return transaction.execute_fetchlast()
 
     def __init__(self, id_):
         r"""Initializes the object
@@ -148,8 +149,9 @@ class QiitaObject(object):
         QiitaDBUnknownIDError
             If `id_` does not correspond to any object
         """
-        if not self._check_id(id_):
-            raise QiitaDBUnknownIDError(id_, self._table)
+        with transaction:
+            if not self._check_id(id_):
+                raise QiitaDBUnknownIDError(id_, self._table)
 
         self._id = id_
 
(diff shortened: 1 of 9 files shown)


qiita_db/base.py
```diff
@@ -130,11 +130,12 @@ def _check_id(self, id_):
...
-        conn_handler = SQLConnectionHandler()
-
-        return conn_handler.execute_fetchone(
-            "SELECT EXISTS(SELECT * FROM qiita.{0} WHERE "
-            "{0}_id=%s)".format(self._table), (id_, ))[0]
+        with TRN:
+            sql = """SELECT EXISTS(
+                        SELECT * FROM qiita.{0}
+                        WHERE {0}_id=%s)""".format(self._table)
+            TRN.add(sql, [id_])
```
ElDeveloper: I am under the impression that these had to be tuples and that lists were interpreted in a very different way.

qiita_db/base.py
```diff
@@ -256,21 +256,20 @@ def check_status(self, status, exclude=False):
...
-                             % set(status).difference(statuses))
-
-        # Get the DB status of the object
-        dbstatus = self.status
-        return dbstatus not in status if exclude else dbstatus in status
+        with TRN:
+            # Get all available statuses
+            sql = "SELECT DISTINCT status FROM qiita.{0}_status".format(
+                self._table)
+            TRN.add(sql, [self._id])
```
ElDeveloper: What is the `[self._id]` doing here? Doesn't seem to be put anywhere in the SQL statement.

qiita_db/base.py
```diff
@@ -256,21 +256,20 @@ def check_status(self, status, exclude=False):
...
+            # We need to access to the results of the last SQL query,
+            # hence indexing using -1
+            avail_status = [x[0] for x in TRN.execute()[-1]]
```
ElDeveloper: This is what I was talking about in your other PR, having the transaction be able to do: `TRN.execute_fetchcolumn()` and then you get the index passed ... does that make sense?

ElDeveloper:
This looks good overall, very simple changes and I like that you are reformatting the SQL to be structured in a consistent way, perhaps you should take some notes and add a section in the CONTRIBUTING.md file. Also minor note, would perhaps be good to have all SQL keywords be all-caps.

Other reviewers, these are very simple changes so it's quick to go through.

qiita_db/base.py
josenavas: Nope, here we just need to pass an iterable so psycopg doesn't fail. It is weird but psycopg always requires an iterable as sql_args, even if there is a single element. Also, the transaction object forces this to be lists so it can do the placeholder replacements.

qiita_db/base.py
josenavas: ooops good call!

qiita_db/base.py
josenavas: Uhm, the difference is that here is not returning a column, here is returning the results of the last query, which is a list of lists... An option might be `def execute_fetchindex(index=-1)` which defaults always to the last one...

qiita_db/base.py
ElDeveloper: In all fairness, this was there already ...

qiita_db/base.py
ElDeveloper: Got it, yes, I see two benefits from having something like that, (1)
centralized documentation on what you are doing and (2) avoids code
duplication :beers:!

josenavas: Ok, will add that function!

ElDeveloper: Thanks!

josenavas:
@ElDeveloper comments addressed!

josenavas:
Re: CONTRIBUTING.md, I've added another item to https://github.com/biocore/qiita/pull/1262

qiita_db/user.py
```diff
@@ -126,31 +125,33 @@ def login(cls, email, password):
...
-            raise IncorrectPasswordError("Password not valid!")
-
-        # pull password out of database
-        conn_handler = SQLConnectionHandler()
-        sql = ("SELECT password, user_level_id FROM qiita.{0} WHERE "
-               "email = %s".format(cls._table))
-        info = conn_handler.execute_fetchone(sql, (email, ))
-
-        # verify user email verification
-        # MAGIC NUMBER 5 = unverified email
```
squirrelo: Not your fault, but could you change this to use the new convert_to_id() function?

squirrelo:
:+1:

ElDeveloper:
:+1:

PR #1318, opened 2015-07-01
WIP: Transaction

josenavas:
This replaces #1262 as the amount of code that needs to be fixed there is too big.

We will create the `Transaction` object and use the singleton pattern across the system. This will work as long as our system continues being single-threads. As long as we have parallel processes in their own python interpreter this will not become an issue.

 - [x] Create the new `Transaction` object ~~(https://github.com/biocore/qiita/pull/1320)~~
 - [x] Fix python patches (https://github.com/biocore/qiita/pull/1337)
 - [x] Fix `__init__.py` (n/a)
 - [x] Fix `analysis.py` (https://github.com/biocore/qiita/pull/1332)
 - [x] Fix `base.py` ~~(https://github.com/biocore/qiita/pull/1323)~~
 - [x] Fix `commands.py` (https://github.com/biocore/qiita/pull/1335)
 - [x] Fix `data.py` ~~(https://github.com/biocore/qiita/pull/1327)~~
 - [x] Fix `environment_manager.py`(https://github.com/biocore/qiita/pull/1337)
 - [x] Fix `exceptions.py`(n/a)
 - [x] Fix `investigation.py` (n/a)
 - [x] Fix `job.py` (https://github.com/biocore/qiita/pull/1332)
 - [x] Fix `logger.py` (https://github.com/biocore/qiita/pull/1334)
 - [x] Fix `meta_util.py` (https://github.com/biocore/qiita/pull/1334)
 - [x] Fix `metadata_template/*` (https://github.com/biocore/qiita/pull/1326)
 - [x] Fix `ontology.py` (https://github.com/biocore/qiita/pull/1334)
 - [x] Fix `parameters.py` (https://github.com/biocore/qiita/pull/1334)
 - [x] Fix `reference.py` (https://github.com/biocore/qiita/pull/1333)
 - [x] Fix `search.py` (https://github.com/biocore/qiita/pull/1335)
 - [x] Fix `study.py` (https://github.com/biocore/qiita/pull/1333)
 - [x] Fix `user.py` ~~(https://github.com/biocore/qiita/pull/1323)~~
 - [x] Fix `util.py` ~~(https://github.com/biocore/qiita/pull/1324)~~
 - [x] Fix `portals.py` (https://github.com/biocore/qiita/pull/1340)
 - [ ] Add mention to the changelog (https://github.com/biocore/qiita/pull/1357)
 - [x] Add notes to CONTRIBUTING.md (https://github.com/biocore/qiita/issues/1355)
[...]

PR #1324, opened 2015-07-03
URGENT: Fix util

josenavas:
This PR fixes the util.py file to use transactions.
Since some function signatures changed, I had to fix other parts of the code to make the tests to pass.
Any future PR will be based on this one, hence that is why it is urgent (the other ones will be in parallel, but they'll rely on the changes in util.py).

Some comments:
 - `purge_filepaths` is tricky. It is not used anywhere but it has the potential of leaving the DB out of sync with the FS if it is executed inside a bigger transaction. I haven't put any check on it, but one option will be to make it use it's own transaction. Since each Transaction object has it's own connection, this will not be an issue. However, I'd like to hear what others think.
 - `get_preprocessed_params_tables` was missing tests and it was incorrect.
 - `find_repeated` was not used anywhere, I removed it as we can use skbio's `find_duplicates`

### DIFF
--- a/qiita_db/sql_connection.py
+++ b/qiita_db/sql_connection.py
@@ -1021,6 +1021,8 @@ class Transaction(object):
                 cmd()
             except Exception as e:
                 error_msg.append(str(e))
+        # The functions in these two lines are mutually exclusive. When one of
+        # them is executed, we can restore both of them.
         self._post_commit_funcs = []
         self._post_rollback_funcs = []
         if error_msg:
--- a/qiita_db/sql_connection.py
+++ b/qiita_db/sql_connection.py
@@ -674,6 +674,8 @@ class Transaction(object):
         self._results = []
         self._contexts_entered = 0
         self._connection = None
+        self._post_commit_funcs = []
+        self._post_rollback_funcs = []
 
     def _open_connection(self):
         # If the connection already exists and is not closed, don't do anything
@@ -1012,6 +1014,20 @@ class Transaction(object):
         """
         return self.execute()[idx]
 
+    def _cmds_executor(self, cmds, func_str):
+        error_msg = []
+        for cmd in cmds:
+            try:
+                cmd()
+            except Exception as e:
+                error_msg.append(str(e))
+        self._post_commit_funcs = []
+        self._post_rollback_funcs = []
+        if error_msg:
+            raise RuntimeError(
+                "An error occurred during the post %s commands:\n%s"
+                % (func_str, "\n".join(error_msg)))
+
     @_checker
     def commit(self):
         """Commits the transaction and reset the queries
@@ -1029,6 +1045,8 @@ class Transaction(object):
         except Exception:
             self._connection.close()
             raise
+        # Execute the post commit commands
+        self._cmds_executor(self._post_commit_funcs, "commit")
 
     @_checker
     def rollback(self):
@@ -1047,10 +1065,44 @@ class Transaction(object):
         except Exception:
             self._connection.close()
             raise
+        # Execute the post rollback commands
+        self._cmds_executor(self._post_rollback_funcs, "rollback")
 
     @property
     def index(self):
         return len(self._queries) + len(self._results)
+
+    @_checker
+    def add_post_commit_func(self, func):
+        """Adds a post commit command
+
+        Parameters
+        ----------
+        func : function
+            The function to add for the post commit commands
+
+        Notes
+        -----
+        func should not accept any parameter, i.e. it should allow to be
+        invoked as `func()`
+        """
+        self._post_commit_funcs.append(func)
+
+    @_checker
+    def add_post_rollback_func(self, func):
+        """Adds a post rollback command
+
+        Parameters
+        ----------
+        func : function
+            The function to add for the post rollback commands
+
+        Notes
+        -----
+        func should not accept any parameter, i.e. it should allow to be
+        invoked as `func()`
+        """
+        self._post_rollback_funcs.append(func)
 
 # Singleton pattern, create the transaction for the entire system
 TRN = Transaction()
--- a/qiita_db/sql_connection.py
+++ b/qiita_db/sql_connection.py
@@ -963,6 +963,7 @@ class Transaction(object):
         --------
         execute_fetchlast
         execute_fetchindex
+        execute_fetchflatten
         """
         try:
             return self._execute()
@@ -986,6 +987,7 @@ class Transaction(object):
         --------
         execute
         execute_fetchindex
+        execute_fetchflatten
         """
         return self.execute()[-1][0][0]
 
@@ -1011,14 +1013,42 @@ class Transaction(object):
         --------
         execute
         execute_fetchlast
+        execute_fetchflatten
         """
         return self.execute()[idx]
 
-    def _funcs_executor(self, cmds, func_str):
+    @_checker
+    def execute_fetchflatten(self, idx=-1):
+        """Executes the transaction and returns the flattened results of the
+        `idx` query
+
+        This is a convenient function that is equivalen to
+        `chain.from_iterable(self.execute()[idx])`
+
+        Parameters
+        ----------
+        idx : int, optional
+            The index of the query to return the result. It defaults to -1, the
+            last query.
+
+        Returns
+        -------
+        list of objects
+            The flattened results of the `idx` query
+
+        See Also
+        --------
+        execute
+        execute_fetchlast
+        execute_fetchindex
+        """
+        return list(chain.from_iterable(self.execute()[idx]))
+
+    def _funcs_executor(self, funcs, func_str):
         error_msg = []
-        for cmd in cmds:
+        for f in funcs:
             try:
-                cmd()
+                f()
             except Exception as e:
                 error_msg.append(str(e))
         # The functions in these two lines are mutually exclusive. When one of
--- a/qiita_db/sql_connection.py
+++ b/qiita_db/sql_connection.py
@@ -1046,9 +1046,9 @@ class Transaction(object):
 
     def _funcs_executor(self, funcs, func_str):
         error_msg = []
-        for f in funcs:
+        for f, args, kwargs in funcs:
             try:
-                f()
+                f(*args, **kwargs)
             except Exception as e:
                 error_msg.append(str(e))
         # The functions in these two lines are mutually exclusive. When one of
@@ -1105,7 +1105,7 @@ class Transaction(object):
         return len(self._queries) + len(self._results)
 
     @_checker
-    def add_post_commit_func(self, func):
+    def add_post_commit_func(self, func, *args, **kwargs):
         """Adds a post commit function
 
         The function added will be executed after the next commit in the
@@ -1117,16 +1117,15 @@ class Transaction(object):
         ----------
         func : function
             The function to add for the post commit functions
-
-        Notes
-        -----
-        func should not accept any parameter, i.e. it should allow to be
-        invoked as `func()`
+        args : tuple
+            The arguments of the function
+        kwargs : dict
+            The keyword arguments of the function
         """
-        self._post_commit_funcs.append(func)
+        self._post_commit_funcs.append((func, args, kwargs))
 
     @_checker
-    def add_post_rollback_func(self, func):
+    def add_post_rollback_func(self, func, *args, **kwargs):
         """Adds a post rollback function
 
         The function added will be executed after the next rollback in the
@@ -1138,13 +1137,12 @@ class Transaction(object):
         ----------
         func : function
             The function to add for the post rollback functions
-
-        Notes
-        -----
-        func should not accept any parameter, i.e. it should allow to be
-        invoked as `func()`
+        args : tuple
+            The arguments of the function
+        kwargs : dict
+            The keyword arguments of the function
         """
-        self._post_rollback_funcs.append(func)
+        self._post_rollback_funcs.append((func, args, kwargs))
 
 # Singleton pattern, create the transaction for the entire system
 TRN = Transaction()
--- a/qiita_db/sql_connection.py
+++ b/qiita_db/sql_connection.py
@@ -1014,7 +1014,7 @@ class Transaction(object):
         """
         return self.execute()[idx]
 
-    def _cmds_executor(self, cmds, func_str):
+    def _funcs_executor(self, cmds, func_str):
         error_msg = []
         for cmd in cmds:
             try:
@@ -1045,8 +1045,8 @@ class Transaction(object):
         except Exception:
             self._connection.close()
             raise
-        # Execute the post commit commands
-        self._cmds_executor(self._post_commit_funcs, "commit")
+        # Execute the post commit functions
+        self._funcs_executor(self._post_commit_funcs, "commit")
 
     @_checker
     def rollback(self):
@@ -1065,8 +1065,8 @@ class Transaction(object):
         except Exception:
             self._connection.close()
             raise
-        # Execute the post rollback commands
-        self._cmds_executor(self._post_rollback_funcs, "rollback")
+        # Execute the post rollback functions
+        self._funcs_executor(self._post_rollback_funcs, "rollback")
 
     @property
     def index(self):
@@ -1074,12 +1074,17 @@ class Transaction(object):
 
     @_checker
     def add_post_commit_func(self, func):
-        """Adds a post commit command
+        """Adds a post commit function
+
+        The function added will be executed after the next commit in the
+        transaction, unless a rollback is executed. This is useful, for
+        example, to perform some filesystem clean up once the transaction is
+        committed.
 
         Parameters
         ----------
         func : function
-            The function to add for the post commit commands
+            The function to add for the post commit functions
 
         Notes
         -----
@@ -1090,12 +1095,17 @@ class Transaction(object):
 
     @_checker
     def add_post_rollback_func(self, func):
-        """Adds a post rollback command
+        """Adds a post rollback function
+
+        The function added will be executed after the next rollback in the
+        transaction, unless a commit is executed. This is useful, for example,
+        to restore the filesystem in case a rollback occurs, avoiding leaving
+        the database and the filesystem in an out of sync state.
 
         Parameters
         ----------
         func : function
-            The function to add for the post rollback commands
+            The function to add for the post rollback functions
 
         Notes
         -----
--- a/qiita_db/util.py
+++ b/qiita_db/util.py
@@ -638,8 +638,8 @@ def insert_filepaths(filepaths, obj_id, table, filepath_table,
             # alrady exists on the DB
             db_path = partial(join, base_fp)
             new_filepaths = [
-                (db_path("%s_%s" % (obj_id, basename(path))), id)
-                for path, id in filepaths]
+                (db_path("%s_%s" % (obj_id, basename(path))), id_)
+                for path, id_ in filepaths]
             # Move the original files to the controlled DB directory
             for old_fp, new_fp in zip(filepaths, new_filepaths):
                     move(old_fp[0], new_fp[0])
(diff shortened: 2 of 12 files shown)


qiita_db/util.py
```diff
@@ -579,18 +590,20 @@ def get_mountpoint_path_by_id(mount_id):
...
+        TRN.add(sql, [mount_id])
+        mountpoint, subdirectory = TRN.execute_fetchindex()[0]
+        return join(get_db_files_base_dir(), mountpoint, subdirectory)
+
+
+def insert_filepaths(filepaths, obj_id, table, filepath_table,
+                     move_files=True):
+    r"""Inserts `filepaths` in the database.
+
+    Since the files live outside the database, the directory in which the files
```
antgonza: mvoes -> moves

qiita_db/util.py
```diff
@@ -604,107 +617,128 @@ def insert_filepaths(filepaths, obj_id, table, filepath_table, conn_handler,
...
-            if fp_type is 'directory':
-                rmtree(fp)
-            else:
-                remove(fp)
+    used in any place
+
+    Raises
+    ------
+    IOError
+        If any error occurs while removing the fileapths from the file system
```
antgonza: file system -> filesystem

qiita_db/util.py
```diff
@@ -604,107 +617,128 @@ def insert_filepaths(filepaths, obj_id, table, filepath_table, conn_handler,
...
+
+    Notes
+    -----
+    This function can potentially leave the DB and the file system out of
```
antgonza: file system -> filesystem

antgonza:
Minor comments. Now, what about extending the idea of transactions so we only execute code once we know the transaction is done. For example, if something fails and you are using purge_filepaths do not move them until the full transaction is done. This could also be helpful for patches. Also, the codes looks fine but I think the biggest worry is not about the changes you made but about the possible missing lines that needed changing. Is there a way to verify that things that calls to any function that has SQL are actually being run in transaction?

josenavas:
Thanks @antgonza. Once everything is changed, a quick search in the code base for SQLConnectionHandler will tell us if there is anything executed outside a transaction.

I agree with you re executing some code once the transaction is done (e.g. moving files). This is a tricky question and it really depends on the operation performed.

For example, in purge_filepaths, if the transaction successfully completes and there is a problem removing files, it is not a big deal, as the only "bad" thing is that the file system is dirty (which is easy to write a function for cleaning it up in a cron job).

However, in move_filepaths_to_uploads_folder, the function is usually executed when unlinking files from raw data. If there is a problem moving files but the transaction is committed, those files are "lost" from the user point of view: they live under the raw data folder, but we've lost the filename and the user cannot access them. In this case, you want to rollback the operation.

The only idea that comes to my mind is this:
We add a list of tuples to the Transaction object, where each tuple is of the form `[(src_path, dest_path)]`. During the code we move the files w/o a problem, but we add them to this list. If the transaction at some point performs a rollback, it is in charge of moving back those files, so the DB and the filesystem are in sync.
Also, we add 2 list of strings, which will be "paths to remove on commit" and "paths to remove on rollback". In the former, the transaction will remove those files if it commits (e.g. for purge_filepaths) and in the later it will remove those files in rollback (e.g. creating a qiime_mapping_file).This way, we can make sure that all the filepaths that are in the DB are present in the filesystem, while we are allowing the filesystem to have files that they might not be in the DB (which I think is not a big deal).

What do you think @antgonza, is this a good solution?

qiita_db/util.py

qiita_db/util.py

qiita_db/util.py

qiita_db/util.py
```diff
@@ -604,107 +617,128 @@ def insert_filepaths(filepaths, obj_id, table, filepath_table, conn_handler,
...
+        dd_id, mp = get_mountpoint(table)[0]
+        base_fp = join(get_db_files_base_dir(), mp)
+
+        if move_files:
+            # Generate the new fileapths. Format: DataId_OriginalName
+            # Keeping the original name is useful for checking if the RawData
+            # alrady exists on the DB
+            db_path = partial(join, base_fp)
+            new_filepaths = [
+                (db_path("%s_%s" % (obj_id, basename(path))), id)
```
ElDeveloper: We usually avoid `id` as a var name, but I see this was like this in the original version of the code.

qiita_db/util.py
```diff
@@ -604,107 +617,128 @@ def insert_filepaths(filepaths, obj_id, table, filepath_table, conn_handler,
...
-
-        # Remove the data
-        fp = join(get_mountpoint_path_by_id(dd_id), fp)
-        if exists(fp):
-            if fp_type is 'directory':
-                rmtree(fp)
-            else:
-                remove(fp)
+    used in any place
+
```
ElDeveloper: Thanks for improving the docstring! :+1:

qiita_db/util.py
```diff
@@ -604,107 +617,128 @@ def insert_filepaths(filepaths, obj_id, table, filepath_table, conn_handler,
...
+    Raises
+    ------
+    IOError
+        If any error occurs while removing the fileapths from the filesystem
+
+    Notes
+    -----
+    This function can potentially leave the DB and the filesystem out of
+    sync if purge_filepaths is execute inside a bigger transaction. Thus,
```
ElDeveloper: execute -> executed

qiita_db/util.py
```diff
@@ -604,107 +617,128 @@ def insert_filepaths(filepaths, obj_id, table, filepath_table, conn_handler,
...
+    care should be taken when using this function and do not include it in
+    a bigger transactions, as it can leave the database pointing to files that
```
ElDeveloper: I think this second sentence repeats the first sentence. Also if I understand correctly, I would note that if this error came up, you would likely have to the syncing yourself i.e. go through each of the files listed in the error message and make sure that the files are removed or ignored.

qiita_db/util.py
```diff
@@ -926,11 +985,13 @@ def get_processed_params_tables():
...
-    sql = ("SELECT * FROM information_schema.tables WHERE table_schema = "
-           "'qiita' AND SUBSTR(table_name, 1, 17) = 'processed_params_'")
-
-    conn = SQLConnectionHandler()
-    return sorted([x[2] for x in conn.execute_fetchall(sql)])
+    with TRN:
+        sql = """SELECT table_name FROM information_schema.tables
+                 WHERE table_schema = 'qiita'
+                    AND SUBSTR(table_name, 1, 17) = 'processed_params_'
+                 ORDER BY table_name"""
```
ElDeveloper: Hmmm, I like that we use ORDER BY instead of sorted, but are these two statements really the same? I guess if the tests didn't catch anything ... we should be good?

qiita_db/util.py
```diff
@@ -941,20 +1002,21 @@ def get_lat_longs():
...
-                AND column_name IN ('latitude', 'longitude');"""
-    tables_gen = (t[0] for t in conn.execute_fetchall(sql))
+    with TRN:
+        sql = """SELECT DISTINCT table_name
+                 FROM information_schema.columns
+                 WHERE SUBSTR(table_name, 1, 7) = 'sample_'
+                    AND table_schema = 'qiita'
+                    AND column_name IN ('latitude', 'longitude');"""
+        TRN.add(sql)
+        tables_gen = (t[0] for t in TRN.execute_fetchindex())
```
ElDeveloper: So far this pattern is repeated in a few places, can it be synthesized into a `Transaction` method or would it be overkill?

qiita_db/util.py
```diff
@@ -1025,31 +1066,30 @@ def check_access_to_analysis_result(user_id, requested_path):
...
+                 FROM qiita.analysis_job aj JOIN (
+                    SELECT analysis_id FROM qiita.analysis A
+                    JOIN qiita.analysis_status stat
+                    ON A.analysis_status_id = stat.analysis_status_id
+                    WHERE stat.analysis_status_id = 6
+                    UNION
+                    SELECT analysis_id FROM qiita.analysis_users
+                    WHERE email = %s
+                    UNION
+                    select analysis_id FROM qiita.analysis WHERE email = %s
```
ElDeveloper: select -> SELECT ... for consistency with these changes :P

ElDeveloper:
Looks good @josenavas, just a few tiny comments and a question.

qiita_db/util.py
josenavas: I am working right now on a way of adding something to the Transaction object that will make sure that all files that are in the filepath table are in the filesystem.

However, consistency in the other way around is not ensured, i.e. files that are present in the FS are not in the DB. For example, in purge_filepaths these files have been removed from the DB but for some error they're not removed from the FS.

qiita_db/util.py
josenavas: Good point, done.

qiita_db/util.py

qiita_db/util.py
josenavas: I think so, why shouldn't they?

qiita_db/util.py
josenavas: Uhmmm... not sure, we were already doing this before. What others think? I have mixed feelings.

qiita_db/util.py
josenavas: Good catch!

josenavas:
Thanks @ElDeveloper and @antgonza 

I've added some functionality to the Transaction object so we can make sure that functions like purge_filepaths do not break the DB.

qiita_db/util.py
ElDeveloper: I was thinking on edge cases with non-alphanumeric characters ... but
this is referring to table names, so I think it's all good!

qiita_db/util.py
ElDeveloper: Hmm, shouldn't be considered a blocker comment, though we should keep
this in mind if we keep seeing this in other pull requests.

qiita_db/sql_connection.py
```diff
@@ -1012,6 +1014,20 @@ def execute_fetchindex(self, idx=-1):
...
+        error_msg = []
+        for cmd in cmds:
+            try:
+                cmd()
+            except Exception as e:
+                error_msg.append(str(e))
+        self._post_commit_funcs = []
+        self._post_rollback_funcs = []
+        if error_msg:
+            raise RuntimeError(
```
ElDeveloper: I can't find a test that checks for a case where :hankey: hits the fan, can you add one?

qiita_db/sql_connection.py
```diff
@@ -1012,6 +1014,20 @@ def execute_fetchindex(self, idx=-1):
         """
         return self.execute()[idx]
 
+    def _cmds_executor(self, cmds, func_str):
```
ElDeveloper: For consistency I think you want to cal cmds funcs and so on ...

qiita_db/sql_connection.py
```diff
@@ -1012,6 +1014,20 @@ def execute_fetchindex(self, idx=-1):
...
+        error_msg = []
+        for cmd in cmds:
+            try:
+                cmd()
+            except Exception as e:
+                error_msg.append(str(e))
+        self._post_commit_funcs = []
```
ElDeveloper: I'm having trouble understanding the logic behind this and the next line, would you mind explaining me why this is the case?

qiita_db/sql_connection.py
```diff
@@ -1047,10 +1065,44 @@ def rollback(self):
...
             raise
+        # Execute the post rollback commands
+        self._cmds_executor(self._post_rollback_funcs, "rollback")
 
     @property
     def index(self):
         return len(self._queries) + len(self._results)
 
+    @_checker
+    def add_post_commit_func(self, func):
```
ElDeveloper: Since this is some really nifty functionality, I would probably describe this feature as part of the docstring ...

qiita_db/sql_connection.py
josenavas: Yeah, I started calling it cmds and then I change to functions, but I didn't update everything.
Sorry about that!

qiita_db/sql_connection.py
josenavas: Once the commit/rollback is done, the transaction object should remain clean. Hence, once the functions have been executed, we are resetting these attributed. These two lists are mutually exclusive, only one of them is executed, so when one of them is executed, we can reset both of them.

qiita_db/sql_connection.py
josenavas: Sure!

qiita_db/sql_connection.py
josenavas: Done!

qiita_db/sql_connection.py
ElDeveloper: Ahhh, got it! Would be useful to mention in a comment that these are
mutually exclusive and that if one is run, we are safe to clear them
both.

josenavas:
Thanks @ElDeveloper

qiita_db/sql_connection.py
josenavas: Done!

2015-07-03 19:35 GMT-07:00 Yoshiki Vázquez Baeza <notifications@github.com>:

> In qiita_db/sql_connection.py
> <https://github.com/biocore/qiita/pull/1324#discussion_r33884088>:
>
> > @@ -1012,6 +1014,20 @@ def execute_fetchindex(self, idx=-1):
> >          """
> >          return self.execute()[idx]
> >
> > +    def _cmds_executor(self, cmds, func_str):
> > +        error_msg = []
> > +        for cmd in cmds:
> > +            try:
> > +                cmd()
> > +            except Exception as e:
> > +                error_msg.append(str(e))
> > +        self._post_commit_funcs = []
>
> Ahhh, got it! Would be useful to mention in a comment that these are
> mutually exclusive and that if one is run, we are safe to clear them both.
> … <#14e56ead0b77bf15_>
> On (Jul-03-15|19:29), josenavas wrote: > @@ -1012,6 +1014,20 @@ def
> execute_fetchindex(self, idx=-1): > """ > return self.execute()[idx] > > +
> def _cmds_executor(self, cmds, func_str): > + error_msg = [] > + for cmd in
> cmds: > + try: > + cmd() > + except Exception as e: > +
> error_msg.append(str(e)) > + self._post_commit_funcs = [] Once the
> commit/rollback is done, the transaction object should remain clean. Hence,
> once the functions have been executed, we are resetting these attributed.
> These two lists are mutually exclusive, only one of them is executed, so
> when one of them is executed, we can reset both of them. --- Reply to this
> email directly or view it on GitHub:
> https://github.com/biocore/qiita/pull/1324/files#r33884073
>
> —
> <https://github.com/biocore/qiita/pull/1324/files#r33884088>.
>



-- 
Jose Navas

qiita_db/sql_connection.py
```diff
@@ -1012,6 +1014,22 @@ def execute_fetchindex(self, idx=-1):
         """
         return self.execute()[idx]
 
+    def _funcs_executor(self, cmds, func_str):
```
ElDeveloper: Thanks, I think you still missed changing cmds here and in the tests.

qiita_db/util.py
squirrelo: I would advocate for a `flatten=[True/False]` parameter added to the fetch calls, but I know y'all don't like changing what's returned through parameters. If we don't use the parameter, do we just want a flatten util function instead? then it's `flatten(TRN.execute_fetchindex())`. Flattening does show up a lot in the codebase, so best to get this sorted out now.

josenavas: We don't need a flatten function, with `chain.form_iterable` is enough

ElDeveloper: I do agree that we don't explicitly need a flatten, but if it's a thing
we do over and over and over, having a centralized implementation
*might* be useful and is overall cleaner (IMHO).

squirrelo:
The `_clean_up` function seems to be missing the func calls for post-rollback and post-commit (line 750).

josenavas:
Those calls are inside the commit and rollback functions

qiita_db/util.py
josenavas: I will add a new function in transaction.

squirrelo:
Ah, ok. I get the centralization logic here.

qiita_db/sql_connection.py
```diff
@@ -1012,6 +1014,22 @@ def execute_fetchindex(self, idx=-1):
...
+        error_msg = []
+        for cmd in cmds:
+            try:
+                cmd()
```
squirrelo: Will these function calls ever need parameters passed? If so, we should have a way to do this (kwargs or whatever is easiest).

qiita_db/sql_connection.py
```diff
@@ -1047,10 +1067,54 @@ def rollback(self):
...
+        committed.
+
+        Parameters
+        ----------
+        func : function
+            The function to add for the post commit functions
+
+        Notes
+        -----
+        func should not accept any parameter, i.e. it should allow to be
```
squirrelo: As stated above, this may need parameters. I can see this being useful for passing in a list of files or some other list of items that need to be worked on.

qiita_db/sql_connection.py
josenavas: Please read the documentation. There are ways or making this viable always, check the purge_filepath function.

qiita_db/sql_connection.py
josenavas: As stated above, see the purge_filepaths function

squirrelo: This does not concern that specific function. I could see this functionality being useful if, inside a transaction, we are moving files, and if the transaction fails, the file list is passed on rollback to move them all back to where they need to be. This way we have both filesystem and database integrity, and the ability to rollback files by tracking that list. So maybe not as useful for post commit, but definitely rollback.

josenavas: You're talking about passing parameters you can do this (copying here from purge_filepaths):
```
from functools import partial

func = partial(remove, fp)
TRN.add_post_commit_func(func)
```

Your specific case on moving files:
```
func = partial(move, src, dest)
TRN.add_post_rollback_func(func)
```

squirrelo: Creating a partial seems like a hacky way to get around just passing in parameters. It would be cleaner if you could instead do, using your moving files example (and being extra explicit):
`TRN.add_post_rollback_func(move, args={src: SOURCE, {dest=DEST}})`

Then have transaction call be
`cmd(**args)`

This is much more readable, and again less hacky then having to wrap things in partials.

josenavas: I don't think having a partial is hacky, and both of them are valid options.

Tie breaker @biocore/qiita-dev ?

ElDeveloper: partials are fine IMHO, though kwargs/args is more standard. Until we have a real use-case, we will only really be guessing ...

antgonza: I think they are the same thing just done 2 different ways. @squirrelo, if you feel strongly about this "extra" functionality, what about adding an issue so we don't block this PR to be merged?

squirrelo: Another concern with using partials this way is they are not actually partials any more, they are full function calls. This does not fit with how partials are standardly used.

josenavas: Please add an issue if you feel strong as @antgonza suggested...

antgonza: Not sure what you mean, I think is fine based on the [documentation](https://docs.python.org/2/library/functools.html#partial-objects)

squirrelo: Issue #1325 added and put in as part of #1318

squirrelo: @antgonza based on [documentation with examples](https://docs.python.org/2/library/functools.html#functools.partial) this is not the standard way of using it.

antgonza: Got it. However, I still is fine, it's impossible to cover all possible and valid examples in the documentation.

antgonza:
I think this looks ready ... :+1:

qiita_db/sql_connection.py
josenavas: Good call. I think I've fixed them all now...

josenavas:
Ready for another review.
I've created the function `exectue_fetchflatten` so we can centralize in a single function the loop that we were doing in multiple places.

Given that @squirrelo felt strong about it, I end up fixing https://github.com/biocore/qiita/issues/1325. However, **I did not implment @squirrelo  proposal**. His proposal was hackish and not standard, so I've actually implemented a standard way of doing this.

ElDeveloper:
:+1:

PR #1327, opened 2015-07-05
Fixing qiita_db/data.py

josenavas:
Fixes the objects under data.py

A couple of comments:
 - There was a lot of instance methods under the base class that were calling `check_subclass`. This is not needed as the base class cannot be instantiated, so I removed those calls.
 - `RawData.create` had a `if filepaths` as if they're optional, but it is actually a required parameter.

### DIFF
--- a/qiita_db/data.py
+++ b/qiita_db/data.py
@@ -386,7 +386,7 @@ class RawData(BaseData):
             raw_data_count = TRN.execute_fetchlast()
             if raw_data_count == 1 and RawData(raw_data_id).get_filepath_ids():
                 raise QiitaDBError(
-                    "Raw data (%d) can't be remove because it has linked "
+                    "Raw data (%d) can't be removed because it has linked "
                     "files. To remove it, first unlink files." % raw_data_id)
 
             # delete
@@ -1039,11 +1039,10 @@ class PreprocessedData(BaseData):
         ValueError
             If the status is not known.
         """
-        with TRN:
-            if status not in ('not submitted', 'submitting', 'success',
-                              'failed'):
-                raise ValueError("Unknown status: %s" % status)
+        if status not in ('not submitted', 'submitting', 'success', 'failed'):
+            raise ValueError("Unknown status: %s" % status)
 
+        with TRN:
             sql = """UPDATE qiita.{0}
                      SET submitted_to_vamps_status = %s
                      WHERE preprocessed_data_id=%s""".format(self._table)
@@ -1080,11 +1079,10 @@ class PreprocessedData(BaseData):
         ValueError
             If the state is not known
         """
+        if (state not in ('not_processed', 'processing', 'processed') and
+                not state.startswith('failed')):
+            raise ValueError('Unknown state: %s' % state)
         with TRN:
-            if (state not in ('not_processed', 'processing', 'processed') and
-                    not state.startswith('failed')):
-                raise ValueError('Unknown state: %s' % state)
-
             sql = """UPDATE qiita.{0} SET processing_status=%s
                      WHERE preprocessed_data_id=%s""".format(self._table)
             TRN.add(sql, [state, self.id])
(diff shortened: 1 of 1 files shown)


qiita_db/data.py
```diff
@@ -367,48 +361,47 @@ def delete(cls, raw_data_id, prep_template_id):
...
+
+            # Check to how many prep templates the raw data is still linked.
+            # If last one, check that are no linked files
+            sql = """SELECT COUNT(*) FROM qiita.prep_template
+                     WHERE raw_data_id = %s"""
+            TRN.add(sql, [raw_data_id])
+            raw_data_count = TRN.execute_fetchlast()
+            if raw_data_count == 1 and RawData(raw_data_id).get_filepath_ids():
+                raise QiitaDBError(
+                    "Raw data (%d) can't be remove because it has linked "
```
ElDeveloper: It was like this before but ... remove->removed

qiita_db/data.py
```diff
@@ -1034,13 +1039,16 @@ def update_vamps_status(self, status):
         ValueError
             If the status is not known.
         """
-        if status not in ('not submitted', 'submitting', 'success', 'failed'):
-            raise ValueError("Unknown status: %s" % status)
+        with TRN:
+            if status not in ('not submitted', 'submitting', 'success',
```
ElDeveloper: Does this need to be inside a transaction? Intuition says no, but maybe I'm missing something.

qiita_db/data.py
```diff
@@ -1070,14 +1080,15 @@ def processing_status(self, state):
         ValueError
             If the state is not known
         """
-        if (state not in ('not_processed', 'processing', 'processed') and
-                not state.startswith('failed')):
-            raise ValueError('Unknown state: %s' % state)
+        with TRN:
+            if (state not in ('not_processed', 'processing', 'processed') and
```
ElDeveloper: Same as above.

qiita_db/data.py
```diff
@@ -1398,44 +1410,47 @@ def processing_info(self):
...
+        with TRN:
+            # Get processed date and the info for the dynamic table
+            sql = """SELECT processed_date, processed_params_table,
+                processed_params_id FROM qiita.{0}
+                WHERE processed_data_id=%s""".format(self._table)
+            TRN.add(sql, [self.id])
+            static_info = TRN.execute_fetchindex()[0]
+
+            # Get the info from the dynamic table, including reference used
+            sql = """SELECT * from qiita.{0}
```
ElDeveloper: from -> FROM

ElDeveloper:
:+1: Looks good, just one question!

qiita_db/data.py
```diff
@@ -229,19 +222,19 @@ def _set_link_filepaths_status(self, status):
...
-            raise ValueError(msg)
-
-        conn_handler = SQLConnectionHandler()
-        conn_handler.execute(
-            "UPDATE qiita.{0} SET link_filepaths_status = %s "
-            "WHERE {1} = %s".format(self._table, self._data_filepath_column),
-            (status, self._id))
+        with TRN:
+            if (status not in ('idle', 'linking', 'unlinking') and
+                    not status.startswith('failed')):
```
antgonza: Not sure why flake8 is not complaining about this one ...

antgonza:
One minor comment.

qiita_db/data.py
squirrelo: flake8 treats if statements differently for indentation, since the things in the if statement are already indented once. This means the next line of the if needs to be indented twice so you can tell the difference between lines in the if and the continuation of the if.

josenavas:
I did not forget about this, I will address the reviews and finish the rest of the PR today :smile:

qiita_db/data.py

qiita_db/data.py
josenavas: That's is correct, flake 8 will fail if I indent back this line.

qiita_db/data.py
josenavas: You're correct, there is not DB access in that line, so it can be outside of the transaction started here
Good catch!

qiita_db/data.py
josenavas: Fixed

qiita_db/data.py

josenavas:
@antgonza @ElDeveloper comments addressed

PR #1328, opened 2015-07-05
Portals convert to/from id

squirrelo:
Change the behavior of the convert functions. This is needed because they are now being used to see if portals, and possibly other things in the future, actually exist in the table. Since this is no longer an `IncompetentQiitaDeveloperError`, a new `QiitaDBLookupError` was created for this case.


squirrelo:
Ready for review

josenavas:
:+1:

ElDeveloper:
:+1:

PR #1330, opened 2015-07-06
Portals add/delete checks

squirrelo:
This adds a check to make sure a deleted study is not used in an analysis on the portal, and another to make sure an analysis is not added to a portal without access to all data needed in the analysis.

### DIFF
--- a/qiita_db/portal.py
+++ b/qiita_db/portal.py
@@ -265,6 +265,7 @@ class Portal(QiitaObject):
             Trying to delete from QIITA portal
         QiitaDBError
             Some studies given do not exist in the system
+            Some studies are already used in an analysis on the portal
         QiitaDBWarning
             Some studies already do not exist in the given portal
         """
@@ -274,7 +275,7 @@ class Portal(QiitaObject):
 
         conn_handler = SQLConnectionHandler()
         # Make sure study not used in analysis in portal
-        sql = """SELECT DISTINCT study_id from qiita.study_processed_data
+        sql = """SELECT DISTINCT study_id FROM qiita.study_processed_data
                  JOIN qiita.analysis_sample USING (processed_data_id)
                  JOIN qiita.analysis_portal USING (analysis_id)
                  WHERE portal_type_id = %s AND study_id IN %s"""
--- a/qiita_db/portal.py
+++ b/qiita_db/portal.py
@@ -278,12 +278,12 @@ class Portal(QiitaObject):
                  JOIN qiita.analysis_sample USING (processed_data_id)
                  JOIN qiita.analysis_portal USING (analysis_id)
                  WHERE portal_type_id = %s AND study_id IN %s"""
-        analysed = [x[0] for x in conn_handler.execute_fetchall(sql,
-                                                                [self.id])]
+        analysed = [x[0] for x in conn_handler.execute_fetchall(
+            sql, [self.id, tuple(studies)])]
         if analysed:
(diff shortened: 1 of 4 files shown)


squirrelo:
This is ready for review.

qiita_db/portal.py
```diff
@@ -338,12 +350,28 @@ def add_analyses(self, analyses):
...
             Some given analyses do not exist in the system,
             or are default analyses
+            Portal does not contain all studies used in analyses
         QiitaDBWarning
             Some analyses already exist in the given portal
         """
         self._check_analyses(analyses)
 
         conn_handler = SQLConnectionHandler()
+        # Make sure new portal has access to all studies in analysis
```
josenavas: This check is only needed if the portal is not `QIITA`.

josenavas:
Just one commen.

qiita_db/portal.py
squirrelo: I'd rather have as few if statement checks for "if QIITA" as possible, since they are harder to maintain. This is a very cheap check, so should not be an issue.

josenavas: @biocore/qiita-dev comments?

@squirrelo an SQL query is not very cheap compared with an if statement....

qiita_db/portal.py
```diff
@@ -272,8 +272,20 @@ def remove_studies(self, studies):
             raise ValueError('Can not remove from main QIITA portal!')
         self._check_studies(studies)
 
-        # Clean list of studies down to ones associated with portal already
         conn_handler = SQLConnectionHandler()
+        # Make sure study not used in analysis in portal
+        sql = """SELECT DISTINCT study_id from qiita.study_processed_data
```
ElDeveloper: from -> FROM

qiita_db/portal.py
```diff
@@ -272,8 +272,20 @@ def remove_studies(self, studies):
...
+                 JOIN qiita.analysis_sample USING (processed_data_id)
+                 JOIN qiita.analysis_portal USING (analysis_id)
+                 WHERE portal_type_id = %s AND study_id IN %s"""
+        analysed = [x[0] for x in conn_handler.execute_fetchall(
+            sql, [self.id, tuple(studies)])]
+        if analysed:
+            raise QiitaDBError("The following studies are used in an analysis "
```
ElDeveloper: Can you add a description in the docstring of why this exception could get raised?

qiita_db/portal.py
```diff
@@ -338,12 +350,28 @@ def add_analyses(self, analyses):
...
             or are default analyses
+            Portal does not contain all studies used in analyses
         QiitaDBWarning
             Some analyses already exist in the given portal
         """
         self._check_analyses(analyses)
 
         conn_handler = SQLConnectionHandler()
+        # Make sure new portal has access to all studies in analysis
+        sql = """SELECT DISTINCT analysis_id from qiita.analysis_sample
```
ElDeveloper: from -> FROM

qiita_db/portal.py
ElDeveloper: I would go with the if statement, though I don't really have a strong preference. Can see both points.

qiita_db/portal.py
squirrelo: This already exists?

squirrelo:

josenavas:
:+1:

PR #1326, opened 2015-07-05
Fix metadata template object for transaction

josenavas:
This PR fixes the files under `qiita_db/metadata_template`

Few comments:
 - The functions `to_file` and `create_qiime_mapping_file` were missing the `encoding` parameter for `to_csv`
 - The `SampleTemplate.delete` function was checking for `RawData` to allow deletion of the Sample Template. This is incorrect as it should be checking for `PrepTemplate`.

PR #1332, opened 2015-07-06
Transactionize the analysis and job objects

josenavas:
The analysis and job object now use transactions.

I've been removing some SQL queries to use `convert_to_id` in a couple of places.

### DIFF
--- a/qiita_db/analysis.py
+++ b/qiita_db/analysis.py
@@ -28,7 +28,7 @@ import pandas as pd
 from skbio.util import find_duplicates
 
 from qiita_core.exceptions import IncompetentQiitaDeveloperError
-from .sql_connection import SQLConnectionHandler
+from .sql_connection import TRN
 from .base import QiitaStatusObject
 from .data import ProcessedData
 from .study import Study
@@ -74,7 +74,7 @@ class Analysis(QiitaStatusObject):
     _table = "analysis"
     _analysis_id_column = 'analysis_id'
 
-    def _lock_check(self, conn_handler):
+    def _lock_check(self):
         """Raises QiitaDBStatusError if analysis is not in_progress"""
         if self.check_status({"queued", "running", "public", "completed",
                               "error"}):
@@ -100,11 +100,14 @@ class Analysis(QiitaStatusObject):
         set of int
             All analyses in the database with the given status
         """
-        conn_handler = SQLConnectionHandler()
-        sql = ("SELECT analysis_id FROM qiita.{0} a JOIN qiita.{0}_status ans "
-               "ON a.analysis_status_id = ans.analysis_status_id WHERE "
-               "ans.status = %s".format(cls._table))
-        return {x[0] for x in conn_handler.execute_fetchall(sql, (status,))}
+        with TRN:
+            sql = """SELECT analysis_id
+                     FROM qiita.{0} a
+                        JOIN qiita.{0}_status ans
+                            ON a.analysis_status_id = ans.analysis_status_id
+                     WHERE ans.status = %s""".format(cls._table)
+            TRN.add(sql, [status])
+            return set(TRN.execute_fetchflatten())
 
     @classmethod
     def create(cls, owner, name, description, parent=None, from_default=False):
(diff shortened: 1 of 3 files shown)


josenavas:
This have the same failure as https://github.com/biocore/qiita/pull/1326
Once it gets merged I will pull from upstream/transaction and the failure will go away, so this is safe to review

antgonza:
This looks fine. :+1:

qiita_db/analysis.py
```diff
@@ -100,11 +100,14 @@ def get_by_status(cls, status):
         set of int
             All analyses in the database with the given status
         """
-        conn_handler = SQLConnectionHandler()
-        sql = ("SELECT analysis_id FROM qiita.{0} a JOIN qiita.{0}_status ans "
-               "ON a.analysis_status_id = ans.analysis_status_id WHERE "
-               "ans.status = %s".format(cls._table))
-        return {x[0] for x in conn_handler.execute_fetchall(sql, (status,))}
+        with TRN:
+            sql = """SELECT analysis_id
```
squirrelo: This indentation is odd, and this SQL should also be changed to use `USING`. Standard SQL indentation is:
```SQL
SELECT analysis_id  FROM qiita.{0}
JOIN qiita.{0}_status USING (analysis_status_id)
WHERE ans.status = %s
```

qiita_db/analysis.py
```diff
@@ -363,13 +379,17 @@ def data_types(self):
...
         """
-        sql = ("SELECT DISTINCT data_type from qiita.data_type d JOIN "
-               "qiita.processed_data p ON p.data_type_id = d.data_type_id "
-               "JOIN qiita.analysis_sample a ON p.processed_data_id = "
-               "a.processed_data_id WHERE a.analysis_id = %s ORDER BY "
-               "data_type")
-        conn_handler = SQLConnectionHandler()
-        return [x[0] for x in conn_handler.execute_fetchall(sql, (self._id, ))]
+        with TRN:
+            sql = """SELECT DISTINCT data_type
```
squirrelo: Same as above for indentation and use of `USING`

qiita_db/analysis.py
```diff
@@ -393,28 +414,25 @@ def all_associated_filepath_ids(self):
...
-
-        job_filepaths = {row[0]
-                         for row in conn_handler.execute_fetchall(sql,
-                                                                  [self._id])}
-
-        filepaths = filepaths.union(job_filepaths)
-
-        return filepaths
+        with TRN:
+            sql = """SELECT f.filepath_id
```
squirrelo: Same here and the SQL on line 426

qiita_db/analysis.py
```diff
@@ -425,20 +443,24 @@ def biom_tables(self):
...
-        if not tables:
-            return {}
-        ret_tables = {}
-        _, base_fp = get_mountpoint(self._table)[0]
-        for fp in tables:
-            ret_tables[fp[0]] = join(base_fp, fp[1])
-        return ret_tables
+        with TRN:
+            fptypeid = convert_to_id("biom", "filepath_type")
+            sql = """SELECT dt.data_type, f.filepath
```
squirrelo: Same here

qiita_db/analysis.py
```diff
@@ -449,43 +471,65 @@ def mapping_file(self):
...
-               "WHERE af.analysis_id = %s AND f.filepath_type_id = %s")
-        mapping_fp = conn_handler.execute_fetchone(sql, (self._id, fptypeid))
-        if not mapping_fp:
-            return None
-
-        _, base_fp = get_mountpoint(self._table)[0]
-        return join(base_fp, mapping_fp[0])
+        with TRN:
+            fptypeid = convert_to_id("plain_text", "filepath_type")
+            sql = """SELECT f.filepath
```
squirrelo: Ok there's a lot of `USING` that needs to be done. Last comment on it.

squirrelo:
Some SQL cleanup that needs to be done (not your fault but good to do since this pull request is going in to clean things up anyway). Other than that, looks good.

qiita_db/analysis.py
josenavas: @squirrelo can you point to the "standard" that you're talking about? There is no standard, only coding guidelines (that we can user or not) but None of the ones that I've seen use your indentation. Usually is common practice to put the FROM in the next line under the SELECT, not following it (quick googling will show this to you...)

I can agree on adding the JOIN in a single line but it didn't fit. Will change for USING.

qiita_db/analysis.py
josenavas: Done

qiita_db/analysis.py
josenavas: Done

qiita_db/analysis.py
josenavas: Done

qiita_db/analysis.py
josenavas: Done

josenavas:
@squirrelo  comments addressed

squirrelo:
:+1:

PR #1333, opened 2015-07-07
Transactionize the Study, StudyPerson and Reference

josenavas:
Modifies the Study, StudyPerson and Reference objects to use transactions

Few comments:
 - `Study.create` was not checking for duplicates. A check and a test have been added.
 - ~~`StudyPerson.create` was not raising duplicate error in case of duplicate, just returned the existing object. Changed for consistency with the rest of the system~~ If it doesn't return the already existent object there isn't a way to retrieve the existent one from the DB. Created an issue to solve later https://github.com/biocore/qiita/issues/1336
 - `StudyPerson.exists` was missing tests. Added.

### DIFF
--- a/qiita_db/reference.py
+++ b/qiita_db/reference.py
@@ -12,7 +12,7 @@ from .base import QiitaObject
 from .exceptions import QiitaDBDuplicateError
 from .util import (insert_filepaths, convert_to_id,
                    get_mountpoint)
-from .sql_connection import SQLConnectionHandler
+from .sql_connection import TRN
 
 
 class Reference(QiitaObject):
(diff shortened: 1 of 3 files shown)


qiita_db/reference.py
```diff
@@ -113,54 +115,62 @@ def exists(cls, name, version):
...
 
     @property
     def name(self):
-        conn_handler = SQLConnectionHandler()
-        return conn_handler.execute_fetchone(
-            "SELECT reference_name FROM qiita.{0} WHERE "
-            "reference_id = %s".format(self._table), (self._id,))[0]
-        _, basefp = get_mountpoint('reference')[0]
+        with TRN:
+            sql = """SELECT reference_name FROM qiita.{0}
```
squirrelo: these can be done with the `convert_from_id` function, correct?

qiita_db/reference.py
```diff
@@ -113,54 +115,62 @@ def exists(cls, name, version):
...
 
     @property
     def version(self):
-        conn_handler = SQLConnectionHandler()
-        return conn_handler.execute_fetchone(
-            "SELECT reference_version FROM qiita.{0} WHERE "
-            "reference_id = %s".format(self._table), (self._id,))[0]
-        _, basefp = get_mountpoint('reference')[0]
+        with TRN:
+            sql = """SELECT reference_version FROM qiita.{0}
```
squirrelo: Same, `convert_from_id`?

qiita_db/reference.py
```diff
@@ -113,54 +115,62 @@ def exists(cls, name, version):
...
     def taxonomy_fp(self):
-        conn_handler = SQLConnectionHandler()
-        rel_path = conn_handler.execute_fetchone(
-            "SELECT f.filepath FROM qiita.filepath f JOIN qiita.{0} r ON "
-            "r.taxonomy_filepath=f.filepath_id WHERE "
-            "r.reference_id=%s".format(self._table), (self._id,))[0]
-        _, basefp = get_mountpoint('reference')[0]
-        return join(basefp, rel_path)
+        with TRN:
+            sql = """SELECT f.filepath
```
squirrelo: More SQL cleanup for `USING`

qiita_db/study.py
```diff
@@ -934,13 +963,13 @@ def iter(cls):
             Yields a `StudyPerson` object for each person in the database,
             in order of ascending study_person_id
         """
-        conn = SQLConnectionHandler()
-        sql = "select study_person_id from qiita.{} order by study_person_id"
-        results = conn.execute_fetchall(sql.format(cls._table))
+        with TRN:
+            sql = """SELECT study_person_id FROm qiita.{}
```
squirrelo: FROm -> FROM

qiita_db/reference.py
josenavas: This is not converting from any id, is accessing to the reference_name attribute of the reference class.

qiita_db/reference.py
josenavas: Same as above

qiita_db/reference.py
josenavas: Please read carefully the SQL query... using does not work as the column does not have the same name in both tables.

qiita_db/study.py

josenavas:
@squirrelo I've addressed your comments

@antgonza can you review this one?

qiita_db/study.py
```diff
@@ -160,18 +161,17 @@ def _lock_non_sandbox(self, conn_handler):
...
-                  JOIN qiita.study_processed_data spd
-                    USING (processed_data_id)
-                WHERE spd.study_id = %s"""
-        pd_statuses = conn_handler.execute_fetchall(sql, (self._id,))
-
-        return infer_status(pd_statuses)
+        with TRN:
+            # Get the status of all its processed data
+            sql = """SELECT processed_data_status
+                    FROM qiita.processed_data_status pds
```
antgonza: none blocking but you are missing a space.

josenavas: fixed!

qiita_db/study.py
```diff
@@ -617,11 +635,14 @@ def investigation(self):
         -------
         Investigation id
         """
-        conn_handler = SQLConnectionHandler()
-        sql = ("SELECT investigation_id FROM qiita.investigation_study WHERE "
-               "study_id = %s")
-        inv = conn_handler.execute_fetchone(sql, (self._id, ))
-        return inv[0] if inv is not None else inv
+        with TRN:
+            sql = """SELECT investigation_id FROM qiita.investigation_study
```
antgonza: Will it be better to sort by investigation_id and limit 1?

antgonza:
minor comments

qiita_db/study.py
josenavas: I don't think it is needed. If it exists, it will be only one. The `inv[0][0]` is because if it doesn't exist, `TRN.execute_fetchindex` returns an empty list, but if there is a value it will return `[[inv_id]]`, and doesn't matter how I change the query, since this is because the transaction internally is doing a fetchall.

josenavas:
Addressed/answered

qiita_db/study.py
antgonza: K, thanks.

antgonza:
:+1:

qiita_db/reference.py
squirrelo: Ah, got it.

qiita_db/study.py
squirrelo: A study can be part of more than one investigation though, right? Hence the `investigation_study` join table.

squirrelo:
One question.

qiita_db/study.py
josenavas: I don't know. To be fair we have not been using investigation, and I think that right now is not needed given that we can have multiple prep templates in a single study.

Investigation was useful in the old database, were we had to create a study for each prep template... but right now we can create a single study and have everything there. And if multiple studies belong to "an investigation" that is a meta-analysis.

I think this is another discussion and this code is duplicating the functionality of the previous code, so if there are no more comments, we should merge this and keep moving forward with the transactionization of the system.

squirrelo: Agreed. Should open an issue discussing if we need the investigation object.

squirrelo:
:+1:

PR #1335, opened 2015-07-07
Transactionize the search engine and commands

josenavas:
I discovered an issue with the Transaction object: it was always passing an empty list to psycopg2. This is not a problem unless we are passing `%` chars in the SQL query, which are passed in the search engine for the `LIKE`-alike queries.

I just added a small check in the Transaction object that instead of passing an empty list, it will pass None, which doesn't create any problem.

### DIFF
--- a/qiita_db/commands.py
+++ b/qiita_db/commands.py
@@ -24,7 +24,7 @@ from .metadata_template import (SampleTemplate, PrepTemplate,
                                 load_template_to_dataframe)
 from .parameters import (PreprocessedIlluminaParams, Preprocessed454Params,
                          ProcessedSortmernaParams)
-from .sql_connection import SQLConnectionHandler
+from .sql_connection import TRN
 
 with standard_library.hooks():
     from configparser import ConfigParser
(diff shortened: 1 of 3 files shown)


qiita_db/commands.py
```diff
@@ -80,25 +80,25 @@ def get_optional(name):
...
-    pi_name, pi_email, pi_affiliation = pi_name_email.split(',', 2)
-    infodict['principal_investigator_id'] = StudyPerson.create(
-        pi_name.strip(), pi_email.strip(), pi_affiliation.strip())
+        pi_name_email = infodict.pop('principal_investigator')
+        pi_name, pi_email, pi_affiliation = pi_name_email.split(',', 2)
+        infodict['principal_investigator_id'] = StudyPerson.create(
+            pi_name.strip(), pi_email.strip(), pi_affiliation.strip())
 
-    return Study.create(User(owner), title, efo_ids, infodict)
+        return Study.create(User(owner), title, efo_ids, infodict)
```
antgonza: should this return be outside the TRN? I guess it doesn't matter but rather ask ...

qiita_db/commands.py
```diff
@@ -235,25 +238,26 @@ def load_processed_data_cmd(fps, fp_types, processed_params_table_name,
...
-    fp_types = [fp_types_dict[x] for x in fp_types]
+    with TRN:
+        fp_types_dict = get_filepath_types()
+        fp_types = [fp_types_dict[x] for x in fp_types]
 
-    if preprocessed_data_id is not None:
-        preprocessed_data = PreprocessedData(preprocessed_data_id)
-    else:
-        preprocessed_data = None
+        if preprocessed_data_id is not None:
```
antgonza: Non blocking but it will be nice IMOO. This block and the next:
preprocessed_data = None if preprocessed_data_id is None else PreprocessedData(preprocessed_data_id)

qiita_db/commands.py
```diff
@@ -348,77 +352,63 @@ def update_preprocessed_data_from_cmd(sl_out_dir, study_id, ppd_id=None):
...
+            raise ValueError("Study %s does not have any preprocessed data")
+
+        if ppd_id:
+            if ppd_id not in ppds:
+                raise ValueError(
+                    "The preprocessed data %d does not exist in "
+                    "study %d. Available preprocessed data: %s"
+                    % (ppd_id, study_id, ', '.join(map(str, ppds))))
+            ppd = PreprocessedData(ppd_id)
         else:
```
antgonza: This else needs one more indentation, no?

qiita_db/search.py
```diff
@@ -192,27 +192,31 @@ def __call__(self, searchstr, user):
...
+            for sid in study_ids:
+                TRN.add(sample_sql.format(sid))
+                study_res = TRN.execute_fetchindex()
+                if study_res:
+                    # only add study to results if actually has samples
+                    # in results
+                    results[sid] = study_res
+            self.results = results
+            self.meta_headers = meta_headers
+            return results, meta_headers
```
antgonza: Same question about the return within/out the TRN ...

qiita_db/search.py
```diff
@@ -370,32 +374,34 @@ def filter_by_processed_data(self, datatypes=None):
...
+                    # skip processed data if it doesn't fit the given datatypes
+                    if datatypes is not None and datatype not in datatypes:
+                        continue
+                    filter_samps = proc_data.samples.intersection(
+                        study_sample_ids)
+                    if filter_samps:
+                        proc_data_samples[proc_data_id] = sorted(filter_samps)
+                        study_proc_ids[study_id][datatype].append(proc_data_id)
+
+            return study_proc_ids, proc_data_samples, samples_meta
```
antgonza: Same about return within TRN

antgonza:
A few questions.

qiita_db/commands.py
josenavas: Nope, that return is executing multiple access to the DB at different points: `User(owner)` and in `Study.create...`. Also, if something during the creation goes wrong, I want the entire transaction to rollback, so no modifications in the DB are performed (there are some `StudyPerson.create` above)

qiita_db/commands.py
josenavas: Good call, done.

qiita_db/search.py
josenavas: It doesn't matter in this case. I've done it inside for consistency with other parts of the code, but it exits the context anyways...

qiita_db/search.py
josenavas: Same, if you feel strong I can change though...

josenavas:
@antgonza I think I've answered your questions and addressed your comments.

qiita_db/commands.py
antgonza: What about this one?

qiita_db/search.py
antgonza: That's fine, just wanted to be sure ...

qiita_db/commands.py
josenavas: Oops, I did not see this one.

This is GitHub messing up. That else (in my version) is actually the one in line 362 in this line. If you check line 363 here you will see that is line 364 in my version...

josenavas:
@squirrelo can you review this one?

qiita_db/commands.py
antgonza: Thanks.

antgonza:
:+1:

qiita_db/search.py
squirrelo: returns in the context are fine in my opinion, especially since it keeps everything contained in a single transaction block in the file.

squirrelo:
:+1:

PR #1337, opened 2015-07-07
Transactionaize patches and environment mgr

josenavas:
Modifies the patches and the environment manager to use transactions.

Few notes:
 - the environment manager is tricky. There are a couple of places in which the SQLConnectionHandler should still be used because it is connecting to the DB with admin privileges, and is changing the connection type to set autocommit = True. Since autocommit is true, there is no point of having the transaction so using the old conn handler there. Also the Transaction does not support admin access.
 - in qiita_pet/test/tornado_test_base.py there was a bit of code duplication. I removed that code and used the function from qiita_db.

### DIFF
--- a/qiita_db/environment_manager.py
+++ b/qiita_db/environment_manager.py
@@ -55,34 +55,36 @@ def _check_db_exists(db, conn_handler):
     return (db,) in dbs
 
 
-def create_layout_and_patch(conn, verbose=False):
+def create_layout_and_patch(verbose=False):
     r"""Builds the SQL layout and applies all the patches
 
     Parameters
     ----------
-    conn : SQLConnectionHandler
-        The handler connected to the DB
     verbose : bool, optional
         If true, print the current step. Default: False.
     """
-    if verbose:
-        print('Building SQL layout')
-    # Create the schema
-    with open(LAYOUT_FP, 'U') as f:
-        conn.execute(f.read())
+    with TRN:
+        if verbose:
+            print('Building SQL layout')
+        # Create the schema
+        with open(LAYOUT_FP, 'U') as f:
+            TRN.add(f.read())
+        TRN.execute()
 
-    if verbose:
-        print('Patching Database...')
-    patch(verbose=verbose)
+        if verbose:
+            print('Patching Database...')
+        patch(verbose=verbose)
 
 
-def _populate_test_db(conn):
+def _populate_test_db():
     print('Populating database with demo data')
-    with open(POPULATE_FP, 'U') as f:
-        conn.execute(f.read())
+    with TRN:
+        with open(POPULATE_FP, 'U') as f:
+            TRN.add(f.read())
+        TRN.execute()
 
 
-def _add_ontology_data(conn):
+def _add_ontology_data():
     print ('Loading Ontology Data')
     if not exists(reference_base_dir):
         mkdir(reference_base_dir)
@@ -101,18 +103,21 @@ def _add_ontology_data(conn):
             raise IOError("Error: Could not fetch ontologies file from %s" %
                           url)
 
-    with gzip.open(fp, 'rb') as f:
-        conn.execute(f.read())
+    with TRN:
+        with gzip.open(fp, 'rb') as f:
+            TRN.add(f.read())
+        TRN.execute()
 
 
-def _insert_processed_params(conn, ref):
-    sortmerna_sql = """INSERT INTO qiita.processed_params_sortmerna
-                       (reference_id, sortmerna_e_value, sortmerna_max_pos,
-                        similarity, sortmerna_coverage, threads)
-                       VALUES
-                       (%s, 1, 10000, 0.97, 0.97, 1)"""
-
-    conn.execute(sortmerna_sql, [ref._id])
+def _insert_processed_params(ref):
+    with TRN:
+        sortmerna_sql = """INSERT INTO qiita.processed_params_sortmerna
+                           (reference_id, sortmerna_e_value, sortmerna_max_pos,
+                            similarity, sortmerna_coverage, threads)
+                           VALUES
+                           (%s, 1, 10000, 0.97, 0.97, 1)"""
+        TRN.add(sortmerna_sql, [ref._id])
+        TRN.execute()
 
 
 def _download_reference_files(conn):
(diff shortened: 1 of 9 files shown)


josenavas:
@antgonza @squirrelo this is ready for review.

qiita_db/support_files/patches/python_patches/15.py
```diff
@@ -4,23 +4,26 @@
...
+                JOIN qiita.analysis_filepath afp
+                    ON f.filepath_id = afp.filepath_id"""
+    TRN.add(sql)
+    filepaths = TRN.execute_fetchindex()
 
-filepaths = conn_handler.execute_fetchall(
-    'SELECT f.* from qiita.filepath f JOIN qiita.analysis_filepath afp ON '
-    'f.filepath_id = afp.filepath_id')
+    # retrieve relative filepaths as dictionary for matching
+    mountpoints = {m[1].rstrip('/\\'): m[0] for m in get_mountpoint(
```
antgonza: nice!

antgonza:
:+1: if tests pass

josenavas:
errors were with my friend flake8

qiita_db/environment_manager.py
```diff
@@ -101,18 +103,21 @@ def _add_ontology_data(conn):
             raise IOError("Error: Could not fetch ontologies file from %s" %
```
squirrelo: This part should probably also be in the TRN, since it can fail and cause rollback.

qiita_db/environment_manager.py
```diff
@@ -142,11 +147,11 @@ def _download_reference_files(conn):
             except:
                 raise IOError("Error: Could not fetch %s file from %s" %
```
squirrelo: Same here. If it's explicitly in the TRN it will always cause a rollback.

qiita_db/environment_manager.py
```diff
@@ -346,41 +352,40 @@ def patch(patches_dir=PATCHES_DIR, verbose=False):
...
     patches found in the patches directory.
     """
-    conn = SQLConnectionHandler()
+    with TRN:
+        TRN.add("SELECT current_patch FROM settings")
+        current_patch = TRN.execute_fetchlast()
+        current_sql_patch_fp = join(patches_dir, current_patch)
+        corresponding_py_patch = partial(join, patches_dir, 'python_patches')
+
+        sql_glob = join(patches_dir, '*.sql')
```
squirrelo: should this also be natsorted?

squirrelo: woops, misread code. Disregard.

qiita_db/environment_manager.py
josenavas: Not needed. If this is the first function, the transaction didn't even started. If another function inside a transaction is calling this, that other context will catch it and rollback.

qiita_db/environment_manager.py
josenavas: Same as above

qiita_db/support_files/patches/python_patches/6.py
```diff
@@ -6,29 +6,28 @@
...
-
-    fp = join(fp_base, '%d_prep_%d_%s.txt' % (pt.study_id, prep_template_id,
-              strftime("%Y%m%d-%H%M%S")))
-    pt.to_file(fp)
-    pt.add_filepath(fp)
+        fp = join(fp_base,
+                  '%d_prep_%d_%s.txt' % (pt.study_id, prep_template_id,
+                                         strftime("%Y%m%d-%H%M%S")))
+        pt.to_file(fp)
+        pt.add_filepath(fp)
```
squirrelo: Missing an execute at the end

qiita_db/support_files/patches/python_patches/23.py
```diff
@@ -1,20 +1,19 @@
...
-# Get all the prep templates
-sql = """SELECT prep_template_id from qiita.prep_template"""
-prep_ids = {p[0] for p in conn_handler.execute_fetchall(sql)}
-for prep_id in prep_ids:
-    PrepTemplate(prep_id).generate_files()
+    # Get all the prep templates
+    TRN.add("SELECT DISTINCT prep_template_id from qiita.prep_template")
+    prep_ids = TRN.execute_fetchflatten()
+    for prep_id in prep_ids:
+        PrepTemplate(prep_id).generate_files()
```
squirrelo: missing an execute at the end

qiita_db/support_files/patches/python_patches/7.py
```diff
@@ -3,18 +3,16 @@
...
-for prep_template_id in conn_handler.execute_fetchall(
-        "SELECT prep_template_id FROM qiita.prep_template"):
-    prep_template_id = prep_template_id[0]
-    pt = PrepTemplate(prep_template_id)
-    study_id = pt.study_id
-
-    for _, fpt in pt.get_filepaths():
-        pt.create_qiime_mapping_file(fpt)
+        for _, fpt in pt.get_filepaths():
+            pt.create_qiime_mapping_file(fpt)
```
squirrelo: missing an execute at the end.

squirrelo:
few comments.

qiita_db/support_files/patches/python_patches/6.py
josenavas: It is not needed. No SQL query remaining to execute in this block

qiita_db/support_files/patches/python_patches/7.py
josenavas: same as above

qiita_db/support_files/patches/python_patches/23.py
josenavas: same as below...

josenavas:
@squirrelo comments answered

qiita_db/support_files/patches/python_patches/6.py
squirrelo: Some patches have the execute, some don't, and things like `add_filepath` are actually SQL just hidden. Better to standardize on all execute or all not now.

josenavas: Some patches have the execute because there are remaining SQL queries to execute. Each function/patch is responsible of executing their SQL queries. Thus, `add_filepath` is executing the SQL query inside. Anyways, if there is any SQL query remaining, it will be executed when we exit the last context.

Again is not needed

squirrelo:
:+1:

PR #1340, opened 2015-07-08
Transactionize portals and decorator

josenavas:
It looks like I messed up something with my local branches but here I include 2 things:
 - The Portal has been transactionized
 - A new decorator in qiita_core.util that will allow to decorate functions to execute them as a single transaction.


josenavas:
@squirrelo can you review this one as you did the portals?
@antgonza as a second reviewer would be appreciated.

antgonza:
:+1:

squirrelo:
:+1:

PR #1347, opened 2015-07-08
Transactionize QIITA-pet

josenavas:
Although I've changed a lot of files, I've been just decorating the functions that need to be executed as a transactions.

The files that I'm missing to transactionize are:
preprocessing_handlers.py
compute.py
vamps_handlers.py
ebi_handlers.py
analysis_handlers.py

The reason why is that they depend on Qiita-ware and we need to do some changes in qiita-ware, which we will talk on today's meeting.

Please test this in your local machines before merging.


antgonza:
:+1:

squirrelo:
AuthCreateHandler post method needs a decorator. Other then that, is good.

josenavas:
@squirrelo can you merge?

PR #1362, opened 2015-07-10
Httperror issue

josenavas:
This will reduce the amount of information in our log system.
Otherwise we can be logging in an entire miseq run.
403 is not needed. to be in the logging table


adamrp:
:+1: after tests pass

PR #1364, opened 2015-07-10
New user default analysis bug

josenavas:
This morning @rob-knight found an issue in which a new created user that goes to the create analysis from selected samples part receives an error and anything else in the system related with analysis works.

The issue is that the default analysis was not linked with the portals.

I've fixed the user creation, added a specific test and added a patch to update any existing case of this in the database.

This is priority super high, I've scheduled a system update for tonight, and I already posted the announcement in the live system.


squirrelo:
:+1:

Issue #1369, opened 2015-07-10
Provide a better name for the beta diversity results


josenavas:
right now we just see index.html, but we need to see weighted/unweighted

PR #1373, opened 2015-07-11
Improve names in the interface for bdiv results

josenavas:
Fixes https://github.com/biocore/qiita/issues/1369

I agree that this is not the best solution, but we need to refactor the analysis pipeline to improve a lot of different points, so I don't think it is worth to spend more time on the current code.

This modifies the name on the result page, so instead of showing index.html for both un/weighted unifrac, it shows the folder name:

screen shot 2015-07-10 at 6 48 24 pm

The testres.htm and subres.html are test files to check that it actually works as expected, the interesting part is that it does not show index.html but it shows which method was applied.

### DIFF
--- a/qiita_pet/handlers/analysis_handlers.py
+++ b/qiita_pet/handlers/analysis_handlers.py
@@ -12,7 +12,7 @@ Qitta analysis handlers for the Tornado webserver.
 from __future__ import division
 from future.utils import viewitems
 from collections import defaultdict
-from os.path import join, sep, commonprefix
+from os.path import join, sep, commonprefix, basename, dirname
 from json import dumps
 
 from tornado.web import authenticated, HTTPError, StaticFileHandler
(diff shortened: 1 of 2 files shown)


squirrelo:
:+1:

rob-knight:
That solution will do for the Sloan meeting - when is the broader
refactoring anticipated?

On Jul 10, 2015, at 6:52 PM, josenavas <notifications@github.com> wrote:

Fixes #1369 <https://github.com/biocore/qiita/issues/1369>

I agree that this is not the best solution, but we need to refactor the
analysis pipeline to improve a lot of different points, so I don't think it
is worth to spend more time on the current code.

This modifies the name on the result page, so instead of showing index.html
for both un/weighted unifrac, it shows the folder name:

[image: screen shot 2015-07-10 at 6 48 24 pm]
<https://cloud.githubusercontent.com/assets/2501478/8631767/cf69a868-2734-11e5-9556-e550a8691824.png>
------------------------------
You can view, comment on, or merge this pull request online at:

  https://github.com/biocore/qiita/pull/1373
Commit Summary

   - Showing the folder name rather than the file name if the filename is
   index.html

File Changes

   - *M* qiita_pet/handlers/analysis_handlers.py
   <https://github.com/biocore/qiita/pull/1373/files#diff-0> (10)
   - *M* qiita_pet/templates/analysis_results.html
   <https://github.com/biocore/qiita/pull/1373/files#diff-1> (5)

Patch Links:

   - https://github.com/biocore/qiita/pull/1373.patch
   - https://github.com/biocore/qiita/pull/1373.diff

—
<https://github.com/biocore/qiita/pull/1373>.

josenavas:
Not sure, I think it is related with the plug-in system but we will need to discuss in the next meeting.

PR #1372, opened 2015-07-11
added delete buttons on show_analyses page

RNAer:
Each analysis can be deleted on that page.

my first commit to qiita, yay! Thanks Jose for answering my questions - it is tremendously helpful!

PR #1374, opened 2015-07-11
Showing a better message when there are no samples available

josenavas:
When there are no analyses available it shows a message rather than empty table:

screen shot 2015-07-10 at 7 15 59 pm

The link "create analysis" goes to the study listing page.

Fixes https://github.com/biocore/qiita/issues/1368

### DIFF
--- a/qiita_pet/templates/show_analyses.html
+++ b/qiita_pet/templates/show_analyses.html
@@ -1,47 +1,66 @@
 {% extends sitebase.html %}
-{%block content %}
-<table class="table">
-  <tr>
-    <th>Analysis Name</th>
-    <th>Analysis Description</th>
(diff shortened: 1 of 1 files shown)


josenavas:
This PR will have a merge conflict with https://github.com/biocore/qiita/pull/1372. merge that one first and I will resolve the merge conflicts so @RNAer don't have to deal with this in his first PR :smile:

squirrelo:
:+1:

josenavas:
I'm not sure if @RNAer is going to fix #1372 today. If by the time i want to merge everything it has not been fixed, I will pull down and fix the issues, since there is only one comment from @squirrelo and a few flake8 errors.

PR #1378, opened 2015-07-14
ENH: Add configuration file argument to qiita CLI

ElDeveloper:
With the portals merged in master, changing the desired configuration file for
the webserver is a bit cumbersome, with this proposed change we should be able
to just pass the configuration file as part of the command call, for example:

```bash
qiita --config-fp=portal_A.cfg pet webserver start

qiita --config-fp=portal_B.cfg pet webserver start
```

I believe this is somewhat common in other command line interfaces and it feels
more natural than having to change an environment variable before calling a
command. However I'm unsure about the unintended sideffects that reloading a
module might cause.

### DIFF
--- a/scripts/qiita
+++ b/scripts/qiita
@@ -18,6 +18,8 @@ import tornado.ioloop
 from psycopg2 import OperationalError
 from future.utils import viewitems
 from os.path import join, abspath, dirname
+from os import environ
+from imp import reload
 
 from qiita_db.util import (get_filepath_types, get_filetypes,
                            get_data_types,
@@ -39,6 +41,7 @@ from qiita_core.configuration_manager import ConfigurationManager
 from qiita_ware.commands import ebi_actions, submit_EBI as _submit_EBI
 from qiita_ware.processing_pipeline import generate_demux_file
 from qiita_ware.context import system_call, ComputeError
+from qiita_core import qiita_settings
 from moi import r_client
 
 
@@ -54,8 +57,18 @@ else:
 
 
 @click.group()
-def qiita():
-    pass
+@click.option('--config-fp', default=ConfigurationManager().conf_fp,
+              show_default=True,
+              type=click.Path(exists=True, dir_okay=False, readable=True,
+                              resolve_path=True),
+              help="Qiita configuration file to use globally.")
+def qiita(config_fp):
+    """Command line utility to interact with Qiita and its main components"""
+
+    # with the updated envrionment variable, reload the config files
+    if config_fp != ConfigurationManager().conf_fp:
+        environ['QIITA_CONFIG_FP'] = config_fp
+        reload(qiita_settings)
 
 
 @qiita.group()


josenavas:
Uhm... what about the modules that have been already loaded? Those might be using the previous configuration as they're already loaded, which means that can be accessing to another DB. What do you think?

ElDeveloper:
Good point, I'll have a look.

antgonza:
what's the status of this PR?

ElDeveloper:
I haven't figured out a good solution to this, however if others think
this is a feature we want, I'll investigate further.

josenavas:
I'm unsure if we are going to be able to fix this. Should we just update the documentation to point out that the configuration file is retrieved from the global variable? Anyways, for having multiple servers in a single machine, the recommended way is to use screen, so no problems on exporting different values for the configuration file.

wasade:
I don't think the recommended way should be via screen -- I can't think of
any established system service that follows that model. Why not just have a
`-c /path/to/config`?

josenavas:
That is what we are trying to do here, but the main problem is that the imported modules load the config file before we even parse the command line interface.

wasade:
Guess it should be pulled out of init?

josenavas:
I don't think that will help. I think the best approach is to create a different command file that is used only for starting qiita. The problem is that for building the CLI we are pulling out data from the DB, so the config file will be parsed anyways.

wasade:
Well.... One possibility is to have init check if -c exists in sys.argv...

antgonza:
I think this is becoming more difficult than the benefit we will get from implementing. IMOO do not implement. What do other think?

josenavas:
Agree on not implementing now, we can revisit later.

wasade:
@antgonza or @josenavas, before this gets tabled, can one of you work with Jeff to see if it is possible to get qiita setup as a system service for all the portals in its present form? If it is possible, then I think this is fine to defer. If it is not possible, or is awkward, then this needs to be resolved.

josenavas:
I don't understand what you mean @wasade 

We currently have the different portals working in a single machine...

antgonza:
If we really do not want to use screen we could perhaps do:
```
export QIITA_CONFIG_FP=your_qiita_file.conf
nohup qiita pet webserver start &
export QIITA_CONFIG_FP=your_other_qiita_file.conf
nohup qiita pet webserver start &
export QIITA_CONFIG_FP=your_other_other_qiita_file.conf
nohup qiita pet webserver start &
```

wasade:
@josenavas, if the system reboots, qiita does not automatically restart. 

@antgonza, that's a start. Recommend discussing with Jeff though.

antgonza:
@ElDeveloper it seems like we will close for now, could you?

wasade:
Here's a proof of concept for just gutting `-c` from `__init__` which enables processing the config from either `-c` or `$QIITA_CONFIG_FP`, and which pulls the argument out of `argv` if necessary so that it doesn't cause issues with click.

```python
### example __init__.py
import sys
import os

if '-c' in sys.argv:
    idx = sys.argv.index('-c')
    if not os.path.exists(os.path.expandvars(sys.argv[idx+1])):
        raise IOError("Config file argument must be a valid file path.")
    conf_fp = sys.argv.pop(idx+1)
    _ = sys.argv.pop(idx)
else:
    conf_fp = os.environ['QIITA_CONFIG_FP']

print conf_fp
print sys.argv

__all__ = ['conf_fp']

### example click script
from foo import conf_fp

import click

@click.command()
def blah():
    click.echo('from command: %s' % conf_fp)

if __name__ == '__main__':
    blah()
```

```bash
(python27)09:00:50 scripts$ python runfoo.py
/Users/mcdonadt/ResearchWork/software/qiita/qiita_core/support_files/config_test.txt
['runfoo.py']
from command: /Users/mcdonadt/ResearchWork/software/qiita/qiita_core/support_files/config_test.txt
(python27)09:00:54 scripts$ python runfoo.py -c blahblah
  File "runfoo.py", line 1, in <module>
    from foo import conf_fp
  File "/Users/mcdonadt/not_backed_up/testfoo/foo/__init__.py", line 7, in <module>
    raise IOError("Config file argument must be a valid file path.")
IOError: Config file argument must be a valid file path.
(python27)09:01:02 scripts$ python runfoo.py -c runfoo.py
runfoo.py
['runfoo.py']
from command: runfoo.py
```

ElDeveloper:
Ahhhhh, the good 'ol switcharoo, though I thin in order for this to wor,
the first part would have be at the top of this file and not in the
__init__.py file, right?

wasade:
@ElDeveloper, needs to work for CLI calls as well as `from qiita import blah`

wasade:
(and there are dirty side effects, so something better than `-c` would be advisable)

antgonza:
@ElDeveloper, what do you want to do with this?

ElDeveloper:
I don't think it will be possible to get this to work, so closing.

PR #1379, opened 2015-07-17
added gz support for sff

### LINKED ISSUE #1299: gzipped sff files are not supported
`process_sff.py` does support gzipped files, but we are computing incorrectly the output file name.

The offending line is this one: https://github.com/biocore/qiita/blob/master/qiita_ware/processing_pipeline.py#L321

The code in QIIME used for compute the output file name is this one:
```python
        sff_fp = join(sff_dir, filename)
        base_filename = splitext(filename)[0]
        if filename.endswith('.gz'):
            base_filename = splitext(base_filename)[0]
        base_output_fp = join(output_dir, base_filename)
```

RNAer:
This fixes #1299 

and corrected a typo "assertIn" in another test.

### DIFF
--- a/qiita_ware/processing_pipeline.py
+++ b/qiita_ware/processing_pipeline.py
@@ -319,6 +319,9 @@ def _get_preprocess_fasta_cmd(raw_data, prep_template, params):
         prepreprocess_cmds = []
         for sff in sffs:
             base = splitext(basename(sff))[0]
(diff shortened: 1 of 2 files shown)


antgonza:
:+1:

josenavas:
When this gets deployed, we should change the configuration files to support this type of files.

PR #1381, opened 2015-07-17
Log file creation

### LINKED ISSUE #1380: Have standard log files
We don't have a log file for qiita, which is necessary for diagnosing some system problems and automating things like bot catching.

squirrelo:
Fixes #1380

Creates a log file if a log file path is given, using the tornado standard logging at `warning` settings.

### DIFF
--- a/scripts/qiita
+++ b/scripts/qiita
@@ -454,8 +454,8 @@ def start(port):
     from qiita_pet.webserver import Application
     from tornado.options import options, parse_command_line
 
-    if _CONFIG.write_log_file:
-        options.log_file_prefix = _CONFIG.log_file
+    if _CONFIG.log_path:
+        options.log_file_prefix = _CONFIG.log_path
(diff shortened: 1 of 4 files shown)


scripts/qiita
```diff
@@ -452,7 +452,12 @@ def start(port):
     # having to do that. The solution will (perhaps) be to move this subcommand
     # into an entirely different script.
     from qiita_pet.webserver import Application
+    from tornado.options import options, parse_command_line
 
+    if _CONFIG.log_path:
```
josenavas: I just check the file and the variable _CONFIG is only used here. This is incorrect, as we need to do `from qiita_core.settings import qiita_config`.

squirrelo: Fair enough, will fix.

squirrelo:
Changes made, should be good to go.

josenavas:
:+1:

PR #1383, opened 2015-07-20
print qiita db patch number by qiita-test-install

### LINKED ISSUE #1077: qiita_test_install should include current patch number
In general this information is useful.

RNAer:
fix #1077

### DIFF
--- a/scripts/qiita-test-install
+++ b/scripts/qiita-test-install
@@ -1,9 +1,10 @@
 #!/usr/bin/env python
 from os import environ
-from os.path import join, dirname, abspath
+from os.path import join, dirname, abspath, splitext
 from sys import platform, version as python_version, exit, executable, stdout
 from unittest import TestLoader, TextTestRunner, TestCase
 from smtplib import SMTP, SMTP_SSL
+from qiita_db.sql_connection import TRN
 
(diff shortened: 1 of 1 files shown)


scripts/qiita-test-install
```diff
@@ -1,9 +1,10 @@
 #!/usr/bin/env python
 from os import environ
-from os.path import join, dirname, abspath
+from os.path import join, dirname, abspath, splitext
 from sys import platform, version as python_version, exit, executable, stdout
 from unittest import TestLoader, TextTestRunner, TestCase
 from smtplib import SMTP, SMTP_SSL
+from qiita_db.sql_connection import TRN
```
josenavas: This import can file in some misconfigured systems. Check [this line](https://github.com/RNAer/qiita/blob/i1077/scripts/qiita-test-install#L178) to see how DB accesses should be done in this file. Sorry for not pointing this out in our previous discussion, I did not remember that this file is a bit special compared with the other ones.

I will suggest not using the TRN object and actually use the SQL connection handler directly, as is done in such line.

RNAer:
updated. thanks!

josenavas:
Thanks! :+1:

PR #1405, opened 2015-08-06
Clean messaging

squirrelo:
Replaces newlines with HTML line breaks and linkifies all message strings.

### DIFF
--- a/qiita_pet/handlers/study_handlers/description_handlers.py
+++ b/qiita_pet/handlers/study_handlers/description_handlers.py
@@ -34,7 +34,8 @@ from qiita_ware.metadata_pipeline import (
     create_templates_from_qiime_mapping_file)
 from qiita_ware.exceptions import QiitaWareError
 from qiita_pet.handlers.base_handlers import BaseHandler
-from qiita_pet.handlers.util import check_access, convert_text_html
+from qiita_pet.util import convert_text_html
+from qiita_pet.handlers.util import check_access
 from qiita_pet.handlers.study_handlers.listing_handlers import (
     ListStudiesHandler)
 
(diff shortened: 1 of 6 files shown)


qiita_pet/util.py
```diff
@@ -68,6 +69,15 @@ def clean_str(item):
     return str(item).replace(" ", "_").replace(":", "")
 
 
+def convert_text_html(message):
+    """Linkify URLs and turn newlines into <br/> for HTML"""
+    lines = message.split('\n')
```
josenavas: I think this entire function can be replaced by:
`return tornado_linkify(message.replace('\n', '<br/>'))`

squirrelo: Replace needs to happen after linkifying, but it works. Done.

qiita_pet/handlers/study_handlers/description_handlers.py
```diff
@@ -211,6 +213,7 @@ def process_sample_template(self, study, user, callback):
                          else 'parsing the sample template')
             msg = html_error_message % (error_msg, basename(fp_rsp),
                                         str(e))
+            msg = convert_text_html(msg)
```
josenavas: I'm going to comment only here but applies to all. There is something weird when passing the `html_error_message` variable contents to the `convert_text_html`, as not all the html is interpreted correctly:
screen shot 2015-08-07 at 2 06 42 pm
Note that the only part of the error message that needs to be passed to `convert_text_html`, which is the variable `e`.
Suggested change:
```python
msg = html_error_message % (error_msg, basename(fp_rsp), convert_text_html(str(e)))
```

squirrelo: Seems linkify also escapes the HTML, interesting. Will fix.

squirrelo:
Change made, good to go.

josenavas:
:+1:

PR #1406, opened 2015-08-06
Issue 1343 - reformat large UTF-8 error

squirrelo:
This breaks it down into each column and the rows that are an issue:

screen shot 2015-08-06 at 1 33 09 pm

### DIFF
--- a/qiita_db/metadata_template/util.py
+++ b/qiita_db/metadata_template/util.py
@@ -273,7 +273,7 @@ def load_template_to_dataframe(fn, strip_whitespace=True, index='sample_name'):
                     cell.encode('utf-8')
                 except UnicodeError:
                     errors[headers[col]].append(row)
-        lines = ['%s: row(s) %s' % (header, ', '.join(rows))
+        lines = ['%s: row(s) %s' % (header, ', '.join(map(str, rows)))
                  for header, rows in viewitems(errors)]
         raise QiitaDBError('Non UTF-8 characters found:\n' +
                            '\n'.join(lines))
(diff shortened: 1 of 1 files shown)


qiita_db/metadata_template/util.py
```diff
@@ -265,15 +266,17 @@ def load_template_to_dataframe(fn, strip_whitespace=True, index='sample_name'):
...
                 try:
                     cell.encode('utf-8')
                 except UnicodeError:
-                    errors.append('row %d, header %s' % (row, headers[col]))
-        raise QiitaDBError('Non UTF-8 characters found at ' +
-                           '; '.join(errors))
+                    errors[headers[col]].append(row)
+        lines = ['%s: row(s) %s' % (header, ', '.join(map(str, rows)))
+                 for header, rows in viewitems(errors)]
+        raise QiitaDBError('Non UTF-8 characters found:\n' +
```
josenavas: `Non UTF-8 characters found in columns:\n`

squirrelo:
Fixed, ready for another review.

josenavas:
Thanks @squirrelo ! :+1: 

Not part of this PR, but do people think that it will useful to be able to provide a link in the error message to an error page where a comprehensive list of errors is provided and potentially marking the character that is not UTF-8? Something like what SQL does when you have incorrect syntax.

antgonza:
@josenavas I like the idea, wanna add an issue? Merging as is a separate issue.

josenavas:
Sure @antgonza , see https://github.com/biocore/qiita/issues/1414

PR #1407, opened 2015-08-07
Fixing sql_connection.py module documentation

### LINKED ISSUE #982: sql_connection.py incorrect documentation
example documentation doesn't work. two problems:

```python
from qiita_db.sql_connection import SQLConnectionHandler
conn_handler = SQLConnectionHandler # doctest: +SKIP
conn_handler.create_queue("example_queue") # doctest: +SKIP
```
should become 
```python
from qiita_db.sql_connection import SQLConnectionHandler
conn_handler = SQLConnectionHandler() # doctest: +SKIP
conn_handler.create_queue("example_queue") # doctest: +SKIP
```
otherwise it will fail. 

Any of the execute commands fail. For instance

```python
from qiita_db.sql_connection import SQLConnectionHandler
conn_handler = SQLConnectionHandler # doctest: +SKIP
conn_handler.create_queue("example_queue") # doctest: +SKIP
conn_handler.add_to_queue(
    "example_queue", "INSERT INTO qiita.qiita_user (email, name, password,"
    "phone) VALUES (%s, %s, %s, %s)",
    ['insert@foo.bar', 'Toy', 'pass', '111-111-11112']) # doctest: +SKIP
conn_handler.add_to_queue(
    "example_queue", "UPDATE qiita.qiita_user SET user_level_id = 1, "
    "phone = '222-222-2221' WHERE email = %s",
    ['insert@foo.bar']) # doctest: +SKIP
conn_handler.execute_queue("example_queue")
```
fails with 
```python
QiitaDBExecutionError: 
Error running SQL query in queue example_queue: INSERT INTO qiita.qiita_user (email, name, password,phone) VALUES (%s, %s, %s, %s)
ARGS: ['insert@foo.bar', 'Toy', 'pass', '111-111-11112']
Error: permission denied for schema qiita
LINE 1: INSERT INTO qiita.qiita_user (email, name, password,phone) V...
```

### LINKED ISSUE #1356: Modify sql_connection handler documentation to reflect the changes with Transactions
We need to rewrite the examples and I don't want this to hold the merge/testing of the new system. Thus, opening an issue to keep track of it although I'm compromising to have it done by sunday.

josenavas:
It also enables the doctests in this module.

Fixes #1356 
Fixes #982


qiita_db/sql_connection.py
```diff
@@ -22,57 +22,101 @@
...
-    ['{0}', '{1}']) # doctest: +SKIP
-conn_handler.execute_queue("example_queue") # doctest: +SKIP
-conn_handler.execute_fetchall(
-    "SELECT * from qiita.qiita_user WHERE email = %s", ['insert@foo.bar'])
-[['insert@foo.bar', 1, 'pass', 'Toy', None, None, '222-222-2221', None, None,
-  None]] # doctest: +SKIP
+
+>>> from qiita_db.sql_connection import TRN
+
+Transaction blocks are used through the variable TRN available in this module.
```
antgonza: Can you rephrase this as is not clear?

qiita_db/sql_connection.py
```diff
@@ -22,57 +22,101 @@
...
+...         TRN.add("SELECT 43")
+...         res = TRN.execute()
+...     # The transaction is still not committed
+...     TRN.add("SELECT 44")
+...     res = TRN.execute()
+>>> # The transactions committed here
+>>> res
+[[[42]], [[43]], [[44]]]
+
+You can also use results from a previous command in the transaction in a later
```
antgonza: Could you add a more complex example where we can se why we need the 3 ints?

qiita_db/sql_connection.py
```diff
@@ -22,57 +22,101 @@
...
+
+>>> with TRN:
+...     idx = TRN.index
+...     TRN.add("SELECT 42")
+...     TRN.add("SELECT 42 + %s", ["{%d:0:0}" % idx])
+...     res = TRN.execute()
+>>> res
+[[[42]], [[84]]]
+
+Multiple auxiliary functions exist for easy SQL result retrieval:
```
antgonza: Is there a way to add headers? Cause I can see having separate sections:
- Transactions
- Querying
- Retrival
- Etc

antgonza:
A few comments.

qiita_db/sql_connection.py
josenavas: Any suggestion?

qiita_db/sql_connection.py
josenavas: I don't think we can add headers per se in here, as the headers are a controlled set in the documentation (e.g. here the header is "Examples"). However I can try to add a unordered list...

josenavas:
Comments addressed, thanks @antgonza

antgonza:
Any idea why the test failure?

josenavas:
The error test is in the doctests. I've activated them for this file, but I've been having problems running them locally, I will take a look and see if I can run them locally and fix them. The turn around time from travis is quite big...

josenavas:
I just fixed the doctest. I'm worried about one thing here. The tests are passing, however, these tests are not safe as the unit tests are, in the sense that they don't fail if the database is not a test database.

I have ideas on how to fix it, however, this is not possible until our code supports python 3.4. I need the [FAIL_FAST](https://docs.python.org/3/library/doctest.html#doctest.FAIL_FAST) flag for doctest, so after the first failure no other doctest is executed.

The first call in the doctests here destroys the database, as happens in the unittest. I think I should skip the doctests for now (by adding SKIP), but I would like to hear from other before doing it, as it is time consuming.

antgonza:
:+1: could you add an issue about the tests?

josenavas:
Sure, added here: https://github.com/biocore/qiita/issues/1415

@squirrelo @wasade @ElDeveloper anyone available for review?

wasade:
going through

qiita_db/sql_connection.py
```diff
@@ -22,57 +22,139 @@
...
+>>> drop_and_rebuild_tst_database()  # doctest: +SKIP
+
+* Querying
+
+In order to perform any query in the database you first need to import the
+`TRN` variable available in this module:
+
+>>> from qiita_db.sql_connection import TRN
+
+The `TRN` variable is an instance of the Transaction object. All queries should
```
wasade: Transaction -> `Transaction`

qiita_db/sql_connection.py
```diff
@@ -22,57 +22,139 @@
...
+be executed through this transaction object to avoid committing anything to the
```
wasade: ~~transaction~~

qiita_db/sql_connection.py
```diff
@@ -22,57 +22,139 @@
...
+database that is partially executed. To perform any query you first need to
```
wasade: What about: "...to ensure atomicity." instead of "...to avoid committing ..."

qiita_db/sql_connection.py
```diff
@@ -22,57 +22,139 @@
...
+add the query to the transaction and then execute the transaction:
+
+>>> with TRN:
+...     TRN.add("SELECT 42")
+...     res = TRN.execute()
+>>> res
+[[[42]]]
```
wasade: A curious developer may wonder why 3 layers of nesting...

qiita_db/sql_connection.py
```diff
@@ -22,57 +22,139 @@
...
+...     TRN.add("SELECT 42")
+...     res = TRN.execute_fetchindex()
+>>> res
+[[42]]
+
+Getting the results of the specified SQL query
+>>> with TRN:
+...     TRN.add("SELECT 42")
+...     TRN.add("SELECT 43")
+...     res = TRN.execute_fetchindex(0)
```
wasade: It would be nice to explicitly note that the value `0` here corresponds to the first query in the transaction

qiita_db/sql_connection.py
```diff
@@ -22,57 +22,139 @@
...
+
+>>> with TRN:
+...     sql = ("SELECT param_set_name, preprocessed_params_id "
+...            "FROM qiita.preprocessed_sequence_454_params "
+...            "WHERE barcode_type = %s ORDER BY preprocessed_params_id")
+...     TRN.add(sql, ["golay_12"])
+...     sql = ("SELECT * FROM qiita.preprocessed_data "
+...            "WHERE preprocessed_params_table = %s "
+...            "AND preprocessed_params_id = %s")
+...     TRN.add(sql, ["preprocessed_sequence_454_params", "{0:0:1}"])
```
wasade: The indexing here is not intuitive since the a `SELECT *` is being used. Possible to change the `SELECT *` to project only a specific column or columns?

qiita_db/sql_connection.py
```diff
@@ -22,57 +22,139 @@
...
+...     TRN.add("SELECT 44")
+...     res = TRN.execute()
+>>> # The transactions committed here
+>>> res
+[[[42]], [[43]], [[44]]]
+
+You can also use results from a previous command in the transaction in a later
+command without partially executing the transaction by using placeholders. The
+placeholders are of the form {int:int:int}, where the first integer is the
+index of the query in the transaction, the second integer is the row in the
```
wasade: The row is difficult to understand since the cursor is not assured (AFAIK) to return rows in any particular order. In other words, the database is unordered unless a requested order is specified. I can see how this works when a single row is returned, but I can't help but thing anything besides `0` is meaningful unless the query includes an explicit ordering AND the developer knows in advance what the returned row of interest will be

wasade:
Few comments, :+1: otherwise

antgonza:
@josenavas ping

qiita_db/sql_connection.py

qiita_db/sql_connection.py

qiita_db/sql_connection.py
josenavas: Really good suggestion!

qiita_db/sql_connection.py
josenavas: Added a comment explaining the 3 levels of nesting.

qiita_db/sql_connection.py

qiita_db/sql_connection.py
josenavas: @wasade the indexing is used to index the previous query (hence the 0 index) so it's actually selecting the `preprocessed_params_id` from the first row in the previous query

wasade: Okay... so this is extremely confusing and not something that, in my opinion, is intuitive. I suggest adding a few more examples of the placeholders with detailed documentation working through the examples, and/or update the placeholders to be expressive.

qiita_db/sql_connection.py
josenavas: You're correct, the order in the results is not assured unless a requested order is specified.

qiita_db/sql_connection.py
```diff
@@ -22,57 +22,144 @@
...
+...     res = TRN.execute_fetchflatten(0)
+>>> res
+[42, 43, 44]
+
+* Transactions
+
+Transaction blocks are created through the same `TRN` variable exported in this
+module. You can add as many SQL commands as you want and execute all of them at
+once, and it will return the results of all the SQL commands. `TRN` should be
+used as a context so manager, and it autocommits the transaction once the last
```
RNAer: "context so manager"?

qiita_db/sql_connection.py
```diff
@@ -22,57 +22,144 @@
...
+context is exited, as long as no error was generated inside the context, in
+which case a rollback is executed.
+
```
RNAer: it is not clear to me what rollback is executed. It returns to the previous state before any added commands were run?

qiita_db/sql_connection.py
```diff
@@ -22,57 +22,144 @@
...
+...     TRN.add("SELECT 42")
+...     TRN.add("SELECT 42 + %s", ["{0:0:0}"])
+...     res = TRN.execute()
+>>> res
+[[[42]], [[84]]]
+
+>>> with TRN:
+...     sql = ("SELECT param_set_name, preprocessed_params_id "
+...            "FROM qiita.preprocessed_sequence_454_params "
+...            "WHERE barcode_type = %s ORDER BY preprocessed_params_id")
```
RNAer: Doesn't this violate the SQL coding guidelines advocating the usage of `"""`?

qiita_db/sql_connection.py
josenavas: Good catch! Fixed.

qiita_db/sql_connection.py
josenavas: Yes, by definition of transaction, any command added to the current transaction is undone. Note that everything that has been already committed it will not be included in the rollback. Ex:

```python
with TRN:
    # (A) do stuff

with TRN:
    # (B) do stuff
    raise ValueError
```
In the above example, the stuff in (A) is committed, but the stuff in (B) is rollbacked. On the other hand:
```python
with TRN:
    with TRN:
        # (A) do stuff

    with TRN:
        # (B) do stuff
        raise ValueError
```
neither (A) or (B) are committed, so both of them will be rollbacked.

qiita_db/sql_connection.py
josenavas: You're correct. I removed this example because it is using the placeholders and we no longer need them.

josenavas:
Ok this is ready for another review. I removed the placeholders code, updated the portions of the code that were affected (only 2 functions) and added the Transaction to the CHANGELOG (it looks like the previous PR didn't got merged in master?)

antgonza:
:+1:

josenavas:
@RNAer or @wasade do you mind taking another look and merging if there are no more comments?

RNAer:
looking at it now

RNAer:
:+1:

antgonza:
This has 2 :+1: so I'm gonna merge.

PR #1410, opened 2015-08-07
Allow UTF8 characters in the study title, alias, description and abstract

### LINKED ISSUE #1300: Studies with weird characters in their description break the interface
We found a similar issue that the one with the PI. If there is a weird character it breaks the GUI. The solution should be the same to the one applied to the PI.

josenavas:
Fixes #1300 
the limitations was in WTForms. It was already solved for the PI and Lab Person, but it was not fixed for the rest of the attributes of the study that this character can occur. It also adds a specific test.

Pubmed id should never include non-ascii characters, so I think it is fine leaving it as it is.

screen shot 2015-08-07 at 12 57 46 pm

### DIFF
--- a/qiita_pet/handlers/study_handlers/edit_handlers.py
+++ b/qiita_pet/handlers/study_handlers/edit_handlers.py
@@ -78,11 +78,13 @@ class StudyEditorForm(Form):
         if study:
             study_info = study.info
 
-            self.study_title.data = study.title
-            self.study_alias.data = study_info['study_alias']
+            self.study_title.data = study.title.decode('utf-8')
+            self.study_alias.data = study_info['study_alias'].decode('utf-8')
             self.pubmed_id.data = ",".join(study.pmids)
-            self.study_abstract.data = study_info['study_abstract']
-            self.study_description.data = study_info['study_description']
+            self.study_abstract.data = study_info[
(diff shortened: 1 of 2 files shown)


qiita_pet/handlers/study_handlers/edit_handlers.py
```diff
@@ -78,11 +78,13 @@ def __init__(self, study=None, **kwargs):
         if study:
             study_info = study.info
 
-            self.study_title.data = study.title
-            self.study_alias.data = study_info['study_alias']
+            self.study_title.data = study.title.decode('utf-8')
+            self.study_alias.data = study_info['study_alias'].decode('utf-8')
             self.pubmed_id.data = ",".join(study.pmids)
```
squirrelo: This is still a freetext field, so it does have the possibility of having non UTF-8 characters. Better to add the decode.

josenavas: Ok, although we should start checking that those are valid PMIDS. However, we should pospone this checking until the new publication table is implemented. I was planning to do so, but I don't want to have a merge conflict with your open PR.

antgonza:
:+1:

josenavas:
@squirrelo can you merge if you're happy with the code?

PR #1412, opened 2015-08-09
Fix #1401

### LINKED ISSUE #1401: download mapping file and biom tables associated with an analysis
It would be nice for the user to have the ability to download the mapping file and the biom tables associated with an analysis.

antgonza:
This fixes #1401. Note that I moved is_local_connection out of BaseUIModule and added to util so it could be used everywhere. Additionally, I changed the parameter of this func from self to self.request.headers['host'] so it expects a string and it can be tested easier.


qiita_pet/test/test_util.py
```diff
@@ -28,6 +28,14 @@ def test_generate_param_str(self):
                "<b>sortmerna_coverage:</b> 0.97")
         self.assertEqual(obs, exp)
 
+    def test_is_local_connection(self):
+        self.assertTrue(is_local_connection('127.0.0.1'))
+        self.assertTrue(is_local_connection('localhost'))
+        self.assertTrue(is_local_connection('127.0.0.1:21174'))
+
+        self.assertFalse(is_local_connection('10.0.0.1'))
```
wasade: How do you know that `10.0.0.1` isn't local?

antgonza: That's a good point but is also the assumption made since the beginning. Note that this is a "nice" behavior as we can test as localhost and as external connection from the same machine.

wasade: It'd be more correct then if the function name was `is_localhost` as that
has a specific meaning. Should also verify that `::1` is resolved as
expected.

On Mon, Aug 10, 2015 at 10:05 AM, Antonio Gonzalez <notifications@github.com

qiita_pet/templates/show_analyses.html
```diff
@@ -51,6 +55,16 @@
       {% end %}
         </td>
         <td>
+          {% if mappings[_id] is not None %}
+            {% raw download_link_or_path(is_local_request, mappings[_id]['mapping'], mappings[_id]['id'], 'mapping file') %}
```
josenavas: We have tried to remove as much python code from the templates as possible. Can you move this to the handler itself? At the end, what the template wants is for each analysis have a tuple with 2 elements, the mapping file (either link or filepath) and the list of biom tables...

qiita_pet/templates/show_analyses.html
```diff
@@ -21,6 +22,8 @@
   <table class="table">
     <tr>
       <th>Analysis Name</th>
+      <th>Mapping File</th>
```
josenavas: Can you move the mapping file and the biom file columns to be at the end, before the delete button? Or at least after the analysis description columns, it makes a better organization.

Ideally, I think we should put this columns in an "expand" as it is done for the processed data in the study list, so we can reduce how crowded this page looks. What does others think? cc @biocore/qiita-dev

josenavas:
Couple of comments, thanks @antgonza

antgonza:
Addressed both comments.

josenavas:
Thanks @antgonza :+1: 

@wasade can you review/merge given that you've already done some review in here?

qiita_pet/handlers/analysis_handlers.py
```diff
@@ -215,8 +217,27 @@ def get(self):
...
+        for analysis in analyses:
+            _id = analysis.id
+            mapping = analysis.mapping_file
+            if mapping is not None:
+                mappings[_id] = download_link_or_path(
+                    is_local_request, mapping,
+                    get_filepath_id('analysis', mapping), 'mapping file')
+            else:
+                mappings[_id] = ''
+            bioms[_id] = '\n'.join(
```
wasade: non-blocking, but this is unnecessarily complex...

wasade:
One comment, :+1: otherwise

qiita_pet/handlers/analysis_handlers.py
antgonza: if non-blocking, that means that is OK to merge, right? If you feel like I should make it less complex, could you give a suggestion? Perhaps separate the join and the list creation?

wasade:
```python
get_analysis_fp = partial(get_filepath_id, 'analysis')  # can be used on line 230 as well
dl_or_p = partial(download_link_or_path, is_local_request) # can be used on line 228 as well

links = []
for l, f in viewitems(analysis.biom_tables):
    links.append(dl_or_p(f, get_analysis_fp(f), l)
bioms[_id] = '\n'.join(links)
```

josenavas: Can links be a list comprehension? There is no need to the re-allocs of doing the `append` and this is not that complex...
```python
links = [dl_or_p(f, get_analysis_fp(f), l) for l, f in viewitems(analysis.biom_tables)]
```

wasade: sure, but that spans lines and multiple line comprehensions are semantically complex

antgonza: What about?
```
        is_local_request = is_localhost(self.request.headers['host'])
        gfi = partial(get_filepath_id, 'analysis')
        dlop = partial(download_link_or_path, is_local_request)
        mappings = {}
        bioms = {}
        for analysis in analyses:
            _id = analysis.id
            mapping = analysis.mapping_file
            if mapping is not None:
                mappings[_id] = dlop(mapping, gfi(mapping), 'mapping file')
            else:
                mappings[_id] = ''
            links = [dlop(f, gfi(f), l)
                     for l, f in viewitems(analysis.biom_tables)]
            bioms[_id] = '\n'.join(links)
```

josenavas: It is fine with me.

wasade:
will merge following tests

antgonza:
@wasade ping

PR #1416, opened 2015-08-12
fix #1080

### LINKED ISSUE #1080: remove remove_add_study_template
description_handlers.py has remove_add_study_template, which basically allowed you to update a sample template by deleting its prep template, raw data and recreating it without recreating those other deleted object. IMOO this should not be allowed and should be removed.

antgonza:


qiita_pet/handlers/study_handlers/description_handlers.py
```diff
@@ -188,12 +188,19 @@ def process_sample_template(self, study, user, callback):
         msg_level = "success"
         is_mapping_file = looks_like_qiime_mapping_file(fp_rsp)
 
+        if is_mapping_file and data_type == "":
```
squirrelo: You made data_type an int explicitly above, so how does this if statement work?

antgonza: Early. Fixed back ...

wasade: Is `data_type` only ever a `str`? If so, then this can be reduced to `if is_mapping_file and not data_type`

antgonza: @josenavas?

josenavas: Yes, it is always a string since that is what tornado returns. The "non selected" value is coded as an empty string `""`, I think @wasade proposal can be done as I think that is the standard way of checking against an empty string.

antgonza: K fixing.

josenavas:
@antgonza I'm not exactly sure how your changes broke this, but it looks like the raised ValueError is no longer caught:

gif

antgonza:
Thanks for catching. Fixing now.

josenavas:
Thanks @antgonza 

:+1:

antgonza:
@squirrelo wanna do a final pass so we can merge?

antgonza:
@squirrelo ping or someone else that could do a quick review?

PR #1417, opened 2015-08-12
fix #1213

### LINKED ISSUE #1213: add citing qiita section in the documentation
Currently mention that they should cite qiita.microbio.me

antgonza:


qiita_pet/templates/sitebase.html
```diff
@@ -33,15 +33,15 @@
...
       function overlay_check() {
         // from http://stackoverflow.com/q/5916900/379593
-        var ua=navigator.userAgent,tem,M=ua.match(/(opera|chrome|safari|firefox|msie|trident(?=\/))\/?\s*(\d+)/i) || []; 
+        var ua=navigator.userAgent,tem,M=ua.match(/(opera|chrome|safari|firefox|msie|trident(?=\/))\/?\s*(\d+)/i) || [];
         if(/trident/i.test(M[1])){
-            tem=/\brv[ :]+(\d+)/g.exec(ua) || []; 
+            tem=/\brv[ :]+(\d+)/g.exec(ua) || [];
             return 'IE '+(tem[1]||'');
-            }   
+            }
```
squirrelo: since this editing is happening, please align this with the if statement. same for the one below.

qiita_pet/templates/sitebase.html
```diff
@@ -239,7 +239,8 @@
       {% block content %}{% end %}
     </div>
 
-    <div id="footer">
+    <div id="footer" style="font-size: 12px; text-align: center; z-index: 1000;">
+      Thank you for using Qiita. For citation point to <a href="http://qiita.microbio.me">http://qiita.microbio.me</a>.
```
squirrelo: Either `For citations` or `To cite qiita` for grammar.

antgonza: K

qiita_pet/templates/sitebase.html
```diff
@@ -239,7 +239,8 @@
       {% block content %}{% end %}
     </div>
 
-    <div id="footer">
+    <div id="footer" style="font-size: 12px; text-align: center; z-index: 1000;">
+      Thank you for using Qiita. For citations point to <a href="http://qiita.microbio.me">http://qiita.microbio.me</a>.
```
wasade: What about with using a different portal?

antgonza: I think qiita's reference should not change.

wasade: Interesting, Chrome thinks this page is in Portuguese...

It may be the case that in the future there are publications built on top of portals as a resource paper. But that can be resolved then I guess.

On an aside, any reason qiita is not using `https`?

antgonza: https: None except that we do not have it running there. If we decide to do that we can always redirect from http -> https so http at this point is safer.

PR #1420, opened 2015-08-13
Fix nan and bool values

squirrelo:
This is a small pull request that expands what we consider NaN and bool values. This comes after looking at the AG18 load and realizing that everything was considered a string, since the blanks all were 'unknown' for all columns.

### DIFF
--- a/qiita_db/metadata_template/constants.py
+++ b/qiita_db/metadata_template/constants.py
@@ -71,13 +71,14 @@ ALL_RESTRICTIONS = [SAMPLE_TEMPLATE_COLUMNS, PREP_TEMPLATE_COLUMNS,
                     PREP_TEMPLATE_COLUMNS_TARGET_GENE]
 
 # This is what we consider as "NaN" cell values on metadata import
-NA_VALS = ['', 'no_data', 'unknown', 'Unspecified', 'unspecified']
+NA_VALUES = ['', 'no_data', 'unknown', 'Unspecified', 'unspecified']
 
 # These are what will be considered 'True' bool values on metadata import
-META_TRUE = ['Yes', 'yes', 'YES', 'Y', 'y', 'True', 'true', 'TRUE', 't', 'T']
+TRUE_VALUES = ['Yes', 'yes', 'YES', 'Y', 'y', 'True', 'true', 'TRUE', 't', 'T']
 
 # These are what will be considered 'False' bool values on metadata import
-META_FALSE = ['No', 'no', 'NO', 'N', 'n', 'False', 'false', 'FALSE', 'f', 'F']
+FALSE_VALUES = ['No', 'no', 'NO', 'N', 'n', 'False', 'false', 'FALSE',
+                'f', 'F']
 
 
 # A set holding all the controlled columns, useful to avoid recalculating it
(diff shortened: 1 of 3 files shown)


qiita_db/metadata_template/constants.py
```diff
@@ -70,6 +70,15 @@
 ALL_RESTRICTIONS = [SAMPLE_TEMPLATE_COLUMNS, PREP_TEMPLATE_COLUMNS,
                     PREP_TEMPLATE_COLUMNS_TARGET_GENE]
 
+# This is what we consider as "NaN" cell values on metadata import
+NA_VALS = ['', 'no_data', 'unknown', 'Unspecified']
```
josenavas: `NA_VALS` -> `NAN_VALS`

@biocore/qiita-dev in my opinion, N/A has a different meaning in metadata and we should keep it (i.e. is not the same having a null, no_data, unknown, and Unspecified that an N/A). Hence the name change here.

Can you also add 'unspecified'?

squirrelo: I can add that. `XXQIITAXX` should probably be considered NaN as well.

josenavas: Uhm, I don't think that is correct. I think we want to maintain `XXQIITAXX` as those are indicative that they need to change those columns.

squirrelo: Ok, left. Not changing the NA_VALS variable name though, since the actual ` pd.read_csv` function calls it `na_values`

squirrelo:
Changes made

josenavas:
:+1:

antgonza:
Failure

squirrelo:
Fixed

qiita_db/metadata_template/constants.py
```diff
@@ -70,6 +70,15 @@
 ALL_RESTRICTIONS = [SAMPLE_TEMPLATE_COLUMNS, PREP_TEMPLATE_COLUMNS,
                     PREP_TEMPLATE_COLUMNS_TARGET_GENE]
 
+# This is what we consider as "NaN" cell values on metadata import
+NA_VALS = ['', 'no_data', 'unknown', 'Unspecified', 'unspecified']
```
antgonza: IMOO for easy search/replace the 3 list should be: META_* or *_VALS. Could you change?

antgonza:
@squirrelo ping!

qiita_db/metadata_template/constants.py
squirrelo: Search/replace has already bit us before, so I'm hesitant to make a change solely on that basis.

antgonza: I think this this is a change needed for consistency in the variable names. BTW you mentioned that """Ok, left. Not changing the NA_VALS variable name though, since the actual pd.read_csv function calls it na_values""" but didn't keep it for the others ...

squirrelo: True. In that case, changing to the names used in the function for consistency.

squirrelo:
Rename done, ready for review again

antgonza:
Thanks, could you either make NA_VALS -> NA_VALUES or the other ones TRUE/FALSE_VALS?

squirrelo:
It is?

antgonza:
Thanks. My computer is playing mind games with me! Gonna wait for test.

PR #1422, opened 2015-08-17
Add patch to convert nulls and boolan cols

squirrelo:


squirrelo:
This patch is ready to go.

josenavas:
Testing it locally right now

qiita_db/support_files/patches/python_patches/30.py
```diff
@@ -0,0 +1,63 @@
...
+
+        # Update now boolean columns to bool in database
+        TRN.add("SELECT * FROM qiita.{}".format(table))
+        col_vals = zip(*TRN.execute_fetchindex())
+        cols = [x[0] for x in colinfo]
+        for col, vals in zip(cols, col_vals):
+            if set(vals) == {None}:
+                # Ignore columns that are all NULL
+                continue
+            if all(v in bool_vals for v in vals):
```
josenavas: change this line by:
```python
if all([v in bool_vals for v in vals]):
```

and this error disappears:
```
$ qiita-env patch
  File "/Users/jose/qiime_software/QiiTa/scripts/qiita-env", line 183, in <module>
    env()
  File "/Users/jose/.virtualenvs/qiita/lib/python2.7/site-packages/click/core.py", line 664, in __call__
  File "/Users/jose/.virtualenvs/qiita/lib/python2.7/site-packages/click/core.py", line 644, in main
    rv = self.invoke(ctx)
  File "/Users/jose/.virtualenvs/qiita/lib/python2.7/site-packages/click/core.py", line 991, in invoke
  File "/Users/jose/.virtualenvs/qiita/lib/python2.7/site-packages/click/core.py", line 837, in invoke
  File "/Users/jose/.virtualenvs/qiita/lib/python2.7/site-packages/click/core.py", line 464, in invoke
  File "/Users/jose/qiime_software/QiiTa/scripts/qiita-env", line 112, in patch
    _patch()
  File "/Users/jose/qiime_software/QiiTa/qiita_db/environment_manager.py", line 401, in patch
    execfile(py_patch_fp)
  File "/Users/jose/qiime_software/QiiTa/qiita_db/support_files/patches/python_patches/30.py", line 56, in <module>
    if all(v in bool_vals for v in vals):
  File "/Users/jose/qiime_software/QiiTa/qiita_db/support_files/patches/python_patches/30.py", line 56, in <genexpr>
    if all(v in bool_vals for v in vals):
NameError: global name 'bool_vals' is not defined
```

qiita_db/support_files/patches/python_patches/30.py
```diff
@@ -0,0 +1,63 @@
...
+nans = tuple(NA_VALUES)
+false_vals = tuple(FALSE_VALUES)
+true_vals = tuple(TRUE_VALUES)
+
+with TRN:
+    sql = """SELECT table_name
+             FROM information_schema.tables
+             WHERE table_schema='qiita'
+                AND (table_name SIMILAR TO 'sample\_[0-9]+'
+                OR table_name SIMILAR TO 'prep\_[0-9]+')"""
```
josenavas: Missing one indentation here (the OR is inside the parenthesis)

qiita_db/support_files/patches/python_patches/30.py
```diff
@@ -0,0 +1,63 @@
...
+    TRN.add(sql)
+    tables = [x[0] for x in TRN.execute_fetchindex()]
```
josenavas: use `fetchflatten`

qiita_db/support_files/patches/python_patches/30.py
```diff
@@ -0,0 +1,63 @@
...
+                # Ignore columns that are all NULL
+                continue
+            if all(v in bool_vals for v in vals):
+                # Every value in the column should be bool, so do it
+                TRN.add(alter_sql.format(table, col),
+                        [false_vals, true_vals])
+                if "sample" in table:
+                    TRN.add(ssc_update_sql, [table_id, col])
+                else:
+                    TRN.add(pc_update_sql.format(), [table_id, col])
```
josenavas: Why are you calling format here?

josenavas:
@squirrelo you're missing regenerating all the files.
For each sample template, you need to call `generate_files()`
For each prep template that is NOT related with any of the previous sample templates, you need to call `generate_files()`. Note that `SampleTemplate.generate_files` calls PrepTemplate.generate_files`, so that's why you only need to call this function to the prep templates that are not related with previous sample templates.

There are also some comments in the code.

qiita_db/support_files/patches/python_patches/30.py
```diff
@@ -0,0 +1,63 @@
...
+        TRN.add(cols_sql, [table])
+        colinfo = TRN.execute_fetchindex()
+        for col, ctype in colinfo:
+            if ctype == 'character varying':
+                TRN.add(null_sql.format(table, col), [nans])
+        TRN.execute()
+
+        # Update now boolean columns to bool in database
+        TRN.add("SELECT * FROM qiita.{}".format(table))
+        col_vals = zip(*TRN.execute_fetchindex())
```
josenavas: I was worried about these lines of codes you can't assume that the result of the last query and the results of cols are in the same order, so the `zip(cols, col_vals)` that you are doing in line 52 might not be pairing the columns with the actual values.

qiita_db/support_files/patches/python_patches/30.py
squirrelo: That makes no sense, but cool find.

qiita_db/support_files/patches/python_patches/30.py
squirrelo: They are by nature of how SQL works, but I can be explicit in column order in line 49 to make extra paranoid sure.

qiita_db/support_files/patches/python_patches/30.py
squirrelo: leftover form previous iteration.

squirrelo:
I still don't understand why we have file versions of what's in the database, and it has never been adequately explained to me. Generating files from the database is already taken care of, so why are we not doing that on the fly?

josenavas:
In short: generating files is time consuming, having them pre-computed is faster.
We can discuss this in another moment but please, add now the re-generation of the files as the system expects them to be the most up to date.

qiita_db/support_files/patches/python_patches/30.py
```diff
@@ -0,0 +1,63 @@
...
+        # Change NaN values to NULL in database
+        TRN.add(cols_sql, [table])
+        colinfo = TRN.execute_fetchindex()
+        for col, ctype in colinfo:
+            if ctype == 'character varying':
+                TRN.add(null_sql.format(table, col), [nans])
+        TRN.execute()
+
+        # Update now boolean columns to bool in database
+        cols = [x[0] for x in colinfo]
```
josenavas: I just noticed this: here you're retrieving all the columns, however, we are only interested in the ones that are 'character varying'. Shouldn't the `cols_sql` command retrieve only the `varchar` columns?

Note that all the values that you have in the `NA_VALUES`, `TRUE_VALUES` and `FALSE_VALUES` are strings (varchars) so the if statement in line 56 will return true iif is a varchar column.

Given that some template can be quite big and have a fair amount of rows, reducing the amount of loop iterations is important.

squirrelo:
Already added, but if it's expensive we should really rethink it. It should be relatively lightweight and fast considering it's essentially dumping a table to stringIO or file.

josenavas:
@squirrelo if you added it, is not reflected in the code in this PR, forgot to push?

squirrelo:
guess so.

josenavas:
Re lightweight and fast: it depends on the size of the sample template and the number of columns, specially in generating the QIIME mapping file.

You need to query the entire sample template, merge it with the entire prep template and write it to a file. Let's say that we have 20k samples. The amount of data retrieved is not trivial...

squirrelo:
No, it's not a small amount of data, but the formatting of it is extremely fast, especially if we have SQL do the actual tab formatting, and that way we never worry about the files and DB being out of sync. Agreed we should have this discussion at the next meeting though.

qiita_db/support_files/patches/python_patches/30.py
squirrelo: Not sure what you mean by logic in line 56 since it will only return true if every value in a column matches that list, but data_type change made.

josenavas: in line 56 you are asking if all the values in the current column are in the set of `bool_vals`. `bool_vals` is a set of string, so only the `varchar` columns will return make that if statement to evaluate `True`

squirrelo:
Ready for another round of review.

qiita_db/support_files/patches/python_patches/30.py
```diff
@@ -0,0 +1,75 @@
...
+                    st_update.add(table_id)
+                    TRN.add(ssc_update_sql, [table_id, col])
+                else:
+                    pr_update.add(table_id)
+                    TRN.add(pc_update_sql, [table_id, col])
+
+        TRN.execute()
+        for stid in st_update:
+            SampleTemplate(int(stid)).generate_files()
+        for prid in pr_update:
```
josenavas: This is not correct, please read carefully my previous comment about re-generating the files for the prep templates:
>For each prep template that is NOT related with any of the previous sample templates, you need to call generate_files(). Note that SampleTemplate.generate_files calls PrepTemplate.generate_files`, so that's why you only need to call this function to the prep templates that are not related with previous sample templates.

That is, if the prep template does not belong to the same study that any of the previous sample template.

Also, note that you've these 2 for loops over-indented and they're inside the for loop, which means that you are re-generating the same files up to `len(tables)` times...

qiita_db/support_files/patches/python_patches/30.py
```diff
@@ -0,0 +1,75 @@
+from qiita_db.sql_connection import TRN
+from qiita_db.metadata_template.constants import (NA_VALUES, TRUE_VALUES,
+                                                  FALSE_VALUES)
+from qiita_db.metadata_template.prep_template import PrepTemplate
```
josenavas: `from qiita_db.metadata_template import SampleTemplate, PrepTemplate` and remove line 5

qiita_db/support_files/patches/python_patches/30.py
squirrelo: This was fixed before you commented, please reload the page before commenting.

squirrelo:
ready for review.

qiita_db/support_files/patches/python_patches/30.py
```diff
@@ -0,0 +1,79 @@
...
+                    st_update.add(table_id)
+                    TRN.add(ssc_update_sql, [table_id, col])
+                else:
+                    pr_update.add(table_id)
+                    TRN.add(pc_update_sql, [table_id, col])
+
+    TRN.execute()
+    for stid in st_update:
+        SampleTemplate(int(stid)).generate_files()
+        for rd_id in Study(int(stid)).raw_data():
```
josenavas: This is incorrect. Please `Study.prep_templates`. There might be PrepTemplates that do not have RawData, so will not detect them through this route.

qiita_db/support_files/patches/python_patches/30.py
```diff
@@ -0,0 +1,79 @@
...
+                        SET column_type = 'bool'
+                        WHERE prep_template_id = %s AND column_name = %s"""
+
+    for table in tables:
+        table_id = table.split("_")[1]
+        # Change NaN values to NULL in database
+        TRN.add(cols_sql, [table])
+        cols = TRN.execute_fetchflatten()
+        for col in cols:
+                TRN.add(null_sql.format(table, col), [nans])
```
josenavas: Line over-indented

josenavas:
A couple more comments

qiita_db/support_files/patches/python_patches/30.py
squirrelo: This was a direct copy paste from the sample template `generate_files` function, so I'm leaving it since that's how the function itself checks if prep templates need to be updated. Doing it that way has the potential to skip updating prep templates that are not found by RawData but are by prep_templates.

josenavas: @squirrelo duplicating bugs is not a good practice. That means that we found a bug in a previous template, and the reason is that THE STRUCTURE OF THE DATABASE CHANGED SINCE THEN! Please change as this is a bug.

squirrelo: So I should fix the SampleTemplate and PrepTemplate code as well? Because that could potentially become a large fix and is well outside the scope of this patch pull request.

josenavas: Please change, as it is a bug and it is NOT a large fix.

josenavas:
:+1: @antgonza @ElDeveloper can any of you review/merge this PR? This is blocking the release.

qiita_db/support_files/patches/python_patches/30.py
```diff
@@ -0,0 +1,78 @@
+from qiita_db.sql_connection import TRN
+from qiita_db.metadata_template.constants import (NA_VALUES, TRUE_VALUES,
+                                                  FALSE_VALUES)
+from qiita_db.metadata_template import SampleTemplate, PrepTemplate
+from qiita_db.study import Study
+
+bool_vals = set(TRUE_VALUES + FALSE_VALUES + [None])
```
wasade: `None` is not a `bool`

qiita_db/support_files/patches/python_patches/30.py
```diff
@@ -0,0 +1,78 @@
...
+true_vals = tuple(TRUE_VALUES)
+
+st_update = set()
+pr_update = set()
+
+with TRN:
+    sql = """SELECT table_name
+             FROM information_schema.tables
+             WHERE table_schema='qiita'
+                AND (table_name SIMILAR TO 'sample\_[0-9]+'
```
wasade: are there study numbers that start with a `0`?

wasade:
just 2 comments, i don't believe they block. 

strongly recommend doing a dump of the db (e.g., `pg_dump -d qiita -f foo.sql`) prior to applying as this is a massive change.

qiita_db/support_files/patches/python_patches/30.py
squirrelo: It is considered for the bool col check though, since a bool col can have True, False, NULL.

qiita_db/support_files/patches/python_patches/30.py
squirrelo: I don't think so, but this works fine regardless because it makes sure that only integers follow the underscore, so the 0 is needed in the brackets.

josenavas:
@wasade can you merge if your concerns have been addressed?

josenavas:
@wasade told me it is fine to merge, but he is not on his computer, so merging. Thanks!

PR #1429, opened 2015-08-20
Solving np.nan problems

josenavas:
when pandas parses blanks, it always represent them as np.nan. This is a problem since the np.nan are Floats, but we might want to insert them in other data types (e.g. bool or varchar). This PR forces the template to represent all np.nan as None, so they'll be inserted as 'Null' in the DB.

In order to use the `DataFrame.where` function, the DataFrame should not have duplicate columns. We where allowing them in the load function, but raising an error in the `create` step. I think it is ok to already raise an error in the load function, so we execute less work. I've changed that here and updated the relevant test.

### DIFF
--- a/qiita_db/metadata_template/util.py
+++ b/qiita_db/metadata_template/util.py
@@ -321,6 +321,9 @@ def load_template_to_dataframe(fn, strip_whitespace=True, index='sample_name'):
                       'because all their values are empty: '
                       '%s' % ', '.join(dropped_cols), QiitaDBWarning)
 
+    # Pandas represents data with np.nan rather than Nones, change it
+    template = template.where((pd.notnull(template)), None)
+
     return template
 
 
--- a/qiita_db/metadata_template/util.py
+++ b/qiita_db/metadata_template/util.py
@@ -328,7 +328,9 @@ def load_template_to_dataframe(fn, strip_whitespace=True, index='sample_name'):
                       'because all their values are empty: '
                       '%s' % ', '.join(dropped_cols), QiitaDBWarning)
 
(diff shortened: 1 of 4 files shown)


qiita_db/metadata_template/util.py
```diff
@@ -321,6 +328,9 @@ def load_template_to_dataframe(fn, strip_whitespace=True, index='sample_name'):
                       'because all their values are empty: '
                       '%s' % ', '.join(dropped_cols), QiitaDBWarning)
 
+    # Pandas represents data with np.nan rather than Nones, change it
```
ElDeveloper: This should instead allude to psycopg and the conversion from None to NULL.

ElDeveloper:
Aside from the missing test cases that we discussed offline, and the one minor comment I had, this looks good.

qiita_db/metadata_template/util.py
```diff
@@ -321,6 +328,9 @@ def load_template_to_dataframe(fn, strip_whitespace=True, index='sample_name'):
...
+    template = template.where((pd.notnull(template)), None)
```
ElDeveloper: What about using `fillna`? http://pandas.pydata.org/pandas-docs/stable/generated/pandas.DataFrame.fillna.html

josenavas: I tried, it doesn't work.

qiita_db/metadata_template/util.py
josenavas: Done

ElDeveloper:
:+1:

PR #1430, opened 2015-08-21
Portal stats 1427

### LINKED ISSUE #1427: portal stats reflect system wide results
We should make it specific to the portal.

adamrp:
Fixes #1427 

Note that if you switch portals, you have to clear the redis keys manually for this to work properly. Maybe we should clear redis keys automatically when webserver is launched?

### DIFF
--- a/qiita_db/meta_util.py
+++ b/qiita_db/meta_util.py
@@ -183,17 +183,16 @@ def get_lat_longs():
     with TRN:
         sql = """SELECT DISTINCT table_name
                  FROM information_schema.columns
-                 WHERE SUBSTR(table_name, 1, 7) = 'sample_'
+                 WHERE table_name SIMILAR TO 'sample_[0-9]+'
                     AND table_schema = 'qiita'
-                    AND column_name IN ('latitude', 'longitude');"""
-        TRN.add(sql)
+                    AND column_name IN ('latitude', 'longitude')
+                    AND SPLIT_PART(table_name, '_', 2)::int IN %s;"""
+        TRN.add(sql, [tuple(portal_table_ids)])
 
         sql = "SELECT latitude, longitude FROM qiita.{0}"
         idx = TRN.index
 
         portal_tables = TRN.execute_fetchflatten()
-        portal_tables = [x for x in TRN.execute_fetchflatten()
-                         if int(x[7:]) in portal_table_ids]
 
         for table in portal_tables:
             TRN.add(sql.format(table))
--- a/qiita_db/meta_util.py
+++ b/qiita_db/meta_util.py
@@ -14,6 +14,7 @@ Methods
     :toctree: generated/
 
     get_accessible_filepath_ids
+    get_lat_longs
 """
 # -----------------------------------------------------------------------------
 # Copyright (c) 2014--, The Qiita Development Team.
@@ -24,11 +25,15 @@ Methods
 # -----------------------------------------------------------------------------
 from __future__ import division
 
+from itertools import chain
+
+from qiita_core.qiita_settings import qiita_config
 from .study import Study
 from .data import RawData, PreprocessedData, ProcessedData
 from .analysis import Analysis
 from .sql_connection import TRN
 from .metadata_template import PrepTemplate, SampleTemplate
+from .portal import Portal
 
 
 def _get_data_fpids(constructor, object_id):
@@ -163,3 +168,34 @@ def get_accessible_filepath_ids(user):
             filepath_ids.update(analysis.all_associated_filepath_ids)
 
         return filepath_ids
+
+
+def get_lat_longs():
+    """Retrieve the latitude and longitude of all the samples in the DB
+
+    Returns
+    -------
+    list of [float, float]
+        The latitude and longitude for each sample in the database
+    """
+    portal_table_ids = Portal(qiita_config.portal).get_studies()
+
+    with TRN:
+        sql = """SELECT DISTINCT table_name
+                 FROM information_schema.columns
+                 WHERE SUBSTR(table_name, 1, 7) = 'sample_'
+                    AND table_schema = 'qiita'
+                    AND column_name IN ('latitude', 'longitude');"""
+        TRN.add(sql)
+
+        sql = "SELECT latitude, longitude FROM qiita.{0}"
+        idx = TRN.index
+
+        portal_tables = TRN.execute_fetchflatten()
+        portal_tables = [x for x in TRN.execute_fetchflatten()
+                         if int(x[7:]) in portal_table_ids]
+
+        for table in portal_tables:
+            TRN.add(sql.format(table))
+
+        return list(chain.from_iterable(TRN.execute()[idx:]))
--- a/qiita_db/test/test_meta_util.py
+++ b/qiita_db/test/test_meta_util.py
@@ -7,15 +7,27 @@
 # -----------------------------------------------------------------------------
 
 from unittest import TestCase, main
+from datetime import datetime
 
+import pandas as pd
+
+from qiita_core.qiita_settings import qiita_config
 from qiita_core.util import qiita_test_checker
-from qiita_db.meta_util import get_accessible_filepath_ids
+from qiita_db.meta_util import get_accessible_filepath_ids, get_lat_longs
 from qiita_db.study import Study
 from qiita_db.user import User
+from qiita_db.portal import Portal
+from qiita_db.metadata_template.sample_template import SampleTemplate
 
 
 @qiita_test_checker()
 class MetaUtilTests(TestCase):
+    def setUp(self):
+        self.old_portal = qiita_config.portal
+
+    def tearDown(self):
+        qiita_config.portal = self.old_portal
+
     def _set_processed_data_private(self):
         self.conn_handler.execute(
             "UPDATE qiita.processed_data SET processed_data_status_id=3")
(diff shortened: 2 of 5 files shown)


ElDeveloper:
@adamrp, there's a few failures, would you mind having a look?

josenavas:
@adamrp if you need a new study just use the 'Study.create' function, no need to create a new study manually in the populate_test_db.sql.

Re: redis. Where is redis involved in all this?

adamrp:
@josenavas it is not clear to me how to create a study in a different portal using `Study.create` (without modifying `qiita_config` directly).

redis enters into the equation [on these lines](https://github.com/biocore/qiita/blob/master/qiita_pet/handlers/stats.py#L19-L44)

adamrp:
@josenavas to expand on my previous comment: the current `Study.create` just uses what's in `qiita_config` (and also adds to `QIITA` portal if a different portal is specified in the `qiita_config`). There is also not a setter method for the portal(s) the study is associated with. I can modify the `Study.create` function if we think this is a better approach than creating the study in `populate_test_db.sql`.

@ElDeveloper the failures are largely due to previously written tests that use the `Study.create` function assuming that the studies that they create will be assigned particular IDs, and using those assumed IDs in the tests themselves. I think this is probably not the best test design (and I'm sure I'm guilty of it, too); instead of assuming the ID, the tests should probably actually retrieve the ID that was assigned.

ElDeveloper:
Ah, I see, thanks for the information!

josenavas:
@adamrp you can assign studies to different portals using [this](https://github.com/biocore/qiita/blob/master/qiita_db/portal.py#L229) function. This will fix your problem with the tests, and reduce the amount of code you touch.

Thanks for pointing the redis thing. In such case, the redis keys need to be portal dependent. It is not ok to clear the redis keys. Those lines needs to be fixed in this PR too.

adamrp:
@josenavas okay, thanks, sounds good!

qiita_db/util.py
```diff
@@ -969,14 +969,21 @@ def get_processed_params_tables():
         return TRN.execute_fetchflatten()
 
 
-def get_lat_longs():
+def get_lat_longs(portal):
```
josenavas: The API shouldn't change, the portals is something that is transparent to most of the API. Check other parts of the code (the Study class, user, etc...) to see how the portals stuff is handled completely transparent from the API point of view.

qiita_pet/handlers/stats.py
```diff
@@ -18,10 +20,12 @@ def _get_stats(self, callback):
         # check if the key exists in redis
         lats = r_client.lrange('stats:sample_lats', 0, -1)
         longs = r_client.lrange('stats:sample_longs', 0, -1)
-        if not (lats or longs):
+        if not (lats and longs):
             # if we don't have them, then fetch from disk and add to the
             # redis server with a 24-hour expiration
-            lat_longs = get_lat_longs()
+            lat_longs = get_lat_longs(Portal(qiita_config.portal))
```
josenavas: How was this working before w/o the float cast?

josenavas:
Also, there has been a benchmark to see how long it takes to retrieve the stats from the DB? If it happens fast I don't think we should keep storing them in the redis db....

qiita_db/util.py
adamrp: This was needed to avoid circular import (qiita_db/util.py and qiita_db/portal.py). I could move `get_lat_longs` to meta_util.py

qiita_pet/handlers/stats.py
adamrp: It was "working" (i.e., not raising exceptions or anything), but the information that was being displayed was wrong.

adamrp:
Re: use of Redis: I am not aware of a benchmark that was run. This was the design that @biocore/qiita-dev (I think specifically @wasade) suggested when this code was first written.

wasade:
it doesn't happen fast. we put it in place as the request was extremely slow with AG (similar mechanism for getting pin locations for participants). the database request is effectively a full table scan across N tables where N is the number of studies...

josenavas:
Thanks @wasade 

@adamrp yes please move that function to meta_util.py as we would not like to have any API that needs the explicit portal, since that should be transparent through the config_file. There was a fair amount of work to avoid that.

adamrp:
@josenavas @ElDeveloper Ready for another round of review here -- thanks!

josenavas:
Real failures in here

squirrelo:
Just for reference, the test_study.py has a standard way of testing changing portals and seeing results. Take a look at the first lines of SetUp and TearDown functions. Then you can do something like [this](https://github.com/biocore/qiita/blob/master/qiita_db/test/test_study.py#L191) to actually change the portal in the test. I don't think we should be able to change the portals any other way than this and the add_to_portal function.

adamrp:
@squirrelo okay, thanks -- I did think of changing the portal directly in the `qiita_config`, but thought it wasn't a very clean solution. At any rate, the `add_studies` method that @josenavas suggested works great here.

adamrp:
Sorry about those test failures -- fixed now

qiita_db/meta_util.py
```diff
@@ -163,3 +168,34 @@ def get_accessible_filepath_ids(user):
...
+                 WHERE SUBSTR(table_name, 1, 7) = 'sample_'
+                    AND table_schema = 'qiita'
+                    AND column_name IN ('latitude', 'longitude');"""
+        TRN.add(sql)
+
+        sql = "SELECT latitude, longitude FROM qiita.{0}"
+        idx = TRN.index
+
+        portal_tables = TRN.execute_fetchflatten()
+        portal_tables = [x for x in TRN.execute_fetchflatten()
```
josenavas: This for loop can be removed if we modify the original SQL query to do the filtering for us:

```python
sql = """SELECT DISTINCT table_name
         WHERE SUBSTR(table_name, 1, 7) = 'sample_'
             AND table_name != 'sample_template_filepath'
             AND table_schema = 'qiita'
             AND column_name IN ('latitude', 'longitude');
             AND SPLIT_PART(table_name, '_', 2)::int IN %s
TRN.add(sql, [tuple(portal_table_ids)])
portal_tables = TRN.execute_fetchflatten()
```
And then do the `for` loop to add the sql for each table.

qiita_db/test/test_meta_util.py
```diff
@@ -89,6 +95,86 @@ def test_get_accessible_filepath_ids(self):
...
+        md_ext = pd.DataFrame.from_dict(md, orient='index')
+        SampleTemplate.create(md_ext, study)
+
+        old_portal = qiita_config.portal
+        qiita_config.portal = 'EMP'
+
+        try:
+            obs = get_lat_longs()
+        finally:
+            qiita_config.portal = old_portal
```
josenavas: This construction should go in the `tearDown` function for the tests.

josenavas:
Just a couple of comments @adamrp

qiita_db/test/test_meta_util.py
adamrp: good idea, I'll add `self.old_portal = qiita_config.portal` to `setUp` and `qiita_config.portal = self.old_portal` to `tearDown`. Or do you mean it should go in the class decorator (i.e., [here](https://github.com/biocore/qiita/blob/master/qiita_core/util.py#L76-L83))

qiita_db/meta_util.py
adamrp: I can do that. I'm curious if this is better than doing the `for` loop in Python for a reason that I'm not thinking of?

qiita_db/test/test_meta_util.py
josenavas: I think it should go in the `setUp` and `tearDown` in this test class. No need to modify the class decorator.

The use cases are still not enough to generalize that behavior for all test classes.

adamrp: :+1:

qiita_db/meta_util.py
josenavas: I though that avoiding the python loop will be faster, as is one less loop. Using the live system, both take the same time:
```
In [33]: %timeit old_get_lat_longs()
1 loops, best of 3: 1.24 s per loop

In [34]: %timeit new_get_lat_longs()
1 loops, best of 3: 1.24 s per loop
```

I think the main problem is the for loop in line 198... Feel free to not change it if you don't feel like.

adamrp: I made the change -- you may be right, since terrance/phillip database server is probably more efficient than the VM on which the the python code would be running. Not sure, but it's an easy enough change to make.

adamrp:
Was there anything more that needs to be done to merge this PR?

qiita_db/meta_util.py
```diff
@@ -163,3 +168,33 @@ def get_accessible_filepath_ids(user):
...
+    -------
+    list of [float, float]
+        The latitude and longitude for each sample in the database
+    """
+    portal_table_ids = Portal(qiita_config.portal).get_studies()
+
+    with TRN:
+        sql = """SELECT DISTINCT table_name
+                 FROM information_schema.columns
+                 WHERE table_name similar to 'sample_[0-9]+'
```
josenavas: `similar to` -> `SIMILAR TO`

josenavas:
Just a small comment, otherwise :+1: 
It needs another dev - @antgonza or @ElDeveloper are you available?

ElDeveloper:
:+1:, just one comment from @josenavas, aside from that looks good.

adamrp:
@josenavas @ElDeveloper fixed! Thanks for the review.

PR #1432, opened 2015-08-23
Int instantiation 1423

### LINKED ISSUE #1423: Instantiating with string value works
Currently we can instantiate objects with strings of integers or straight integers and, because of how psycopg2 works, we can still use the objects either way. However, some internal functions require it to be an integer for processing. We need to add a check to make sure the passed id is actually an integer, and not a string.

adamrp:
Fixes #1423

### DIFF
--- a/qiita_db/base.py
+++ b/qiita_db/base.py
@@ -39,7 +39,7 @@ class QiitaObject(object):
 
     Parameters
     ----------
-    id_: object
+    id_: int, long, str, or unicode
         The object id on the storage system
 
     Attributes
@@ -166,13 +166,30 @@ class QiitaObject(object):
 
         Parameters
         ----------
-        id_: the object identifier
+        id_: int, long, str, or unicode
+            the object identifier
 
         Raises
         ------
         QiitaDBUnknownIDError
             If `id_` does not correspond to any object
         """
+        # Most IDs in the database are numerical, but some (e.g., IDs used for
+        # the User object) are strings. Moreover, some integer IDs are passed
+        # as strings (e.g., '5'). Therefore, explicit type-checking is needed
+        # here to accommodate these possibilities.
+        if not isinstance(id_, (int, long, str, unicode)):
+            raise TypeError("id_ must be a numerical or text type (not %s) "
+                            "when instantiating "
+                            "%s" % (id_.__class__.__name__,
+                                    self.__class__.__name__))
+
+        if isinstance(id_, (str, unicode)):
+            if id_.isdigit():
+                id_ = int(id_)
+        elif isinstance(id_, long):
+            id_ = int(id_)
+
         with TRN:
             self._check_subclass()
             if not self._check_id(id_):


josenavas:
Just looking at the code (probably the errors are about this):

The user object doesn't work with ints, is a special case...

adamrp:
No, if you were looking at the code, you would see that it currently allows int, long, and str. The failures are about unicode type. Fixing it now.

josenavas:
Thanks @adamrp I will review it later, after the release.

qiita_db/base.py
```diff
@@ -166,13 +166,26 @@ def __init__(self, id_):
...
             If `id_` does not correspond to any object
         """
+        id_type = type(id_)
+        if not isinstance(id_, (int, long, str, unicode)):
+            raise TypeError("id_ must be a numerical or text type (not %s) "
+                            "when instantiating %s" % (str(id_type),
+                                                       str(self.__class__)))
+
+        if id_type in (str, unicode):
+            if id_.isdigit():
```
squirrelo: This is a nice compromise for User object. Could you put in a comment about why it's needed?

squirrelo:
1 non-blockig comment :+1:

qiita_db/util.py
```diff
@@ -863,7 +863,7 @@ def convert_to_id(value, table, text_col=None):
                                      % (value, table))
         # If there was a result it was a single row and and single value,
         # hence access to [0][0]
-        return _id[0][0]
+        return int(_id[0][0])
```
josenavas: I'm curious about this change... why is it needed? which is the specific case that was failing? the id in the database should already be an integer, so psycopg2 should return an int.

adamrp: The docstring says it returns an `int`, but that was not actually always the case. I thought this was the easier/better change (as opposed to changing the docstring to say that it returns `object`, or something to that effect).

qiita_db/base.py
adamrp: Yes -- no problemo

qiita_db/util.py
josenavas: Well, that is what is confusing. In which cases it was not returning an int? AFAIK it should always return an int. If not, there might be another bug in some other place, so "patching" it in this way is not the way to go...

adamrp: I don't recall at the moment what was specifically raising the error -- I'll remove the change and run the tests to find out what was failing.

adamrp:
@josenavas it turns out that the change was not actually needed, so I removed it -- good catch :smile: 

Should be ready for review again now

qiita_db/base.py
```diff
@@ -166,13 +166,30 @@ def __init__(self, id_):
...
+        # the User object) are strings. Moreover, some integer IDs are passed
+        # as strings (e.g., '5'). Therefore, explicit type-checking is needed
+        # here to accommodate these possibilities.
+        id_type = type(id_)
+        if not isinstance(id_, (int, long, str, unicode)):
+            raise TypeError("id_ must be a numerical or text type (not %s) "
+                            "when instantiating %s" % (str(id_type),
+                                                       str(self.__class__)))
+
+        if id_type in (str, unicode):
```
josenavas: Should we keep using here `isinstance`?
```python
if isinstance(id_, (str, unicode)):
    #do stuff
elif isinstance(id_, long):
    #do stuff
```

qiita_db/base.py
```diff
@@ -166,13 +166,30 @@ def __init__(self, id_):
...
         """
+        # Most IDs in the database are numerical, but some (e.g., IDs used for
+        # the User object) are strings. Moreover, some integer IDs are passed
+        # as strings (e.g., '5'). Therefore, explicit type-checking is needed
+        # here to accommodate these possibilities.
+        id_type = type(id_)
+        if not isinstance(id_, (int, long, str, unicode)):
+            raise TypeError("id_ must be a numerical or text type (not %s) "
+                            "when instantiating %s" % (str(id_type),
+                                                       str(self.__class__)))
```
josenavas: Shouldn't this be `self.__class__.__name__`?

qiita_db/base.py
adamrp: Yeah, probably more consistent that way. The behavior is slightly different (since `isinstance` returns true if it is one of the specified classes or any subclass), but I don't think it matters here.

qiita_db/base.py
adamrp: Good idea, that is cleaner.

adamrp:
Ready for review again

josenavas:
:+1: @squirrelo ca you review again and merge if you're happy with it? There has been a few commits after your last :+1:

Issue #1160, opened 2015-05-06
Allow users to specify their own EBI credentials


wasade:
So that users can submit with alternative credentials instead of the Knight lab set.

antgonza:
@wasade each deploy can have their own EBI credentials for submission, will this suffice or do you mean that the user should be able to supply their own credentials through the GUI, the system should validate them and then submit?

wasade:
Latter

antgonza:
Thanks for the clarification. Is this a nice to have or did a user requested for this? I'm just not sure how useful will this be compared to the user uploading directly ...

wasade:
Right now, our submissions say that the submitting center is "Colorado Center for Microbial Ecology, University of Colorado at Boulder" (at least as of [AG round 17](http://www.ebi.ac.uk/ena/data/view/ERP010416)). That is an issue in and of itself that is separate from this issue. The greater context though is that a user may want to have their data associated with them, not with UCSD, or CU. At this time, a user using `http://qiita.microbio.me` will implicitly have their data associated to us, and _we_ own the accessions not the submitter.

I don't know what you mean by "user uploading directly", do you mean having the user go the EBI and manually upload files processed by qiita? Or do you mean requiring that a user do a custom deploy of qiita in order to specify credentials?

antgonza:
CCME, UCB: Agree we should change the credentials to microbio.me or something like that ... 

direct upload: I was thinking in the former but either should work. IMOO having your own credentials is something advanced ...

wasade:
Can you even manually do that via EBI? Do we provide the files that allow them to do that? If you can do it, then yes, one solution is to provide the files they need to do the upload manually however we then loose the knowledge of what the subsequent produced accession is. 

Note: the present system means that any user who submits via `qiita.microbio.me`, who subsequently discovers a problem with the data (as is quite common given the version history seen with NCBI accessions), will need to contact us and ask us to login to EBI and do whatever tinkering is necessary. The users cannot get access to the accessions to modify them as they are owned by our account, and we obviously cannot distribute those credentials.

antgonza:
K, adding group discussion label so we can discuss next meeting.

rob-knight:
SRA has concept of study center, sequencing center, analysis center. We
should let users choose the other two if not sequenced here but remember
choices are supposed to be from predefined list. And it's an advanced
feature.

On Aug 13, 2015, at 12:10 PM, Antonio Gonzalez <notifications@github.com>
wrote:

[...]

PR #1437, opened 2015-08-24
WIP: ebi-improvements -- initial cleanup

antgonza:
The plan for this branch it to contain the new EBI submissions code. The main problem is that the EBI submission was created in parallel with Qiita (this implies that it doesn't really comply with how Qiita works now) and based on the old DB. 

The main goals are:
- Make the EBI submission a multiple step vs. a single step process. This implies that every step will be part of a the single class and that each step will be called via the GUI within a networkx/moi step, which should allow us to save on which step we are and recover if needed.
- Remove the file (of templates) dependency of submissions by querying the DB directly. Currently, we need to write template files and then process them. By removing this we are making sure that we only write what's on the DB. Additionally, this will imply that we will remove any template manipulation (adding samples, defining no-data values, etc) from the EBI code.
- Modify the click command so you can only pass the prep template id, action. BTW I'm not sure we actually need this if the GUI is working fine. What do you think? Should I open an issue to discuss?

The first step (this PR) is to consolidate	qiita_ware/commands.py and qiita_ware/ebi.py so all validations are made in ebi.py against the db at __init__. This required deletion of __init__ parameters, unnecessary code/tests for some table manipulation. Note that this is not fully done but there are already a bunch of changes here.  

Things still to do (this will be growing as needed):
- [ ] Understand/Fix what `lib_strategies = {'metagenome': 'POOLCLONE',` is doing.
- [ ] Change CCME-COLORADO and confirm with EBI

### DIFF
--- a/qiita_ware/ebi.py
+++ b/qiita_ware/ebi.py
@@ -85,8 +85,7 @@ class EBISubmission(object):
     Submit a preprocessed data to EBI
 
     The steps for EBI submission are:
-    1. Validate that the submission status is different than
-    self.valid_ebi_actions and that there is a valid investigation_type
+    1. Validate that we have all required info to submit
     2. Generate per sample demultiplexed files
     3. Generate XML files for submission
     4. Submit sequences files
@@ -97,20 +96,25 @@ class EBISubmission(object):
     preprocessed_data_id : int
         The preprocesssed data id
     action : str
-        The action to perfom, it has to be one of the self.valid_ebi_actions
+        The action to perfom, it has to be one of the
+        EBISubmission.valid_ebi_actions
 
     Parameters
     ----------
     preprocessed_data_id : str
     """
-    def __init__(self, preprocessed_data_id, action, **kwargs):
-        self.valid_ebi_actions = ('ADD', 'VALIDATE', 'MODIFY')
-        self.valid_ebi_submission_states = ('submitting', 'success')
+
+    valid_ebi_actions = ('ADD', 'VALIDATE', 'MODIFY')
+    valid_ebi_submission_states = ('submitting', 'success')
+
+    def __init__(self, preprocessed_data_id, action):
+        valid_ebi_actions = EBISubmission.valid_ebi_actions
+        valid_ebi_submission_states = EBISubmission.valid_ebi_submission_states
 
         # Step 1: variable setup and validations
-        if action not in self.valid_ebi_actions:
+        if action not in valid_ebi_actions:
             raise ValueError("Not a valid action (%s): %s" % (
-                ', '.join(self.valid_ebi_actions), action))
+                ', '.join(valid_ebi_actions), action))
 
         ena_ontology = Ontology(convert_to_id('ENA', 'ontology'))
         ppd = PreprocessedData(preprocessed_data_id)
@@ -118,7 +122,7 @@ class EBISubmission(object):
         pt = PrepTemplate(ppd.prep_template)
 
         status = ppd.submitted_to_insdc_status()
-        if status in self.valid_ebi_submission_states:
+        if status in valid_ebi_submission_states:
             raise ValueError("Cannot resubmit! Current status is: %s" % status)
 
         self.preprocessed_data_id = preprocessed_data_id
@@ -140,7 +144,7 @@ class EBISubmission(object):
                              "one of the user-defined terms in the ENA "
                              "ontology")
         ts = datetime.now().strftime('%Y_%m_%d_%H:%M:%S')
-        self.ebi_dir = '%s_%s' % (preprocessed_data_id, ts)
+        self.ebi_dir = 'ebi_submission_%s_%s' % (preprocessed_data_id, ts)
         self.sequence_files = []
         self.study_xml_fp = None
         self.sample_xml_fp = None
--- a/qiita_ware/ebi.py
+++ b/qiita_ware/ebi.py
@@ -18,6 +18,9 @@ from qiita_ware.exceptions import EBISumbissionError
 from qiita_db.logger import LogEntry
 from qiita_db.ontology import Ontology
 from qiita_db.util import convert_to_id
+from qiita_db.study import Study
+from qiita_db.data import PreprocessedData
+from qiita_db.metadata_template import PrepTemplate
 
 
 class InvalidMetadataError(Exception):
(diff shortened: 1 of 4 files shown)


antgonza:
Note that the errors are due to the ebi_actions variable and their import into the qiita script. I can fix but will like to hear your opinions about this code. Thus, please review.

qiita_ware/ebi.py
```diff
@@ -77,56 +80,74 @@ def iter_file_via_list_of_dicts(input_file):
...
-    new_investigation_type: str, optional
-        'metagenome', and 'mimarks-survey' are specially recognized and used to
-        set other attributes in the submission, but any string is valid. This
-        value is required only if `'Other'` is passed in `investigation_type`.
     """
-    def __init__(self, preprocessed_data_id, study_title, study_abstract,
-                 investigation_type, empty_value='no_data',
-                 new_investigation_type=None, pmids=None, **kwargs):
+    def __init__(self, preprocessed_data_id, action, **kwargs):
+        self.valid_ebi_actions = ('ADD', 'VALIDATE', 'MODIFY')
```
josenavas: This makes more sense to be class attributes, rather than instance attributes.

qiita_ware/ebi.py
```diff
@@ -77,56 +80,74 @@ def iter_file_via_list_of_dicts(input_file):
...
+        self.valid_ebi_submission_states = ('submitting', 'success')
```
josenavas: Same here

qiita_ware/ebi.py
```diff
@@ -77,56 +80,74 @@ def iter_file_via_list_of_dicts(input_file):
...
+            raise ValueError("Not a valid action (%s): %s" % (
+                ', '.join(self.valid_ebi_actions), action))
+
+        ena_ontology = Ontology(convert_to_id('ENA', 'ontology'))
+        ppd = PreprocessedData(preprocessed_data_id)
+        s = Study(ppd.study)
+        pt = PrepTemplate(ppd.prep_template)
+
+        status = ppd.submitted_to_insdc_status()
+        if status in self.valid_ebi_submission_states:
```
josenavas: I'm confused here... what is the EBI action "Modify" means? If it is tailored to change a successful EBI submission (e.g. we have successfully submitted, but something needs to be changed), this line will raise a ValueError. Am I missing something?

qiita_ware/ebi.py
```diff
@@ -77,56 +80,74 @@ def iter_file_via_list_of_dicts(input_file):
...
+            self.investigation_type = 'Other'
+            self.new_investigation_type = it
+        else:
+            # This should never happen
+            raise ValueError("Unrecognized investigation type: '%s'. This "
+                             "term is neither one of the official terms nor "
+                             "one of the user-defined terms in the ENA "
+                             "ontology")
+        ts = datetime.now().strftime('%Y_%m_%d_%H:%M:%S')
+        self.ebi_dir = '%s_%s' % (preprocessed_data_id, ts)
```
josenavas: probably better to also include a string that denotes that this is a EBI submission:
```python
self.ebi_dir = "ebi_submission_%s_%s" % (preprocessed_data_id, ts)
```

qiita_ware/ebi.py
```diff
@@ -77,56 +80,74 @@ def iter_file_via_list_of_dicts(input_file):
...
-                             " to specify a value for new_investigation_type.")
-
-        ontology = Ontology(convert_to_id('ENA', 'ontology'))
-        if ontology.term_type(self.investigation_type) == 'not_ontology':
-            raise ValueError("The investigation type must be part of ENA's "
-                             "ontology, '%s' is not valid" %
-                             self.investigation_type)
+        self.pmids = s.pmids
 
         # dicts that map investigation_type to library attributes
```
josenavas: I know that you didn't modify any of this part of the code, but it looks like there is a lot of hard-coding in here. Do you think it is worth to make a note in the PR list so we don't forget to actually take a look in here and see what is going on?

qiita_ware/ebi.py
```diff
@@ -77,56 +80,74 @@ def iter_file_via_list_of_dicts(input_file):
...
-        for a particular metadata field is missing
-    new_investigation_type: str, optional
-        'metagenome', and 'mimarks-survey' are specially recognized and used to
-        set other attributes in the submission, but any string is valid. This
-        value is required only if `'Other'` is passed in `investigation_type`.
     """
-    def __init__(self, preprocessed_data_id, study_title, study_abstract,
-                 investigation_type, empty_value='no_data',
-                 new_investigation_type=None, pmids=None, **kwargs):
+    def __init__(self, preprocessed_data_id, action, **kwargs):
```
josenavas: what are you expecting in `kwargs`?

josenavas:
Done with a first review.

qiita_ware/ebi.py
antgonza: at this point is simply to make sure test pass, my expectation is to remove it on a later stage when I remove the other unnecessary code.

qiita_ware/ebi.py
antgonza: Agree and this will allow me to make all tests pass.

qiita_ware/ebi.py
antgonza: The current code doesn't allow any modification/validation, really. The only thing that changes is what's added in the text. The "right" way to do it will be to only update the metadata on modify (you can not modify seq files automatically).

qiita_ware/ebi.py

qiita_ware/ebi.py
antgonza: To be honest I have no idea what's doing, will fix or remove on a next iteration of this code.

qiita_ware/ebi.py
antgonza: realized that I already fix this, removing ...

adamrp:
@antgonza small thing, but can you also change [the value for the default EBI center name in the example configuration](https://github.com/biocore/qiita/blob/master/qiita_core/support_files/config_test.cfg#L125) as part of this PR?

qiita_ware/ebi.py
adamrp: The current EBI code is overcomplicated because it had to accommodate submissions that were drawing some data from the Qiita DB, some data from files, and some data from the CLI, so it is very likely that there's a bunch of code that can be removed (or should be modified to pull from the DB as much as possible).

antgonza:
Yup, what's the denomination for UCSD? Also, do we need to tell EBI?

adamrp:
Yes, you're right, I think input from @rob-knight (and possibly the UCSD sequencing center?) is needed here regarding what to call the center, and we should confirm with Marc at EBI that we will be using that name going forward.

adamrp:
The only discussion on the topic before this (that I'm aware of) is in #1160

antgonza:
OK, I'm gonna be adding things that we need to remember on the first comment of this PR so we do not forget. The idea is that this section will grow as needed but I will be marking as done the things that are accomplish as PRs move along. If this sounds right and both @josenavas and @adamrp feel confortable with this initial step it will be great if you can merge to the `ebi-improvements` branch.

adamrp:
Sounds like a good approach to me -- thanks!

adamrp:
I think @ackermag is the person most likely able to answer the question about POOLCLONE

josenavas:
@antgonza I merged this PR in the `ebi-improvements` branch. Now, you should create a new PR from `ebi-improvements` to `master` and list there the missing bits (the list in here will be harder to keep track off).

You can then keep doing PR to the `ebi-improvements` branch and it will automatically update the PR as they get merged.

PR #1439, opened 2015-08-24
BUG: Fix problem with default basedir

ElDeveloper:
If the base data directory was not set, a default one would be selected,
by using '..' to refer to a directory below, which in turn would
inconsistently flag users as lacking permissions:

```
QiitaPetAuthorizationError: User demo@microbio.me is not authorized to
access
/Users/yoshikivazquezbaeza/.virtualenvs/qiita-0.2.0-rc1/lib/python2.7/site-packages/qiita_db/support_files/test_data/job/1_summarize_taxa_through_plots.py_output_dir/taxa_summary_plots/bar_charts.html
```

Although the paths were the same, this is verified through string
matching thus 'lol/foo/../jk' and 'lol/jk' would not match (see example
structure below) even though they refer to the same folder:

lol/
├── foo
│   └── file.txt
├── jk
└── plel

### DIFF
--- a/qiita_core/configuration_manager.py
+++ b/qiita_core/configuration_manager.py
@@ -144,8 +144,8 @@ class ConfigurationManager(object):
     def _get_main(self, config):
         """Get the configuration of the main section"""
         self.test_environment = config.getboolean('main', 'TEST_ENVIRONMENT')
-        default_base_data_dir = join(dirname(abspath(__file__)),
-                                     '..', 'qiita_db', 'support_files',
+        install_dir = dirname(dirname(abspath(__file__)))
+        default_base_data_dir = join(install_dir, 'qiita_db', 'support_files',
                                      'test_data')
         self.base_data_dir = config.get('main', 'BASE_DATA_DIR') or \
             default_base_data_dir
--- a/qiita_db/support_files/patches/31.sql
+++ b/qiita_db/support_files/patches/31.sql
@@ -1,0 +1,3 @@
+-- August 24, 2015
+-- Delete all occurrences of '..' in the base_data_dir entry
+SELECT 42;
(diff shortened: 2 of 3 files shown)


josenavas:
:+1: once test pass

ElDeveloper:
After chatting offline with @antgonza, it seems like we will also need a DB patch. The reason is that the settings table might already have a path of this form `/Users/yoshikivazquezbaeza/.virtualenvs/qiita-0.2.0-rc1/lib/python2.7/site-packages/qiita_core/../qiita_db/support_files/test_data` (notice the `..`), and we really just want to change it to `/Users/yoshikivazquezbaeza/.virtualenvs/qiita-0.2.0-rc1/lib/python2.7/site-packages/qiita_db/support_files/test_data`.

The only way to look for this would be to query for the `base_data_dir` path and if it ends in `../qiita_db/support_files/test_data`, then change it to the correct path. However it may be that a user could have replicated the a similar structure but with a different path, thus fixing this problem would imply removing any occurrences of .. and the preceding folder name. Does that make sense or am I overthinking this?

josenavas:
We will not need to modify anything in the DB besides the settings table... and given that if we change the path in the DB it will point to the same folder, it will not hurt.

wasade:
@ElDeveloper, I think this is touching on the broader issue of path
canonicalization (removing all the .. from the paths, not just start or
end)

ElDeveloper:
@wasade, yeah that's what I meant.

ElDeveloper:
For some reason it works in an IPython session but when I call `qiita-env patch` the program just hangs, anyone has ideas as to why this may be happening?

ElDeveloper:
Ok, this should be ready for review, thanks for the help @josenavas!

josenavas:
:+1:

Issue #1413, opened 2015-08-10
fix to not getting email for confirmation


antgonza:
As @josenavas mentioned, most of the emails we are getting in the qiita.help account is about not getting the account confirmation email. From what I can see is that the reports normally come from academic emails; my guess is: more stringent filters. I think a solution will be to ask users to add the email address to their contact list before registering.

josenavas:
Agree, there isn't much that we can do here, just adding a message on the sign-up page.

One thing is that most of those academic email systems don't even put the message in the spam folder, they just get completely filtered before getting to the user. I'm not sure if adding the email address to their contact list before registering will suffice.

Another option will be to get a "ucsd" and configure Qiita to use the SMTP server from the university, rather than the one in the linux distribution...

antgonza:
Perhaps the UCSD or the qiita.help account will be good options too.

josenavas:
I think the qiita.help can work. I think we also have the configuration for setting up the SMTP using google commented out in the configuration file, so that would be easy to test...

wasade:
CU had an open relay for registered systems on campus, and UCSD may as well
which would just solve the problem. It may be good to verify that the DNS
entry for qiita.microbio.me is properly describing it as an MX record. It
isn't necessary, but I could totally understand why mail servers would just
blindly filter out mail coming from a host w/o an MX record (if that's
actually the case).

antgonza:
Currently we are using donotreply@qiita.mirobio.me to send emails, if we change to qiita.help@gmail.com and use google we are set and will not need to rely on the UCSD system. What do you think?

josenavas:
:+1:

antgonza:
We continue getting reports about this. Note that @josenavas and I tried to configure Gmail in the system  unsuccessfully. Currently, checking if we can do something with UCSD.

antgonza:
It seems like we finally solved all issues. At this point if a user doesn't get an email is because their email server has some restrictions, like not accepting emails from .me servers (remember that they are sent from @qiita.micriobio.me), which I think we can live with. Closing for now.

PR #1440, opened 2015-08-25
Change the example email address, remove colorado

adamrp:
tangentially related to #1413

### DIFF
--- a/qiita_core/support_files/config_test.cfg
+++ b/qiita_core/support_files/config_test.cfg
@@ -63,7 +63,7 @@ USER =
 PASSWORD =
 
 # The email to have messages sent from
-EMAIL = donotreply@qiita.colorado.edu
+EMAIL = example@domain.com
 
 # ----------------------------- Redis settings --------------------------------
 [redis]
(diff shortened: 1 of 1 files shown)


antgonza:
:+1: when tests pass.

ElDeveloper:
:+1:

josenavas:
Should we actually put something that we do not control? I'm wondering if somebody has a custom installation for their lab we might not want to receive all the internal emails...

antgonza:
Excellent point! However, that email is only used as the email that emails as sent as. I'm OK either way.

adamrp:
I don't feel strongly, and would be fine changing it to something like "example@domain.com" or something. The `EBI_ORGANIZATION_PREFIX` is also currently set to qiime; do we want to change this, too, so that the example configuration doesn't contain other information specific to the knight lab or UCSD?

antgonza:
I think these change will make it more generic, which is good ...

ElDeveloper:
Yes, I also think those sound like good changes!

Issue #1029, opened 2015-04-02
Split-up tests in ReadOnly vs ReadWrite on qiita_db


josenavas:
I'm listing everything here, just in case that somebody wants to work in something that requires no brain. Please submit a PR per single file. Add your name next to the file that you're currently working in.
This will reduce development time since tests are run faster.

Some general rules
 - If the code tested potentially modifies the database, it should go under the ReadWrite category, in case that the code is not working correctly it does not impact other tests
 - If both tests sets are going to share a lot of data, useful to create a single base class that initializes both test classes. 

 - [ ] qiita_db/test/test_analysis.py
 - [x] qiita_db/test/test_artifact.py (Jose)
 - [ ] qiita_db/test/test_base.py
 - [ ] qiita_db/test/test_commands.py
 - [ ] qiita_db/test/test_data.py
 - [ ] qiita_db/test/test_job.py
 - [ ] qiita_db/test/test_logger.py
 - [ ] qiita_db/test/test_meta_util.py
 - [x] qiita_db/test/test_metadata_template.py (Jose - PR #1033)
 - [ ] qiita_db/test/test_ontology.py
 - [ ] qiita_db/test/test_parameters.py
 - [x] qiita_db/test/test_processing_job.py (Jose)
 - [ ] qiita_db/test/test_reference.py
 - [ ] qiita_db/test/test_search.py
 - [ ] qiita_db/test/test_setup.py
 - [x] qiita_db/test/test_software.py (Jose)
 - [ ] qiita_db/test/test_sql.py
 - [ ] qiita_db/test/test_sql_connection.py
 - [ ] qiita_db/test/test_study.py
 - [ ] qiita_db/test/test_user.py
 - [ ] qiita_db/test/test_util.py

souravsingh:
I would like to work on this issue.How do I start?

antgonza:
@souravsingh that will be great, thanks!

Some background: currently we working on improving the time it takes to run the qiita tests as it is slowly becoming an issue and this is one of the likely culprits. Basically, every time that we run the DB tests we are resetting the database to a fresh install. To facilitate this task we decided to create a decorator, used [here](https://github.com/biocore/qiita/blob/master/qiita_db/test/test_analysis.py#L28), to do this operation at the end of each test. Also, we have this decorator even in those cases were we are only querying the DB (AKA we don't need to reset).

[...]

PR #1450, opened 2015-08-26
EBI step 2

antgonza:
In this PR I'm removing all kwargs from methods, generating step 2 of the pipeline (generate per sample files) and adding tests for it.

### DIFF
--- a/qiita_ware/commands.py
+++ b/qiita_ware/commands.py
@@ -6,12 +6,12 @@
 # The full license is in the file LICENSE, distributed with this software.
 # -----------------------------------------------------------------------------
 
+from os import makedirs
 from os.path import join, isdir
-from os import makedirs, remove
+from shutil import rmtree
 from functools import partial
-from tempfile import mkdtemp
-from gzip import open as gzopen
 from tarfile import open as taropen
+from tempfile import mkdtemp
 from moi.job import system_call
 
 from qiita_db.study import Study
@@ -20,9 +20,7 @@ from qiita_db.metadata_template import PrepTemplate, SampleTemplate
 from qiita_core.qiita_settings import qiita_config
 
 from qiita_ware.ebi import EBISubmission
-from qiita_ware.demux import to_per_sample_ascii
 from qiita_ware.exceptions import ComputeError
-from qiita_ware.util import open_file
 
 
 def submit_EBI(preprocessed_data_id, action, send, fastq_dir_fp=None):
@@ -44,53 +42,22 @@ def submit_EBI(preprocessed_data_id, action, send, fastq_dir_fp=None):
     If fastq_dir_fp is passed, it must not contain any empty files, or
     gzipped empty files
     """
+    # step 1
+    ebi_submission = EBISubmission(preprocessed_data_id, action)
 
-    preprocessed_data = PreprocessedData(preprocessed_data_id)
-    preprocessed_data_id_str = str(preprocessed_data_id)
-    study = Study(preprocessed_data.study)
-    sample_template = SampleTemplate(study.sample_template)
-    prep_template = PrepTemplate(preprocessed_data.prep_template)
+    # step 2
+    ebi_submission.preprocessed_data.update_insdc_status('demuxing samples')
+    try:
+        demux_samples = ebi_submission.generate_demultiplexed_fastq()
+    except:
+        if isdir(ebi_submission.ebi_dir):
+            rmtree(ebi_submission.ebi_dir)
 
-    investigation_type = None
-    new_investigation_type = None
-
-    if send:
-        # If we intend actually to send the files, then change the status in
-        # the database
-        preprocessed_data.update_insdc_status('submitting')
-
-    if fastq_dir_fp is not None:
-        # If the user specifies a FASTQ directory, use it
-
-        # Set demux_samples to None so that MetadataTemplate.to_file will put
-        # all samples in the template files
-        demux_samples = None
-    else:
-        # If the user does not specify a FASTQ directory, create one and
-        # re-serialize the per-sample FASTQs from the demux file
-        fastq_dir_fp = mkdtemp(prefix=qiita_config.working_dir)
-        demux = [path for _, path, ftype in preprocessed_data.get_filepaths()
-                 if ftype == 'preprocessed_demux'][0]
-
-        # Keep track of which files were actually in the demux file so that we
-        # can write those rows to the prep and samples templates
-        demux_samples = set()
-
-        with open_file(demux) as demux_fh:
-            for samp, iterator in to_per_sample_ascii(demux_fh,
-                                                      list(sample_template)):
-                demux_samples.add(samp)
-                sample_fp = join(fastq_dir_fp, "%s.fastq.gz" % samp)
-                wrote_sequences = False
-                with gzopen(sample_fp, 'w') as fh:
-                    for record in iterator:
-                        fh.write(record)
-                        wrote_sequences = True
-
-                if not wrote_sequences:
-                    remove(sample_fp)
-
+    # other steps
     output_dir = fastq_dir_fp + '_submission'
+    sample_template = ebi_submission.sample_template
+    prep_template = ebi_submission.prep_template
+    preprocessed_data = ebi_submission.preprocessed_data
 
     samp_fp = join(fastq_dir_fp, 'sample_metadata.txt')
     prep_fp = join(fastq_dir_fp, 'prep_metadata.txt')
@@ -112,19 +79,12 @@ def submit_EBI(preprocessed_data_id, action, send, fastq_dir_fp=None):
         raise IOError('The output folder already exists: %s' %
                       output_dir)
 
-    with open(samp_fp, 'U') as st, open(prep_fp, 'U') as pt:
-        submission = EBISubmission.from_templates_and_per_sample_fastqs(
-            preprocessed_data_id_str, study.title,
-            study.info['study_abstract'], investigation_type, st, pt,
-            fastq_dir_fp, new_investigation_type=new_investigation_type,
-            pmids=study.pmids)
-
-    submission.write_all_xml_files(study_fp, sample_fp, experiment_fp, run_fp,
-                                   submission_fp, action)
+    ebi_submission.write_all_xml_files(study_fp, sample_fp, experiment_fp,
+                                       run_fp, submission_fp, action)
 
     if send:
-        submission.send_sequences()
-        study_accession, submission_accession = submission.send_xml()
+        ebi_submission.send_sequences()
+        study_accession, submission_accession = ebi_submission.send_xml()
 
         if study_accession is None or submission_accession is None:
             preprocessed_data.update_insdc_status('failed')
--- a/qiita_ware/commands.py
+++ b/qiita_ware/commands.py
@@ -12,13 +12,13 @@ from shutil import rmtree
 from functools import partial
 from tarfile import open as taropen
 from tempfile import mkdtemp
-from moi.job import system_call
 
+from moi.job import system_call
 from qiita_db.study import Study
 from qiita_db.data import PreprocessedData
 from qiita_db.metadata_template import PrepTemplate, SampleTemplate
+from qiita_db.logger import LogEntry
 from qiita_core.qiita_settings import qiita_config
-
 from qiita_ware.ebi import EBISubmission
 from qiita_ware.exceptions import ComputeError
 
@@ -54,6 +54,8 @@ def submit_EBI(preprocessed_data_id, action, send, fastq_dir_fp=None):
             rmtree(ebi_submission.ebi_dir)
         ebi_submission.preprocessed_data.update_insdc_status(
             'failed: %s' % str(e))
+        LogEntry.create('Runtime', str(e),
+                        info={'ebi_submission': preprocessed_data_id})
         raise
 
     # other steps
--- a/qiita_ware/commands.py
+++ b/qiita_ware/commands.py
@@ -49,9 +49,12 @@ def submit_EBI(preprocessed_data_id, action, send, fastq_dir_fp=None):
     ebi_submission.preprocessed_data.update_insdc_status('demuxing samples')
     try:
         demux_samples = ebi_submission.generate_demultiplexed_fastq()
-    except:
+    except Exception as e:
         if isdir(ebi_submission.ebi_dir):
             rmtree(ebi_submission.ebi_dir)
+        ebi_submission.preprocessed_data.update_insdc_status(
+            'failed: %s' % str(e))
+        raise
 
     # other steps
     output_dir = fastq_dir_fp + '_submission'
--- a/qiita_ware/commands.py
+++ b/qiita_ware/commands.py
@@ -21,6 +21,7 @@ from qiita_db.logger import LogEntry
 from qiita_core.qiita_settings import qiita_config
 from qiita_ware.ebi import EBISubmission
 from qiita_ware.exceptions import ComputeError
+from traceback import format_exc
 
 
 def submit_EBI(preprocessed_data_id, action, send, fastq_dir_fp=None):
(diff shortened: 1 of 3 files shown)


qiita_ware/commands.py
```diff
@@ -44,53 +42,22 @@ def submit_EBI(preprocessed_data_id, action, send, fastq_dir_fp=None):
...
-                        wrote_sequences = True
-
-                if not wrote_sequences:
-                    remove(sample_fp)
-
+    # step 1
+    ebi_submission = EBISubmission(preprocessed_data_id, action)
+
+    # step 2
+    ebi_submission.preprocessed_data.update_insdc_status('demuxing samples')
```
josenavas: Does this even work? I though that the insdc_status was a controlled vocab, except for the `failed: ...` status.

antgonza: Not sure, but at this point I'm just setting this section as it should work and then cleaning as needed. Note that the issue is that there is no tests for this file ... I'm gonna need to add it in a future iteration. Already added it to the joy list: https://github.com/biocore/qiita/pull/1451

qiita_ware/commands.py
```diff
@@ -44,53 +42,22 @@ def submit_EBI(preprocessed_data_id, action, send, fastq_dir_fp=None):
...
+    try:
+        demux_samples = ebi_submission.generate_demultiplexed_fastq()
+    except:
```
josenavas: shouldn't this exception update the insdc status and raise an error? I think you actually want to raise the same error, so probably:
```python
except Exception as e:
    # Your if statement
    ebi_submission.preprocessed_data.update_insdc_status('failed: str(e)')
    raise
```
Note that I used `raise` instead of `raise e`. The reason is because by using `raise` you don't modify the traceback with this raise in here:
```python
In [4]: try:
   ...:     f()
   ...: except Exception as e:
   ...:     raise
   ...: 
ValueError                                Traceback (most recent call last)
<ipython-input-4-4a4820f11212> in <module>()
      1 try:
----> 2     f()
      3 except Exception as e:
      4     raise
      5 

<ipython-input-1-033e74257ed1> in f()
      1 def f():
----> 2     raise ValueError()
      3 

ValueError: 

In [5]: try:
   ...:     f()
   ...: except Exception as e:
   ...:     raise e
   ...: 
ValueError                                Traceback (most recent call last)
<ipython-input-5-b78dd80ee2b2> in <module>()
      2     f()
      3 except Exception as e:
----> 4     raise e
      5 

ValueError: 
```

josenavas:
@antgonza I'm a bit confused about this. IIRC, we agreed on using the ParallelWrapper right? Are you just cleaning up the `submit_EBI` to just have the 5 steps now and then change it later to make usage of the parallel wrapper? I just want to make sure that we are in the same page.

antgonza:
5 steps: Yup, same page.

qiita_ware/commands.py
antgonza: yup

josenavas:
Thanks @antgonza

qiita_ware/commands.py
josenavas: Thanks, I'm unsure if this function even needs to exist... We probably only want to use the parallel wrapper directly and this function should disappear. However, we will see as the code evolves...

qiita_ware/ebi.py
```diff
@@ -981,3 +950,45 @@ def send_xml(self):
...
+                        for record in i:
+                            fh.write(record)
+                            wrote_sequences = True
+
+                    if wrote_sequences:
+                        demux_samples.add(s)
+                    else:
+                        remove(sample_fp)
+
+        else:
```
josenavas: Just to make sure that I understand what is going on here. Is this else covering the case in which the per sample FASTQ's have been already generated? If so, can you add a comment to make it clear?

qiita_ware/test/test_ebi.py
```diff
@@ -405,6 +410,72 @@ def test_generate_curl_command(self):
                          '%20ebi_access_key%3D"')
         self.assertEqual(obs, exp_with_cert)
 
+    def test_generate_demultiplexed_fastq(self):
+        # generating demux file for testing
+        if not isdir(self.temp_dir):
+            makedirs(self.temp_dir)
```
josenavas: Can you add a `tearDown` function to the tests class and make sure that this directory gets removed after the execution of the test? Is fine for me if you want to add it to the todo list and do it later, but we should make sure that we  are green and we clean the environment :smile:

qiita_ware/test/test_ebi.py
```diff
@@ -405,6 +410,72 @@ def test_generate_curl_command(self):
...
+
+        fna_fp = join(self.temp_dir, 'seqs.fna')
+        demux_fp = join(self.temp_dir, 'demux.seqs')
+        with open(fna_fp, 'w') as f:
+            f.write(FASTA_EXAMPLE)
+        with File(demux_fp, "w") as f:
+            to_hdf5(fna_fp, f)
```
josenavas: Same here. All the files created in the test should be cleaned up.

josenavas:
Few comments.

qiita_ware/test/test_ebi.py
```diff
@@ -37,6 +43,18 @@ def setUp(self):
...
+        self.ppd_to_remove = []
+
+    def tearDown(self):
+        for f in self.files_to_remove:
+            if isdir(f):
+                rmtree(f)
+            else:
+                remove(f)
+
+        for _id in self.ppd_to_remove:
```
josenavas: I don't think this is needed.

Actually, I just noticed that this test file is unsafe!!! Can you put the class decorator `@qiita_test_checker()` to the test class, so we avoid to run the test if this a production environment? That decorator will automatically drop the schema after each test, so you don't need to delete the processed data, only the files that it generates.

qiita_ware/test/test_ebi.py
josenavas: I just realized that these files go to the folder that you're also marking to remove.

qiita_ware/test/test_ebi.py
```diff
@@ -405,6 +423,78 @@ def test_generate_curl_command(self):
...
+        exp_demux_samples = set(
+            ['1.SKD6.640190', '1.SKM6.640187', '1.SKD9.640182',
+             '1.SKB2.640194', '1.SKM8.640201', '1.SKM4.640180',
+             '1.SKM2.640199', '1.SKB3.640195', '1.SKB6.640176'])
+        ppd = PreprocessedData.create(Study(1),
+                                      "preprocessed_sequence_illumina_params",
+                                      1, [(demux_fp, 6)],
+                                      prep_template=PrepTemplate(1))
+        self.ppd_to_remove.append(ppd.id)
+        [self.files_to_remove.append(fps) for _, fps, _ in ppd.get_filepaths()]
```
josenavas: I think this construction is really strange, you're defining a list of None that you're never using...

Can you use this instead?
```python
self.files_to_remove.extend([fps for _, fps, _ in ppd.get_filepaths()])
```

qiita_ware/ebi.py
antgonza: Done as a Note.

qiita_ware/test/test_ebi.py
antgonza: Good point, made the change but note that I divided the test in readonly and write/read cause without that decorator the test took 0.246s vs. with all the tests with the decorator it took 31.447s so with that split it takes 1.194s.

qiita_ware/test/test_ebi.py
antgonza: Done

qiita_ware/test/test_ebi.py
antgonza: Done

qiita_ware/test/test_ebi.py
antgonza: Done but note that saying that's strange doesn't mean that is wrong. There are different ways to do the same thing.

josenavas:
Thanks @antgonza :+1: 
Splitting the tests between read-only and read-write is the way to go to speed up our test suite. I've only done it for the metadata template classes but we have an issue (https://github.com/biocore/qiita/issues/1029) to keep track of that. We may need to extend that list to all our tests

antgonza:
Can this get another review? @ElDeveloper @squirrelo @adamrp

qiita_ware/ebi.py
```diff
@@ -143,8 +147,9 @@ def __init__(self, preprocessed_data_id, action):
                              "term is neither one of the official terms nor "
                              "one of the user-defined terms in the ENA "
                              "ontology")
-        ts = datetime.now().strftime('%Y_%m_%d_%H:%M:%S')
-        self.ebi_dir = 'ebi_submission_%s_%s' % (preprocessed_data_id, ts)
+
+        self.ebi_dir = '%s/ebi_submission_%s' % (qiita_config.working_dir,
```
ElDeveloper: Preferably, this should be using `path.join` instead of string formatting.

qiita_ware/commands.py
```diff
@@ -6,12 +6,12 @@
...
+from os import makedirs
 from os.path import join, isdir
-from os import makedirs, remove
+from shutil import rmtree
 from functools import partial
-from tempfile import mkdtemp
-from gzip import open as gzopen
 from tarfile import open as taropen
+from tempfile import mkdtemp
 from moi.job import system_call
```
adamrp: add blank line before this `moi` import

qiita_ware/commands.py
```diff
@@ -6,12 +6,12 @@
...
```
adamrp: delete this blank line

qiita_ware/commands.py
```diff
@@ -20,9 +20,7 @@
 from qiita_core.qiita_settings import qiita_config
```
adamrp: delete this blank line

qiita_ware/commands.py
```diff
@@ -44,53 +42,25 @@ def submit_EBI(preprocessed_data_id, action, send, fastq_dir_fp=None):
...
-                with gzopen(sample_fp, 'w') as fh:
-                    for record in iterator:
-                        fh.write(record)
-                        wrote_sequences = True
-
-                if not wrote_sequences:
-                    remove(sample_fp)
-
+    # step 1
+    ebi_submission = EBISubmission(preprocessed_data_id, action)
```
adamrp: I think we might want to change the signature on the `EBISubmission` constructor. What needs to be supplied for an `ADD` action versus an `UPDATE` action is quite different (and an `UPDATE` cannot be performed using only information in a prep template)

antgonza: Agree on comment about requirements for UPDATE. However, I think sending the preprocessed_data_id is the way to go, as from there we know which submission is being updated and retrieve all study and sample information.

qiita_ware/ebi.py
```diff
@@ -981,3 +950,52 @@ def send_xml(self):
...
+
+        Returns
+        -------
+        demux_samples
+            List of successful demultiplexed samples
+
+        Notes
+        -----
+        - As a performace feature, this method will check if self.ebi_dir
+        already exist and, if it does, the script will assume that in a
```
ElDeveloper: exist -> exists

qiita_ware/ebi.py
```diff
@@ -981,3 +950,52 @@ def send_xml(self):
...
+                            fh.write(record)
+                            wrote_sequences = True
+
+                    if wrote_sequences:
+                        demux_samples.add(s)
+                    else:
+                        remove(sample_fp)
+
+        else:
+            demux_samples = set()
```
ElDeveloper: I think this whole block could be simplified using `glob.glob`.

ElDeveloper: Just to be clear what I mean is that you could do `for f in glob.glob(path.join(self.ebi_dir, '*.fastq.gz')`.

qiita_ware/ebi.py
```diff
@@ -981,3 +950,52 @@ def send_xml(self):
...
+        """Generates demultiplexed fastq
+
+        Returns
+        -------
+        demux_samples
+            List of successful demultiplexed samples
+
+        Notes
+        -----
+        - As a performace feature, this method will check if self.ebi_dir
```
ElDeveloper: This seems like a fine performance feature, but what would be a way to bypass this? i.e. a case where you know that there's something wrong with the generate per-sample FASTQ files and (luckily) your submission failed before? Maybe I am missing something but it seems like the only solution would be to delete the files manually?

qiita_ware/commands.py
```diff
@@ -44,53 +42,25 @@ def submit_EBI(preprocessed_data_id, action, send, fastq_dir_fp=None):
...
+
+    # step 2
+    ebi_submission.preprocessed_data.update_insdc_status('demuxing samples')
+    try:
+        demux_samples = ebi_submission.generate_demultiplexed_fastq()
+    except Exception as e:
+        if isdir(ebi_submission.ebi_dir):
+            rmtree(ebi_submission.ebi_dir)
+        ebi_submission.preprocessed_data.update_insdc_status(
+            'failed: %s' % str(e))
```
ElDeveloper: Not sure what are other's thought, but I think the full traceback should be logged in the Log table. Will specially be useful for debugging obscure problems.

qiita_ware/commands.py
adamrp: Yep, that makes sense. During the submission process, we should add to the database the EBI aliases for everything that is submitted, so that in the case of an update, we can retrieve them easily.

qiita_ware/commands.py
adamrp: I was thinking the same thing. :+1:

qiita_ware/ebi.py
adamrp: :+1:

qiita_ware/ebi.py
```diff
@@ -404,10 +380,6 @@ def add_sample_prep(self, sample_name, platform, file_type, file_path,
             raise KeyError("Sample %s: sample has not yet been associated "
                            "with the submission.")
 
-        if self.samples[sample_name]['prep']:
-            raise KeyError("Sample %s: multiple rows in prep with this sample "
-                           "id!" % sample_name)
-
```
adamrp: Why remove this check?

qiita_ware/ebi.py
antgonza: If this fails the code will remove the folder so I really do not see how that can happen. Anyway, if you feel strongly about this, I can put a parameter 'regenerate' by default FALSE ...

qiita_ware/ebi.py
antgonza: Cause this is impossible if you get the sample names from the DB. I think this check was there for loading from files ...

qiita_ware/test/test_ebi.py
```diff
@@ -11,18 +11,25 @@
...
-from os.path import join
+from os import close, remove, path, makedirs
+from os.path import join, isdir
+from shutil import rmtree
 from tempfile import mkstemp, gettempdir
 from unittest import TestCase, main
 from xml.dom import minidom
 from xml.etree import ElementTree as ET
 from functools import partial
+from h5py import File
```
adamrp: add blank line before this import

qiita_ware/ebi.py
adamrp: Good point :smile:

adamrp:
A few comments, but I think the code looks good. We should think carefully about what we want to do for `UPDATE`s, because I think they are the trickiest part here.

- For updates of metadata where none of the samples have been added or removed, I think it is relatively straightforward just to change the `action` that is passed.
- For all other updates (where samples are added or removed, or sequence files need to be updated), I believe we cannot actually do the update automatically. However, if we store the proper information in the database (i.e., the EBI aliases for everything that we submit), then we can automatically send an email requesting changes.

antgonza:
@adamrp, thanks. I have added your comment about update to the joy list: https://github.com/biocore/qiita/pull/1451

antgonza:
Ping??

qiita_ware/commands.py
```diff
@@ -44,53 +42,27 @@ def submit_EBI(preprocessed_data_id, action, send, fastq_dir_fp=None):
...
+        LogEntry.create('Runtime', str(e),
```
ElDeveloper: Instead of just saving the exception message, I would save the whole traceback (for debugging purposes), seems to be something like this:

```
>>> from traceback import format_exc
>>> try:
...    1/0
... except:
...    x = format_exc()
... 
>>> print x
  File "<stdin>", line 2, in <module>
ZeroDivisionError: integer division or modulo by zero
```

ElDeveloper:
The whole thing looks good to me, just one more comment and this should be ready for merge.

adamrp:
:+1:

PR #1452, opened 2015-08-26
Clarify data_type and prep_template parameters

### LINKED ISSUE #1443: Preprocessed data confusing documentation
The documentation for Preprocessed data says that data_type and prep_template are [optional](https://github.com/biocore/qiita/blob/master/qiita_db/data.py#L700) but the code verifies that neither of those are None ... should fix the documentation or remove those validations.

adamrp:
Fixes #1443


antgonza:
After more tests, I realized that if you create a PreprocessedData with just data_type and you ask for the prep_template from the object you will get an error that doesn't make sense. Could you also fix as part of this issue as IMOO is related?

```
  File "qiita_ware/test/test_ebi.py", line 432, in test_generate_demultiplexed_fastq
    print ppd.prep_template
  File "/Users/antoniog/svn_programs/qiita/qiita_db/data.py", line 849, in prep_template
    return TRN.execute_fetchlast()
  File "/Users/antoniog/svn_programs/qiita/qiita_db/sql_connection.py", line 510, in wrapper
    return func(self, *args, **kwargs)
  File "/Users/antoniog/svn_programs/qiita/qiita_db/sql_connection.py", line 782, in execute_fetchlast
    return self.execute()[-1][0][0]
```

josenavas:
In my opinion this is bringing a different question. Do we want to accept preprocessed data w/o prep templates? It is fine to have it w/o raw data, but I think a prep template is required since it contains further information (e.g. which data type it contains: 16S, 18S, ITS, metabolomics....)

antgonza:
IMOO no, others?

josenavas:
After discussion in today's meeting, we want to make the prep template required for any type of data.

adamrp:
This makes sense to me. So the solution here is to make `prep_template` a required parameter for `PreprocessedData.create`, right? And this means that I can remove the `data_type` parameter and all of the logic in the body of the code that aims to ensure agreement between the two?

josenavas:
I think that makes sense.

adamrp:
some tests are likely to fail, but I need to go AFK for a bit and will just let travis do its thing

antgonza:
@adamrp ping.

antgonza:
Closing for now: to many changes since this PR was opened and I think this has been superseded. Please reopen, if you feel otherwise.

PR #1453, opened 2015-08-26
Unify extend update gui

josenavas:
Users can now extend and update the sample template in a single button, rather than having to do it in two steps (first extending, then updating).

The extend functionality was not available for the prep template. This has been added here. The update prep template button in the GUI now also extends the prep template (same behavior as update sample template).

There has been minor improvements in the way the warning messages are shown to the user when updating templates.

I did not see any issue related with the changes here, please point me to any issue if I'm missing any.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,8 @@
 Version 0.2.0-dev (Changes since version 0.2.0 go here)
 -------------------------------------------------------
 
+* Users can now update and extend the templates at once using the "Update" button, rather than in two steps.
+
 Version 0.2.0 (2015-08-25)
 --------------------------
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,9 +3,7 @@
 Version 0.2.0-dev (Changes since version 0.2.0 go here)
 -------------------------------------------------------
 
-* Users can now update and extend sample templates by using the "Update sample templates" button, rather than having to do 2 steps with the previous design.
-* Users can now update and extend prep templates, using the "Update prep template" button in the prep template tab.
-* The 'extend' functionality has been moved from the sample template to the base metadata template object, so now both sample and prep templates can be extended.
+* Users can now change values and add samples and/or columns to sample and prep templates using the <kbd>Update</kbd> button (see the prep template and sample template tabs).
 
 Version 0.2.0 (2015-08-25)
 --------------------------
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,9 @@
 Version 0.2.0-dev (Changes since version 0.2.0 go here)
 -------------------------------------------------------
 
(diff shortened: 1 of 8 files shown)


CHANGELOG.md
```diff
@@ -3,6 +3,10 @@
 Version 0.2.0-dev (Changes since version 0.2.0 go here)
 -------------------------------------------------------
 
+* Users can now update and extend sample templates by using the "Update sample templates" button, rather than having to do 2 steps with the previous design.
```
antgonza: Are these really 3 new features? Perhaps a way to rephrase it is:
* The extent and update in templates (prep and sample) have been merged. Now you can do both via the GUI by clicking the Update button.

antgonza:
Agree that raising an error if nothing changed is not good but IMOO not raising a warning is also bad cause the user could have uploaded an older version and not realizing it.

qiita_db/metadata_template/test/test_prep_template.py
```diff
@@ -1241,6 +1241,77 @@ def test_can_be_updated_on_new(self):
                                  self.data_type)
         self.assertTrue(pt.can_be_updated({'barcode'}))
 
+    def test_extend_add_samples(self):
```
antgonza: Could you also add a test for update/extend at the same time? I couldn't find it in the code.

CHANGELOG.md
```diff
@@ -3,6 +3,10 @@
...
+* Users can now update and extend prep templates, using the "Update prep template" button in the prep template tab.
+* The 'extend' functionality has been moved from the sample template to the base metadata template object, so now both sample and prep templates can be extended.
+
```
ElDeveloper: It seems like the three points in this section could be synthesized in something like:

* Users can now change values and add columns to sample and prep templates using the <kbd>Update</kbd> button (see the prep template and sample template tabs).

Or something along those lines ...

qiita_db/metadata_template/base_metadata_template.py
```diff
@@ -1065,6 +1057,20 @@ def categories(self):
 
         return cols
 
+    def extend(self, md_template):
+        """Adds the given template to the current one
+
+        Parameters
+        ----------
+        md_template : DataFrame
+            The metadata template contents indexed by sample Ids
```
ElDeveloper: Ids -> ids

ElDeveloper:
Looks good to me! :+1:

qiita_db/metadata_template/base_metadata_template.py

CHANGELOG.md
josenavas: Thanks, done

qiita_db/metadata_template/test/test_prep_template.py
josenavas: Done

josenavas:
Ready for another review.

antgonza:
:+1: if tests pass

PR #1462, opened 2015-09-01
[WIP] Artifact db changes

josenavas:
This is an initial set of changes in the DB to represent an "Artifact" (i.e. data) instead of having the previous division of RawData, (pre)processed data. There are still a few missing things (including some peripheral tables), but I would like to get the feedback on the DB schema design before spending too much time.

Ask any question that may rise from looking into this!


josenavas:
@biocore/qiita-dev This is ready for review - tests are going to fail since this is only changing the DB - no python code is modified.

Few considerations:
 - The amount of DB changes in this PR is really big. I recommend reviewers to get some coffee and pay attention to this PR as a lot of things can go bad.
 - If you want to run the patch in your system, do a pgdump first so you don't loose your "release" data.
 - There was a lot of manual SQL generation. I think the SQL, the DBS and the HTML are in sync, but it will be worth to check those just in case.

qiita_db/support_files/patches/33.sql
```diff
@@ -0,0 +1,435 @@
...
+	can_be_submitted_to_vamps           bool DEFAULT 'FALSE' NOT NULL,
+    submitted_to_vamps                  bool DEFAULT 'FALSE' NOT NULL,
+    CONSTRAINT pk_artifact PRIMARY KEY ( artifact_id )
+ ) ;
+CREATE INDEX idx_artifact_0 ON qiita.artifact ( visibility_id ) ;
+CREATE INDEX idx_artifact_1 ON qiita.artifact ( artifact_type_id ) ;
+CREATE INDEX idx_artifact ON qiita.artifact ( command_id ) ;
+COMMENT ON TABLE qiita.artifact IS 'Represents data in the system';
+COMMENT ON COLUMN qiita.artifact.visibility_id IS 'If the artifact is sandbox, awaiting_for_approval, private or public';
+COMMENT ON COLUMN qiita.artifact.file_status IS 'If it is linking, unlinking or idle';
```
antgonza: So the idea is that an artifact row is created before the file is generated? Not sure if that's possible. However, I can see that the unlinking happens before the row is deleted.

qiita_db/support_files/patches/33.sql
```diff
@@ -0,0 +1,435 @@
...
+    artifact_id                         bigserial  NOT NULL,
+    generated_timestamp                 timestamp  NOT NULL,
+    command_id                          bigint  ,
+    command_parameters_id               bigint  ,
+    visibility_id                       bigint  NOT NULL,
+    file_status                         varchar  NOT NULL,
+    artifact_type_id                    integer  ,
+    can_be_submitted_to_ebi             bool DEFAULT 'FALSE' NOT NULL,
+	can_be_submitted_to_vamps           bool DEFAULT 'FALSE' NOT NULL,
+    submitted_to_vamps                  bool DEFAULT 'FALSE' NOT NULL,
```
antgonza: Should this be here or should it be a different table?

qiita_db/support_files/patches/33.sql
antgonza: Looking below perhaps the file_status should be in the artifact_filepath table.

qiita_db/support_files/patches/33.sql
```diff
@@ -0,0 +1,435 @@
...
+    CONSTRAINT pk_artifact PRIMARY KEY ( artifact_id )
+ ) ;
+CREATE INDEX idx_artifact_0 ON qiita.artifact ( visibility_id ) ;
```
antgonza: Should the index be each element or the combination of the 3?

qiita_db/support_files/patches/33.sql
```diff
@@ -0,0 +1,435 @@
...
+COMMENT ON COLUMN qiita.artifact.file_status IS 'If it is linking, unlinking or idle';
+ALTER TABLE qiita.artifact ADD CONSTRAINT fk_artifact_type FOREIGN KEY ( artifact_type_id ) REFERENCES qiita.artifact_type( artifact_type_id )    ;
+
+-- Artifact filepath table - relates an artifact with its files
+CREATE TABLE qiita.artifact_filepath (
+    artifact_id          bigint  NOT NULL,
+    filepath_id          bigint  NOT NULL,
+    CONSTRAINT idx_artifact_filepath PRIMARY KEY ( artifact_id, filepath_id )
+ ) ;
+CREATE INDEX idx_artifact_filepath_artifact ON qiita.artifact_filepath ( artifact_id ) ;
```
antgonza: Should the index be on each one element or both? As there are only 2 elements not sure on the advantage of indexing by both.

qiita_db/support_files/patches/33.sql
```diff
@@ -0,0 +1,435 @@
...
+
+-- Parent artifact table - keeps track of the procenance of a given artifact.
+-- If an artifact doesn't have a parent it means that it was uploaded by the user.
+CREATE TABLE qiita.parent_artifact (
+    artifact_id          bigint  NOT NULL,
+    parent_id            bigint  NOT NULL,
+    CONSTRAINT idx_parent_artifact PRIMARY KEY ( artifact_id, parent_id )
+ ) ;
+CREATE INDEX idx_parent_artifact_artifact ON qiita.parent_artifact ( artifact_id ) ;
+CREATE INDEX idx_parent_artifact_parent ON qiita.parent_artifact ( parent_id ) ;
```
antgonza: Same comment about 2 elements and indexing.

qiita_db/support_files/patches/33.sql
```diff
@@ -0,0 +1,435 @@
...
+    name                 varchar  NOT NULL,
+    version              varchar  NOT NULL,
+    description          varchar  NOT NULL,
+    CONSTRAINT pk_software PRIMARY KEY ( software_id )
+ ) ;
+
+-- soft_command table - holds the information of a command in a given software
+-- this table should be renamed to command once the command table in the
+-- analysis table is merged with this one
+CREATE TABLE qiita.soft_command (
```
antgonza: I rather have software command.

qiita_db/support_files/patches/33.sql
```diff
@@ -0,0 +1,435 @@
...
+ ) ;
+CREATE INDEX idx_soft_command ON qiita.soft_command ( software_id ) ;
+ALTER TABLE qiita.soft_command ADD CONSTRAINT fk_soft_command_software FOREIGN KEY ( software_id ) REFERENCES qiita.software( software_id );
+
+-- Publication table - holds the minimum information for a given publication
+-- It is useful to keep track of the publication of the studies and the software
+-- used for processing artifacts
+CREATE TABLE qiita.publication (
+    doi                  varchar  NOT NULL,
+    pubmed_id            integer  ,
```
antgonza: This info already exists on study, are you gonna delete from there? Also note the data type difference.

qiita_db/support_files/patches/33.sql
```diff
@@ -0,0 +1,435 @@
...
+    WHERE visibility = 'public';
+
+-- Artifact table - holds an abstract data object from the system
+CREATE TABLE qiita.artifact (
+    artifact_id                         bigserial  NOT NULL,
+    generated_timestamp                 timestamp  NOT NULL,
+    command_id                          bigint  ,
+    command_parameters_id               bigint  ,
+    visibility_id                       bigint  NOT NULL,
+    file_status                         varchar  NOT NULL,
```
squirrelo: I don't see any controlled vocab for this column.

qiita_db/support_files/patches/33.sql
squirrelo: It's an index for that specific column for FK reasons, so it should be fine.

qiita_db/support_files/patches/33.sql
squirrelo: If you index both at once, the index will combine the columns and only work if you are searching both at once. Not useful for when you have FK, as in this case.

qiita_db/support_files/patches/33.sql
```diff
@@ -0,0 +1,435 @@
...
+    parameters_table     varchar  NOT NULL,
+    CONSTRAINT pk_soft_command PRIMARY KEY ( command_id )
+ ) ;
+CREATE INDEX idx_soft_command ON qiita.soft_command ( software_id ) ;
+ALTER TABLE qiita.soft_command ADD CONSTRAINT fk_soft_command_software FOREIGN KEY ( software_id ) REFERENCES qiita.software( software_id );
+
+-- Publication table - holds the minimum information for a given publication
+-- It is useful to keep track of the publication of the studies and the software
+-- used for processing artifacts
+CREATE TABLE qiita.publication (
```
squirrelo: This table can be wrapped into software_publication by adding DOI as a column and making both DOI and pmid nullable. Since we can add PMID, DOI, or both, this removes sone joins and tidys up the DB.

qiita_db/support_files/patches/33.sql
```diff
@@ -0,0 +1,435 @@
...
+    software_id          bigint  NOT NULL,
+    publication_doi      varchar  NOT NULL,
+    CONSTRAINT idx_software_publication_0 PRIMARY KEY ( software_id, publication_doi )
+ ) ;
+CREATE INDEX idx_software_publication_software ON qiita.software_publication ( software_id ) ;
+CREATE INDEX idx_software_publication_publication ON qiita.software_publication ( publication_doi ) ;
+ALTER TABLE qiita.software_publication ADD CONSTRAINT fk_software_publication FOREIGN KEY ( software_id ) REFERENCES qiita.software( software_id )    ;
+ALTER TABLE qiita.software_publication ADD CONSTRAINT fk_software_publication_0 FOREIGN KEY ( publication_doi ) REFERENCES qiita.publication( doi )    ;
+
+-- Add remaining FK to the artifact table. Creating here since the target tables
```
squirrelo: I don't like this addition of FK so far after table creation, as it can lead to some unwanted circular references or other issues. Better to rearrange table creation to make sure the links are logical and one way.

qiita_db/support_files/patches/33.sql
```diff
@@ -0,0 +1,435 @@
...
+ALTER TABLE qiita.artifact ADD CONSTRAINT fk_artifact_visibility FOREIGN KEY ( visibility_id ) REFERENCES qiita.visibility( visibility_id )    ;
+ALTER TABLE qiita.artifact ADD CONSTRAINT fk_artifact_soft_command FOREIGN KEY ( command_id ) REFERENCES qiita.soft_command( command_id )    ;
+
+-- Create a function to infer the visibility of the artifact from the
+-- raw data
+CREATE FUNCTION infer_rd_status(rd_id bigint, st_id bigint) RETURNS bigint AS $$
+    DECLARE
+        result bigint;
+    BEGIN
+        CREATE TEMP TABLE irds_temp
```
squirrelo: Why the temp table as opposed to just querying into a variable?

qiita_db/support_files/patches/33.sql
```diff
@@ -0,0 +1,435 @@
...
+        IF EXISTS(SELECT * FROM irds_temp WHERE processed_data_status_id = 2) THEN
+            result := 2;
+        ELSIF EXISTS(SELECT * FROM irds_temp WHERE processed_data_status_id = 3) THEN
+            result := 3;
+        ELSIF EXISTS(SELECT * FROM irds_temp WHERE processed_data_status_id = 1) THEN
+            result := 1;
+        ELSE
+            result := 4;
+        END IF;
+        DROP TABLE irds_temp;
```
squirrelo: Temp tables don't need drop, they drop automatically.

qiita_db/support_files/patches/33.sql
```diff
@@ -0,0 +1,435 @@
...
+        END IF;
+        DROP TABLE ippds_temp;
+        RETURN result;
+    END;
+$$ LANGUAGE plpgsql;
+
+-- Populate the software and soft_command tables so we can assignt the
+-- correct values to the preprocessed and processed tables
+INSERT INTO qiita.software (name, version, description) VALUES
+    ('QIIME', '1.9.1', 'Quantitative Insigts Into Microbial Ecology (QIIME) is an open-source bioinformatics pipeline for performing microbiome analysis from raw DNA sequencing data');
```
squirrelo: Insigts -> Insights

qiita_db/support_files/patches/33.sql
```diff
@@ -0,0 +1,435 @@
...
+    END;
+$$ LANGUAGE plpgsql;
+
+-- Create a function to infer the visibility of the artifact from the
+-- preprocessed data
+CREATE FUNCTION infer_ppd_status(ppd_id bigint) RETURNS bigint AS $$
+    DECLARE
+        result bigint;
+    BEGIN
+        CREATE TEMP TABLE ippds_temp
```
squirrelo: Same temp table question

qiita_db/support_files/patches/33.sql
```diff
@@ -0,0 +1,435 @@
...
+    FOREIGN KEY ( artifact_id ) REFERENCES qiita.artifact(artifact_id);
+
+-- We need to modify the ebi run accession table to point to the artifact table
+-- rather than to the preprocessed data table
+ALTER TABLE qiita.ebi_run_accession ADD artifact_id bigint;
+CREATE INDEX idx_ebi_run_accession_artifact_id ON qiita.ebi_run_accession (artifact_id);
+ALTER TABLE qiita.ebi_run_accession ADD CONSTRAINT fk_ebi_run_accesion_artifact
+    FOREIGN KEY ( artifact_id ) REFERENCES qiita.artifact(artifact_id);
+
+-- We need to modufy tge analysis_sample table to point to the artifact table
```
squirrelo: modufy tge -> modify the

qiita_db/support_files/patches/33.sql
josenavas: That is because if an error occurs, it is stored in this variable. That is what we've been doing so far, and we still have not found a good alternative for this.

qiita_db/support_files/patches/33.sql
josenavas: I though about that, but there is no other information. That other table will only have 2 columns: artifact_id and submitted_to_vamps, which shows that it is only an attribute of the artifact. Unless we have more information about the vamps submission, I don't think it is worth having another table.

qiita_db/support_files/patches/33.sql
josenavas: Those indices are created automatically by dbschema. I just renamed them to avoid name conflicts - but as @squirrelo says, those are there due to the FK.

qiita_db/support_files/patches/33.sql
josenavas: That cannot be in the artifact_filepath table. However, your first comment is really good. The only reason we had the linking status was because we needed to hack at some point the creation of a raw data w/o files. I think we can remove this column and don't allow artifacts w/o files.

qiita_db/support_files/patches/33.sql
josenavas: Same as above.

qiita_db/support_files/patches/33.sql
josenavas: Same as above.

qiita_db/support_files/patches/33.sql
josenavas: Changed done. Note that I would prefer to have only command, as the comment reflects, but that needs changes on the analysis pipeline. I think we can rename once the refactor on the analysis pipeline is done.

qiita_db/support_files/patches/33.sql
josenavas: Thanks for noticing the type difference. You are correct. The idea is that we remove that from the study and we use this table to store all the publications in the DB.

The reason why I'm not doing such movement here is because in the current database we have the PMID and not the DOI, si it will require some manual intervention to actually get all the DOI information from the different PMIDS.

Alternatively, I can put the PMID in the DOI for the time being and we can fix later. What do you think?

qiita_db/support_files/patches/33.sql
josenavas: See my comment below, in @antgonza comment.

qiita_db/support_files/patches/33.sql
antgonza: Agree, no artifacts without files

qiita_db/support_files/patches/33.sql
josenavas: Good point - I was biased by the order in what dbschema generated the stuff. I re-ordered the table creation.

qiita_db/support_files/patches/33.sql
josenavas: I'm unsure on how to perform the functionality done in this function w/o using a temp table. Can you provide an alternative that executes the same functionality outlined here?

qiita_db/support_files/patches/33.sql
josenavas: The table drops on commit. However, anything is commited until the end, so if you execute this function more than time, the table already exists. Thus, you need to drop to allow the subsequent executions of the function.

qiita_db/support_files/patches/33.sql
josenavas: Same answer.

qiita_db/support_files/patches/33.sql
josenavas: Done

qiita_db/support_files/patches/33.sql
josenavas: hahaha changed....

qiita_db/support_files/patches/33.sql
antgonza: Fine with me to have both but my guess is that pubmed id is not gonna work as int. Checking the DB currently there are DOI and the word test in that table.

josenavas: I'm changing the pubmed_id to be a varchar - that was a mistake (sorry I made not this clear in my previous comment).

The main issue is that the PK in this table is the doi, so it cannot be null. So in order to transfer the information from the current DB, we either use the current PMID as doi so we can transfer such information.

I'm sorry, but I don't understand this:
> Checking the DB currently there are DOI and the word test in that table.

antgonza: I just wanted to let you know that I checked the information in the table study_pmid in the main deploy and the table contains entries with only numbers, only integers, a combination of both and one of them is a DOI entry.

josenavas: Oh, ok! Thanks!

I will connect the study with this table and transfer the data here. I will insert the same value in both columns.

josenavas:
Comments addressed / answered.

antgonza:
Looks good to me :+1:

qiita_db/support_files/patches/33.sql
squirrelo: It looks to me as if you're querying out all processed data made from a raw data, and setting the result based on that. If so, you can do this in a single query, with no need for the function:
```sql
SELECT CASE MIN(processed_data_status_id)
	WHEN 1 THEN 1
	WHEN 2 THEN 2
	WHEN 3 THEN 3
	ELSE 4
END
FROM qiita.processed_data
    JOIN qiita.preprocessed_processed_data USING (processed_data_id)
    JOIN qiita.prep_template_preprocessed_data USING (preprocessed_data_id)
    JOIN qiita.prep_template USING (prep_template_id)
    JOIN qiita.study_processed_data USING (processed_data_id)
WHERE raw_data_id = rd_id AND study_id = st_id;
```

squirrelo: Sorry, that should be on the above comment.

josenavas: Your solution is not correct - the order of the if statements is important. I first need to check for 2, then for 3, then for 1 and if none of them are present, then 4. That's why I can't use a min and I have to use an exist to check each of them independently. I can use multiple queries, but since it is a huge joint, I think is better to retrieve the data once and just query it.

qiita_db/support_files/patches/33.sql
squirrelo: new column, nullable, called `err_msg`?

squirrelo: This way we don't have one column pulling double duty, which will make logic much cleaner.

josenavas: No worries, this columns has been removed.

2015-10-28 11:02 GMT-07:00 Joshua Shorenstein <notifications@github.com>:

> In qiita_db/support_files/patches/33.sql
> <https://github.com/biocore/qiita/pull/1462#discussion_r43294359>:
>
> > +ALTER TABLE qiita.visibility RENAME COLUMN processed_data_status_description TO visibility_description;
> > +UPDATE qiita.visibility
> > +    SET visibility_description = 'Only visible to the owner and shared users'
> > +    WHERE visibility = 'private';
> > +UPDATE qiita.visibility
> > +    SET visibility_description = 'Visible to everybody'
> > +    WHERE visibility = 'public';
> > +
> > +-- Artifact table - holds an abstract data object from the system
> > +CREATE TABLE qiita.artifact (
> > +    artifact_id                         bigserial  NOT NULL,
> > +    generated_timestamp                 timestamp  NOT NULL,
> > +    command_id                          bigint  ,
> > +    command_parameters_id               bigint  ,
> > +    visibility_id                       bigint  NOT NULL,
> > +    file_status                         varchar  NOT NULL,
>
> new column, nullable, called err_msg?
>
> —
> <https://github.com/biocore/qiita/pull/1462/files#r43294359>.
>



-- 
Jose Navas

qiita_db/support_files/patches/33.sql
squirrelo: What do you gain by having two tables over one though? From that conversation I'm not sure it's necessary. You can PK on all 3 columns instead of just the two, and you can do the data move gracefully as you don't need a DOI to attach a PMID to.

josenavas: What if a single publications is used for a study and a method, or an analysis (we will support this)?

josenavas: And note that we're requiring the DOI, not the PMID.

squirrelo: Requiring DOI is a bad assumption, since most biologists go off PMID, and DOI as a secondary if no PMID is available.

josenavas: DOI is more general than pubmed id. Every publication with pubmed_id will have a DOI, while the other direction is not true. Can somebody else weight in here?

squirrelo: Here is a minimal example:
screen shot 2015-10-28 at 11 49 27

This allows anything to link to a publication (or multiple) using a join table. Also, with a CHECK added to the DOI and PMID columns, you can make sure that at least one of them exists.

josenavas: DOI are naturally identifiers...

PR #1464, opened 2015-09-03
Ebi step 3

antgonza:
Sorry for the large amount of changes. Have no other idea of how to divide nicely, except making the DB changes in a different PR but that's small compared to the rest so decided to leave it here.

General comments:
- Added scientific name (required for EBI submission) to the populate_test_db.sql so we can test 
- Remember that commands.submit_EBI is a work in progress and is just so we have an idea of how the calls will look
- Removed iter_file_via_list_of_dicts, add_sample, add_sample_prep and add_samples_from_templates as these were required for from-file submission, which we will not support anymore as everything is being loaded from the DB
- Removed _generate_library_descriptor as this was using/forcing to have some library design fields that are not required anymore
- Removed all write_*_xml as now we simply call the write/generate functions
- Removed write_all_xml_files cause this will defeat the idea of having networkx call/submit jobs. See point 2 of this list.
- Replaced gzopen -> GzipFile cause this will allow us to set mtime (float to calculate gzip's timestamp) for testing. FYI: gzip incorporates the timestamp in the file so we need to know/define the timestamp for testing.
- Removed all qiita_ware/test/test_data/* as now the files are generated and deleted on the fly.

Who wants to review?


qiita_db/metadata_template/test/test_sample_template.py
```diff
@@ -1862,12 +1864,9 @@ def test_to_dataframe(self):
             'water_content_soil', 'elevation', 'temp', 'tot_nitro',
             'samp_salinity', 'altitude', 'env_biome', 'country', 'ph',
             'anonymized_name', 'tot_org_carb', 'description_duplicate',
-            'env_feature'})
+            'env_feature', 'scientific_name'})
 
     def test_check_restrictions(self):
-        obs = self.tester.check_restrictions([SAMPLE_TEMPLATE_COLUMNS['EBI']])
```
josenavas: Instead of deleting this lines, can you just check that is returning an empty set? I think this was the original idea of this test but got lost when scientific_name become required. This way we will have a test for a template that is not missing anything.

qiita_ware/ebi.py
```diff
@@ -143,33 +126,76 @@ def __init__(self, preprocessed_data_id, action):
...
-        self.submission_xml_fp = None
+            error_msgs.append("Unrecognized investigation type: '%s'. This "
+                              "term is neither one of the official terms nor "
+                              "one of the user-defined terms in the ENA "
+                              "ontology." % it)
+
+        self.ebi_dir = join(qiita_config.working_dir,
+                            'ebi_submission_%d' % preprocessed_data_id)
+        get_output_fp = partial(join, self.ebi_dir, 'xml_dir')
+        self.xml_dir = get_output_fp('')
```
josenavas: you don't need to call it with an empty string:
```python
In [6]: get_output_fp = partial(join, './', 'xml_dir')

In [7]: get_output_fp()
Out[7]: './xml_dir'
```

qiita_ware/ebi.py
```diff
@@ -143,33 +126,76 @@ def __init__(self, preprocessed_data_id, action):
...
-                          'mimarks-survey': 'AMPLICON'}
-        lib_selections = {'mimarks-survey': 'PCR'}
-        lib_sources = {}
-
-        self.library_strategy = lib_strategies.get(key, "OTHER")
-        self.library_source = lib_sources.get(key, "METAGENOMIC")
-        self.library_selection = lib_selections.get(key, "unspecified")
-
-        # This will hold the submission's samples, keyed by the sample name
+        # ignore data type for the restrictions as we assume that this was
```
josenavas: This comment is not accurate because you're actually checking the data type on line 153.

Anyways, I don't think the logic here is correct. In here we should be checking the columns required by EBI. In the current code, you're checking all the required columns, for multiple reasons - however I think that the correct thing to do is checking only the EBI columns.

Also, you don't need to actually check the data type - the templates have a property that will return the columns that are required for their datatype:
```python
In [11]: st = SampleTemplate(1)

In [12]: st.columns_restrictions
Out[12]: 
{'EBI': Restriction(columns={'physical_specimen_location': 'varchar', 'collection_timestamp': 'timestamp', 'scientific_name': 'varchar', 'taxon_id': 'integer'}, error_msg='EBI submission disabled'),
 'qiita_main': Restriction(columns={'description': 'varchar', 'physical_specimen_remaining': 'bool', 'dna_extracted': 'bool', 'longitude': 'float8', 'sample_type': 'varchar', 'latitude': 'float8', 'host_subject_id': 'varchar'}, error_msg='Processed data approval disabled')}

In [13]: pt = PrepTemplate(1)

In [14]: pt.columns_restrictions
Out[14]: 
{'EBI': Restriction(columns={'center_name': 'varchar', 'platform': 'varchar', 'primer': 'varchar', 'experiment_design_description': 'varchar', 'library_construction_protocol': 'varchar'}, error_msg='EBI submission disabled'),
 'demultiplex': Restriction(columns={'primer': 'varchar', 'barcode': 'varchar'}, error_msg='Demultiplexing disabled. You will not be able to preprocess your raw data'),
 'demultiplex_multiple': Restriction(columns={'primer': 'varchar', 'barcode': 'varchar', 'run_prefix': 'varchar'}, error_msg='Demultiplexing with multiple input files disabled. If your raw data includes multiple raw input files, you will not be able to preprocess your raw data')}
```

Given this, I think the correct path here is:
```python
restrictions = self.sample_template.columns_restrictions['EBI'].columns.keys()
restrictions.extend(self.prep_template.columns_restrictions['EBI'].columns.keys()
```

antgonza: I don't have a strong feeling about this just thought it was easier to remove than add. If nobody opposes I will only use EBI restrictions.

josenavas: Actually, thinking a bit more about this, you may be able to take advantage of the [`check_restrictions`](https://github.com/biocore/qiita/blob/master/qiita_db/metadata_template/base_metadata_template.py#L1311) function from the templates:

```python
pt_missing = self.prep_template.check_restrictions([self.prep_template.column_restrictions['EBI'])
st_missing = self.sample_template.check_restrictions([self.sample_template.column_restrictions['EBI'])
```
And then let the user know where are the columns missing (i.e. not only telling them which columns are missing but where):
```python
if st_missing:
    # add error_msg for sample template
if pt_missing:
    # add error_msg for prep tempalte
```

qiita_ware/ebi.py
```diff
@@ -143,33 +126,76 @@ def __init__(self, preprocessed_data_id, action):
...
+        missing_columns = restrictions.difference(existing_columms)
+        # testing if there are any missing columns
+        if missing_columns:
+            error_msgs.append(
+                "You are missing some columns in study #%d to have a valid "
+                "submission #%d: %s." % (s.id, preprocessed_data_id,
+                                         ', '.join(list(missing_columns))))
+
+        # generating all samples from sample template
+        # self.samples = dict(self.sample_template.items())
```
josenavas: Remove commented code

qiita_ware/ebi.py
```diff
@@ -143,33 +126,76 @@ def __init__(self, preprocessed_data_id, action):
...
         self.samples = {}
+        self.samples_prep = {}
+        nvp = []
+        for k, v in self.sample_template.items():
```
josenavas: minor, but can you use `viewitems`?

qiita_ware/ebi.py
```diff
@@ -143,33 +126,76 @@ def __init__(self, preprocessed_data_id, action):
...
+            # validating required fields
+            if 'platform' not in sample_prep:
+                nvp.append(k)
+            else:
+                platform = sample_prep['platform'].upper()
+                if platform not in self.valid_platforms:
+                    nvp.append(k)
+
+            self.samples[k] = v
+            self.samples_prep[k] = self.prep_template[k]
```
josenavas: really extremely minor `self.samples_prep[k] = sample_prep`
The reason is that you're doing one less dictionary look-up, but again really minor

qiita_ware/ebi.py
```diff
@@ -143,33 +126,76 @@ def __init__(self, preprocessed_data_id, action):
...
+
+        if nvp:
+            error_msgs.append(
+                "These samples from study #%d and preprocessed data #%d do "
```
josenavas: I _think_ it might be more useful to provide the prep template id, as the prep template is the one who should include the correct value for platform

qiita_ware/ebi.py
```diff
@@ -328,6 +316,7 @@ def generate_sample_xml(self):
                                              "/sra_1_3/SRA.sample.xsd"})
 
         for sample_name, sample_info in sorted(viewitems(self.samples)):
+            sample_info = dict(sample_info)
```
josenavas: I'm curious here - what is the original type of sample_info and why are you casting it here to a dict?

antgonza: object => it's simply so I can get all the elements and their values. Additionally this is helpful as I can remove the ones that are "required" and know which ones are left so they can be added as sample_attributes.

qiita_ware/ebi.py
```diff
@@ -455,20 +376,20 @@ def generate_experiment_xml(self):
 
         Returns
         -------
-        xml.etree.Element
-            The root elelement of the generated ``ElementTree``
+        str
+            string with experiment XML values
         """
         study_alias = self._get_study_alias()
         experiment_set = ET.Element('EXPERIMENT_SET', {
```
josenavas: I know you have not modified this but, some of the values in the parameter dictionary are repeated in multiple places. Actually, it looks like the only thing that changes is the string after `/sra_1_3/` (i.e. `SRA.experiment.xsd` or `SRA.sample.xsd`, `SRA.run.xsd`, etc). Is it possible to create some globals and/or a function that returns this dictionary? It will reduce code duplication

qiita_ware/ebi.py
```diff
@@ -455,20 +376,20 @@ def generate_experiment_xml(self):
...
             "xmlns:xsi": "http://www.w3.org/2001/XMLSchema-instance",
             "xsi:noNamespaceSchemaLocation": "ftp://ftp.sra.ebi.ac.uk/meta/xsd"
                                              "/sra_1_3/SRA.experiment.xsd"})
-        for sample_name, sample_info in sorted(self.samples.items()):
+        for sample_name, sample_prep in sorted(self.samples_prep.items()):
             sample_alias = self._get_sample_alias(sample_name)
-
             experiment_alias = self._get_experiment_alias(sample_name)
+            sample_prep = dict(self.samples_prep[sample_name])
```
josenavas: Same question re casting to dict - shouldn't it be cheaper to create it as a dict from the beginning?

qiita_ware/ebi.py
```diff
@@ -482,53 +403,81 @@ def generate_experiment_xml(self):
...
+                library_source.text = sample_prep.pop('library_source')
+            if 'library_selection' in sample_prep:
+                library_selection = ET.SubElement(library_descriptor,
+                                                  "LIBRARY_SELECTION")
+                library_selection.text = sample_prep.pop('library_selection')
+            if 'library_layout' in sample_prep:
+                library_layout = ET.SubElement(library_descriptor,
+                                               "LIBRARY_LAYOUT")
+                library_layout.text = sample_prep.pop('library_layout')
+                # no idea what SINGLE means
```
josenavas: That's an interesting comment... does anybody know what we have to have single here?

qiita_ware/ebi.py
antgonza: Well, those are created as objects in qiita_db so converting to dicts for code simplicity

qiita_ware/ebi.py
antgonza: Yup, added the comment for that reason. @adamrp, any ideas?

qiita_ware/ebi.py
```diff
@@ -482,53 +403,81 @@ def generate_experiment_xml(self):
...
+            library_name.text = self._get_library_name(sample_name)
+
+            lcp = ET.SubElement(library_descriptor,
+                                "LIBRARY_CONSTRUCTION_PROTOCOL")
+            lcp.text = escape(clean_whitespace(
+                sample_prep.pop('library_construction_protocol')))
+
+            # these are not requiered field but present add them in the right
+            # format
+            if 'library_strategy' in sample_prep:
```
josenavas: There is a fair amount of code duplication in here, it can be collapsed:
```python
keys = ['library_strategy', 'library_source', 'library_selection', 'library_layout']
for k in keys:
    if k in sample_prep:
        sub_element = ET.SubElement(library_description, k.upper())
        sub_element.text = sample_prep.pop(k)
```
Also, the list `keys` can be defined as a class attribute, since it is a constant.

qiita_ware/ebi.py
```diff
@@ -482,53 +403,81 @@ def generate_experiment_xml(self):
...
 
             self._generate_spot_descriptor(design, platform)
 
             platform_element = ET.SubElement(experiment, 'PLATFORM')
             platform_info = ET.SubElement(platform_element,
                                           platform.upper())
-            instrument_model = ET.SubElement(platform_info,
-                                             'INSTRUMENT_MODEL')
+            instrument_model = ET.SubElement(platform_info, 'INSTRUMENT_MODEL')
             instrument_model.text = 'unspecified'
```
josenavas: `instrument_model` hardcoded to unspecified? Isn't this saying that instrument model is actually a required value that we are always putting 'unspecified'? what if the user did define this in the prep template?

antgonza: Good point, forgot to move this one to the if feature in sample_prep ..., will fix.

qiita_ware/ebi.py
```diff
@@ -482,53 +403,81 @@ def generate_experiment_xml(self):
...
+        for sample_name, sample_prep in viewitems(self.samples_prep):
+            sample_prep = dict(sample_prep)
 
             experiment_alias = self._get_experiment_alias(sample_name)
 
-            file_type = sample_info['prep']['file_type']
-            file_path = sample_info['prep']['file_path']
+            # We only submit fastq
+            file_type = 'fastq'
+            file_path = join(self.ebi_dir, "%s.fastq.gz" % sample_name)
```
josenavas: I personally don't like this. The sample fastq file has been generated somewhere else, which means that the code to generate the filepath is duplicated. Should this be stored in some EBI attribute so we just query it instead of re-generating again here? The main concern is that if that code changes, we will need to change it in 2 places

antgonza: I did it this way cause this is how it was done before but I can implement your suggestions really quickly. The idea will be to add a `sample_demux_fps` variable that is filled in `generate_demultiplexed_fastq`.

qiita_ware/ebi.py
```diff
@@ -591,231 +521,59 @@ def generate_submission_xml(self, action):
...
         """
-        if xml_gen_fn_arg is None:
-            xml_element = xml_gen_fn()
-        else:
-            xml_element = xml_gen_fn(xml_gen_fn_arg)
+        if not isdir(self.xml_dir):
+            makedirs(self.xml_dir)
 
-        xml = minidom.parseString(ET.tostring(xml_element))
+        xml = minidom.parseString(text)
```
josenavas: I think this is redundant... we are doing `ET.tostring` at the end of each `generate_FOO_xml`. Shouldn't it be better if in those functions we do a `ET.toxml(encoding='UTF-8') and here we just write that string out?

I don't understand why we need to generate another XML object here just to write it out again....

antgonza: Good question, perhaps so the text test are built easier? @adamrp, do you know?

qiita_ware/ebi.py
```diff
@@ -993,6 +744,8 @@ def generate_demultiplexed_fastq(self, rewrite_fastq=False):
                     if wrote_sequences:
                         demux_samples.add(s)
                     else:
+                        del(self.samples[s])
```
josenavas: Can you explain the logic behind these `del` calls? Why are they useful?

antgonza: Will add a note but they are required as if we do not delete the sample from the sample/prep template the methods will try to submit them even if they do not have a fastq.gz file AKA the no sequences error.

qiita_ware/test/test_ebi.py
```diff
@@ -35,20 +34,16 @@
...
-        ebi_test_file = partial(join, self.path)
-
-        self.sample1_fp = ebi_test_file('sample1.fastq.gz')
-        self.sample2_fp = ebi_test_file('sample2.fastq.gz')
-        self.sample3_fp = ebi_test_file('sample3.fastq.gz')
-
         self.temp_dir = gettempdir()
         self.files_to_remove = []
 
+        if not isdir(self.temp_dir):
```
josenavas: Why you have to specifically test here if such directory exists? Also I think that using `gettempdir()` in line 37 is not correct, what you actually want to use is [`tempfile.mkdtemp`](https://docs.python.org/2/library/tempfile.html#tempfile.mkdtemp)

qiita_ware/test/test_ebi.py
```diff
@@ -35,20 +34,16 @@
...
+            makedirs(self.temp_dir)
+        self.files_to_remove.append(self.temp_dir)
```
josenavas: If gettempdir is set to `/tmp` you will be trying to remove `/tmp` from the system

qiita_ware/test/test_ebi.py
```diff
@@ -136,83 +125,49 @@ def test_add_dict_as_tags_and_values(self):
...
+        exp = ''.join([l.strip() for l in SAMPLEXML.splitlines()])
+        self.assertEqual(obs, exp)
 
-        e._generate_library_descriptor(elm, 'sample', 'libconsprot')
-        exp = ''.join([l.strip() for l in GENLIBDESC.splitlines()])
-        obs = ET.tostring(elm)
+    def test_generate_experiment_xml(self):
+        submission = EBISubmission(2, 'ADD')
+        # removing samples so test text is easier to read
+        keys_to_del = ['1.SKD6.640190', '1.SKM6.640187', '1.SKD9.640182',
```
josenavas: Are you always deleting the same keys? If so, you can either create that list on the setup or actually remove those keys already in the setup

qiita_ware/test/test_ebi.py
```diff
@@ -413,63 +226,129 @@ def test_generate_curl_command(self):
...
-        # The 3rd parameter 1 is the PreprocessedIlluminaParams id, which is
-        # not relevant for this test
+        ppd = PreprocessedData.create(Study(1),
+                                      "preprocessed_sequence_illumina_params",
+                                      1, [(demux_fp, 6)], prep_template)
+
+        return ppd
+
+    def generate_new_prep_template_and_write_demux_files(self):
+        """Creates a new prep template and writes demux test file to avoid
```
josenavas: Minor, but these type of docs should fit in a single line in order to show correctly in the nosetest -v output

qiita_ware/test/test_ebi.py
```diff
@@ -413,63 +226,129 @@ def test_generate_curl_command(self):
...
+        duplication of code"""
+
+        # ignoring warnings generated when adding templates
+        simplefilter("ignore")
```
josenavas: I would prefer that instead of ignoring the warnings, we put a npt.assert_warns in the call. This way, if anything ever changes in the future it will fail, so we know that we are modifying some test case and or code that affects here

qiita_ware/test/test_ebi.py
antgonza: Nop, only in 2 tests.

qiita_ware/test/test_ebi.py
antgonza: I'm not sure about the advantages of doing this vs. simply ignoring based on the tests on this block, could you elaborate?

qiita_ware/test/test_ebi.py
josenavas: The errors in your build is related with this and with the previous comment - use `mkdtemp` and that will go away

qiita_ware/test/test_ebi.py
josenavas: I'm just trying to be safer - if we know that a warning should be raised here, we should check that is raised. If any future change in the code makes that warning to not be raised I'm not sure if that can break some of the functionality here...

antgonza: It will not brake anything so that's why I'm not sure on the benefit.

josenavas: Ok, in that case fine as it is - thanks for double-checking :smile:

josenavas:
Thanks for the titanic effort here @antgonza ! I have done one pass and I think your errors will go away after addressing some of my comments

qiita_ware/ebi.py
antgonza: this is not required, just checked, so will do the same that in others: if exists, add, if not, ignore.

josenavas: Sounds good!

antgonza:
Tests are passing. Ready for a second pass and another reviewe: @adamrp, @ElDeveloper  ??

qiita_db/support_files/populate_test_db.sql
```diff
@@ -156,38 +157,39 @@ CREATE TABLE qiita.sample_1 (
...
+INSERT INTO qiita.sample_1 (sample_id, season_environment, assigned_from_geo, texture, taxon_id, depth, host_taxid, common_name, water_content_soil, elevation, temp, tot_nitro, samp_salinity, altitude, env_biome, country, ph, anonymized_name, tot_org_carb, description_duplicate, env_feature, physical_specimen_location, physical_specimen_remaining, dna_extracted, sample_type, collection_timestamp, host_subject_id, description, latitude, longitude, scientific_name) VALUES
+    ('1.SKM7.640188', 'winter', 'n', '63.1 sand, 17.7 silt, 19.2 clay', '1118232', 0.15, '3483', 'root metagenome', 0.101, 114, 15, 1.3, 7.44, 0, 'ENVO:Temperate grasslands, savannas, and shrubland biome', 'GAZ:United States of America', 6.82, 'SKM7', 3.31, 'Bucu Roots', 'ENVO:plant-associated habitat', 'ANL', TRUE, TRUE, 'ENVO:soil', '2011-11-11 13:00', '1001:B6', 'Cannabis Soil Microbiome', 60.1102854322, 74.7123248382, 'scientific name'),
```
ElDeveloper: Is 'scientific name' a valid `scientific_name`?

qiita_ware/commands.py
```diff
@@ -60,46 +58,35 @@ def submit_EBI(preprocessed_data_id, action, send, fastq_dir_fp=None):
...
-
-    # Get specific output directory and set filepaths
-    get_output_fp = partial(join, output_dir)
-    study_fp = get_output_fp('study.xml')
-    sample_fp = get_output_fp('sample.xml')
-    experiment_fp = get_output_fp('experiment.xml')
-    run_fp = get_output_fp('run.xml')
-    submission_fp = get_output_fp('submission.xml')
-
-    if not isdir(output_dir):
```
ElDeveloper: Just to confirm, this block of code is already replicated somewhere else (in your previous PR IIRC), right?

antgonza: Yes, before we had different folders for different things and now we have 1 single folder.

qiita_ware/ebi.py
```diff
@@ -105,21 +74,36 @@ class EBISubmission(object):
...
     def __init__(self, preprocessed_data_id, action):
         valid_ebi_actions = EBISubmission.valid_ebi_actions
         valid_ebi_submission_states = EBISubmission.valid_ebi_submission_states
+        error_msgs = []
 
-        # Step 1: variable setup and validations
         if action not in valid_ebi_actions:
-            raise ValueError("Not a valid action (%s): %s" % (
-                ', '.join(valid_ebi_actions), action))
+            error_msg = "Not a valid action (%s): %s" % (
```
ElDeveloper: Perhaps worth it to add more information to this message? Maybe something like `"%s is not a valid EBI submission action, valid actions are: %s"`

antgonza: I can modify if you feel strongly about it. Note that all the information is there.

qiita_ware/ebi.py
```diff
@@ -143,33 +128,72 @@ def __init__(self, preprocessed_data_id, action):
...
+        # getting the restrictions
+        st_missing = self.sample_template.check_restrictions(
+            [self.sample_template.columns_restrictions['EBI']])
+        pt_missing = self.prep_template.check_restrictions(
+            [self.prep_template.columns_restrictions['EBI']])
+        # testing if there are any missing columns
+        if st_missing:
+            error_msgs.append(
+                "You are missing some columns in your sample template for "
+                "study #%d to have a valid submission #%d: %s." % (
```
ElDeveloper: Would be nice to know what the second number in this error message stands for. Maybe rephrasing the second part of the sentence would make sense.

qiita_ware/ebi.py
```diff
@@ -143,33 +128,72 @@ def __init__(self, preprocessed_data_id, action):
...
+                    s.id, preprocessed_data_id, ', '.join(list(st_missing))))
+        if pt_missing:
+            error_msgs.append(
+                "You are missing some columns in your prep template for "
+                "study #%d to have a valid submission #%d: %s." % (
```
ElDeveloper: Same as above.

qiita_ware/ebi.py
```diff
@@ -105,21 +74,36 @@ class EBISubmission(object):
     Parameters
     ----------
     preprocessed_data_id : str
+
+    Raises
+    ------
+    EBISumbissionError
+        If either the required columns for EBI and Qiita deploy are not
+        present in the sample and prep templates
```
ElDeveloper: It seems like there's a few missing reasons why this error can be raised, would you mind listing them?

qiita_ware/ebi.py
```diff
@@ -482,53 +398,70 @@ def generate_experiment_xml(self):
...
+                sample_prep.pop('library_construction_protocol')))
+
+            # these are not requiered field but present add them in the right
+            # format
+            for field in self.experiment_library_fields:
+                if field in sample_prep:
+                    element = ET.SubElement(library_descriptor, field.upper())
+                    element.text = sample_prep.pop(field)
+                    if field == 'library_layout':
+                        # no idea what SINGLE means
```
ElDeveloper: :cold_sweat: @adamrp EBI whisperer any ideas? Else should we reach to our sequence archivist friends across the pond? Not blocking of course, not the best to have things like this.

qiita_db/support_files/populate_test_db.sql
antgonza: Excellent question. When I implemented I wasn't sure. Anyway, EBI accepts different taxonomies (5) for this field and only two are restricted to "NCBI TaxID where possible". In other words: it's valid but not perfect.

qiita_ware/ebi.py
```diff
@@ -543,46 +476,31 @@ def generate_run_xml(self):
...
-    def generate_submission_xml(self, action):
+    def generate_submission_xml(self, date_to_hold=date.today()):
         """Generates the submission XML file
 
         Parameters
         ----------
-        action : {'ADD', 'VALIDATE', 'UPDATE'}
-            What action to take when communicating with EBI
+        date_to_hold : date, optional
+            Date to be added to submission. Useful for testing.
```
ElDeveloper: This is more than the date added to the submission, this determines the date until which the submission will be held private at EBI's repository i.e. 365 days after this date. Would you mind adding that information in the other generate_*_xml methods?

antgonza: Didn't now. As you can imagine wasn't in the previous documentation. Will add.

qiita_db/support_files/populate_test_db.sql
josenavas: If you want to be perfect, use "root metagenome" which is the scientific_name correspondent to the taxon_id "1118232" present in this template. Of course not blocking IMO.

qiita_ware/ebi.py
```diff
@@ -591,231 +509,59 @@ def generate_submission_xml(self, action):
...
-        Notes
-        -----
-        xml_gen_fn_arg is needed for generating the submission XML
         """
-        if xml_gen_fn_arg is None:
-            xml_element = xml_gen_fn()
-        else:
-            xml_element = xml_gen_fn(xml_gen_fn_arg)
+        if not isdir(self.xml_dir):
+            makedirs(self.xml_dir)
```
ElDeveloper: Minor, but skbio.util.create_dir would help you avoid the `if` statement.

qiita_ware/test/test_ebi.py
```diff
@@ -413,63 +230,128 @@ def test_generate_curl_command(self):
...
+                    "neither one of the official terms nor one of the "
+                    "user-defined terms in the ENA ontology.\nYou are missing "
+                    "some columns in your prep template for study #1 to have "
+                    "a valid submission #3: platform, primer, "
+                    "experiment_design_description, "
+                    "library_construction_protocol.\nThese samples from study "
+                    "#1, preprocessed data #3 and prep template #2 do not "
+                    "have a valid platform: 1.SKD6.640190, 1.SKM6.640187, "
+                    "1.SKD9.640182")
+        with self.assertRaisesRegexp(EBISumbissionError, exp_text):
```
ElDeveloper: There doesn't seem to be a regexp in the exp_text, shouldn't this be a simple assertRaises with an extra check for the error message to be correct?

antgonza: There is no easy way to do that. The suggested way in the Interwebs is as is implemented.

qiita_ware/test/test_ebi.py
```diff
@@ -498,31 +380,197 @@ def test_generate_demultiplexed_fastq(self):
...
+        <TAG>altitude</TAG><VALUE>0.0</VALUE>
+      </SAMPLE_ATTRIBUTE>
+      <SAMPLE_ATTRIBUTE>
+        <TAG>anonymized_name</TAG><VALUE>SKB2</VALUE>
+      </SAMPLE_ATTRIBUTE>
+      <SAMPLE_ATTRIBUTE>
+        <TAG>assigned_from_geo</TAG><VALUE>n</VALUE>
+      </SAMPLE_ATTRIBUTE>
+      <SAMPLE_ATTRIBUTE>
+        <TAG>collection_timestamp</TAG><VALUE>2011-11-11 13:00:00</VALUE>
```
ElDeveloper: :D https://www.youtube.com/watch?v=VBbY8kQx4wk

qiita_ware/test/test_ebi.py
```diff
@@ -530,7 +578,7 @@ def test_generate_demultiplexed_fastq(self):
       <STUDY_TITLE>
         Identification of the Microbiomes for Cannabis Soils
       </STUDY_TITLE>
-      <STUDY_TYPE existing_study_type="Metagenomics"/>
+      <STUDY_TYPE existing_study_type="Metagenomics" />
```
ElDeveloper: Any idea where this space is coming from?

antgonza: Yes, from the tool doing the format. Note that doesn't add/remove anything.

ElDeveloper:
@antgonza, I think I'm done with my review. Looks good so far, but it would definitely be helpful if @adamrp reviewed this as well, there's a lot going on.

qiita_ware/ebi.py
ElDeveloper: If you wouldn't mind, modifying would be great!

qiita_ware/test/test_ebi.py
ElDeveloper: Ah ok makes sense, perhaps worth documenting that  we are using this as
a convenience and that there's no regex in the message? Side-note I
would have just done with:

with self.assertRaises(ValueError) as e:
  # code that raises the value error
self.assertEqual(str(e.exception), 'Something went wrong')

qiita_ware/test/test_ebi.py
ElDeveloper: You mean the xml library? Why was that not there before?

qiita_ware/test/test_ebi.py
antgonza: Have you tried that? AFAIK it doesn't work as e is what assertRaises returns and not what the actual error is raising. Tricky!!

qiita_ware/test/test_ebi.py
antgonza: Cause before the test was testing the XML (the test was converting test to XML and my guess is that the space was added there) and now we are testing the string that the XML generates.

qiita_ware/test/test_ebi.py
ElDeveloper: Yes I have, there's another instance of this somewhere in the codebase
but I cannot find it, this SO answer however shows what I meant:
http://stackoverflow.com/a/9965090/379593

qiita_ware/test/test_ebi.py
ElDeveloper: Ahhh, I see, got it, thanks for the explanation!

qiita_ware/ebi.py
```diff
@@ -47,39 +48,7 @@ class NoXMLError(Exception):
 def clean_whitespace(s):
     """Standardizes whitespace so that there is only ever one space separating
```
mortonjt: Since this isn't a private function would it be worthwhile to add the Parameters section in the numpydocs? (To clarify that `s` is a `str`)

mortonjt:
I did notice that in `ebi.py` is missing proper doc strings for `send_sequences()` and `send_xml()`.  Do you guys mind if I fix this in a future PR?

qiita_ware/ebi.py
antgonza: Yup.

qiita_ware/test/test_ebi.py
antgonza: Changed it.

antgonza:
@mortonjt thanks for offering but those will be added in step 4 of this branch.

qiita_ware/ebi.py
antgonza: Looking at this ftp://ftp.sra.ebi.ac.uk/meta/xml/checklist.xml there is no requirement for this. Thus, removing.

qiita_ware/ebi.py
antgonza: Well after checking the documentation and trying to do ElementTree.toxml, this method doesn't exist. @josenavas have you used before? Perhaps I'm missing something. Anyway, after further review there is a ElementTree.write method that could allow us to not use minidom but this will mean that we will need to test/build ElementTree objects. At this point I really do not see the benefit of all these changes to not import minidom.

antgonza:
Addressed all issues and questions. Can someone do a final review/merge?

josenavas:
Sure, going over it again!

qiita_db/support_files/populate_test_db.sql
```diff
@@ -156,38 +157,39 @@ CREATE TABLE qiita.sample_1 (
...
+INSERT INTO qiita.sample_1 (sample_id, season_environment, assigned_from_geo, texture, taxon_id, depth, host_taxid, common_name, water_content_soil, elevation, temp, tot_nitro, samp_salinity, altitude, env_biome, country, ph, anonymized_name, tot_org_carb, description_duplicate, env_feature, physical_specimen_location, physical_specimen_remaining, dna_extracted, sample_type, collection_timestamp, host_subject_id, description, latitude, longitude, scientific_name) VALUES
+    ('1.SKM7.640188', 'winter', 'n', '63.1 sand, 17.7 silt, 19.2 clay', '1118232', 0.15, '3483', 'root metagenome', 0.101, 114, 15, 1.3, 7.44, 0, 'ENVO:Temperate grasslands, savannas, and shrubland biome', 'GAZ:United States of America', 6.82, 'SKM7', 3.31, 'Bucu Roots', 'ENVO:plant-associated habitat', 'ANL', TRUE, TRUE, 'ENVO:soil', '2011-11-11 13:00', '1001:B6', 'Cannabis Soil Microbiome', 60.1102854322, 74.7123248382, '1118232'),
```
josenavas: It looks like my previous comment here was a bit confusing, sorry about that. The scientific_name should be 'root metagenome'.

'1118232' is the taxon_id (which is already present in the template).

IMO, not blocking - but would like to hear from @ElDeveloper

ElDeveloper: Not blocking at all.

qiita_ware/ebi.py
```diff
@@ -46,40 +47,20 @@ class NoXMLError(Exception):
...
-
-def iter_file_via_list_of_dicts(input_file):
-    """Iterates over a TSV file, yielding dicts keyed by the column headers
+    tokens
 
     Parameters
     ----------
-    input_file : file
-        The file to iterate over
+    text : str
```
josenavas: `text` -> `s`

qiita_ware/ebi.py
josenavas: I know you didn't modify this, but can you reword this sentence so it fits in a single line in order to follow the numpy standard?

qiita_ware/ebi.py
```diff
@@ -105,21 +86,44 @@ class EBISubmission(object):
     Parameters
     ----------
     preprocessed_data_id : str
+
+    Raises
+    ------
+    EBISumbissionError
+        - If the action is not in EBISubmission.valid_ebi_actions
```
josenavas: I _think_ you don't need to make a list in here to create a list (i.e. remove the `-`)

antgonza: I added them to make them clearer while reading in text format. However, I can remove if you feel strongly about it and _you are sure!_

qiita_ware/ebi.py
```diff
@@ -105,21 +86,44 @@ class EBISubmission(object):
...
     valid_ebi_submission_states = ('submitting', 'success')
+    valid_platforms = ['LS454', 'ILLUMINA', 'UNKNOWN']
+    xmlns_xsi = "http://www.w3.org/2001/XMLSchema-instance"
+    xsi_noNSL = "ftp://ftp.sra.ebi.ac.uk/meta/xsd/sra_1_3/SRA.%s.xsd"
+    experiment_library_fields = [
+        'library_strategy', 'library_source', 'library_selection',
+        'library_layout']
 
     def __init__(self, preprocessed_data_id, action):
         valid_ebi_actions = EBISubmission.valid_ebi_actions
```
josenavas: I think this slip through the cracks in the previous review. Is there any reason why you are assigning `EBISubmission.valid_ebi_actions` and `EBISubmission.valid_ebi_submission_states` to a variable? By them being class attributes, they're easily accessible using `self.valid_ebi_actions` and `self.valid_ebi_submission_states`.

qiita_ware/ebi.py
```diff
@@ -143,33 +148,72 @@ def __init__(self, preprocessed_data_id, action):
...
+            error_msgs.append(
+                "You are missing some columns in your sample template for "
+                "study #%d, preprocessed data #%d. The missing columns: %s."
+                % (s.id, preprocessed_data_id, ', '.join(list(st_missing))))
+        if pt_missing:
+            error_msgs.append(
+                "You are missing some columns in your prep template for "
+                "study #%d, preprocessed data #%d. The missing columns: %s."
+                % (s.id, preprocessed_data_id, ', '.join(list(pt_missing))))
+
```
mortonjt: I did notice that these two error messages are nearly identical.  Do you think it would be a good idea to create a helper function (just to clean this up a little in the future)?

qiita_ware/ebi.py
antgonza: The only reason is to avoid really long/awkward if statements.

qiita_ware/ebi.py
antgonza: I think is fine this way, not sure what the modification will give us ... if you feel strongly about it, could you explain the concrete benefits?

mortonjt: Its just a good coding practice.  Whenever there is redundant code, it tends to be good practice to collapse that code into helper functions - that way if there is a change in these messages, you can just modify the helper function rather than modifying many, many messages.

In this case, there are only two messages that are redundant.  However, if the number of redundant messages grows substantially, then it can be overwhelming to make any changes to these messages.

I don't know how common this error message is, so its completely your call.

antgonza: Not redundant. Note that your explanation makes sense but not sure it makes sense for 1 line ... I guess it depends on what you want: perform an operation that is one line or call a 1 line function with one line.

qiita_ware/ebi.py
```diff
@@ -143,33 +148,72 @@ def __init__(self, preprocessed_data_id, action):
...
-        self.library_selection = lib_selections.get(key, "unspecified")
-
-        # This will hold the submission's samples, keyed by the sample name
+        # getting the restrictions
+        st_missing = self.sample_template.check_restrictions(
+            [self.sample_template.columns_restrictions['EBI']])
+        pt_missing = self.prep_template.check_restrictions(
+            [self.prep_template.columns_restrictions['EBI']])
+        # testing if there are any missing columns
+        if st_missing:
```
josenavas: Should these errors include the prep template id as the error in line 209?

I'm also noticing that these errors have a common pattern: provide the study, prep template and preprocessed data id and then the error. Can this duplication be reduced by modifying this error messages to just include the error (e.g. her will be
```python
error_msgs.append("Missing column in the sample template: %s" % ', '.join(list(st_missin)))
```
and then on line 214 do something like:
```python
if error_msgs:
    error_msgs = "Errors found during EBI submission for study #%d, preprocessed data #%d and prep template #%d:\n" % (s.id, preprocessed_data_id, self.prep_template.id, '\n'.join(error_msgs))
```
I think it will be easy to maintain and the user will receive less duplicated information

antgonza: This makes sense. Making the change.

qiita_ware/ebi.py
```diff
@@ -591,231 +528,57 @@ def generate_submission_xml(self, action):
...
-        xml_gen_fn_arg is needed for generating the submission XML
         """
-        if xml_gen_fn_arg is None:
-            xml_element = xml_gen_fn()
-        else:
-            xml_element = xml_gen_fn(xml_gen_fn_arg)
-
-        xml = minidom.parseString(ET.tostring(xml_element))
+        create_dir(self.xml_dir)
+        xml = minidom.parseString(text)
```
josenavas: it looks like we still do not know why we are parsing the XML here.
I can already foresee a problem with this and I think we already hit it. I think that when an EBI submission failed in the past due to non-ascii (i.e. UTF-8) characters are because the above calls `tostring` have ascii as default (see [here](https://docs.python.org/2/library/xml.etree.elementtree.html?highlight=tostring#xml.etree.ElementTree.tostring)), so if we actually have an UTF-8 character, it will fail in the other functions and will never hit line 578, which is the one that "in theory" is writing the file correctly (i.e in UTF-8).
I think we should remove the parsing here and change the tostring calls in the other functions to include the correct encoding, or the other functions just return the minidom object and the only place where a `tostring` or a `toxml` call happens is in here. Whatever is easier works.

antgonza: Do you think is worth doing those changes here or in another PR? This is already pretty long and we can add to the joy list ...

mortonjt: Big :+1: .  This needs a special test case to make sure that we catch this

qiita_ware/ebi.py
```diff
@@ -970,6 +726,12 @@ def generate_demultiplexed_fastq(self, rewrite_fastq=False):
         already exists and, if it does, the script will assume that in a
         previous execution this step was performed correctly and will simply
         read the file names from self.ebi_dir
+        - When the object is created (init), samples, samples_prep and
```
josenavas: No need of creating a list here (remove `-`)

qiita_ware/test/test_ebi.py
josenavas: Ok!

qiita_ware/ebi.py
josenavas: Will add to the joy list, understand that this PR is getting big and the full work in here is not definitive.

qiita_ware/ebi.py
josenavas: It is not required by numpy doc, but is not explicitly prohibited. Happy to leave as it is unless somebody else feels strong about it.

qiita_ware/ebi.py
josenavas: `self.` only adds 5 letters, I will suggest changing it as it is reducing the number of variables (I had to scroll up to see where the `valid_ebi_actions` was coming from as I expected to see just `self.valid_ebi_actions` - so I needed to check what is different).

These type of variable aliasing makes the code maintenance a bit harder.

josenavas:
thanks @antgonza! few more comments.

I've added a couple of concerns to the joy list, as I think that this PR is getting to big and they can be fixed later and/or the final code will change once your work continues moving forward. In short, the issues that I'm concerned are:
 - UTF-8 char errors (add test)
 - Add a test in which the prep template has less samples than the sample template
 - `EBISubmission.__init__` is getting long and complex, so probably we want to revisit it. I can see this function changing once we move all the steps so I don't want to discuss it here.

antgonza:
Thanks for adding. The other issues (not on the list) have been addressed.

qiita_ware/ebi.py
```diff
@@ -543,46 +485,33 @@ def generate_run_xml(self):
...
                 'quality_scoring_system': 'phred',
                 'checksum_method': 'MD5',
                 'checksum': md5}
             )
 
-        return run_set
+        return ET.tostring(run_set)
 
-    def generate_submission_xml(self, action):
+    def generate_submission_xml(self, date_to_hold=date.today()):
```
adamrp: IMHO, the `date_to_hold` parameter should be renamed. To me, this name implies that the release should be held until today.

Also, this should default to `None`, then in the body of the function it should be set to `date.today()` (or today + 365 days) if it is `None`. Otherwise, this will not behave as you expect it to. As it is now, `date.today()` will be called only once, when the `def` is first executed (so, only once each time Qiita is launched, I assume), and the `date_to_hold` will be the same for all submissions, regardless of the actual submission date.

qiita_ware/ebi.py
adamrp: The reason for this was to get the XML to look pretty -- see [this SO post](http://stackoverflow.com/questions/749796/pretty-printing-xml-in-python). Of course, there are a lot of caveats on that post, so there may be a better way around it. There is the `lxml` package, but it requires another dependency as I recall, since it relies on some libraries that some (most?) systems didn't seem to have by default. [See the installation instructions here](http://lxml.de/installation.html).

qiita_ware/test/test_ebi.py
```diff
@@ -557,116 +605,251 @@ def test_generate_demultiplexed_fastq(self):
...
+EXPERIMENTXML = """
 <EXPERIMENT_SET xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:no\
 NamespaceSchemaLocation="ftp://ftp.sra.ebi.ac.uk/meta/xsd/sra_1_3/SRA.\
 experiment.xsd">
-  <EXPERIMENT alias="%(organization_prefix)s_ppdid_2:test1" center_name=\
-"CCME-COLORADO">
-    <TITLE>%(organization_prefix)s_ppdid_2:test1</TITLE>
-    <STUDY_REF refname="%(organization_prefix)s_ppdid_2"/>
+  <EXPERIMENT alias="%(organization_prefix)s_ppdid_2:1.SKB2.640194" \
+center_name="CCME-COLORADO">
```
adamrp: Is the center_name now in the qiita_config? If so, then the expected text should be changed to use the center_name that is specified there rather than assuming CCME-COLORADO

adamrp: [Yes it is, I think](https://github.com/biocore/qiita/blob/master/qiita_core/support_files/config_test.cfg#L125)

adamrp:
A few comments on the code, but it looks good to me. It's great how much code can be deleted now that we are not relying at all on files and can instead use the data model! :smile:

qiita_ware/ebi.py
```diff
@@ -543,46 +485,37 @@ def generate_run_xml(self):
...
-            The root elelement of the generated ``ElementTree``
+        str
+            string with run XML values
 
-        Raises
-        ------
-        NoXMLError
-            If one of the necessary XML files has not been generated
+        Notes
+        _____
```
adamrp: these should be dashes (`-`) and not underscores (`_`) for numpydoc

qiita_ware/ebi.py
josenavas: Thanks @adamrp That SO post points to the usage of `toprettyxml` which is not what we are using here (we're just using `toxml`). My previous experience of `toprettyxml` is that depending on the version it will slightly change the plain text output by including different spaces and/or newlines. IMO I don't think we should worry about producing pretty XML files and since the only usage of these is to send it through the wire to EBI, the condensed version will be more efficient than the "pretty" one. What do others think?

antgonza: Agree.

antgonza:
Ping! @adamrp @josenavas @mortonjt

qiita_ware/ebi.py
```diff
@@ -105,36 +84,58 @@ class EBISubmission(object):
...
+    EBISumbissionError
+        - If the action is not in EBISubmission.valid_ebi_actions
+        - If the preprocesssed submitted_to_insdc_status is in
+        EBISubmission.valid_ebi_submission_states
+        - If the prep template investigation type is not in the
+        ena_ontology.terms or not in the ena_ontology.user_defined_terms
+        - If the submission is missing required EBI fields either in the sample
+        or prep template
+        - If the sample preparation metadata doesn't have a platform field or
+        it isn't a EB.Submission.valid_platforms
```
josenavas: Really minor but `EB.Submission.valid_platforms` -> `EBISubmission.valid_platforms`

josenavas:
Las commit was modifying only documentation - no reason to wait for the tests :smile: thanks @antgonza !!

PR #1467, opened 2015-09-10
cover some of the issue in the joy list

antgonza:
As discussed yesterday, this PR is solving some of the standing issues in the PR https://github.com/biocore/qiita/pull/1451. This should be easy and quick to review.

### DIFF
--- a/qiita_ware/ebi.py
+++ b/qiita_ware/ebi.py
@@ -7,7 +7,6 @@ from os.path import basename, join, split, isdir, isfile
 from os import environ, close, makedirs, remove, listdir
 from datetime import date, timedelta
 from xml.etree import ElementTree as ET
-from xml.dom import minidom
 from xml.sax.saxutils import escape
 from gzip import GzipFile
 from functools import partial
@@ -272,8 +271,8 @@ class EBISubmission(object):
 
         Returns
         -------
-        str
-            string with study XML values
+        Element
+            Object with study XML values
         """
         study_set = ET.Element('STUDY_SET', {
             'xmlns:xsi': self.xmlns_xsi,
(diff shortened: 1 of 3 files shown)


qiita_ware/ebi.py
```diff
@@ -272,8 +271,8 @@ def generate_study_xml(self):
 
         Returns
         -------
-        str
-            string with study XML values
+        Element
+            Object with study XML values
```
mortonjt: Dumb question.  What exactly is `Element`?  Is it an xml element corresponding to `xml.etree`?  It would be nice to include the object type in the docs

antgonza: I can do that but AFAIK is not standard (however, this is not a standard type). What about xml.etree.ElementTree.Element?

mortonjt: :+1:

mortonjt: Actually, we commonly use `pd.DataFrame` instead of `pandas.DataFrame` or `np.DataFrame` instead of `numpy.DataFrame`.

So just `ET.Element` should be good enough.

josenavas: The reason we use `np` and `pd` instead of `numpy` and `pandas` is because `np` and `pd` are standard abbreviations. Is `ET` a standard abbreviation of `xml.etree.ElementTree`?

antgonza: It seems to be: https://docs.python.org/2/library/xml.etree.elementtree.html

josenavas: Thanks for checking! In such case, :+1: to `ET.Element`

qiita_ware/test/test_ebi.py
```diff
@@ -304,6 +337,13 @@ def test_init_exceptions(self):
             EBISubmission(ppd.id, 'ADD')
         self.assertEqual(exp_text, str(e.exception))
 
+    def test_prep_with_less_samples_than_sample_template(self):
```
josenavas: Uhm... is this test incomplete? At least, I would test that some of the internal values set up during `__init__` is correct. For example, that the `self.samples` actually includes the subset that we want in here.

josenavas:
Thanks @antgonza, done with the review.

qiita_ware/test/test_ebi.py
antgonza: I can test that.

adamrp:
:+1:

PR #1471, opened 2015-09-14
Adding enc file

antgonza:
Decided to close the last 2 PRs cause I think the encryption keys weren't being store correctly. This PR merges the master branch with ebi-improvements and adds the enc file.


antgonza:
Not sure why coveralls is not updating but this is ready to go!

PR #1477, opened 2015-09-17
step 4

antgonza:
This is the new PR for step 4 and 5.
However, now step 4 and 5 only generate the commands that we need to send so they are sent to moi. Note that before they were being submitted directly from these scripts. 

After this the next step will be to (1) add the new-functional encrypted configuration file and (2) add the calls to moi and test in commands.py.

This should be a quick review.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,6 +17,10 @@ before_install:
   - export PATH=/home/travis/miniconda3/bin:$PATH
   # Update conda itself
   - conda update --yes conda
+  # Downloading and setting up ascp for EBI testing
+  - curl -s -O ftp://ftp.microbio.me/pub/qiita/ascp-install-3.5.4.102989-linux-64-qiita.sh
+  - chmod +x ascp-install-3.5.4.102989-linux-64-qiita.sh
+  - ./ascp-install-3.5.4.102989-linux-64-qiita.sh
 install:
   # install a few of the dependencies that pip would otherwise try to install
   # when intalling scikit-bio
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,9 +7,6 @@ env:
     - TEST_ADD_STUDIES=False
     - TEST_ADD_STUDIES=True
 before_install:
-  - openssl aes-256-cbc -K $encrypted_e698cf0e691c_key -iv $encrypted_e698cf0e691c_iv
-    -in qiita_core/support_files/config_test_travis.cfg.enc -out qiita_core/support_files/config_test_travis.cfg
-    -d
   - redis-server --version
   - wget http://repo.continuum.io/miniconda/Miniconda3-3.7.3-Linux-x86_64.sh -O miniconda.sh
   - chmod +x miniconda.sh
@@ -18,9 +15,7 @@ before_install:
   # Update conda itself
   - conda update --yes conda
   # Downloading and setting up ascp for EBI testing
-  - curl -s -O ftp://ftp.microbio.me/pub/qiita/ascp-install-3.5.4.102989-linux-64-qiita.sh
-  - chmod +x ascp-install-3.5.4.102989-linux-64-qiita.sh
-  - ./ascp-install-3.5.4.102989-linux-64-qiita.sh
+  - if [ ${TRAVIS_PULL_REQUEST} == "false" ]; then sh setup_acsp.sh; fi
 install:
   # install a few of the dependencies that pip would otherwise try to install
   # when intalling scikit-bio
@@ -31,8 +26,9 @@ install:
   - pip install sphinx sphinx-bootstrap-theme coveralls ipython[all]==2.4.1
   - travis_retry pip install .
 script:
-  - export QIITA_CONFIG_FP=`pwd`/qiita_core/support_files/config_test_travis.cfg
-  - export MOI_CONFIG_FP=`pwd`/qiita_core/support_files/config_test_travis.cfg
+  - export MOI_CONFIG_FP=`pwd`/qiita_core/support_files/config_test.cfg
+  - if [ ${TRAVIS_PULL_REQUEST} == "false" ]; then export QIITA_CONFIG_FP=`pwd`/qiita_core/support_files/config_test_travis.cfg; fi
+  - if [ ${TRAVIS_PULL_REQUEST} == "false" ]; then export MOI_CONFIG_FP=`pwd`/qiita_core/support_files/config_test_travis.cfg; fi
   - ipython profile create qiita-general --parallel
   - qiita-env start_cluster qiita-general
   - qiita-env make --no-load-ontologies
--- a/qiita_ware/commands.py
+++ b/qiita_ware/commands.py
@@ -79,23 +79,29 @@ def submit_EBI(preprocessed_data_id, action, send, fastq_dir_fp=None):
         old_ascp_pass = environ.get('ASPERA_SCP_PASS', '')
         environ['ASPERA_SCP_PASS'] = qiita_config.ebi_seq_xfer_pass
         try:
-            # seqs_cmds = ebi_submission.generate_send_sequences_cmd()
-            # run commands in moi
-            ebi_submission.generate_send_sequences_cmd()
+            seqs_cmds = ebi_submission.generate_send_sequences_cmd()
+            # send seqs_cmds to moi and get the reply
+            # temporarly setting to seqs_cmds
+            seqs_cmds_moi = seqs_cmds
+        except:
+            LogEntry.create('Fatal', seqs_cmds_moi,
+                            info={'ebi_submission': preprocessed_data_id})
+            ebi_submission.preprocessed_data.update_insdc_status('failed')
         finally:
             environ['ASPERA_SCP_PASS'] = old_ascp_pass
 
         # step 5: sending xml and parsing answer
-        # execute in moi and parse result
-        # xmls_cmds = ebi_submission.generate_curl_command()
-        ebi_submission.generate_curl_command()
-        result = ''
-        LogEntry.create('Runtime', result,
+        xmls_cmds = ebi_submission.generate_curl_command()
+        # send xmls_cmds to moi and get the reply
+        # temporarly setting to xmls_cmds
+        xmls_cmds_moi = xmls_cmds
+        LogEntry.create('Runtime', xmls_cmds_moi,
                         info={'ebi_submission': preprocessed_data_id})
-        study_acc, submission_acc = ebi_submission.parse_EBI_reply()
+        study_acc, submission_acc = ebi_submission.parse_EBI_reply(
+            xmls_cmds_moi)
 
         if study_acc is None or submission_acc is None:
-            LogEntry.create('Fatal', result,
+            LogEntry.create('Fatal', xmls_cmds_moi,
                             info={'ebi_submission': preprocessed_data_id})
             ebi_submission.preprocessed_data.update_insdc_status('failed')
             raise ComputeError("EBI Submission failed!")
--- a/qiita_ware/commands.py
+++ b/qiita_ware/commands.py
@@ -20,6 +20,7 @@ from qiita_core.qiita_settings import qiita_config
 from qiita_ware.ebi import EBISubmission
 from qiita_ware.exceptions import ComputeError
 from traceback import format_exc
+from os import environ
 
 
 def submit_EBI(preprocessed_data_id, action, send, fastq_dir_fp=None):
@@ -70,22 +71,41 @@ def submit_EBI(preprocessed_data_id, action, send, fastq_dir_fp=None):
     ebi_submission.write_xml_file(ebi_submission.generate_submission_xml(),
                                   ebi_submission.submission_xml_fp)
 
-    # other steps
     if send:
-        ebi_submission.send_sequences()
-        study_accession, submission_accession = ebi_submission.send_xml()
+        # step 4: sending sequences
 
-        if study_accession is None or submission_accession is None:
+        # Set the ASCP password to the one in the Qiita config, but remember
+        # the old pass so that we can politely reset it
+        old_ascp_pass = environ.get('ASPERA_SCP_PASS', '')
+        environ['ASPERA_SCP_PASS'] = qiita_config.ebi_seq_xfer_pass
+        try:
+            # seqs_cmds = ebi_submission.generate_send_sequences_cmd()
+            # run commands in moi
+            ebi_submission.generate_send_sequences_cmd()
+        finally:
+            environ['ASPERA_SCP_PASS'] = old_ascp_pass
+
+        # step 5: sending xml and parsing answer
+        # execute in moi and parse result
+        # xmls_cmds = ebi_submission.generate_curl_command()
+        ebi_submission.generate_curl_command()
+        result = ''
+        LogEntry.create('Runtime', result,
+                        info={'ebi_submission': preprocessed_data_id})
+        study_acc, submission_acc = ebi_submission.parse_EBI_reply()
+
+        if study_acc is None or submission_acc is None:
+            LogEntry.create('Fatal', result,
+                            info={'ebi_submission': preprocessed_data_id})
             ebi_submission.preprocessed_data.update_insdc_status('failed')
-
             raise ComputeError("EBI Submission failed!")
         else:
             ebi_submission.preprocessed_data.update_insdc_status(
-                'success', study_accession, submission_accession)
+                'success', study_acc, submission_acc)
     else:
-        study_accession, submission_accession = None, None
+        study_acc, submission_acc = None, None
 
-    return study_accession, submission_accession
+    return study_acc, submission_acc
 
 
 def submit_VAMPS(preprocessed_data_id):
--- a/qiita_ware/commands.py
+++ b/qiita_ware/commands.py
@@ -77,27 +77,38 @@ def submit_EBI(preprocessed_data_id, action, send, fastq_dir_fp=None):
         environ['ASPERA_SCP_PASS'] = qiita_config.ebi_seq_xfer_pass
         seqs_cmds = ebi_submission.generate_send_sequences_cmd()
         LogEntry.create('Runtime',
-                        'ASPERA_SCP_PASS changed. Going to try to submit seqs',
-                        info={'ebi_submission': preprocessed_data_id})
+                        ("Submitting sequences for pre_processed_id: "
+                         "%d" % preprocessed_data_id))
         try:
-            # place holder for moi submission, see PR notes in #1477
+            # place holder for moi call see #1477
             pass
         except:
             LogEntry.create('Fatal', seqs_cmds,
                             info={'ebi_submission': preprocessed_data_id})
+        else:
+            LogEntry.create('Runtime',
+                            ('Submission of sequences of pre_processed_id: '
+                             '%d completed successfully' %
+                             preprocessed_data_id))
         environ['ASPERA_SCP_PASS'] = old_ascp_pass
 
         # step 5: sending xml and parsing answer
         xmls_cmds = ebi_submission.generate_curl_command()
-        LogEntry.create('Runtime', 'Submitting xml files',
-                        info={'ebi_submission': preprocessed_data_id})
+        LogEntry.create('Runtime',
+                        ("Submitting XMLs for pre_processed_id: "
+                         "%d" % preprocessed_data_id))
         try:
-            # place holder for moi submission, see PR notes in #1477
+            # place holder for moi call see #1477
             xmls_cmds_moi = xmls_cmds
-            pass
         except:
+            # handle exception
             LogEntry.create('Fatal', seqs_cmds,
                             info={'ebi_submission': preprocessed_data_id})
+        else:
+            LogEntry.create('Runtime',
+                            ('Submission of sequences of pre_processed_id: '
+                             '%d completed successfully' %
+                             preprocessed_data_id))
 
         study_acc, submission_acc = ebi_submission.parse_EBI_reply(
             xmls_cmds_moi)
--- a/qiita_ware/commands.py
+++ b/qiita_ware/commands.py
@@ -73,30 +73,32 @@ def submit_EBI(preprocessed_data_id, action, send, fastq_dir_fp=None):
 
     if send:
         # step 4: sending sequences
-
-        # Set the ASCP password to the one in the Qiita config, but remember
-        # the old pass so that we can politely reset it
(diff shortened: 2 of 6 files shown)


qiita_ware/commands.py
```diff
@@ -70,22 +71,41 @@ def submit_EBI(preprocessed_data_id, action, send, fastq_dir_fp=None):
...
 
-    # other steps
     if send:
-        ebi_submission.send_sequences()
-        study_accession, submission_accession = ebi_submission.send_xml()
+        # step 4: sending sequences
+
+        # Set the ASCP password to the one in the Qiita config, but remember
+        # the old pass so that we can politely reset it
+        old_ascp_pass = environ.get('ASPERA_SCP_PASS', '')
```
josenavas: Why do you have to change this environment variable? I'm not seeing anything in `ebi_submission.generate_send_sequences_cmd()` that needs that environment variable.

qiita_ware/commands.py
```diff
@@ -70,22 +71,41 @@ def submit_EBI(preprocessed_data_id, action, send, fastq_dir_fp=None):
...
+            ebi_submission.generate_send_sequences_cmd()
+        finally:
+            environ['ASPERA_SCP_PASS'] = old_ascp_pass
+
+        # step 5: sending xml and parsing answer
+        # execute in moi and parse result
+        # xmls_cmds = ebi_submission.generate_curl_command()
+        ebi_submission.generate_curl_command()
+        result = ''
+        LogEntry.create('Runtime', result,
```
josenavas: Why is this log entry created? It will only add an empty string so I don't think this is useful...

qiita_ware/commands.py
```diff
@@ -70,22 +71,41 @@ def submit_EBI(preprocessed_data_id, action, send, fastq_dir_fp=None):
...
+                        info={'ebi_submission': preprocessed_data_id})
+        study_acc, submission_acc = ebi_submission.parse_EBI_reply()
 
-        if study_accession is None or submission_accession is None:
+        if study_acc is None or submission_acc is None:
+            LogEntry.create('Fatal', result,
```
josenavas: Same here. Result is an empty string so a bit more information will be useful

qiita_ware/commands.py
```diff
@@ -70,22 +71,41 @@ def submit_EBI(preprocessed_data_id, action, send, fastq_dir_fp=None):
...
+                            info={'ebi_submission': preprocessed_data_id})
             ebi_submission.preprocessed_data.update_insdc_status('failed')
-
             raise ComputeError("EBI Submission failed!")
```
josenavas: I know that you didn't modify this, but a bit more information in this `ComputeError` will be more useful. This type of error messages don't help to debug.

qiita_ware/ebi.py
```diff
@@ -674,25 +644,7 @@ def send_xml(self):
...
-                                            "above."))
-                print ("However, the accession numbers could not be found in "
-                       "the output above.")
-        else:
-            LogEntry.create('Fatal', curl_result)
-            print curl_result
-            print "FAILED"
-
-        return (study_accession, submission_accession)
+        return study_accession, submission_accession
```
josenavas: I think this function should be the one that raises the error if the ebi submission failes (i.e. `success="false"`).
This function has the result so we can actually do some intelligent parsing to raise different type of errors. I know that right now we might not now those different types of errors, but one think that we can do here is raise an error an include the curl_result in the error, so we know what went wrong and we can debug way easier. That error should be both in a LogEntry and in a raised exception (either a new type for EBI or some of the existing ones).

qiita_ware/test/test_ebi.py
```diff
@@ -398,6 +420,26 @@ def test_generate_demultiplexed_fastq(self):
...
+        e.generate_demultiplexed_fastq()
+        self.files_to_remove.append(e.ebi_dir)
+        e.write_xml_file(e.generate_study_xml(), e.study_xml_fp)
+        e.write_xml_file(e.generate_sample_xml(), e.sample_xml_fp)
+        e.write_xml_file(e.generate_experiment_xml(), e.experiment_xml_fp)
+        e.write_xml_file(e.generate_run_xml(), e.run_xml_fp)
+        e.write_xml_file(e.generate_submission_xml(), e.submission_xml_fp)
+        obs = e.generate_send_sequences_cmd()
+
+        exp = ('ascp --ignore-host-key -L- -d -QT -k2 '
```
josenavas: I don't think this is a safe tests. It looks like it is only duplicating the code from the actual function so you're testing the function with the same code - if there is a bug in one of them it will be a bug in both...

josenavas:
Few comments. Also tests are failing...

qiita_ware/commands.py
wasade: Agree, perhaps mentioning the study id + prep id prior to issuing the submission, followed by a message indicating study id + prep id + returned accession details (or error if appropriate). This then gives timing information for free.

qiita_ware/ebi.py
```diff
@@ -616,57 +608,35 @@ def generate_curl_command(
...
+        Notes
+        -----
+        - All 5 XML files (study, sample, experiment, run, and submission) must
+          be generated before executing this function
+        """
+        fastqs = [sfp for _, sfp in viewitems(self.sample_demux_fps)]
+
+        ascp_command = ('ascp --ignore-host-key -L- -d -QT -k2 '
+                        '{0} {1}@{2}:./{3}/'.format(
+                            ' '.join(fastqs), qiita_config.ebi_seq_xfer_user,
```
wasade: This will cause issues if submitting a very large number of fastqs due to the command length, but I don't know how best to work around it unless EBI would accept multiple `ascp` calls.

qiita_ware/ebi.py
```diff
@@ -616,57 +608,35 @@ def generate_curl_command(
...
+        return ascp_command
 
+    def parse_EBI_reply(self, curl_result):
+        """Parse and verify reply from EBI after sending XML files
+
+        Parameters
+        ----------
+        curl_result : str
+            The reply sent by EBI after sending XML files
+        """
```
mortonjt: Don't forget to include `Returns` in the numpydoc :)

qiita_ware/ebi.py
```diff
@@ -616,57 +608,35 @@ def generate_curl_command(
...
-        curl_command_parts = shsplit(curl_command)
-        temp_fd, temp_fp = mkstemp()
-        call(curl_command_parts, stdout=temp_fd)
-        close(temp_fd)
-
-        with open(temp_fp, 'U') as curl_output_f:
-            curl_result = curl_output_f.read()
+    def generate_send_sequences_cmd(self):
+        """Generate the sequences to EBI via ascp command
+
```
mortonjt: Same here.

qiita_ware/ebi.py
wasade: Are you sure though? It complicates the method to be both a parser and to care about errors. I'd vote to keep the parser as simple as necessary. For instance, `parse_fasta` doesn't care if the sequences have random characters as the validation and handling of the parsed records is at the responsibility of the consumer (or another method call issued by the consumer).

qiita_ware/test/test_ebi.py
```diff
@@ -243,6 +244,27 @@ def test_generate_curl_command(self):
                '%%20ebi_access_key%%3D"') % {'xml_dir': submission.xml_dir}
         self.assertEqual(obs, exp)
 
+    def test_parse_EBI_reply(self):
+        e = EBISubmission(2, 'ADD')
+        curl_result = ""
```
wasade: `curl_result = ""` is conceptually very different from `curl_result = 'success="true"'` yet the returns are the same. Should the possible states or detail from `parse_EBI_reply` be expanded?

qiita_ware/test/test_ebi.py
wasade: Agree. It also isn't clear to me where the `*_xml_fp` files factor in here?

qiita_ware/commands.py
antgonza: Because aspera (ascp) expects the password to be an environment variable. Nothing we can do about this.

qiita_ware/commands.py
antgonza: Remember that commands.py is a placeholder for the full pipeline and that first we want to have all the methods working fine so we can add to moi and hopefully test a full submission on the fly. In other words, these lines are gonna change in the next PR but just wanted to give an idea of how it is gonna look.

qiita_ware/commands.py
antgonza: Same than before.

qiita_ware/ebi.py
antgonza: Good point and we haven't hit that problem yet, which is strange. I could submit in batches of 10 files, what do you think?

qiita_ware/ebi.py

qiita_ware/ebi.py
antgonza: Agree with @wasade the best will be to raise the errors in the accessions are None in the commands.py method/run.

qiita_ware/test/test_ebi.py
antgonza: Not sure is needed. The parser will return None in both cases, which are errors. Then commands.py should raise the error with the return of the command.

qiita_ware/test/test_ebi.py
antgonza: Well, I'm not sure how else to test. The method basically builds a command based on the available/successful fastq.gz files. In other words, what do you suggest testing? The expansion? Well, that's tested somewhere else where we validate that only good sequence files are generated.

antgonza:
The error is due: http://docs.travis-ci.com/user/pull-requests/#Security-Restrictions-when-testing-Pull-Requests. Nothing to do with the code. The plan was to always test EBI submissions but this is not possible via encrypted files. Any other ideas? cc: @ElDeveloper

qiita_ware/ebi.py
```diff
@@ -616,57 +608,47 @@ def generate_curl_command(
...
+        ascp_command = ('ascp --ignore-host-key -L- -d -QT -k2 '
+                        '{0} {1}@{2}:./{3}/'.format(
+                            ' '.join(fastqs), qiita_config.ebi_seq_xfer_user,
+                            qiita_config.ebi_seq_xfer_url, self.ebi_dir))
+
+        return ascp_command
 
+    def parse_EBI_reply(self, curl_result):
+        """Parse and verify reply from EBI after sending XML files
+
```
mortonjt: I hate to say this.

But should the documentation for `curl_result` also be included in `Parameters`?

qiita_ware/commands.py
josenavas: Yup but `generate_send_sequences_cmd` is only generating the command, not executing it; so the actual change in the environment should be done when ascp is called.

qiita_ware/commands.py
josenavas: Sounds good

qiita_ware/ebi.py
josenavas: I think that is feasible iif EBI actually accepts multiple `ascp` calls as @wasade points out

qiita_ware/ebi.py
josenavas: Sounds good, but then we need to make sure that the actual `curl_result` is also stored in the error

qiita_ware/test/test_ebi.py
josenavas: Here you should know which are the successful files (since you're able to test that elsewhere). The only reason for the existence of a test for such function is to test that the formatting is correct - but if you're executing the same formatting here then this is not testing anything...

qiita_ware/ebi.py
josenavas: I think the problem is that @antgonza had the sections in the wrong order (first the returns and then the parameters). @antgonza can you re-order?

qiita_ware/commands.py
antgonza: Which is gonna be in commands.py -- next PR, right here.

qiita_ware/ebi.py
antgonza: Agree, those lines are already in commands.py.

qiita_ware/commands.py
josenavas: Ok, I guess we will need to be extremely careful once this branch is ready to be merged against master, since there is a lot of stuff that gets changed later and is getting harder and harder to review.

IMO we should remove those lines from here, as they will never be in here.

qiita_ware/ebi.py
josenavas: That is not correct. Currently it is only storing an empty string. Just make sure to not forget to add later.

ElDeveloper:
> The error is due: http://docs.travis-ci.com/user/pull-requests/#Security-Restrictions-when-testing-Pull-Requests. Nothing to do with the code. The plan was to always test EBI submissions but this is not possible via encrypted files. Any other ideas? cc: @ElDeveloper

:scream:, thanks for the link, the explanation makes a lot of sense. What about the following, the specific tests that need the encrypted information don't get run on PRs (regrettable, I know), but they do get run once they are merged to master which if I understand is possible. The downside to this is that we wouldn't have a full CI system for PRs but it would at the very least alert us once the code is merged (hopefully before it is deployed).

antgonza:
Yes, not ideal!!! I'm gonna comment out the testing lines in .travis.yml while we figure out a way to do this.

qiita_ware/commands.py
josenavas: I'm still strong here. The lines modifying the environment should be surrounding when the call to the command is done, not when generating the command. Does this make sense? If not, it might be useful to have a call later today so I can explain better the concern in here.

antgonza: Will moving those lines within the try will make you feel better? I can do that ... Just remember that commands.py is the last thing that is gonna change

josenavas: If you're already planning removing these lines in a subsequent PR is ok, but just want to point out really well that those lines of code here are not doing anything useful.

antgonza: Just for peace of mind: remember that we have a list of things that we need to make sure we have in the ebi branch. Please add this comment if you feel like you are gonna forget ...

josenavas: Thanks for the remainder. I've added it to the joy list.

wasade: Agree with @josenavas. Reverting the environment on like 91 means the change on line 80 actually has zero effect since the `finally` block is always executed, and is executed prior to the system call. 

Note, moving the `environ` calls to within the `try` change this, and there isn't a specific need as the interaction is safe via the use of `.get`

.travis.yml
```diff
@@ -27,8 +26,9 @@ install:
   - pip install sphinx sphinx-bootstrap-theme coveralls ipython[all]==2.4.1
   - travis_retry pip install .
 script:
-  - export QIITA_CONFIG_FP=`pwd`/qiita_core/support_files/config_test_travis.cfg
-  - export MOI_CONFIG_FP=`pwd`/qiita_core/support_files/config_test_travis.cfg
+  - export MOI_CONFIG_FP=`pwd`/qiita_core/support_files/config_test.cfg
+  - if [ ${TRAVIS_PULL_REQUEST} == "false" ]; then export QIITA_CONFIG_FP=`pwd`/qiita_core/support_files/config_test_travis.cfg; fi
```
josenavas: These 2 if statement are equal. Can you just simple add both `export` commands in the same if clause?

qiita_ware/commands.py
```diff
@@ -70,22 +71,47 @@ def submit_EBI(preprocessed_data_id, action, send, fastq_dir_fp=None):
...
-
-        if study_accession is None or submission_accession is None:
+        # step 4: sending sequences
+
+        # Set the ASCP password to the one in the Qiita config, but remember
+        # the old pass so that we can politely reset it
+        old_ascp_pass = environ.get('ASPERA_SCP_PASS', '')
+        environ['ASPERA_SCP_PASS'] = qiita_config.ebi_seq_xfer_pass
+        try:
+            seqs_cmds = ebi_submission.generate_send_sequences_cmd()
```
josenavas: It looks like `seqs_cmd` is only used to assign the `seqs_cmds_moi`. Probably just better to call it `seqs_cmds_moi` and we avoid one variable aliasing?

qiita_ware/commands.py
```diff
@@ -70,22 +71,47 @@ def submit_EBI(preprocessed_data_id, action, send, fastq_dir_fp=None):
...
+                            info={'ebi_submission': preprocessed_data_id})
             ebi_submission.preprocessed_data.update_insdc_status('failed')
+        finally:
+            environ['ASPERA_SCP_PASS'] = old_ascp_pass
+
+        # step 5: sending xml and parsing answer
+        xmls_cmds = ebi_submission.generate_curl_command()
+        # send xmls_cmds to moi and get the reply
+        # temporarly setting to xmls_cmds
+        xmls_cmds_moi = xmls_cmds
```
josenavas: Same here

qiita_ware/commands.py
wasade: Oh! Sorry, does `generate_send_sequences_cmd` issue the system call? Didn't realize that. If so, then I do still agree that it would be more explicit to move the environment manipulation to as close to the system call as possible.

qiita_ware/commands.py
wasade: I _think_ these are just for logging?

qiita_ware/commands.py
antgonza: Nop it doesn't, just generates the commands. Previously the command was being executed there, which we can do but now I want to add it to moi, which will make more sense.

qiita_ware/commands.py
antgonza: Yes, just for logging.

qiita_ware/test/test_ebi.py
```diff
@@ -239,10 +239,31 @@ def test_generate_curl_command(self):
                '-F "SAMPLE=@%(xml_dir)s/sample.xml" '
                '-F "RUN=@%(xml_dir)s/run.xml" '
                '-F "EXPERIMENT=@%(xml_dir)s/experiment.xml" '
-               '"ebi_dropbox_url/?auth=ERA%%20ebi_seq_xfer_user'
-               '%%20ebi_access_key%%3D"') % {'xml_dir': submission.xml_dir}
+               '"ebi_dropbox_url/?auth=ENA%%20ebi_seq_xfer_user'
+               '%%20ebi_access_key"') % {'xml_dir': submission.xml_dir}
         self.assertEqual(obs, exp)
 
+    def test_parse_EBI_reply(self):
```
josenavas: Just to make it clear. This function is going to be refactored once we have the DB changes to parse all the EBI values, correct?

qiita_ware/commands.py
wasade: Okay, but this PR doesn't include the actual moi call...?

qiita_ware/test/test_ebi.py
antgonza: That's the plan ...

qiita_ware/commands.py
antgonza: That's correct. Basically I'm trying to make it as modular as possible (small PRs) as things change as I start figuring out what the previous version did and if they are wrong fixing them. 

BTW, from the original PR submission:
``After this the next step will be to (1) add the new-functional encrypted configuration file and (2) add the calls to moi and test in commands.py.```

qiita_ware/test/test_ebi.py
wasade: Should the parser be modified? Rather, is there additional information or states that we may want to track?

josenavas:
Just few comments. I'm aware that this is merging in a WIP branch, so once of the travis comments are fixed, :+1: as it looks like that most of the comments that I made are going to go away in a subsequent PR.

Anyways, at the end, I'm planning to do a comprehensive review of the bigger WIP branch before it gets merged to master, to check that all the pieces came up together sanely.

Thanks for working into this @antgonza, I'm aware on how hard is working with this code.

.travis.yml
antgonza: I could do this but it will make a really long single line, which I'm not a fan of.

See: https://groups.google.com/forum/#!topic/travis-ci/uaAP9zEdiCg

qiita_ware/test/test_ebi.py
josenavas: Yes, but we still don't have the DB to support such data - so I think it makes sense to leave as it is in this PR and modify once we modify the DB (PR in progress currently).

.travis.yml
josenavas: Ugh, that's unfortunate. :+1: on leaving as it is

qiita_ware/commands.py
wasade: Okay. In that case, if it isn't possible (or expected) that `generate_send_sequences_cmd()` will except, I recommend:

```python
environ['ASPERA_SCP_PASS'] = qiita_config.ebi_seq_xfer_pass
seqs_cmds = ebi_submission.generate_send_sequences_cmd()
try:
    # place holder for moi submission, see PR notes in #1477
    pass
except:
    LogEntry.create('Fatal', seqs_cmds, info={'ebi_submission': preprocessed_data_id})
environ['ASPERA_SCP_PASS'] = old_ascp_pass
```

This limits the `try` to the single in which we anticipate an exception happening.

wasade: ...actually, it would be great to have an informative log message prior to entering the `try`

antgonza: Informative message? Like what?

wasade:
```python
LogEntry.create('Runtime', "Submitting sequences for pre_processed_id: %d" % preprocessed_id)`
try:
    # place holder for moi call see #1477
    pass 
except:
    # handle exception
    LogEntry.create('Fatal', ...
else:
    LogEntry.create('Runtime', 'Submission of sequences of pre_processed_id: %d completed successfully' % preprocessed_id)
...
```

This allows us to get timing information for free, and to verify that events actually happened. In principle, logging all events that happen is useful particularly as it isn't clear what will and will not be useful in the future. Take, for example, the default verbosity of virtually any server like Apache, nginx, Tornado, vsftp, dhcpd, Torque, slurm, postfix, sendmail, etc.

josenavas:
:+1: I'm gonna wait for the tests to pass and then merge - I need these changes for my PR (https://github.com/biocore/qiita/pull/1481) as the EBI tests are the ones that are currently failing and I want to minimize the merge conflicts.

josenavas:
Thanks @antgonza

Issue #1051, opened 2015-04-08
Fix --insdc option in qiita load_preprocessed 


ElDeveloper:
Currently it is a boolean option, it should really be a "choice" type, in general we need to:
- [ ] Set a patch to correct existing entries as `true` to `submitted` (where should we get the accession numbers from)?
- [ ] Change the option type so we dynamically retrieve the available options from the DB.
- [ ] Rename the option to reflect the type.

adamrp:
The option is "submitted_to_insdc_status", which, as @wasade pointed out on the call today, implies that the variable should take boolean values. So to add a comment to this issue, the option should probably be renamed to something like "insdc_submission_status" to avoid confusion.

antgonza:
CLI will be minimized: https://github.com/biocore/qiita/issues/1999, closing.

PR #1481, opened 2015-09-23
WIP: DB changes for EBI

josenavas:
This PR modifies the DB to store all the information that EBI returns when we are doing a submission.
Storing all this information (in the correct place) will help us to do a better usage of the EBI structure (which actually fits pretty nicely with our structure) as well as being able to do EBI updates in the future.

Tests are currently failing, and I'm fixing them right now.


antgonza:
What about transferring the error status from submitted_to_insdc_status? Currently, you are simply ignoring those.

josenavas:
Is it useful? In the current deployed implementation it only reads `failed`. Do we want to keep track of those too, or just set them up as not submitted?

antgonza:
Good point, we are not storing any useful information and the reason we need these changes ...

josenavas:
@antgonza this PR is ready - there are 2 more things that need to be done but I prefer to do them in another PR to avoid too many changes in this single PR:
- Modify EBI code to use previous accession numbers if they exist
- Fix qiita pet to correctly show the EBI information
I added them to the joy list and I put my name next to them as I already started working in the first one (and we agreed that I will do the second one).

I think this is the best path to move forward, so you can keep working on the next steps in the EBI refactoring. Sounds good?

qiita_db/support_files/patches/32.sql
```diff
@@ -0,0 +1,40 @@
...
+
+ALTER TABLE qiita.prep_template_sample ADD ebi_experiment_accession varchar  ;
+ALTER TABLE qiita.study ADD ebi_study_accession varchar  ;
+ALTER TABLE qiita.study ADD ebi_submission_status varchar  NOT NULL DEFAULT 'not submitted';
+ALTER TABLE qiita.study_sample ADD ebi_sample_accession varchar  ;
+ALTER TABLE qiita.study_sample ADD biosample_accession varchar  ;
+
+CREATE TABLE qiita.ebi_run_accession (
+	sample_id            varchar  	NOT NULL,
+	preprocessed_data_id bigint		NOT NULL,
```
antgonza: Extra tab.

qiita_db/test/test_data.py
```diff
@@ -504,10 +430,10 @@ def test_delete_advanced(self):
             PreprocessedData.delete(ppd.id)
         ppd.update_vamps_status('failed')
 
-        # fails due to EBI submission
-        ppd.update_insdc_status('success', 'AAAA', 'AAAA')
-        with self.assertRaises(QiitaDBStatusError):
-            PreprocessedData.delete(ppd.id)
+        # TODO: fails due to EBI submission
```
antgonza: What are these lines?

qiita_ware/ebi.py
```diff
@@ -646,23 +670,65 @@ def parse_EBI_reply(self, curl_result):
...
-        if 'success="true"' in curl_result:
-            accessions = search('<STUDY accession="(?P<study>.+?)".*?'
-                                '<SUBMISSION accession="(?P<submission>.+?)"',
-                                curl_result)
-            if accessions is not None:
-                study_accession = accessions.group('study')
-                submission_accession = accessions.group('submission')
-
-        return study_accession, submission_accession
+        try:
```
antgonza: There is no difference between this try/except and the next if. I _think_ we should have a way to differentiate between these two errors.

qiita_db/metadata_template/base_metadata_template.py
```diff
@@ -1325,3 +1326,78 @@ def check_restrictions(self, restrictions):
...
+            sql = """SELECT sample_id, {0}
+                     FROM qiita.{1}
+                     WHERE {2}=%s""".format(column, self._table,
+                                            self._id_column)
+            TRN.add(sql, [self.id])
+            result = {sample_id: accession
+                      for sample_id, accession in TRN.execute_fetchindex()}
+        return result
+
+    def _update_accession_numbers(self, column, values):
```
antgonza: Does this function have tests? I couldn't clearly find them ...

qiita_db/metadata_template/test/test_sample_template.py
```diff
@@ -1026,55 +977,31 @@ def test_create_int_prefix(self):
...
-        self.assertEqual(obs, exp)
+                'scientific_name': 'homo sapiens'}}
+        for s_id in exp_sample_ids:
+            self.assertEqual(st[s_id]._to_dict(), exp_dict[s_id])
 
     def test_create_str_prefixes(self):
         """Creates a new SampleTemplate with sample names string prefixed"""
-        new_id = self.new_study.id
         st = SampleTemplate.create(self.metadata_str_prefix, self.new_study)
+        new_id = self.new_study.id
```
antgonza: Out of curiosity, what's the reason of this change?

antgonza:
Thanks for working on this. It looks good but some failures, conflicts and suggestions/questions. Suggest working on this before someone else reviews.

Also, which PR should we merge first (as I'm sure we will have conflicts), this or https://github.com/biocore/qiita/pull/1486? I think #1486 as this one already have conflicts and might be easier to fix all at the same time ... 

Good plan moving forward, I can work on making Qiita having EBI tests and the rest of the joy list.

qiita_db/metadata_template/base_metadata_template.py
josenavas: This function is tested through the setters for the ebi accession numbers: `SampleTemplate.ebi_sample_accession`, `SampleTemplate.biosample_accession` and `PrepTemplate.ebi_experiment_accession`. Those 3 functions are calling this one directly and this function cannot be used form the base class (because it cannot be instantiated).

antgonza: OK, thanks!

qiita_db/metadata_template/test/test_sample_template.py
josenavas: The previous test failed, because now in the `qiita.study_sample` table we are storing more information, but just the link between the study and the samples. This made the line 1045 (old code) to fail. Following @wasade's proposal of making the tests independent of the storage system, I change the tests to be using the objects API. This way, if in the future we want to change the storage technology, we will not need to modify the test.

However, I just noticed that I'm missing here the 2 new properties. Will add.

antgonza: OK, thanks!

qiita_db/support_files/patches/32.sql
josenavas: Weird, it was aligned correctly on my machine - fixed.

qiita_db/test/test_data.py
josenavas: A TODO that I forgot to do :flushed:

qiita_ware/ebi.py
josenavas: Now they rise errors with different messages.

josenavas:
Comments/merge conflicts addressed @antgonza 

The test failures are in Qiita pet, so I would prefer to let them fail and solve them in the next PR in which I'm solving the interface part of it. Does this sound good?

qiita_db/data.py
```diff
@@ -783,16 +772,15 @@ def delete(cls, ppd_id):
             if ppd.status != 'sandbox':
                 raise QiitaDBStatusError(
                     "Illegal operation on non sandboxed preprocessed data")
+            elif ppd.is_submitted_to_ebi:
```
antgonza: As far as I can tell not all these cases are tested, right? Just want to confirm ...

antgonza:
One question. If all that block is tested :+1:

qiita_db/data.py
josenavas: This case already existed before (see line 791 in the old code) and it was already tested (they're the lines that I have commented out in the test - so it is tested here: https://github.com/biocore/qiita/pull/1481/files#diff-e54a9d2a89750a099959730f348c06e1R432)

qiita_db/metadata_template/base_metadata_template.py
```diff
@@ -1325,3 +1326,78 @@ def check_restrictions(self, restrictions):
                 for col in restriction.columns}
 
         return cols.difference(self.categories())
+
+    def _get_accession_numbers(self, column):
+        """Return the accession numbers stored under the given column
```
ElDeveloper: under the given column -> in \`column\` 

As it stands it isn't incorrect, but it reads kinda funny.

qiita_db/metadata_template/base_metadata_template.py
```diff
@@ -1325,3 +1326,78 @@ def check_restrictions(self, restrictions):
...
+                     FROM qiita.{1}
+                     WHERE {2}=%s""".format(column, self._table,
+                                            self._id_column)
+            TRN.add(sql, [self.id])
+            result = {sample_id: accession
+                      for sample_id, accession in TRN.execute_fetchindex()}
+        return result
+
+    def _update_accession_numbers(self, column, values):
+        """Update the accession numbers stored under the given column
```
ElDeveloper: I think the clarity of this docstring would be benefited by making use of backticks to indicate variable names.

qiita_db/metadata_template/base_metadata_template.py
```diff
@@ -1325,3 +1326,78 @@ def check_restrictions(self, restrictions):
...
+                    "%s" % ', '.join(diff))
+
+            # Remove the common samples form the values dictionary
+            values = deepcopy(values)
+            for sample in common_samples:
+                del values[sample]
+
+            if values:
+                sql_vals = ', '.join(
+                    ["('%s', '%s')" % (k, v) for k, v in viewitems(values)])
```
ElDeveloper: Shouldn't these values be formatted by psycopg?

qiita_db/metadata_template/base_metadata_template.py
```diff
@@ -1325,3 +1326,78 @@ def check_restrictions(self, restrictions):
...
+        column : str
+            The column name where the accession number is stored
+        values : dict of {str: str}
+            The accession numbers keyed by sample id
+
+        Raises
+        ------
+        QiitaDBError
+            If values tries to change the accession number of a sample that
+            already has an assigned accession number
```
ElDeveloper: I think this section is missing to mention the QiitaDBWarning.

qiita_db/metadata_template/sample_template.py
```diff
@@ -252,3 +252,59 @@ def generate_files(self):
...
+
+        Parameters
+        ----------
+        values : dict of {str: str}
+            The EBI sample accessions, keyed by sample id
+
+        Raises
+        ------
+        QiitaDBError
+            If value tries to change the accession number of a sample that
```
ElDeveloper: What about of something along the lines of:

\`If a sample in \`value\` already has an accession number\`.

Something similar could be done with the rest of the docstrings.

qiita_db/metadata_template/sample_template.py
```diff
@@ -252,3 +252,59 @@ def generate_files(self):
...
+
+        Parameters
+        ----------
+        values : dict of {str: str}
+            The biosample accessions, keyed by sample id
+
+        Raises
+        ------
+        QiitaDBError
+            If value tries to change the accession number of a sample that
```
ElDeveloper: It seems like this is the only instance where you refer to this as `accession number` in the rest of them you refer to them as `biosample accessions`, probably worth to be consistent with this.

qiita_db/metadata_template/test/test_prep_template.py
```diff
@@ -802,58 +802,27 @@ def test_create_error_cleanup(self):
...
+                          '%s.SKD8.640184' % self.test_study.id,
+                          '%s.SKB7.640196' % self.test_study.id}
+        self.assertEqual(pt._get_sample_ids(), exp_sample_ids)
+        self.assertEqual(len(pt), 3)
+        exp_categories = {'str_column', 'ebi_submission_accession',
+                          'run_prefix', 'barcode', 'primer', 'platform',
+                          'instrument_model', 'experiment_design_description',
+                          'library_construction_protocol', 'center_name',
+                          'center_project_name', 'emp_status'}
+        self.assertEqual(set(pt.categories()), exp_categories)
```
ElDeveloper: If the return value is not a set, why do you cast it to a set? If it's to simplify the testing, then assertItemsEqual would fix this https://docs.python.org/2/library/unittest.html#unittest.TestCase.assertItemsEqual

qiita_db/metadata_template/test/test_prep_template.py
```diff
@@ -961,59 +931,28 @@ def test_create_warning(self):
...
+                          '%s.SKD8.640184' % self.test_study.id,
+                          '%s.SKB7.640196' % self.test_study.id}
+        self.assertEqual(pt._get_sample_ids(), exp_sample_ids)
+        self.assertEqual(len(pt), 3)
+        exp_categories = {'str_column', 'ebi_submission_accession',
+                          'run_prefix', 'primer', 'platform',
+                          'instrument_model', 'experiment_design_description',
+                          'library_construction_protocol', 'center_name',
+                          'center_project_name', 'emp_status'}
+        self.assertEqual(set(pt.categories()), exp_categories)
```
ElDeveloper: Same as above.

qiita_db/metadata_template/test/test_sample_template.py
```diff
@@ -1407,72 +1331,63 @@ def test_extend_add_samples(self):
...
+                          "%s.Sample5" % st.id}
+        self.assertEqual(st._get_sample_ids(), exp_sample_ids)
+        self.assertEqual(len(st), 5)
+        exp_categories = {'collection_timestamp', 'description',
+                          'dna_extracted', 'host_subject_id', 'int_column',
+                          'latitude', 'longitude',
+                          'physical_specimen_location',
+                          'physical_specimen_remaining', 'sample_type',
+                          'scientific_name', 'str_column', 'taxon_id'}
+        self.assertEqual(set(st.categories()), exp_categories)
```
ElDeveloper: Same as above, I will stop marking these in case we reach a resolution that's not applicable for the general case.

qiita_db/study.py
```diff
@@ -673,7 +674,13 @@ def sample_template(self):
...
         SampleTemplate id
         """
-        return self._id
+        with TRN:
+            sql = """SELECT EXISTS(SELECT *
+                                   FROM qiita.study_sample
+                                   WHERE study_id = %s)"""
+            TRN.add(sql, [self.id])
+            exists = TRN.execute_fetchlast()
+        return self._id if exists else None
```
ElDeveloper: I don't generally like the idea of having different return types for a property, not sure how others feel? Would you mind updating the docstring to indicate that the returned value is an int and that it can be None?

antgonza: Good point, this has caused problems in the past. What about returning -1?

qiita_db/study.py
```diff
@@ -789,6 +796,87 @@ def _portals(self):
...
+        value : str
+            The new EBI study accession
+
+        Raises
+        ------
+        QiitDBError
+            If the study already has an EBI study accession
+        """
+        with TRN:
+            if self.ebi_study_accession is not None:
```
ElDeveloper: This can be done outside of the TRN block, right?

qiita_db/study.py
```diff
@@ -789,6 +796,87 @@ def _portals(self):
...
+            TRN.execute()
+
+    @property
+    def ebi_submission_status(self):
+        """The EBI submission status of this study
+
+        Returns
+        -------
+        str
+            The study EBI submission status
```
ElDeveloper: If possible can you list what are the different submission statuses here?

qiita_db/support_files/patches/32.sql
```diff
@@ -0,0 +1,40 @@
+-- September 22, 2015
+-- Update the database schema to be able to store all the information that EBI
+-- returns, and it is needed to perform further modifications/additions
```
ElDeveloper: Not sure if the second sentence in this comment makes sense.

qiita_db/test/test_data.py
```diff
@@ -6,8 +6,6 @@
 # The full license is in the file LICENSE, distributed with this software.
 # -----------------------------------------------------------------------------
 
-# TODO: Modify DB to fix 1084
```
ElDeveloper: :beers:

qiita_ware/ebi.py
```diff
@@ -117,24 +118,31 @@ def __init__(self, preprocessed_data_id, action):
...
         self.preprocessed_data = PreprocessedData(preprocessed_data_id)
-        s = Study(self.preprocessed_data.study)
-        self.sample_template = SampleTemplate(s.sample_template)
+        self.study = Study(self.preprocessed_data.study)
+        self.sample_template = SampleTemplate(self.study.sample_template)
         self.prep_template = PrepTemplate(self.preprocessed_data.prep_template)
 
-        status = self.preprocessed_data.submitted_to_insdc_status()
+        if self.preprocessed_data.is_submitted_to_ebi:
+            error_msg = ("Cannot resubmit! Preprocessed data %d has been "
```
ElDeveloper: been already -> already been

qiita_db/study.py
ElDeveloper: I would prefer raising an error, but that may be rather draconian for a
getter.

qiita_ware/ebi.py
```diff
@@ -680,23 +704,70 @@ def parse_EBI_reply(self, curl_result):
...
+
+        success = root.get('success') == 'true'
+        if not success:
+            raise EBISubmissionError("The EBI submission failed:\n%s"
+                                     % curl_result)
+
+        study_elem = root.findall("STUDY")
+        if len(study_elem) > 1:
+            raise EBISubmissionError(
+                "Multiple study tags found in EBI reply: %d" % len(study_elem))
```
ElDeveloper: It would maybe be useful to print what the study tags are.

qiita_ware/ebi.py
```diff
@@ -680,23 +704,70 @@ def parse_EBI_reply(self, curl_result):
...
-                study_accession = accessions.group('study')
-                submission_accession = accessions.group('submission')
-
-        return study_accession, submission_accession
+        try:
+            root = ET.fromstring(curl_result)
+        except ParseError:
+            raise EBISubmissionError(
+                "The curl result from the EBI submission doesn't look like "
+                "an XML file:\n%s" % curl_result)
```
ElDeveloper: How large can this message be? And if it can be too large and it would be presented to the GUI, perhaps we should store the message in the log table and rephrase the error to suggest they get in touch with an admin?

qiita_ware/ebi.py
```diff
@@ -680,23 +704,70 @@ def parse_EBI_reply(self, curl_result):
 
         Returns
         -------
-        study_accession
-            The study accession number, in case of failure it returns None
-        submission_accession
-            The submission accession number, in case of failure it returns None
+        bool
```
ElDeveloper: This argument is not being returned in the `return` statement at line 769.

qiita_ware/ebi.py
```diff
@@ -680,23 +704,70 @@ def parse_EBI_reply(self, curl_result):
...
+            If the submission was successful or not
+        study_accession : str
```
ElDeveloper: Numpy docs are not supposed to name the outputs of the function i.e. this should only really be `str`.

scripts/qiita
```diff
@@ -165,17 +165,14 @@ def load_raw(fp, fp_type, filetype, prep_template):
               help="id in the paramater table associated with the parameters")
 @click.option('--prep_template_id',
               help="Prep template id associated with data", default=None)
-@click.option('--insdc', is_flag=True,
```
ElDeveloper: Does this close #1051?

ElDeveloper:
Ok, I think I'm done with my review, should be ready to go afterwards.

qiita_db/metadata_template/base_metadata_template.py

qiita_db/metadata_template/base_metadata_template.py
josenavas: Doe

qiita_db/metadata_template/base_metadata_template.py
josenavas: Good call! Added.

qiita_db/metadata_template/base_metadata_template.py
josenavas: excellent point! changed!

qiita_db/metadata_template/sample_template.py
josenavas: Changed. As you can see, that day was not my best day of english... ugh

qiita_db/metadata_template/sample_template.py
josenavas: Yup, I changed everything to accession numbers

qiita_db/metadata_template/test/test_prep_template.py
josenavas: Good point. I always forget the existence of the `assertItemsEqual`. I guess a sticky note in my desk will help :smile:

qiita_db/metadata_template/test/test_prep_template.py
josenavas: Done

qiita_db/metadata_template/test/test_sample_template.py
josenavas: Changed, here and all the appearences :)

qiita_db/study.py
josenavas: I've mixed fillings here... If we use `-1` if there is any part of the code that assumes that the Sample Template exists and it actually doesn't, it can generate some errors that may be obscure to find. Raising an error has the advantage that we will quickly identify the source of the error.

On the other hand, if we want to check if a study has a sample template or not, we will be able to check if the returned id is `-1`, while with the error we will need to have a try/except.

@antgonza @ElDeveloper after these thoughts, which one do you think is the best one moving forward?
If I have to choose, I'm slightly towards returning -1, but not sure.

qiita_db/study.py
josenavas: Since `self.ebi_study_accession` is accession to the database, by including it in the TRN block will make sure that such operation is executed inside the transaction of the setter, avoiding potential lost-write errors in a multi-user environment.

qiita_db/study.py
antgonza: Suggest changing to -1 and then running all tests to see if something brakes.

qiita_db/study.py

qiita_db/support_files/patches/32.sql
josenavas: Yup, it was incomplete. Fixed.

qiita_ware/ebi.py
josenavas: Done

qiita_ware/ebi.py
josenavas: Aaaand that's correct! leftover code

qiita_ware/ebi.py
josenavas: Thanks for the tip, changed

qiita_ware/ebi.py
josenavas: AFAIK, EBISubmissions can currently be done only by admin users. Anyways, I think is a good idea on moving this to the log table and just point the log id in the raised error.

qiita_ware/ebi.py
josenavas: Added. This specific error was not tested so I added a new test case for it.

scripts/qiita
josenavas: Looks like it does!

qiita_db/study.py
ElDeveloper: Got it, thanks for the explanation!

qiita_ware/ebi.py
ElDeveloper: Got it, AFAIK, the future is to move towards a state of the system where
the requirements are clearly specified in the interface and everyone
that can conform to these requirements will be able to submit data - but
I may have missed an intermediated discussion that makes that assumption
no longer hold true.

qiita_db/study.py
```diff
@@ -789,6 +802,87 @@ def _portals(self):
...
+        Parameters
+        ----------
+        value : str {%s}
+            The new EBI submission status
+
+        Raises
+        ------
+        ValueError
+            If the status is not known
+        """.format(', '.join(self._valid_ebi_status))
```
ElDeveloper: I just tested out a minimal example and this doesn't work , I think @antgonza once came up with a way to format data inside a docstring, can't remember where though.

```bash
(ipy)yoshikivazquezbaeza:qiita@db-changes-ebi$ python -c 'from py import fun; print help(fun)'
# should have displayed the help here, but it doesn't :( 
(ipy)yoshikivazquezbaeza:qiita@db-changes-ebi$ cat py.py
def fun(x):
    """Lolercoaster

    Has the following values {%s}
    """.format(['foo', 'bar', 'baz'])
    pass
```

antgonza: http://stackoverflow.com/questions/10307696/how-to-put-a-variable-into-python-docstring

josenavas: Thanks for the link @antgonza fixed.
I think at some point we did it as I had it, but then we find out that it doesn't work.

ElDeveloper: Great, thanks!

josenavas:
@ElDeveloper Only the expected qiita-pet errors here. There was a flake8 error but I fixed in my last push, so I don't think that is worth waiting for that one to be done to merge this. Thanks!

PR #1482, opened 2015-09-30
Update raw data

josenavas:
Adds a command to be able to update the files attached to a raw data object - similar to what we have for update preprocessed data.

This is needed to update the EMP data uploaded at the beginning of Qiita, and this can be useful for any future problems.

@antgonza able to review?

### DIFF
--- a/qiita_db/commands.py
+++ b/qiita_db/commands.py
@@ -18,7 +18,7 @@ from shutil import move
 from .study import Study, StudyPerson
 from .user import User
 from .util import (get_filetypes, get_filepath_types, compute_checksum,
-                   convert_to_id)
+                   convert_to_id, move_filepaths_to_upload_folder)
 from .data import RawData, PreprocessedData, ProcessedData
 from .metadata_template import (SampleTemplate, PrepTemplate,
                                 load_template_to_dataframe)
@@ -355,7 +355,12 @@ def update_raw_data_from_cmd(filepaths, filepath_types, study_id, rd_id=None):
         filepath_types_dict = get_filepath_types()
         filepath_types = [filepath_types_dict[x] for x in filepath_types]
 
-        raw_data.clear_filepaths()
+        fps = raw_data.get_filepaths()
+        sql = "DELETE FROM qiita.raw_filepath WHERE raw_data_id = %s"
+        TRN.add(sql, [raw_data.id])
+        TRN.execute()
+        move_filepaths_to_upload_folder(study_id, fps)
+
         raw_data.add_filepaths(list(zip(filepaths, filepath_types)))
 
     return raw_data
--- a/qiita_db/commands.py
+++ b/qiita_db/commands.py
@@ -334,13 +334,13 @@ def update_raw_data_from_cmd(filepaths, filepath_types, study_id, rd_id=None):
         If rd_id is provided and it does not belong to the given study
     """
     if len(filepaths) != len(filepath_types):
-        raise ValueError("Please provide exactly one filepath_type for each"
+        raise ValueError("Please provide exactly one filepath_type for each "
                          "and every filepath")
     with TRN:
         study = Study(study_id)
         raw_data_ids = study.raw_data()
         if not raw_data_ids:
-            raise ValueError("Study %s does not have any raw data" % study_id)
+            raise ValueError("Study %d does not have any raw data" % study_id)
 
         if rd_id:
             if rd_id not in raw_data_ids:
--- a/qiita_db/commands.py
+++ b/qiita_db/commands.py
@@ -307,6 +307,60 @@ def load_parameters_from_cmd(name, fp, table):
     return constructor.create(name, **params)
 
 
+def update_raw_data_from_cmd(filepaths, filepath_types, study_id, rd_id=None):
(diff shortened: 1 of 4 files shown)


antgonza:
My only concern is that update_raw_data_from_cmd overwrites (deletes) old raw data, should we have a parameter or confirmation for this? Note that I understand why this behavior is required for the current situation but not sure we want to have it in the future.

qiita_db/commands.py
```diff
@@ -307,6 +307,65 @@ def load_parameters_from_cmd(name, fp, table):
...
+
+    Raises
+    ------
+    ValueError
+        If 'filepaths' and 'filepath_types' do not have the same length
+        If the study does not have any raw data
+        If rd_id is provided and it does not belong to the given study
+    """
+    if len(filepaths) != len(filepath_types):
+        raise ValueError("Please provide exactly one filepath_type for each"
```
ElDeveloper: There's a space missing at the end of this sentence.

qiita_db/commands.py
```diff
@@ -307,6 +307,65 @@ def load_parameters_from_cmd(name, fp, table):
...
+                         "and every filepath")
+    with TRN:
+        study = Study(study_id)
+        raw_data_ids = study.raw_data()
+        if not raw_data_ids:
+            raise ValueError("Study %s does not have any raw data" % study_id)
```
ElDeveloper: study_id is an int right so it should really be a %d ... though I know python will just coerce it to a string.

qiita_db/commands.py
```diff
@@ -307,6 +307,65 @@ def load_parameters_from_cmd(name, fp, table):
...
+                raise ValueError(
+                    "The raw data %d does not exist in the study %d. Available"
+                    " raw data: %s"
+                    % (rd_id, study_id, ', '.join(map(str, raw_data_ids))))
+            raw_data = RawData(rd_id)
+        else:
+            raw_data = RawData(sorted(raw_data_ids)[0])
+
+        filepath_types_dict = get_filepath_types()
+        filepath_types = [filepath_types_dict[x] for x in filepath_types]
```
ElDeveloper: In the extreme case that a filepath type didn't exist, this would raise a `KeyError` right?

scripts/qiita
```diff
@@ -279,6 +280,25 @@ def load_parameters(fp, table, name):
...
 
 @db.command()
+@click.option('--fp', required=True, type=click.Path(resolve_path=True,
+              readable=True, exists=True), multiple=True,
+              help='Path to the raw data file. This option can be used '
+              'multiple times if there are multiple raw data files.')
+@click.option('--fp_type', required=True, multiple=True, help='Describes the '
+              'contents of the file. Pass one fp_type per fp.',
+              type=click.Choice(get_filepath_types().keys()))
+@click.option('--study', required=True, type=int,
```
ElDeveloper: Is it safe to assume the type is going to be > 0? if that's the case range options may be useful http://click.pocoo.org/5/options/#range-options

scripts/qiita
```diff
@@ -279,6 +280,25 @@ def load_parameters(fp, table, name):
...
+              help='Study whose raw data will be updated')
+@click.option('--raw_data', required=False, type=int,
```
ElDeveloper: Same as above.

ElDeveloper:
Is it worthwhile to add a note the change log?

qiita_db/commands.py
josenavas: Good catch.

qiita_db/commands.py

qiita_db/commands.py
josenavas: Uhm, you're correct. The main usage of this function is through the CLI and in there, those values can only be the correct ones (since the option is of type Choice).

Do you want me to put a specific try/except?

scripts/qiita
josenavas: Good call - done

scripts/qiita

qiita_db/commands.py
ElDeveloper: I don't have a strong feeling on this, what I worry is when this gets
used elsewhere and something goes wrong, the only error we are going to
get is a `KeyError: 'foo'`.

josenavas:
@ElDeveloper I've addressed all your comments.
@antgonza That is not the route that we used for the `update_preprocessed_data`. Note that here it is not removing the data, is actually moving it to the uploads folder, so we are not loosing any data.

This is part of the CLI which is supposed to be used by advanced users/

qiita_db/commands.py
```diff
@@ -307,6 +307,73 @@ def load_parameters_from_cmd(name, fp, table):
...
+        Describes the contents of the files
+    study_id : int
+        The study_id of the study to be updated
+    rd_id : int, optional
+        The id of the raw data to be updated. If not provided, the raw data
+        with lowest id in the study will be updated
+
+    Returns
+    -------
+    qiita_db.data.RawData
```
ElDeveloper: Can you add a description for this return value?

ElDeveloper:
Thanks @josenavas, it looks good to me now!

antgonza:
:+1:

PR #1484, opened 2015-10-02
Instrument model

antgonza:
This adds the instrument model field as a required field for EBI submission. Most of the changes are due to fixing tests. Notable changes: changed the template documentation page, and travis.yml to remove two lines and make it one (suggestion by @josenavas in previous request and found how to do it).

PR #1486, opened 2015-10-03
Library fields

antgonza:
https://github.com/biocore/qiita/pull/1484 should be merged first.

This hardcodes to "good" values:
- library_source "METAGENOMIC"
- library_selection = "PCR" (used to be unknown but decided to go with PCR) 
- library_layout = "SINGLE"

Also removed some constants no needed anymore in test_ebi.py


ElDeveloper:
Can you pull upstream master here?

antgonza:
Not sure we want to do that here. If we go based on plan that should happen in @josenavas' PR. What do you think?

antgonza:
Got it. Is from ebi-improvements not from master. Done.

josenavas:
:+1: @ElDeveloper time for a quick review in here? It is really easy/fast.

ElDeveloper:
My bad, I meant from ebi-improvements not from master :L

Issue #406, opened 2014-10-14
Add ascp to travis to test EBI


ElDeveloper:
@adamrp mentioned that it would be nice to have this downloaded and installed in travis so we could do a test EBI submission, though note that we would need some test credentials of some sort.

antgonza:
Update: We have now tests and ascp added to travis. Added to qiita_ware/tests/test_commands.py by https://github.com/biocore/qiita/pull/1488. However, tests to EBI can only be ran once because once a submission is created it will exists in the test server until the system is reset. EBI knows about this and is in their TODO list. Note that the tests are there except that are commented and have a pass.

antgonza:
Update from EBI:

------------

Neil has added ROLLBACK submission action:
http://ftp.sra.ebi.ac.uk/meta/xsd/sra_1_5/SRA.submission.xsd

Please avoid using against the production service because accession numbers are assigned and lost.

------------


We should be able to fully test on merge to master and close this issue.

antgonza:
Added and tested here: https://github.com/biocore/qiita/pull/2516

PR #1488, opened 2015-10-06
WIP: Next steps

### LINKED ISSUE #1479: Store EBI submission files with the preprocessed data
Currently the files are stored in the working directory. However, the database don't keep track of any of those files. Since all these files are stored in a single directory, we can store such directory under the `preprocessed_data` mount point and link the EBI submission files to the preprocessed data (since there is a 1-1 relationship).

Assigning to @antgonza since he is currently working on the EBI refactor.

antgonza:
This is based on https://github.com/biocore/qiita/pull/1481 so that should be merged first.

Here I'm fixing the last standing points in the joy list, so far:
- Fix #1479
- In the template documentation changing that sample description is a required field for Qiita deploy to a requirement for EBI, as this is required for submissions.
- Removing 2 cfg variables that are not needed anymore with the new EBI system. Today we got a notification that the old system will be out in 2 months.
- Changes so tests pass in the main branch.


qiita_pet/support_files/doc/source/tutorials/prepare-templates.rst
```diff
@@ -98,6 +96,8 @@ These are the columns required for successfully submit your data to EBI:
 +-------------------------------------+-------------------------------------------+--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
 | ``library_construction_protocol``   | free text                                 | Brief description or reference to the protocol that was used for preparing this amplicon library starting from DNA, usually this includes what genomic region was targeted such as *16S*, *ITS*, *18S*, etc. |
 +-------------------------------------+-------------------------------------------+--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
+| ``description``                     | free text                                 | Description of the sample.                                                                                                                                                                                   |
```
josenavas: Why has this been moved from the sample template to the prep template? IIRC, there was a problem if we had duplicated columns between the sample template and the prep template, and the description was required already in the sample template. Also, I think the description applies to the "biological sample" (i.e. sample template) rather than to the preparation sample.

antgonza: My error, this was meant to be in the required fields for EBI in the sample template, not the prep template, will change.

qiita_ware/ebi.py
```diff
@@ -729,20 +710,20 @@ def generate_curl_command(
...
-            self.submission_xml_fp,
-            self.study_xml_fp,
-            self.sample_xml_fp,
-            self.run_xml_fp,
-            self.experiment_xml_fp,
-            join(ebi_dropbox_url, url)
-        )
+        url = '?auth=ENA%20{0}%20{1}'.format(quote(ebi_seq_xfer_user),
+                                             quote(ebi_seq_xfer_pass))
+        curl_command = 'curl -sS -k'
```
josenavas: Usually is recommended to create a list of strings and use appends and a `''.join()` at the end. IIRC, string concatenation is more expensive than list append.

qiita_ware/ebi.py
```diff
@@ -729,20 +710,20 @@ def generate_curl_command(
...
+            curl_command += ' -F "SUBMISSION=@%s"' % self.submission_xml_fp
+        if self.study_xml_fp is not None:
+            curl_command += ' -F "STUDY=@%s"' % self.study_xml_fp
+        if self.sample_xml_fp is not None:
+            curl_command += ' -F "SAMPLE=@%s"' % self.sample_xml_fp
+        if self.run_xml_fp is not None:
+            curl_command += ' -F "RUN=@%s"' % self.run_xml_fp
+        if self.experiment_xml_fp is not None:
+            curl_command += ' -F "EXPERIMENT=@%s"' % self.experiment_xml_fp
+        curl_command += ' "' + join(ebi_dropbox_url, url) + '"'
```
josenavas: curl_command += ' "%s"' % join(ebi_dropbox_url, url)

qiita_ware/ebi.py
```diff
@@ -885,8 +866,14 @@ def generate_demultiplexed_fastq(self, rewrite_fastq=False, mtime=None):
...
         ppd = self.preprocessed_data
 
-        if not isdir(self.ebi_dir) or rewrite_fastq:
-            makedirs(self.ebi_dir)
+        dir_not_exists = not isdir(self.full_ebi_dir)
+        if dir_not_exists or rewrite_fastq:
+            if dir_not_exists:
+                makedirs(self.full_ebi_dir)
+                filetype_id = convert_to_id('directory', 'filepath_type')
+                insert_filepaths([(self.full_ebi_dir, filetype_id)], -1,
```
josenavas: This is just inserting the filepath in the `filepath` table, but it is not linking it with the preprocessed data.
I recommend to use PreprocessedData.add_filepath

ElDeveloper:
I'm having a hard time figuring out what are the specifics that need review in this PR, maybe once #1490 is merged, it will be easier.

antgonza:
This is ready for final review. Note that this solves all the open issues in the joy list: https://github.com/biocore/qiita/pull/1451. Once this is merged, we just need to fix conflicts between master and the ebi-improvements branch ...

qiita_ware/ebi.py
antgonza: Thought about it but I'm not sure we want this because this will imply that we will need to test which filepaths are folders and which ones are files for display, which will complicate code/functionality. I think we could add this later if we see the necessity of displaying this files in the GUI. What do you think?

antgonza: Other option is not to add it to the DB ...

qiita_ware/test/test_commands.py
```diff
@@ -0,0 +1,198 @@
...
+            to_hdf5(fna_fp, f)
+
+        ppd = PreprocessedData.create(
+            study, "preprocessed_sequence_illumina_params", 1,
+            [(demux_fp, 6)], pt)
+
+        return ppd
+
+    def test_submit_EBI_step_2_failure(self):
+        pass
```
mortonjt: Do we still want these tests?  Or should they be dropped?

qiita_ware/test/test_commands.py
```diff
@@ -0,0 +1,198 @@
...
+        # with self.assertRaises(AttributeError):
+        #     submit_EBI(ppd.id, 'ADD', True)
+
+    def test_submit_EBI_parse_EBI_reply_failure(self):
+        ppd = self.write_demux_files(PrepTemplate(1))
+        with self.assertRaises(ComputeError):
+            submit_EBI(ppd.id, 'ADD', True)
+
+    def test_full_submission(self):
+        pass
```
mortonjt: Same here

antgonza: See: https://github.com/biocore/qiita/issues/406

mortonjt: Gotcha.  Thanks for the link!

qiita_pet/support_files/doc/source/tutorials/prepare-templates.rst
```diff
@@ -37,6 +37,8 @@ These are the columns required for successfully submit your data to EBI:
 | ``physical_specimen_location``   | free text            | Where you would go to find physical sample or DNA, regardless of whether it is still available or not.   |
 +----------------------------------+----------------------+----------------------------------------------------------------------------------------------------------+
 | ``taxon_id``                     | integer              | NCBI's taxon id for the sample                                                                           |
++-------------------------------------+-------------------------------------------+----------------------------------------------------------------------------------+
+| ``description``                  | free text            | Description of the sample.                                                                               |
```
ElDeveloper: The rendering is a little bit off, would you mind having a look at the formatting?

screen shot 2015-10-09 at 10 09 59 am

qiita_ware/commands.py
```diff
@@ -66,34 +60,42 @@ def submit_EBI(preprocessed_data_id, action, send, fastq_dir_fp=None):
...
+        for cmd in ebi_submission.generate_send_sequences_cmd():
+            cmd_pieces = shsplit(cmd)
+            try:
+                call(cmd_pieces, stdout=open(ebi_submission.ascp_reply, 'a'))
+            except:
+                with open(ebi_submission.ascp_reply, 'r') as f:
+                    content = f.read()
+                ebi_submission.study.ebi_submission_status = (
+                    "failed: ASCP - %s" % content)
+                LogEntry.create('Fatal', content,
```
ElDeveloper: Are we in risk of this content being too large? i.e. for a submission with thousands of samples? If that's the case, do we really want that much text in the log table?

qiita_ware/commands.py
```diff
@@ -66,34 +60,42 @@ def submit_EBI(preprocessed_data_id, action, send, fastq_dir_fp=None):
...
+                                info={
+                                    'ebi_submission': preprocessed_data_id,
+                                    'fail': cmd})
+        LogEntry.create('Runtime',
+                        ('Submission of sequences of pre_processed_id: '
+                         '%d completed successfully' %
```
ElDeveloper: Does this activate the fireworks at the top of BRF2? :smiley:

qiita_ware/commands.py
antgonza: The only time I have seen this fail is when the login is incorrect and the reply is something like: "Wong login" or something like that. So, no?

qiita_ware/commands.py
antgonza: Nop, cause this time it actually works well (or at least we hope!) vs. giving you the impression via fireworks that it does.

qiita_ware/commands.py
ElDeveloper: Cool, just wanted to avoid cluttering the logging page via the web
interface.

qiita_ware/commands.py
ElDeveloper: So the lack of fireworks is the celebration of success, as this is real
success? ... so postmodern.

qiita_ware/ebi.py
```diff
@@ -872,50 +853,68 @@ def generate_demultiplexed_fastq(self, rewrite_fastq=False, mtime=None):
...
-
-                    if wrote_sequences:
-                        demux_samples.add(s)
-                    else:
-                        del(self.samples[s])
-                        del(self.samples_prep[s])
-                        del(self.sample_demux_fps[s])
-                        remove(sample_fp)
-
+            try:
```
ElDeveloper: Is the reasoning for this block the fact that absolutely any of these statements can fail and that they have been known to fail before? If at all possible, I would try to change this to catch specific exceptions and the `Exception` class, else, it we would need to include the full traceback. I'm trying to think what to do in a case where we have to debug this because it failed at some point.

antgonza: Good question. While writing tests I realized that one issue is that one posible issue is that the demux file doesn't exist or there are errors while reading it. Thus, that try covers any of those possible issues. I tried to isolate them but was really hard as the return is an iterator and there is no easy way to know if empty or not.

qiita_ware/ebi.py
```diff
@@ -925,4 +924,10 @@ def generate_demultiplexed_fastq(self, rewrite_fastq=False, mtime=None):
                 del(self.samples_prep[ms])
                 del(self.sample_demux_fps[ms])
 
+        if not demux_samples:
+            error_msg = ("All samples were removed from the submission "
+                         "because the demux file is empty or the sample names"
+                         "do not match.")
```
ElDeveloper: There's a space missing at the beginning of this sentence.

antgonza: Good catch!

qiita_ware/ebi.py
ElDeveloper: Do you think it makes sense to add the traceback here? I emphasize this
because if a KeyError comes up, the error message is going to
essentially say KeyError XXX, but we won't know where that problem is
coming from, essentially obscuring the message.

antgonza: Not possible to hit: demux will only retrieve samples in the prep template, if there is no hit, it will return an empty iterator.

qiita_ware/test/test_commands.py
```diff
@@ -0,0 +1,198 @@
+#!/usr/bin/env python
```
ElDeveloper: Can you remove this line? Since our files can't be executable (thanks :nose:tests), the shebang has no use.

antgonza: Yup, copied from another file, also removed it there.

ElDeveloper:
:+1: done with my review!

qiita_ware/commands.py
```diff
@@ -66,34 +60,42 @@ def submit_EBI(preprocessed_data_id, action, send, fastq_dir_fp=None):
...
-                            info={'ebi_submission': preprocessed_data_id})
-        else:
-            LogEntry.create('Runtime',
-                            ('Submission of sequences of pre_processed_id: '
-                             '%d completed successfully' %
-                             preprocessed_data_id))
+        for cmd in ebi_submission.generate_send_sequences_cmd():
+            cmd_pieces = shsplit(cmd)
+            try:
+                call(cmd_pieces, stdout=open(ebi_submission.ascp_reply, 'a'))
```
josenavas: Is it worth using moi's [`moi.job.system_call`](https://github.com/biocore/mustached-octo-ironman/blob/master/moi/job.py#L21) function? It does the stdout and stderr parsing for you and it will reduce code in here.

qiita_ware/commands.py
```diff
@@ -66,34 +60,42 @@ def submit_EBI(preprocessed_data_id, action, send, fastq_dir_fp=None):
...
+                    "failed: ASCP - %s" % content)
+                LogEntry.create('Fatal', content,
+                                info={
+                                    'ebi_submission': preprocessed_data_id,
+                                    'fail': cmd})
+        LogEntry.create('Runtime',
+                        ('Submission of sequences of pre_processed_id: '
+                         '%d completed successfully' %
+                         preprocessed_data_id))
         environ['ASPERA_SCP_PASS'] = old_ascp_pass
```
josenavas: I think the above code should be in a try/except so this line can be in a finally. If for whatever reason there is an exception on the above block of code, the environment variable 'ASPERA_SCP_PASS' will not reset.

ElDeveloper:
:+1: on my end, @josenavas, please merge once your review comments are addressed :athletic_shoe:

qiita_ware/commands.py
```diff
@@ -66,34 +60,42 @@ def submit_EBI(preprocessed_data_id, action, send, fastq_dir_fp=None):
...
         # step 5: sending xml and parsing answer
-        xmls_cmds = ebi_submission.generate_curl_command()
+        xmls_cmds = shsplit(ebi_submission.generate_curl_command())
         LogEntry.create('Runtime',
                         ("Submitting XMLs for pre_processed_id: "
                          "%d" % preprocessed_data_id))
         try:
-            # place holder for moi call see #1477
-            xmls_cmds_moi = xmls_cmds
+            call(xmls_cmds, stdout=open(ebi_submission.curl_reply, 'w'))
```
josenavas: I think it may be useful to store the stderr too, in case that something unexpected happens (not only in EBI end)...

qiita_ware/ebi.py
josenavas: My only concern with this solution is that if we call "purge_filepaths" at some point, these entries in the filepath table will be removed because they're not linked from any other table

qiita_ware/ebi.py
josenavas: I agree with @ElDeveloper that the amount of code in the try/except is too big and the lack of traceback will make it hard to debug. Do you know which are the specific errors that where raise? We can include multiple Exception clauses and just leave one at the end that actually raises the entire traceback - for extreme cases...

antgonza: The only possible issue is parsing the hdf5 and this happens when calling the iterator so I could potentially move it to the `for record in i` ... not sure if it will be better.

qiita_ware/ebi.py
```diff
@@ -872,27 +853,39 @@ def generate_demultiplexed_fastq(self, rewrite_fastq=False, mtime=None):
...
+        if dir_not_exists or rewrite_fastq:
+            makedirs(self.full_ebi_dir)
 
             demux = [path for _, path, ftype in ppd.get_filepaths()
                      if ftype == 'preprocessed_demux'][0]
 
             demux_samples = set()
             with open_file(demux) as demux_fh:
+                if not isinstance(demux_fh, File):
+                    error_msg = ("There was an error parsing the demux file: "
```
josenavas: This error message can be improved, as the error is that file is not a demux file.

qiita_ware/commands.py
```diff
@@ -89,26 +92,31 @@ def submit_EBI(preprocessed_data_id, action, send, fastq_dir_fp=None):
...
+            error = str(e)
+            LogEntry.create('Fatal', error,
+                            info={'ebi_submission': preprocessed_data_id,
+                                  'fail': cmd})
         else:
             LogEntry.create('Runtime',
                             ('Submission of sequences of pre_processed_id: '
                              '%d completed successfully' %
                              preprocessed_data_id))
+        finally:
```
josenavas: If `system_call` raises, `xml_content` is not defined, and `stderr` doesn't contain the value from such call (it will have the value from the above call to `system_call` in the loop.

qiita_ware/ebi.py
```diff
@@ -872,27 +853,38 @@ def generate_demultiplexed_fastq(self, rewrite_fastq=False, mtime=None):
...
+        if dir_not_exists or rewrite_fastq:
+            makedirs(self.full_ebi_dir)
 
             demux = [path for _, path, ftype in ppd.get_filepaths()
                      if ftype == 'preprocessed_demux'][0]
 
             demux_samples = set()
             with open_file(demux) as demux_fh:
+                if not isinstance(demux_fh, File):
+                    error_msg = ("'%s' doesn't look like a demux file" % demux)
```
josenavas: Extra parenthesis

josenavas:
2 comments and tests are still failing. It looks like now you're installing ascp, so why is that test failing?

antgonza:
@josenavas this is ready. After your comments I realized that we weren't doing the same for all possible failures in commands.py, which I fixed. Also, fixed some warnings that were left over cause missing instrument_model.

josenavas:
Thanks @antgonza 
awesome work!

PR #1489, opened 2015-10-07
Ebi: ability to extend a submission

josenavas:
This PR modifies the EBI code to be able to extend an existent EBI submission with new samples, experiments and/or runs.


qiita_ware/ebi.py
```diff
@@ -620,6 +652,51 @@ def write_xml_file(self, element, fp):
...
+        # The study.xml file needs to be generated if and only if the study
+        # does NOT have an ebi_study_accession
+        if not self.study.ebi_study_accession:
+            self.study_xml_fp = get_output_fp('study.xml')
+            self.write_xml_file(self.generate_study_xml(), self.study_xml_fp)
+
+        # The sample.xml file needs to be generated if and only if there are
+        # samples in the current submission that do NOT have an
+        # ebi_sample_accession
+        new_samples = {
```
antgonza: Should this be the prep template and not the sample template?

josenavas: No, the ebi_sample_accessions are stored in the sample template.

antgonza: K, thanks!

josenavas:
The failures here are expected - will fix the interface in an oncoming PR

qiita_ware/ebi.py
```diff
@@ -362,8 +372,10 @@ def generate_sample_xml(self):
             'xmlns:xsi': self.xmlns_xsi,
             "xsi:noNamespaceSchemaLocation": self.xsi_noNSL % "sample"})
 
-        for sample_name, sample_info in sorted(viewitems(self.samples)):
-            sample_info = dict(sample_info)
+        samples = samples if samples is not None else viewkeys(self.samples)
```
ElDeveloper: This is a rather subjective comment, but this line doesn't need to be a one-liner, it should jus be an if statement:

```python
if samples is not None:
    samples = viewkeys(self.samples)
```

qiita_ware/ebi.py
```diff
@@ -350,9 +355,14 @@ def generate_study_xml(self):
...
         return study_set
 
-    def generate_sample_xml(self):
+    def generate_sample_xml(self, samples=None):
         """Generates the sample XML file
 
+        Parameters
+        ----------
+        samples : list of str, optional
+            The list of samples to be included in the sample xml
```
ElDeveloper: Cand you add that if `None` is passed, then all the samples are used?

ElDeveloper: BTW, is an empty list an allowed input? If not, should there be a check?

qiita_ware/ebi.py
```diff
@@ -620,6 +652,51 @@ def write_xml_file(self, element, fp):
         create_dir(self.xml_dir)
         ET.ElementTree(element).write(fp, encoding='UTF-8')
 
+    def generate_xml_files(self):
+        """Generate the XML files"""
```
mortonjt: Would it be desirable to add numpy docs here?

ElDeveloper:
:+1:

qiita_ware/ebi.py
josenavas: You're correct, although the correct if is:
```python
if not samples:
    samples = viewkeys(self.samples)
```

qiita_ware/ebi.py
josenavas: It is already a numpy doc. This function doesn't take parameters, doesn't return and doesn't raise anything specific.

josenavas:
comments addressed

qiita_ware/ebi.py
mortonjt: Gotcha.  Dumb comment

mortonjt:
I don't see any outstanding issues

qiita_ware/ebi.py
ElDeveloper: Awesome, thanks - would be great to have an accompanying test case.

josenavas: Added

2015-10-07 15:45 GMT-07:00 Yoshiki Vázquez Baeza <notifications@github.com>:

> In qiita_ware/ebi.py
> <https://github.com/biocore/qiita/pull/1489#discussion_r41457729>:
>
> >          """Generates the sample XML file
> >
> > +        Parameters
> > +        ----------
> > +        samples : list of str, optional
> > +            The list of samples to be included in the sample xml
>
> Awesome, thanks - would be great to have an accompanying test case.
> … <#150447b4de91d02f_>
> On (Oct-07-15|15:08), Jose Navas wrote: > """Generates the sample XML file
> > > + Parameters > + ---------- > + samples : list of str, optional > + The
> list of samples to be included in the sample xml Good point about the empty
> list - in case of an empty list or None, all the samples should be used.
> --- Reply to this email directly or view it on GitHub:
> https://github.com/biocore/qiita/pull/1489/files#r41454309
>
> —
> <https://github.com/biocore/qiita/pull/1489/files#r41457729>.
>



-- 
Jose Navas

PR #1492, opened 2015-10-08
Fixing update code

josenavas:
There is an error in the update code: if you try to update the sample template with a dataframe (i.e. a new sample template) that has less samples that the one in the database, the current system will fail with the lovely error: `Can only compare identically-labeled DataFrame objects`

However, this should be allowed: if I submit a new sample template with only one row and one column, because I want to update only such value, I should be able to do it. This can minimize the errors introduced during sample template updates.

Another advantage of this, is that then I can use the interface to extend the current sample template by adding only the new samples (and leave the old ones as they're). Since the interface performs the extend and update at the same time, this was failing, but now this is possible.

@antgonza @ElDeveloper possible a quick review here? This will be needed for the AG work


antgonza:
:+1:

PR #1496, opened 2015-10-11
Urgent: Fix ebi handler

antgonza:
We forgot to modify the code in the actual handler and this is broken, it should be a quick review.


wasade:
:+1:, no comments

mortonjt:
I see no problems with it

antgonza:
Thanks guys. I _think_ all tests are gonna pass but let's wait for them to finish before merging. Like 30min from now.

Issue #1497, opened 2015-10-11
Errors while updating prep template


antgonza:
While trying to update a prep template I realized that we do this with [two steps](https://github.com/biocore/qiita/blob/master/qiita_pet/handlers/study_handlers/description_handlers.py#L432): extend, then update. This behavior causes 2 issues:
- [ ] When you extend a single column, for example instrument_model, you get a few warnings that are not accurate: 

```
qiita_db/metadata_template/util.py:141: QiitaDBWarning: Sample names were already prefixed with the study id.
  QiitaDBWarning)
qiita_db/metadata_template/base_metadata_template.py:575: QiitaDBWarning: Some functionality will be disabled due to missing columns:
    Demultiplexing with multiple input files disabled. If your raw data includes multiple raw input files, you will not be able to preprocess your raw data: primer, run_prefix, barcode;
    Demultiplexing disabled. You will not be able to preprocess your raw data: primer, barcode;
    EBI submission disabled: platform, library_construction_protocol, experiment_design_description, center_name, primer.
See the Templates tutorial for a description of these fields.
  QiitaDBWarning)
qiita_db/metadata_template/base_metadata_template.py:716: QiitaDBWarning: The following columns have been added to the existing template: instrument_model
  QiitaDBWarning)
```
- [ ] Then, my guess is that the extend did everything we need, the update doesn't do anything and brakes with:

```
ValueError: Error running SQL query:
Query: UPDATE qiita.prep_1060 AS t SET

                     FROM (VALUES )
                        AS c(sample_id, )
                     WHERE c.sample_id = t.sample_id

Arguments: []
Error: syntax error at or near "FROM"
LINE 3:                      FROM (VALUES )
                             ^
```

antgonza:
@josenavas can this be closed?

josenavas:
I think the second part of this issue was fixed (i.e. the `ValueError`). However, the warnings issue is still not fixed. In order to fix that, I think the validation of the sample/prep information should be slightly changed:
 - Instead of only checking the values uploaded by the user, we need to merge his information with the one already present in the database in order to check if there are any missing columns.

Unsure how easy/fast is to change this.

antgonza:
@josenavas is this fixed?

josenavas:
No - should we create a new milestone (metadata refactor) and add the metadata-related issues to it?

antgonza:
I think that will be great. Can you do it?

antgonza:
This was solved in the October 1st release.

PR #1498, opened 2015-10-12
Fixing error in 1497

josenavas:
Fixes the error described in https://github.com/biocore/qiita/issues/1497.

The warning require more thought since the problem is that for creation and update/extend we are using the same function, but the functionality of such method for update/extend is slightly different (i.e. it should warn if the data is not in the DB already). Doing this efficiently and w/o code duplication is a bit more tricky.

### DIFF
--- a/qiita_db/metadata_template/base_metadata_template.py
+++ b/qiita_db/metadata_template/base_metadata_template.py
@@ -506,7 +506,8 @@ class MetadataTemplate(QiitaObject):
         return "%s%d" % (cls._table_prefix, obj_id)
 
     @classmethod
-    def _clean_validate_template(cls, md_template, study_id, restriction_dict):
+    def _clean_validate_template(cls, md_template, study_id, restriction_dict,
+                                 current_columns=None):
         """Takes care of all validation and cleaning of metadata templates
 
         Parameters
@@ -517,6 +518,8 @@ class MetadataTemplate(QiitaObject):
             The study to which the metadata template belongs to.
         restriction_dict : dict of {str: Restriction}
             A dictionary with the restrictions that apply to the metadata
+        current_columns : iterable of str, optional
+            The current list of metadata columns
 
         Returns
         -------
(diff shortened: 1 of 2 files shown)


josenavas:
After a discussion with @antgonza we came up with a solution for the warnings too.
This now fully fixes #1497

antgonza:
:+1: if tests pass.

josenavas:
@ElDeveloper @mortonjt able to do a quick review?

qiita_db/metadata_template/base_metadata_template.py
```diff
@@ -506,7 +506,8 @@ def _table_name(cls, obj_id):
         return "%s%d" % (cls._table_prefix, obj_id)
 
     @classmethod
-    def _clean_validate_template(cls, md_template, study_id, restriction_dict):
+    def _clean_validate_template(cls, md_template, study_id, restriction_dict,
+                                 current_columns=None):
         """Takes care of all validation and cleaning of metadata templates
 
         Parameters
```
mortonjt: In the Parameters section, I'd personally prefer to have `pd.DataFrame` instead of just `DataFrame`.
Again, this is just personal preference.

qiita_db/metadata_template/base_metadata_template.py
```diff
@@ -1185,6 +1193,8 @@ def update(self, md_template):
                     "There are no differences between the data stored in the "
                     "DB and the new data provided",
                     QiitaDBWarning)
+                return
```
mortonjt: Would it be desirable to add a test for this warning?

mortonjt:
Just a couple of very minor comments.  I think it looks good.

qiita_db/metadata_template/base_metadata_template.py
josenavas: Can you open an issue? I can fix then later in the entire code base. Thanks

qiita_db/metadata_template/base_metadata_template.py
josenavas: See `test_update_equal`

mortonjt: Please correct me if I'm wrong, but it looks like the actual check `npt.assert_warns` is deleted? `test_update_fewer_samples`

josenavas: Yes, that is the test that makes sure that the warning is not raised, but the test for this specific warning is in `test_update_equal`

josenavas: Actually, the one in test_update_fewer_samples is for the warning that missing columns is not raised

qiita_db/metadata_template/base_metadata_template.py
mortonjt: Done #1499

mortonjt:
:+1:

PR #1500, opened 2015-10-14
fixing ebi click command

antgonza:
small left over error, which shows that we need tests for cli, which is already an open issue.


josenavas:
:+1:

mortonjt:
Looks good.  But I don't have merge permissions

PR #1516, opened 2015-10-28
fixing (the) future

antgonza:


antgonza:
@biocore/qiita-dev can we get a couple of reviews here? This fixes the future and new biom issues in qiita.

josenavas:
Shouldn't six be in the setup.py now?

antgonza:
We have been using six for other things AKA I didn't introduce this dependency. My guess is that is being installed by other at installation.

josenavas:
Well, accordingly to the code we where using future.utils.six. If six is now a direct dependency of Qiita, it should be in the setup.py. I think it is not safe to assume that it gets installed by one of our dependencies.

squirrelo:
it’s an python dependency, but would still be good to explicitly install it here.

antgonza:
I can add if that's what we agree on. However, I feel that we should keep relaying on ipython to install it as this way we are sure to not have any version conflicts, what do you think?

josenavas:
I would recommend against it, in case that Ipython drops such dependency, we will be missing it. However, if others agree with this solution ok with me.

wasade:
Isn't six part of future?

antgonza:
Not since 0.14 ...

wasade:
Pretty sure the string io imports can be made from "io" which is a built in
module. See how we did it in biom

antgonza:
Tried io but all tests are using strings and io.StringIO expects unicode. Then realized that in other sections of qiita we were using from six so decided to go that route.

josenavas:
I think you just need to add a `u` in front of the strings so they're unicode.

wasade:
Yes

Jorge-C:
Or even importing unicode_literals from __future__. Good overview of the
advantages / drawbacks of the two approaches here
http://python-future.org/unicode_literals.html

antgonza:
I can move back and add a u or use unicode_literals but not sure what's the benefit over six?

PR #1523, opened 2015-11-01
WIP: Artifact changes

josenavas:
This is the PR from upstream artifact against upstream master, and this will be used to keep track of remaining changes. More details will be added as needed.

 - [x] Modify populate_test_db.sql to be adapted to the new schema (#1524)
 - [x] Add software, command and parameters objects (#1525)
 - [x] Add artifact object (#1527)
 - [x] Remove qiita_db/data.py and qiita_db/test/test_data.py and fix all tests in qiita_db (@josenavas - I'm fixing all these at once)
 - [x] Put in place minimal plugin system
 - [x] Move current target gene pipeline to its own plugin (inside the repo)
 - [x] Put in place minimal execution system environment
 - [ ] **kwargs -> dict in create new parameter
 - [x] Adopt GUI to use new system ~~(I think we should include here the work that @antgonza is doing)~~
 - [ ] `submit_to_VAMPS ` function needs to check if the artifact is being submitted or not in order to raise if it has been already submitted
 - [x] Make sure that the new artifacts that represent preprocessed data have the same ids as this will allow us to have the same names for EBI submission. (https://github.com/biocore/qiita/pull/1565)
 - [ ] Add admin documentation in how the plugin launcher script should look like
 - [x] If you add a raw data from another study, you can see the preprocessed data from the other study (https://github.com/biocore/qiita/pull/1572)
 - [x] Then you can't delete the raw data cause it has children (https://github.com/biocore/qiita/pull/1572)
 - [ ] If you change the processed data (artifact) visibility it doesn't propagate to other artifacts (this might  be OK to leave until the new GUI is ready) https://github.com/biocore/qiita/pull/1578
 - [ ] Add a minimal documentation about the new configuration requirements and how to restart jobs. Don't forget all the things that you helped me with to set up this in my laptop. 
 - [x] There is no way to delete a raw data and return the files to the upload folder (https://github.com/biocore/qiita/pull/1572)
 - [x] Regenerate the new config file


ElDeveloper:
There should be a special consideration for positional arguments, as briefly discussed in #1525.

antgonza:
Merging after discussion in Qiita meeting. We need all the help we can get, so: 
`in nomine Patris et Filii et Spiritus Sancti`

PR #1525, opened 2015-11-02
Software, Command and Parameters objects

josenavas:
This adds the Software, Commands and Parameters objects, which is another item in https://github.com/biocore/qiita/pull/1523.

Disclaimer: these objects will probably change once we have the plugins in place. However, I need them in order to move forward with the artifact object. I did a smaller PR so it is easier to review, rather than adding all the objects at once.

### DIFF
--- a/qiita_db/software.py
+++ b/qiita_db/software.py
@@ -1,0 +1,511 @@
+# -----------------------------------------------------------------------------
+# Copyright (c) 2014--, The Qiita Development Team.
+#
+# Distributed under the terms of the BSD 3-clause License.
+#
+# The full license is in the file LICENSE, distributed with this software.
+# -----------------------------------------------------------------------------
+
+from .base import QiitaObject
+from .sql_connection import TRN
+from .exceptions import (QiitaDBDuplicateError, QiitaDBError,
+                         QiitaDBUnknownIDError)
+from .util import get_table_cols, get_table_cols_w_type
+
+
+class Command(QiitaObject):
+    r"""An executable command available in the system
+
+    Attributes
+    ----------
+    name
+    description
+    cli
+    parameters_table
+
+    Methods
+    -------
+    create
+
+    See Also
+    --------
+    qiita_db.software.Software
+    """
+    _table = "software_command"
+
+    def _check_id(self, id_):
+        """Check that the provided ID actually exists in the database
+
+        Parameters
+        ----------
+        id_ : int
+            The ID to test
+
+        Notes
+        -----
+        This function overwrites the base function, as the sql layout doesn't
+        follow the same conventions done in the other classes.
+        """
+        with TRN:
+            sql = """SELECT EXISTS(
+                        SELECT *
+                        FROM qiita.software_command
+                        WHERE command_id = %s)"""
+            TRN.add(sql, [id_])
+            return TRN.execute_fetchlast()
+
+    @classmethod
+    def create(cls, software, name, description, cli_cmd, parameters_table):
+        r"""Creates a new command in the system
+
+        Parameters
+        ----------
+        software : qiita_db.software.Software
+            The software to which this command belongs to.
+        name : str
+            The name of the command
+        description : str
+            The description of the command
+        cli_cmd : str
+            The CLI used to call this command
+        parameters_table : str
+            The name of the table in which the parameters of the commands are
+            stored
+
+        Returns
+        -------
+        qiita_db.software.Command
+            The newly created command
+        """
+        with TRN:
+            sql = """INSERT INTO qiita.software_command
+                            (name, software_id, description, cli_cmd,
+                             parameters_table)
+                     VALUES (%s, %s, %s, %s, %s)
+                     RETURNING command_id"""
+            sql_params = [name, software.id, description, cli_cmd,
+                          parameters_table]
+            TRN.add(sql, sql_params)
+            c_id = TRN.execute_fetchlast()
+
+        return cls(c_id)
+
+    @property
+    def name(self):
+        """The name of the command
+
+        Returns
+        -------
+        str
+            The name of the command
+        """
+        with TRN:
+            sql = """SELECT name
+                     FROM qiita.software_command
+                     WHERE command_id = %s"""
+            TRN.add(sql, [self.id])
+            return TRN.execute_fetchlast()
+
+    @property
+    def description(self):
+        """The description of the command
+
+        Returns
+        -------
+        str
+            The description of the command
+        """
+        with TRN:
+            sql = """SELECT description
+                     FROM qiita.software_command
+                     WHERE command_id = %s"""
+            TRN.add(sql, [self.id])
+            return TRN.execute_fetchlast()
+
+    @property
+    def cli(self):
+        """The CLI used to call the command
+
+        Returns
+        -------
+        str
+            The CLI used to call the command
+        """
+        with TRN:
+            sql = """SELECT cli_cmd
+                     FROM qiita.software_command
+                     WHERE command_id = %s"""
+            TRN.add(sql, [self.id])
+            return TRN.execute_fetchlast()
+
+    @property
+    def parameters_table(self):
+        """The table in which the parameters of th e command are stored
+
+        Returns
+        -------
+        str
+            The name of the table
+        """
+        with TRN:
+            sql = """SELECT parameters_table
+                     FROM qiita.software_command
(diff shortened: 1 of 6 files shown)


qiita_db/software.py
```diff
@@ -0,0 +1,522 @@
...
+                command.parameters_table,
+                ', '.join(kwargs),
+                ', '.join(['%s'] * len(kwargs)))
+            TRN.add(sql, vals)
+
+            return cls(TRN.execute_fetchlast(), command)
+
+    @classmethod
+    def iter(cls, command):
+        """Iterates over all parameter set of the given command
```
antgonza: set -> sets?

antgonza:
Minor comment. Looks good to me.

josenavas:
Thanks @antgonza

qiita_db/software.py
```diff
@@ -0,0 +1,522 @@
...
+        with TRN:
+            sql = """SELECT parameters_table
+                     FROM qiita.software_command
+                     WHERE command_id = %s"""
+            TRN.add(sql, [self.id])
+            return TRN.execute_fetchlast()
+
+
+class Software(QiitaObject):
+    r"""A software available in the system
```
ElDeveloper: 'A software' sounds rather strange, would it be more appropriate to say "A software package" ...

qiita_db/software.py
```diff
@@ -0,0 +1,522 @@
...
+
+        Parameters
+        ----------
+        name : str
+            The name of the software
+        version : str
+            The version of the software
+        description : str
+            The description of the software
+        publications : list of 2-tuples of str
```
ElDeveloper: Can you add `, optional` at the end of this line?

qiita_db/software.py
```diff
@@ -0,0 +1,522 @@
...
+            TRN.add(sql, [self.id])
+            return TRN.execute_fetchflatten()
+
+    @property
+    def publications(self):
+        """The publications attached to the software
+
+        Returns
+        -------
+        list of 2-tuples of str
```
ElDeveloper: Maybe just to be a bit more clear "list of (str, str) tuples".

qiita_db/software.py
```diff
@@ -0,0 +1,522 @@
...
+            TRN.add(sql, [self.id])
+            return TRN.execute_fetchindex()
+
+    def add_publications(self, publications):
+        """Add publications to the software
+
+        Parameters
+        ----------
+        publications : list of 2-tuples of str
+            A list with the (DOI, pubmed_id) of the publications to be attached
```
ElDeveloper: Maybe worth adding a link to what a PubMedID is? https://www.nlm.nih.gov/bsd/disted/pubmedtutorial/020_830.html

qiita_db/software.py
```diff
@@ -0,0 +1,522 @@
...
+            sql = """INSERT INTO qiita.software_publication
+                            (software_id, publication_doi)
+                        VALUES (%s, %s)"""
+            sql_params = [[self.id, doi] for doi, _ in publications]
+            TRN.add(sql, sql_params, many=True)
+            TRN.execute()
+
+
+class Parameters(object):
+    """Models a specific set of parameters of a command
```
ElDeveloper: It seems like this is a helper class for the `Command` class, would it be useful mentioning that?

qiita_db/software.py
```diff
@@ -0,0 +1,522 @@
...
+
+            extra = set(kwargs).difference(db_cols)
+            if extra:
+                raise QiitaDBError(
+                    "Extra parameters for command %s: %s"
+                    % (command.name, ', '.join(extra)))
+
+            cols = ["{} = %s".format(col) for col in kwargs]
+            sql = "SELECT EXISTS(SELECT * FROM qiita.{0} WHERE {1})".format(
+                table, ' AND '.join(cols))
```
ElDeveloper: Could there be an instance where the command has no parameters? If so wouldn't such a case make this query be malformed?

qiita_db/software.py
```diff
@@ -0,0 +1,522 @@
...
+            return TRN.execute_fetchlast()
+
+    @property
+    def values(self):
+        """The values of the parameter set
+
+        Returns
+        -------
+        dict
+            Dictionary with the parameters values keyed by parameter name
```
ElDeveloper: How are positional arguments going to be handled here?

qiita_db/software.py
```diff
@@ -0,0 +1,522 @@
...
+
+            values = self.values
+
+            result = []
+            for p_name, p_type in sorted(table_cols):
+                if p_type == 'boolean':
+                    if values[p_name]:
+                        result.append("--%s" % p_name)
+                else:
+                    result.append("--%s %s" % (p_name, values[p_name]))
```
ElDeveloper: Would it be useful to always quote the arguments? Not sure if that would bite us somehow, but when we've had command-calls before, its been useful to quote arguments in case one of them includes spaces. Which although chaotic and unforgivable, tend to inevitably happen.

qiita_db/test/test_software.py
```diff
@@ -0,0 +1,170 @@
...
+        cmd = Command(1)
+        obs = list(Parameters.iter(cmd))
+        exp = [Parameters(i, cmd) for i in range(1, 8)]
+        self.assertEqual(obs, exp)
+
+    def test_name(self):
+        self.assertEqual(Parameters(1, Command(1)).name, "Defaults")
+
+    def test_values(self):
+        exp = {'min_per_read_length_fraction': 0.75,
```
ElDeveloper: This just made me realize that we are keeping the values as proper types, do we _really_ need this? We've seen how terrible it has been to do this with metadata, so I'm dubious to introduce this into a brand new ground. Your call, but I would advice to keep strings as ultimately it will not matter for the command call, but 3.00 and 3 are interpreted differently depending on the command line argument parser.

ElDeveloper:
@josenavas, just did a quick pass, sorry for the delay!

qiita_db/software.py

qiita_db/software.py

qiita_db/software.py

qiita_db/software.py
josenavas: Done, thanks for the link!

qiita_db/software.py
josenavas: I don't think it is a helper class for the command class, it just models a set of parameters of a command. I foresee this class being used quite around to pass the parameters used/to be used to process an artifact.

qiita_db/software.py
josenavas: I guess is possible, but I don't know how often we're going to hit such case (i.e. a command that doesn't accept any parameter...)

Can you add it a comment in #1523 PR? I don't think that we have a good solution for that and these object will change a bit when the plugin system is in place (planning tomorrow) - but that is not a blocker for #1527

qiita_db/software.py
josenavas: Good point - can you add a comment in #1523?

qiita_db/software.py
josenavas: Good point - changed

qiita_db/test/test_software.py
josenavas: The difference here is that in theory we know these types in advance so we shouldn't hit the same problems than with the metadata.

qiita_db/software.py
ElDeveloper: Noted, thanks!

qiita_db/software.py
ElDeveloper: I think it is possible, not common, but possible. And I agree, it isn't a blocker.

qiita_db/test/test_software.py
ElDeveloper: :L in theory, but reality always bites us. Since it doesn't make a difference, then we should go the safe route and store these as strings. Perhaps I'm being too pessimistic here, what are others' thoughts?

josenavas: Ok, can you add a comment to #1523?
Those tables already exist in the DB and the changes may be bigger that just changing those types here. Since this may impact qiita_ware and it will be revamped with the plugin system, I would suggest to add those changes later.

Thanks!

ElDeveloper: Sounds good - I didn't think about the DB, so yes, that makes perfect sense.

ElDeveloper:
Thanks @josenavas

PR #1527, opened 2015-11-06
Artifact obj

josenavas:
~~Built on top of #1525, so review/merge that one first.~~

This PR changes the way we've been storing files: instead of adding a prefix with the artifact_id, it creates a folder for the artifact (named with the artifact id) and it moves the files in such folder.

Since moving the files to support this new structure will be expensive, we leave the old mountpoints working in the old style while, the new mountpoints (based on the artifact type) follow the new structure. This means that the code has to be back-compatible (yay!) which adds a bit of complexity in the functions dealing with files.


qiita_db/artifact.py
```diff
@@ -0,0 +1,598 @@
...
+                         QiitaDBOperationNotPermittedError)
+from .util import (convert_to_id, insert_filepaths,
+                   move_filepaths_to_upload_folder, retrieve_filepaths)
+from .software import Parameters, Command
+from .study import Study
+from .metadata_template import PrepTemplate
+
+
+class Artifact(QiitaObject):
+    r"""A piece of data that different commands can be applied on
```
antgonza: Not sure this description is clear enough. What about something like: "Any kind of file stored in the system and its attributes"?

qiita_db/artifact.py
```diff
@@ -0,0 +1,598 @@
...
+    delete
+
+    See Also
+    --------
+    qiita_db.QiitaObject
+    """
+    _table = "artifact"
+
+    @classmethod
+    def create(cls, filepaths, artifact_type, timestamp=None,
```
antgonza: In which cases will the optional variables be None? Worth adding to notes?

qiita_db/artifact.py
```diff
@@ -0,0 +1,598 @@
...
+        # Parents or prep template must be provided, but not both
+        if parents and prep_template:
+            raise QiitaDBArtifactCreationError(
+                "parents or prep_template should be provided but not both")
+        elif not (parents or prep_template):
+            raise QiitaDBArtifactCreationError(
+                "at least parents or prep_template must be provided")
+
+        # If parents is provided, processing parameters should also be provided
+        if parents and not processing_parameters:
```
antgonza: I guess this if and the next one could be elifs from the one above, right?

qiita_db/artifact.py
```diff
@@ -0,0 +1,598 @@
...
+                "provided.")
+
+        # If prep_template is provided, processing_parameters should not be
+        # provided
+        if prep_template and processing_parameters:
+            raise QiitaDBArtifactCreationError(
+                "if prep_template is provided, processing_parameters should "
+                "not be provided.")
+
+        if timestamp is None:
```
antgonza: Should datetime.now() be passed on creation and not defaulting here? AKA should timestamp be a required parameter?

qiita_db/artifact.py
```diff
@@ -0,0 +1,598 @@
...
+            timestamp = datetime.now()
+
+        with TRN:
+            visibility_id = convert_to_id("sandbox", "visibility")
+            artifact_type_id = convert_to_id(artifact_type, "artifact_type")
+
+            if parents:
+                # Check that all parents belong to the same study
+                studies = {p.study.id for p in parents}
```
antgonza: Should this check be part of the checking block above? It seems like we shouldn't even start the transaction if he hit this case.

qiita_db/artifact.py
```diff
@@ -0,0 +1,598 @@
...
+        ------
+        QiitaDBArtifactDeletionError
+            If the artifact is public
+            If the artifact has children
+            If the artifact has been analyzed
+            If the artifact has been submitted to EBI
+            If the artifact has been submitted to VAMPS
+        """
+        with TRN:
+            # This will fail if the artifact with id=artifact_id doesn't exist
```
antgonza: It seems like lines 228 to 241 can be out of the TRN, right?

qiita_db/artifact.py
```diff
@@ -0,0 +1,598 @@
...
+                     WHERE artifact_id = %s"""
+            TRN.add(sql, [artifact_id])
+
+            # If the artifact doesn't have parents, we move the files to the
+            # uploads folder. We also need to nullify the column in the prep
+            # template table
+            if not instance.parents:
+                move_filepaths_to_upload_folder(study.id, filepaths)
+
+                sql = """UPDATE qiita.prep_template
```
antgonza: Out of curiosity, should the nullification be in move_filepaths_to_upload_folder?

qiita_db/software.py
```diff
@@ -0,0 +1,533 @@
...
+    def create(cls, param_set_name, command, **kwargs):
+        r"""Create a new parameter set for the given command
+
+        Parameters
+        ----------
+        param_set_name: str
+            The name of the new parameter set
+        command : qiita_db.software.Command
+            The command to add the new parameter set
+        kwargs : dict
```
antgonza: Does it have to be kwargs? It's basically a dict, no?

qiita_db/support_files/patches/33.sql
```diff
@@ -449,3 +469,33 @@ DROP TABLE qiita.preprocessed_data;
...
+
+-- Rename the id columns from the parameters tables
+ALTER TABLE qiita.processed_params_sortmerna RENAME COLUMN processed_params_id TO parameters_id;
+ALTER TABLE qiita.processed_params_uclust RENAME COLUMN processed_params_id TO parameters_id;
+ALTER TABLE qiita.preprocessed_sequence_454_params RENAME COLUMN preprocessed_params_id TO parameters_id;
+ALTER TABLE qiita.preprocessed_sequence_illumina_params RENAME COLUMN preprocessed_params_id TO parameters_id;
+ALTER TABLE qiita.preprocessed_spectra_params RENAME COLUMN preprocessed_params_id TO parameters_id;
+
+-- Create a function to return the roots of an artifact, i.e. the source artifacts
+CREATE FUNCTION qiita.find_artifact_roots(a_id bigint) RETURNS SETOF bigint AS $$
```
antgonza: I don't see an explicit test for this: find_artifact_roots. Note that I think having a simple test in python should suffice.

qiita_db/test/test_artifact.py
```diff
@@ -0,0 +1,374 @@
...
+    def test_delete(self):
+        test = Artifact.create(self.filepaths_root, "FASTQ",
+                               prep_template=self.prep_template)
+
+        Artifact.delete(test.id)
+
+        with self.assertRaises(QiitaDBUnknownIDError):
+            Artifact(test.id)
+
+    def test_tiemstamp(self):
```
antgonza: test_timestamp

qiita_db/test/test_artifact.py
```diff
@@ -0,0 +1,374 @@
...
+                         datetime(2012, 10, 2, 17, 30, 00))
+
+    def test_processing_parameters(self):
+        self.assertIsNone(Artifact(1).processing_parameters)
+        self.assertEqual(Artifact(2).processing_parameters,
+                         Parameters(1, Command(1)))
+        self.assertEqual(Artifact(3).processing_parameters,
+                         Parameters(2, Command(1)))
+
+    def test_visibility(self):
```
antgonza: This and test_artifact_type seem not idea. Basically, they are testing that the data created in the populate is right (testing getter) but not that the setters/getters work as expected.

qiita_db/test/test_artifact.py
```diff
@@ -0,0 +1,374 @@
...
+                                 QiitaDBOperationNotPermittedError,
+                                 QiitaDBUnknownIDError)
+from qiita_db.metadata_template import PrepTemplate
+from qiita_db.study import Study
+from qiita_db.software import Command, Parameters
+from qiita_db.util import get_mountpoint, get_count
+
+
+@qiita_test_checker()
+class ArtifactTests(TestCase):
```
antgonza: I think a test for adding biom might be useful.

qiita_db/software.py
```diff
@@ -0,0 +1,533 @@
...
+            sql = """SELECT EXISTS(
+                        SELECT *
+                        FROM qiita.software_command
+                        WHERE command_id = %s)"""
+            TRN.add(sql, [id_])
+            return TRN.execute_fetchlast()
+
+    @classmethod
+    def create(cls, software, name, description, cli_cmd, parameters_table):
+        r"""Creates a new command in the system
```
antgonza: Should this error if the software already exists?

antgonza: In fact, should it error if the software and/or the command + params already exists?

antgonza:
A few comments. Do you think it will be worth solving some other failures in this PR or the plan is to have this ready and then try to solve all of them (might be a large PR)?

qiita_db/artifact.py
josenavas: Thanks, changed to `Any kind of file (or group of files) stored in the system and its attributes`as it can potentially represent a group of files, not just a single file.

qiita_db/artifact.py
josenavas: I'm not sure what you mean in here. I think the description in the parameters already explain when they are optional. The only one that doesn't have that kind of explanation is timestamp. The only reason that I'm actually exposing that parameter through the API is in case that we need to upload something manually through the CLI.

qiita_db/artifact.py
josenavas: And you're correct! Changed

qiita_db/artifact.py
josenavas: I don't think so - I was thinking in the case that we process something outside and we upload it through the CLI, we probably want to put the correct timestamp. Or do you think it is better to just remove that parameter and always default to "datetime.now()" since that is the time the artifact gets added to the system?

qiita_db/artifact.py
josenavas: The difference here is that `p.study.id` is already accessing the database, so that's why I started the transaction.

qiita_db/artifact.py
josenavas: Nope, they're already accessing the DB.

qiita_db/artifact.py
josenavas: I'm unsure about that. This is not the only place where move_filepaths_to_upload_folder is used.

qiita_db/software.py
josenavas: Good point - do you mind adding it to the list in #1523 so I tackle it when I rework that code for working with plugins?

qiita_db/support_files/patches/33.sql
josenavas: That's a really good point - I found my-self debugging this later when I was working in the artifact object, so having specific tests would've been better!

qiita_db/software.py
antgonza: Done

qiita_db/test/test_artifact.py
josenavas: Ooops

qiita_db/test/test_artifact.py
josenavas: There are a specific test for the setters. This is just checking that the actual getter works, so better minimizing the amount of code that is done for this in here. Does it make sense?

qiita_db/test/test_artifact.py
josenavas: From the point of view of the system is just a "processed" artifact, so it shouldn't make a difference. Anyways, I added a test. At the end is better to be safe than sorry.

qiita_db/software.py
josenavas: Maybe the tuple (software, cli_cmd)?

As an example using QIIME: split_librares.py. The parameters table did not change between 1.8.0 and 1.9.0, so we can re-use the parameters_table. However, the software would be different, since one will be the 1.8.0 version and the other one would be 1.9.0.

Does it makes sense?

josenavas:
Fixing tests is a parallel task - so I would suggest having multiple PR.

qiita_db/artifact.py
antgonza: It's not clear and might be great to add notes on which cases is fine to leave the defaults for: timestamp=None, prep_template=None, parents=None, processing_parameters=None, can_be_submitted_to_ebi=False, can_be_submitted_to_vamps=False

qiita_db/artifact.py
antgonza: I think always using datetime.now() is the way to go.

qiita_db/artifact.py
antgonza: If that's the case, we should have all access to the DB in a TRN, right? Which I don't think. Note that the the idea is that you are validating that it doesn't have multiple studies, which will be a problem with or without the transaction. My comment is more oriented to have all validations in the same block of if/elifs

qiita_db/artifact.py
antgonza: Same reply than before. In other words, why do we need to have validations within a TRN. Note that the values are stored in variables so the TRN is not getting us anything.

qiita_db/artifact.py
antgonza: Rethinking this, I can see that this might make things more difficult in the long run and move_filepaths_to_upload_folder should only move the files ...

qiita_db/artifact.py

qiita_db/artifact.py

qiita_db/artifact.py
josenavas: Currently all accesses to the DB are in a TRN (you can search for my previous PR a few weeks ago about transactionizing everything). That's why through the code there are `with TRN:` even if it looks like is not doing any SQL call.

I can put all validations in the same block of if/elifs but then all the checks will be done inside the TRN, in order to make sure that all the DB accesses performed in the `create` method are done in a Transaction. I think as it is right now is better, but if others have strong feelings I can transactionize everything.

qiita_db/artifact.py
josenavas: No, they're not stored in variables - we don't anything in memory by the object design. Thus, doing the calls `cls(artifact_id)`, `instance.visibility` and `instance.children` are accessing the database.

If we don't put these in a transaction, we can get the race condition in which a children gets created between the check and the actual deletion.

josenavas:
@ElDeveloper able to do a review in here?

qiita_db/artifact.py
antgonza: OK

qiita_db/artifact.py
antgonza: ok

josenavas:
Anyone in @biocore/qiita-dev able to review this? Note that this is going to be the core object of Qiita, so having reviews here is good to avoid possible issues/conflicts with my next PRs

qiita_db/artifact.py
```diff
@@ -0,0 +1,638 @@
...
+    """
+    _table = "artifact"
+
+    @classmethod
+    def create(cls, filepaths, artifact_type, prep_template=None,
+               parents=None, processing_parameters=None,
+               can_be_submitted_to_ebi=False, can_be_submitted_to_vamps=False):
+        r"""Creates a new artifact in the system
+
+        The parameters depend based on how the artifact was generated:
```
ElDeveloper: depend based -> depend

qiita_db/artifact.py
```diff
@@ -0,0 +1,638 @@
...
+            `processing_parameters` should not be provided.
+            - If the artifact was generated by processing one or more
+            artifacts, the parameters `parents` and `processing_parameters`
+            should be provided and the parameter `prep_template` should not
+            be provided.
+
+        The parameters `can_be_submitted_to_ebi` and
+        `can_be_submiited_to_vamps` defaults to false and they should be
+        provided if and only if the artifact can be submitted to EBI and
+        VAMPS, respectivelly.
```
mortonjt: respectivelly -> respectively

qiita_db/artifact.py
```diff
@@ -0,0 +1,638 @@
...
+            - If the artifact was uploaded by the user, the parameter
+            `prep_template` should be provided and the parameters `parents` and
+            `processing_parameters` should not be provided.
+            - If the artifact was generated by processing one or more
+            artifacts, the parameters `parents` and `processing_parameters`
+            should be provided and the parameter `prep_template` should not
+            be provided.
+
+        The parameters `can_be_submitted_to_ebi` and
+        `can_be_submiited_to_vamps` defaults to false and they should be
```
ElDeveloper: submiited -> submited

qiita_db/artifact.py
ElDeveloper: respectivelly -> respectively

qiita_db/artifact.py
```diff
@@ -0,0 +1,638 @@
...
+            If both `parents` and `prep_template` are provided
+            If none of `parents` and `prep_template` are provided
+            If `parents` is provided but `processing_parameters` is not
+            If both `prep_template` and `processing_parameters` is provided
+            If not all the artifacts in `parents` belong to the same study
+
+        Notes
+        -----
+        The visibility of the artifact is set by default to `sandbox`
+        The timestamps of the artifact is set by default to `datetime.now()`
```
ElDeveloper: timestamps -> timestamp

qiita_db/artifact.py
```diff
@@ -0,0 +1,638 @@
...
+        """
+        # We need at least one file
```
ElDeveloper: :eyes: these comments are a bit redundant as its pretty much the same description as in the error message, but that's OK, just wanted to point this out.

qiita_db/artifact.py
```diff
@@ -0,0 +1,638 @@
...
+            sql = """SELECT data_type
+                     FROM qiita.artifact
+                        JOIN qiita.data_type USING (data_type_id)
+                     WHERE artifact_id = %s"""
+            TRN.add(sql, [self.id])
+            return TRN.execute_fetchlast()
+
+    @property
+    def can_be_submitted_to_ebi(self):
+        """Whether if the artifact can be submitted to EBI or not
```
ElDeveloper: Can you remove "if" from this description.

qiita_db/artifact.py
```diff
@@ -0,0 +1,638 @@
...
+            sql = """SELECT sample_id, ebi_run_accession
+                     FROM qiita.ebi_run_accession
+                     WHERE artifact_id = %s"""
+            TRN.add(sql, [self.id])
+            return {s_id: ebi_acc for s_id, ebi_acc in
+                    TRN.execute_fetchindex()}
+
+    @ebi_run_accessions.setter
+    def ebi_run_accessions(self, values):
+        """Set the ebi run accession attached to this artifact
```
ElDeveloper: ebi -> EBI

qiita_db/support_files/patches/33.sql
```diff
@@ -260,7 +265,7 @@ BEGIN
         VALUES ('Demultiplexed', 'Demultiplexed and QC sequeneces')
         RETURNING artifact_type_id INTO demux_type_id;
     INSERT INTO qiita.artifact_type (artifact_type, description)
-        VALUES ('BIOM table', 'Biom table')
+        VALUES ('BIOM', 'Biom table')
```
ElDeveloper: Is it too late to change Biom -> BIOM? entirely not blocking ... but you know :laughing:

qiita_db/test/test_artifact.py
```diff
@@ -0,0 +1,427 @@
...
+        close(fd)
+        with open(self.fp3, 'w') as f:
+            f.write(">1.sid_r4_0 M02034:17:000000000-A5U18:1:1101:15370:1394 "
+                    "1:N:0:1 orig_bc=CATGAGCT new_bc=CATGAGCT bc_diffs=0\n"
+                    "GTGTGCCAGCAGCCGCGGTAATACGTAGGG\n")
+        self.filepaths_processed = [(self.fp3, 4)]
+
+        # Generate some file for a BIOM
+        fd, self.fp4 = mkstemp(suffix='_table.biom')
+        t = Table(np.array([[1, 2], [3, 4]]), ['a', 'b'], ['x', 'y'])
```
ElDeveloper: Pro tip, you can import example_table from biom and you don't have to create a table :D i.e.

```python
from biom import example_table as et

print et
```

No need to change this, but I thought it would be useful to note.

qiita_db/util.py
```diff
@@ -670,6 +686,41 @@ def str_to_id(x):
...
+    """Retrieves the filepaths for the given object id
+
+    Parameters
+    ----------
+    obj_fp_table : str
+        The name of the table that links the object and the filepath
+    obj_id_column : str
+        The name of the column that represents the object id
+    obj_id : int
+        The object id
```
ElDeveloper: Can you add a Returns section?

qiita_db/artifact.py

qiita_db/artifact.py
josenavas: Good catch! I had to read twice to see what you meant xD

qiita_db/artifact.py
josenavas: Double shot!

qiita_db/artifact.py

qiita_db/artifact.py
josenavas: Yeah, sometimes I first write the comments so I don't forget any case and then add the code...

qiita_db/artifact.py

qiita_db/artifact.py

qiita_db/support_files/patches/33.sql
josenavas: Not a problem!

qiita_db/test/test_artifact.py
josenavas: I totally forgot about the example table! Thanks, I changed (less code to worry about here :D )

qiita_db/util.py
josenavas: Sure!

josenavas:
Thanks @ElDeveloper !!! Comments addressed

qiita_db/test/test_sql.py
```diff
@@ -27,5 +35,91 @@ def test_collection_job_trigger(self):
...
+    def test_find_artifact_roots_is_child_multiple_parents_multiple_root(self):
+        """Correctly returns the roots if the children has multiple roots"""
+        sql = "SELECT * FROM qiita.find_artifact_roots(%s)"
+
+        new_root = self._create_root_artifact()
+
+        # Add a child of 2 roots
+        fd, fp = mkstemp(suffix='_seqs.fna')
+        close(fd)
+        with open(fp, 'w') as f:
```
ElDeveloper: Should these files be deleted at some point? Or are these being cleaned up after they are moved to their final destination?

ElDeveloper:
Thanks @josenavas, just one more comment.

qiita_db/test/test_sql.py
josenavas: Good catch done

ElDeveloper: Thanks!!

PR #1528, opened 2015-11-09
Remove data.py and test_data.py

josenavas:
These files are no longer needed because they've been replaced by the artifact object.

~~This depends in #1527 so review/merge that one first to remove the commits from the history here.~~


josenavas:
Ready for review!

antgonza:
:+1: Can someone review/merge?

PR #1529, opened 2015-11-11
`from qiita_db.XXX import YYY` -> `import qiita_db as qdb`

josenavas:
This PR depends on #1528 so review/merge that one first.

Following our discussion by email, the attributes of the objects are going to return instances of other objects rather than their IDs. In order to do that, we have to avoid circular imports by using `import qiita_db`.

This PR modifies the entire qiita_db code base to use `import qiita_db as qdb` rather than `from qiita_db.XXX import YYY`.

The `user.py` and `test_user.py` files have been already modified so the user object returns instances of other objects. The tests for the user object are passing again.


josenavas:
This is ready for review! The amount of changes here is large but there are really simple (mainly changing the imports).

The code that needs more attention is the user code, that has been fixed to return objects rather than ids.

qiita_db/metadata_template/__init__.py
```diff
@@ -6,15 +6,23 @@
...
 
-from .sample_template import SampleTemplate
-from .prep_template import PrepTemplate
-from .util import load_template_to_dataframe, looks_like_qiime_mapping_file
-from .constants import (TARGET_GENE_DATA_TYPES, SAMPLE_TEMPLATE_COLUMNS,
-                        PREP_TEMPLATE_COLUMNS,
-                        PREP_TEMPLATE_COLUMNS_TARGET_GENE, CONTROLLED_COLS)
+import constants
+import util
+# import base_metadata_template
```
antgonza: This is not needed.

qiita_db/metadata_template/__init__.py
```diff
@@ -6,15 +6,23 @@
...
+import sample_template
+import prep_template
 
+__all__ = ["sample_template", "prep_template", "util", "constants"]
 
-__all__ = ['SampleTemplate', 'PrepTemplate', 'load_template_to_dataframe',
-           'TARGET_GENE_DATA_TYPES', 'SAMPLE_TEMPLATE_COLUMNS',
-           'PREP_TEMPLATE_COLUMNS', 'PREP_TEMPLATE_COLUMNS_TARGET_GENE',
-           'CONTROLLED_COLS', 'looks_like_qiime_mapping_file']
+# from .sample_template import SampleTemplate
```
antgonza: Same here.

qiita_db/base.py
```diff
@@ -214,6 +212,10 @@ def __ne__(self, other):
         r"""Self and other are not equal based on type and database id"""
         return not self.__eq__(other)
 
+    def __hash__(self):
```
antgonza: Why do you need this as part of tthis pr? Does it need a test?

antgonza:
a few comments, nothing mayor, fine to merge after their solved :+1:

qiita_db/analysis.py
```diff
@@ -200,39 +198,39 @@ def delete(cls, _id):
         QiitaDBUnknownIDError
             If the analysis id doesn't exist
         """
-        with TRN:
+        with qdb.sql_connection.TRN:
```
ElDeveloper: I suspect you may have been able to get away by doing `with qdb.sql_connection.TRN as TRN:`.

ElDeveloper:
:+1:

qiita_db/metadata_template/__init__.py
josenavas: Done

qiita_db/metadata_template/__init__.py
josenavas: Done

qiita_db/base.py
josenavas: The hash is useful for doing sets and/or dictionaries. In the following example:
```python
s1 = Study(1)
s2 = Study(1)
s = set(s1, s2)
len(s)
```
w/o the hash function `len(s) = 2`. However, with this hash function `len(s) = 1` which is correct given that s1 and s2 represents the same study. This function was not needed until now because we where moving ids around, but since we are using the objects now, we need to hash correctly these objects.

qiita_db/analysis.py
josenavas: You're probably correct, but I did a search/replace. Do you think it is needed to change or it is ok as it is?

ElDeveloper: No, it is not worth it to re-do the changes, unless someone thinks
otherwise, in any case lets not consider this blocking.

antgonza: I think it's fine as it is ...

PR #1531, opened 2015-11-12
Fixing a bunch of tests

josenavas:
This PR depends on #1530 so review/merge that one first.

Fixes the tests for the artifact, base, job, logger, ontology, parameters, reference, sql_connection objects. It also fixes test_setup.py. 

It should be fast to review as there where almost no changes required to the objects themselves. Most of the changes are related to the import changes in the test files.

PR #1532, opened 2015-11-12
Fixing more tests

josenavas:
This PR depends on #1531, so review/merge that one first.

This PR fixes the tests for software, portal, test_sql.py and util.

The software and portal objects have been modified to return instances of other objects rather than the ids.

### DIFF
--- a/qiita_db/portal.py
+++ b/qiita_db/portal.py
@@ -330,7 +330,7 @@ class Portal(qdb.base.QiitaObject):
         Returns
         -------
         set of qiita_db.analysis.Analysis
-            All analysis belonging to the portal
+            All analyses belonging to the portal
         """
         with qdb.sql_connection.TRN:
             sql = """SELECT analysis_id
--- a/qiita_db/portal.py
+++ b/qiita_db/portal.py
@@ -198,18 +198,21 @@ class Portal(qdb.base.QiitaObject):
             return True
 
     def get_studies(self):
-        """Returns study id for all Studies belonging to the portal
+        """Returns all studies belonging to the portal
 
         Returns
         -------
-        set of int
-            All study ids in the database that match the given portal
+        set of qiita_db.study.Study
+            All studies attached to the portal
         """
         with qdb.sql_connection.TRN:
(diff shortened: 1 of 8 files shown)


antgonza:
@josenavas obviously we need to wait to review the other PRs before we review this but looking at the errors in the build I think there is something not right ... could you take a look? My concern is that the DB is not being created.

antgonza:
:+1:

qiita_db/portal.py
```diff
@@ -322,19 +325,21 @@ def remove_studies(self, studies):
...
     def get_analyses(self):
-        """Returns analysis id for all Analyses belonging to a portal
+        """Returns all analyses belonging to a portal
 
         Returns
         -------
-        set of int
-            All analysis ids in the database that match the given portal
+        set of qiita_db.analysis.Analysis
+            All analysis belonging to the portal
```
ElDeveloper: analysis -> analyses

qiita_db/test/test_sql.py
```diff
@@ -69,8 +66,8 @@ def test_find_artifact_roots_is_child_multiple_parents_one_root(self):
         with open(fp, 'w') as f:
             f.write("test")
         fp = [(fp, 7)]
-        new = Artifact.create(fp, "BIOM", parents=[Artifact(2), Artifact(3)],
-                              processing_parameters=Parameters(1, Command(1)),
+        new = qdb.artifact.Artifact.create(fp, "BIOM", parents=[qdb.artifact.Artifact(2), qdb.artifact.Artifact(3)],
```
ElDeveloper: These lines and the next few changes, look way over the 80 column limit.

ElDeveloper:
:+1: Just two small comments.

qiita_db/portal.py
josenavas: Done

qiita_db/test/test_sql.py
josenavas: And you're correct - for some reason my editor was not yelling at me :D

ElDeveloper:
Thanks @josenavas, let's wait for the tests, but this looks good.

PR #1535, opened 2015-11-18
Artifact qiita_db fixes

josenavas:
This PR depends on https://github.com/biocore/qiita/pull/1532 so review/merge that one first.

This PR should fix all the issues in the Qiita DB submodule.

### DIFF
--- a/qiita_db/search.py
+++ b/qiita_db/search.py
@@ -199,9 +199,9 @@ class QiitaStudySearch(object):
             study_ids = set(qdb.sql_connection.TRN.execute_fetchflatten())
             # strip to only studies user has access to
             if user.level not in {'admin', 'dev', 'superuser'}:
-                study_ids = study_ids.intersection(
-                    qdb.study.Study.get_by_status('public') |
-                    user.user_studies | user.shared_studies)
+                studies = qdb.study.Study.get_by_status('public') | \
+                    user.user_studies | user.shared_studies
+                study_ids = study_ids.intersection([s.id for s in studies])
 
             results = {}
             # run search on each study to get out the matching samples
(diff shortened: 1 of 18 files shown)


antgonza:
I guess this should have fixed the errors but the tests are failing in the same place ...

antgonza:
:+1:

qiita_db/search.py
```diff
@@ -396,16 +401,20 @@ def filter_by_processed_data(self, datatypes=None):
...
-                    proc_data = qdb.data.ProcessedData(proc_data_id)
-                    datatype = proc_data.data_type()
+                for artifact in study.artifacts():
+                    if artifact.artifact_type != 'BIOM':
+                        continue
+                    datatype = artifact.data_type
                     # skip processed data if it doesn't fit the given datatypes
                     if datatypes is not None and datatype not in datatypes:
                         continue
-                    filter_samps = proc_data.samples.intersection(
```
ElDeveloper: Is the `if datatypes is not None ...` needed at all?

ElDeveloper:
Looks good, though I'm not exactly sure why some command line interfaces were deleted altogether, if this is because everything is now an artifact, then it makes sense, otherwise, I'm not sure I quite understand.

:+1:

qiita_db/search.py
josenavas: Yup in order to avoid this issue:
```python
In [1]: a = None

In [2]: '1' in a
TypeError                                 Traceback (most recent call last)
<ipython-input-2-096bed3860b1> in <module>()
----> 1 '1' in a

TypeError: argument of type 'NoneType' is not iterable
```

josenavas:
That is correct @ElDeveloper everything is now an artifact, so there is no point on having those other commands.

PR #1538, opened 2015-11-19
fixing EBI modify submission

antgonza:
The AGP modify submission was failing because the code was trying to use all the samples in the sample template vs. only the samples in that prep template that were successful previously. Additionally, the reply of the curl command was being written over the original curl_reply.xml file. 

Here we are fixing both issues. However, we still can't test modify against EBI cause they are still trying to figure out how to allow this.


josenavas:
:+1:

mortonjt:
Looks good to me.  What's left in this?

antgonza:
Nothing, really, just merge.

PR #1537, opened 2015-11-18
Artifact fix templates

josenavas:
Depends on https://github.com/biocore/qiita/pull/1535

Fixes the metadata template tests.

PR #1540, opened 2015-11-23
WIP: Fixing EBI code

josenavas:
Depends on #1537 so review/merge that one first

This fixes the EBI code so we can do some testing with the plugin system (right now we can't even start the webserver)


qiita_ware/ebi.py
```diff
@@ -105,14 +105,28 @@ def __init__(self, preprocessed_data_id, action):
...
+        if not self.artifact.can_be_submitted_to_ebi:
+            error_msg = ("Artifact %d cannot be submitted to EBI"
+                         % self.artifact.id)
+            LogEntry.create('Runtime', error_msg)
+            raise EBISubmissionError(error_msg)
+
+        self.study = self.artifact.study
+        self.sample_template = self.study.sample_template
+        prep_templates = self.artifact.prep_templates
+        if len(prep_templates) > 1:
```
antgonza: After discussing offline with @josenavas we agreed that this is a redundant verification as can_be_submitted_to_ebi is set to False for those cases where prep_template>1 during creation.

antgonza:
One comment, then :+1:

qiita_db/artifact.py
```diff
@@ -450,6 +450,31 @@ def can_be_submitted_to_ebi(self):
...
     @property
+    def is_submitted_to_ebi(self):
+        """Whether the artifact has been submitted to EBI or not
+
+        Returns
+        -------
+        bool
+            True if the artifact has been submitted to EBI. False otherwise
+
+        raises
```
mortonjt: raises -> Raises ?

qiita_ware/test/test_ebi.py
```diff
@@ -407,11 +408,11 @@ def generate_new_study_with_preprocessed_data(self):
         with File(demux_fp, 'w') as f:
             to_hdf5(fna_fp, f)
 
-        ppd = PreprocessedData.create(
-            study, "preprocessed_sequence_illumina_params", 1,
-            [(demux_fp, 6)], pt)
+        artifact = Artifact.create(
+            [(demux_fp, 6)], "Demultiplexed", prep_template=pt,
```
mortonjt: Would it be worthwhile to add a comment on where this magic number comes from?

qiita_db/test/test_artifact.py
```diff
@@ -327,6 +327,16 @@ def test_can_be_submitted_to_ebi(self):
...
 
+    def test_is_submitted_to_ebi(self):
+        with self.assertRaises(
+                qdb.exceptions.QiitaDBOperationNotPermittedError):
+            qdb.artifact.Artifact(1).is_submitted_to_ebi
+        self.assertTrue(qdb.artifact.Artifact(2).is_submitted_to_ebi)
+        self.assertFalse(qdb.artifact.Artifact(3).is_submitted_to_ebi)
+        with self.assertRaises(
+                qdb.exceptions.QiitaDBOperationNotPermittedError):
+            qdb.artifact.Artifact(4).is_submitted_to_ebi
```
mortonjt: Completely optional - I personally like to create separate tests for assertRaises and other checks.

qiita_ware/ebi.py
```diff
@@ -105,14 +104,20 @@ def __init__(self, preprocessed_data_id, action):
...
+        self.artifact = Artifact(artifact_id)
+        if not self.artifact.can_be_submitted_to_ebi:
+            error_msg = ("Artifact %d cannot be submitted to EBI"
+                         % self.artifact.id)
+            LogEntry.create('Runtime', error_msg)
+            raise EBISubmissionError(error_msg)
+
+        self.study = self.artifact.study
+        self.sample_template = self.study.sample_template
+        self.prep_template = self.artifact.prep_templates[0]
```
mortonjt: Is there a reason that only the first prep template is chosen? 

Also, would it be worthwhile to have a check to make sure that `len(self.artifact.prep_templates[0]) > 0`? (Could make future debugging easier)

qiita_ware/ebi.py
```diff
@@ -922,13 +927,13 @@ def generate_demultiplexed_fastq(self, rewrite_fastq=False, mtime=None):
...
-        ppd = self.preprocessed_data
+        ar = self.artifact
 
         dir_not_exists = not isdir(self.full_ebi_dir)
         if dir_not_exists or rewrite_fastq:
             makedirs(self.full_ebi_dir)
 
-            demux = [path for _, path, ftype in ppd.get_filepaths()
+            demux = [path for _, path, ftype in ar.get_filepaths()
                      if ftype == 'preprocessed_demux'][0]
```
mortonjt: Any reason why only the first `demux` path is chosen? (A comment would be nice).

mortonjt:
A few anal comments.  Otherwise looks good!

qiita_db/artifact.py

qiita_db/test/test_artifact.py
josenavas: Done

qiita_ware/ebi.py
josenavas: Prep templates will be always at least 1. In this specific case, it will be also at most 1. The reason is that if len(prep_templates) > 1 (by design it can't be 0) the property can_be_submitted_to_ebi is False so it will fail on line 108.

qiita_ware/ebi.py
josenavas: There should be only one.

qiita_ware/ebi.py
antgonza: adding a comment to the code will be nice.

qiita_ware/test/test_ebi.py
josenavas: Done

qiita_ware/ebi.py
josenavas: Done

PR #1541, opened 2015-11-23
Plugin remove some data.py imports so we can move forward

josenavas:
Depends on #1540 so review/merge that one first

This PR fixes the submit to VAMPS code and removes all the imports from qiita_db/data.py so we can actually start the webserver and move forward with some tests


antgonza:
This is a simple PR removing imports, nothing major happening. Thus, 1 reviewer should suffice.

PR #1542, opened 2015-11-23
Plugin: Processing job

josenavas:
Depends on https://github.com/biocore/qiita/pull/1541 so review/merge that one first

This PR adds the new object ProcessingJob, both to the DB and qiita_db, which will allow the plugin to keep track of the jobs that are being executed.

### DIFF
--- a/qiita_db/processing_job.py
+++ b/qiita_db/processing_job.py
@@ -1,0 +1,307 @@
+# -----------------------------------------------------------------------------
+# Copyright (c) 2014--, The Qiita Development Team.
+#
+# Distributed under the terms of the BSD 3-clause License.
+#
+# The full license is in the file LICENSE, distributed with this software.
+# -----------------------------------------------------------------------------
+
+from uuid import uuid4, UUID
+
+import qiita_db as qdb
+
+
+class ProcessingJob(qdb.base.QiitaObject):
+    r"""A job sent for processing an artifact
+
+    Attributes
+    ----------
+    user
+    command
+    parameters
+    status
+    log
+    heartbeat
+    step
+
+    Methods
+    -------
+    exists
+    create
+    """
+    _table = 'processing_job'
+
+    @classmethod
+    def exists(cls, job_id):
+        """Check if the job `job_id` exists
+
+        Returns
+        -------
+        bool
+            True if thejob `job_id` exists. False otherwise.
+        """
+        try:
+            UUID(job_id)
+        except ValueError:
+            return False
+
+        with qdb.sql_connection.TRN:
+            sql = """SELECT EXISTS(SELECT *
+                                   FROM qiita.processing_job
+                                   WHERE processing_job_id = %s)"""
+            qdb.sql_connection.TRN.add(sql, [job_id])
+            return qdb.sql_connection.TRN.execute_fetchlast()
+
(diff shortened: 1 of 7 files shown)


qiita_db/processing_job.py
```diff
@@ -0,0 +1,307 @@
...
+# The full license is in the file LICENSE, distributed with this software.
+# -----------------------------------------------------------------------------
+
+from uuid import uuid4, UUID
+
+import qiita_db as qdb
+
+
+class ProcessingJob(qdb.base.QiitaObject):
+    r"""A job sent for processing an artifact
```
antgonza: Can you rewrite?

qiita_db/processing_job.py
```diff
@@ -0,0 +1,307 @@
...
+    _table = 'processing_job'
+
+    @classmethod
+    def exists(cls, job_id):
+        """Check if the job `job_id` exists
+
+        Returns
+        -------
+        bool
+            True if thejob `job_id` exists. False otherwise.
```
antgonza: space between the/job

qiita_db/processing_job.py
```diff
@@ -0,0 +1,307 @@
...
+    -------
+    exists
+    create
+    """
+    _table = 'processing_job'
+
+    @classmethod
+    def exists(cls, job_id):
+        """Check if the job `job_id` exists
+
```
mortonjt: No need for `Parameters` in the docstring?

qiita_db/processing_job.py
```diff
@@ -0,0 +1,307 @@
...
+        qiita_db.software.Parameters
+            The parameters used in the job's command
+        """
+        with qdb.sql_connection.TRN:
+            sql = """SELECT command_id, command_parameters_id
+                     FROM qiita.processing_job
+                     WHERE processing_job_id = %s"""
+            qdb.sql_connection.TRN.add(sql, [self.id])
+            res = qdb.sql_connection.TRN.execute_fetchindex()[0]
+            return qdb.software.Parameters(
```
antgonza: Do we really need to instantiate Command and then Parameters? Wouldn't be easier to simply query that directly in the DB?

antgonza:
A few comments.

mortonjt:
:+1: once the changes are made

qiita_db/processing_job.py
josenavas: Done

qiita_db/processing_job.py
josenavas: Oooops, forgot to add. Done

qiita_db/processing_job.py
josenavas: Done

qiita_db/processing_job.py
josenavas: In order to instantiate the Parameters we need a commands object. The problem is that is not sufficient to know the parameter id, we also need to know to which command this parameter set belongs to.

i.e. Parameter id 1 can be from command 1, command 2, etc...

I think that my next PR will simplify this, as we will not have a table for each set of parameters as we have right now - I had it this way because this is how it would work with a table per command

antgonza: ok

PR #1543, opened 2015-11-23
Plugin general fixes ebi

antgonza:
This fixes all EBI tests. Built on top of https://github.com/biocore/qiita/pull/1540

### DIFF
--- a/qiita_ware/ebi.py
+++ b/qiita_ware/ebi.py
@@ -156,7 +156,7 @@ class EBISubmission(object):
         self.experiment_xml_fp = None
         self.run_xml_fp = None
         self.submission_xml_fp = None
-        self.pmids = self.study.pmids
+        self.publications = self.study.publications
 
         # getting the restrictions
         st_missing = self.sample_template.check_restrictions(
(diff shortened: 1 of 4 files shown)


qiita_ware/ebi.py
```diff
@@ -161,7 +161,7 @@ def __init__(self, artifact_id, action):
         self.experiment_xml_fp = None
         self.run_xml_fp = None
         self.submission_xml_fp = None
-        self.pmids = self.study.pmids
+        self.publications = self.study.publications
```
mortonjt: Redundant?

antgonza: This is a good point. It's not really redundant as self.publications will be a list of the publications and self.study.publications is a query to the DB to retrieve the publications. I could remove but I like the idea that in the init we make sure that everything that is possible looks sane.

josenavas:
:+1:

qiita_ware/ebi.py
mortonjt: Ok. That's fine.  Just wanted to make sure thats on the radar.

PR #1544, opened 2015-11-24
Plugin: revamping processing parameters

josenavas:
We were creating a table for each command that we had in the system, which doesn't make the system too flexible, and that strategy creates a lot of dynamic tables.

With the approach implemented here, we only store some information about the parameters that each command can receive, and we store in a single table the default parameter sets (so we can limit the parameters available in the system - i.e. users can't provide arbitrary parameters if we don't want to). However, this makes extremely easy to allow users to provide their own parameters, since the actual parameters used to process an artifact are stored in the artifact themselves using JSON.

Furthermore, the previous implementation did not cover the case in which we had multiple artifact as input and in which parameters they should be passed.

This also addressed the concern that @ElDeveloper had about positional arguments. Using this structure, Qiita will always see named parameters that can provide to the plugin, and the plugin will know what to do with them (i.e. if they're positional, it will create the CLI call correctly). I think this is the best trade-off that we can get between system generality (i.e. Qiita) while still being able to support any type of tool on the plugins (since they're specialized for those tools).

Another advantage of this is that we no longer need to hack various things in the parameters objects themselves due to Qiime.

The changes in this PR are already big enough that I will remove the qiita_db/parameters.py in another PR, so I can follow the same approach as with the data module: remove the module, remove the tests and remove the imports so the webserver starts.

### DIFF
--- a/qiita_db/software.py
+++ b/qiita_db/software.py
@@ -106,7 +106,7 @@ class Command(qdb.base.QiitaObject):
             The description of the parameters that this command received. The
             format is: {parameter_name: (paramter_type, default)},
             where parameter_name, paramter_type and default are strings. If
-            default is None,
+            default is None.
 
         Returns
         -------
@@ -116,6 +116,7 @@ class Command(qdb.base.QiitaObject):
         Raises
         ------
         QiitaDBError
+            - If parameters is empty
             - If the parameters dictionary is malformed
             - If one of the parameter types is not supported
             - If the default value of a choice parameter is not listed in
@@ -131,6 +132,10 @@ class Command(qdb.base.QiitaObject):
         overwrite it.
         """
         # Perform some sanity checks in the parameters dictionary
+        if not parameters:
+            raise qdb.exceptions.QiitaDBError(
+                "Error creating command %s. At least one parameter should "
+                "be provided." % name)
         sql_param_values = []
         for pname, vals in parameters.items():
             if len(vals) != 2:
@@ -227,7 +232,8 @@ class Command(qdb.base.QiitaObject):
 
         Returns
         -------
-        dict of {parameter_name: [ptype, dflt]}
+        dict
+            Dictionary of {parameter_name: [ptype, dflt]}
         """
         with qdb.sql_connection.TRN:
             sql = """SELECT parameter_name, parameter_type, default_value
@@ -243,7 +249,8 @@ class Command(qdb.base.QiitaObject):
 
         Returns
         -------
-        dict of {parameter_name: ptype}
+        dict
+            Dictionary of {parameter_name: ptype}
         """
         with qdb.sql_connection.TRN:
             sql = """SELECT parameter_name, parameter_type
@@ -259,7 +266,8 @@ class Command(qdb.base.QiitaObject):
 
         Returns
         -------
-        dict of {parameter_name: [ptype, default]}
+        dict
+            Dictionary of {parameter_name: [ptype, default]}
         """
         with qdb.sql_connection.TRN:
             sql = """SELECT parameter_name, parameter_type, default_value
@@ -275,7 +283,8 @@ class Command(qdb.base.QiitaObject):
 
         Returns
         -------
-        generator of qiita_db.software.DefaultParameters
+        generator
+            generator of qiita_db.software.DefaultParameters
         """
         with qdb.sql_connection.TRN:
             sql = """SELECT default_parameter_set_id
@@ -477,7 +486,7 @@ class DefaultParameters(qdb.base.QiitaObject):
         ----------
         command : qiita_db.software.Command
             The command to which the parameter set belongs to
-        kwargs : dict
+        kwargs : dict of {str: str}
             The parameters and their values
 
         Returns
@@ -489,8 +498,8 @@ class DefaultParameters(qdb.base.QiitaObject):
         ------
         qiita_db.exceptions.QiitaDBError
             - If there are missing parameters for the given command
-            - If there are extra parameters in `kwargs` than for the given
-              command
+            - If `kwargs` contains parameters not originally defined in the
+            command
         """
         with qdb.sql_connection.TRN:
             command_params = set(command.optional_parameters)
@@ -639,8 +648,8 @@ class Parameters(object):
             The command to which the parameter set belongs to
         json_str : str, optional
             The json string encoding the parameters
-        values_dict : dict of {str: object}, optinal
-            The dictionary witht the parameter values
+        values_dict : dict of {str: object}, optional
+            The dictionary with the parameter values
 
         Returns
         -------
@@ -652,7 +661,7 @@ class Parameters(object):
         qiita_db.exceptions.QiitaDBError
             - If `json_str` and `values` are both provided
             - If neither `json_str` or `values` are provided
-            - If the provided JSON or values do not encode a parameter set of
+            - If `json_str` or `values` do not encode a parameter set of
             the provided command.
 
         Notes
@@ -707,7 +716,7 @@ class Parameters(object):
 
     @classmethod
     def from_default_params(cls, dflt_params, req_params, opt_params=None):
-        """Creates the parameter set from a dflt_parameter set
+        """Creates the parameter set from a `dflt_params` set
 
         Parameters
         ----------
@@ -717,7 +726,7 @@ class Parameters(object):
             The required parameters values, keyed by parameter name
         opt_params : dict of {str: object}, optional
             The optional parameters to change from the default set, keyed by
-            parameter name. Default: None, user the values in `dflt_params`
+            parameter name. Default: None, use the values in `dflt_params`
 
         Raises
         ------
@@ -757,13 +766,13 @@ class Parameters(object):
     def __init__(self, values, command):
         # Time for some python magic! The __init__ function should not be used
         # outside of this module, users should always be using one of the above
-        # classmethods to instantiate the object. Less test that it is the case
+        # classmethods to instantiate the object. Lets test that it is the case
         # First, we are going to get the current frame (i.e. this __init__)
         # function and the caller to the __init__
         current_frame = inspect.currentframe()
         caller_frame = current_frame.f_back
-        # The file names where the function is defines is stored in the
-        # f_code.co_filename attribute,and in this case it has to be the same
+        # The file names where the function is defined is stored in the
+        # f_code.co_filename attribute, and in this case it has to be the same
         # for both of them. Also, we are restricing that the name of the caller
         # should be either `load` or `from_default_params`, which are the two
         # classmethods defined above
@@ -782,12 +791,12 @@ class Parameters(object):
 
     @property
     def command(self):
-        """The command to which this parametr set belongs to
+        """The command to which this parameter set belongs to
 
         Returns
         -------
         qiita_db.software.Command
-            The command to which this parametr set belongs to
+            The command to which this parameter set belongs to
         """
         return self._command
 
@@ -808,6 +817,6 @@ class Parameters(object):
         Returns
         -------
         str
-            The parameter values as a json string
+            The parameter values as a JSON string
         """
         return dumps(self._values, sort_keys=True)
--- a/qiita_db/software.py
+++ b/qiita_db/software.py
@@ -104,8 +104,8 @@ class Command(qdb.base.QiitaObject):
             The description of the command
         parameters : dict
             The description of the parameters that this command received. The
-            format is: {parameter_name: (paramter_type, default)},
-            where parameter_name, paramter_type and default are strings. If
+            format is: {parameter_name: (parameter_type, default)},
+            where parameter_name, parameter_type and default are strings. If
             default is None.
 
         Returns
@@ -141,7 +141,7 @@ class Command(qdb.base.QiitaObject):
             if len(vals) != 2:
                 raise qdb.exceptions.QiitaDBError(
                     "Malformed parameters dictionary, the format should be "
-                    "{param_name: [paramter_type, default]}. Found: "
+                    "{param_name: [parameter_type, default]}. Found: "
                     "%s for parameter name %s" % (vals, pname))
 
             ptype, dflt = vals
@@ -629,7 +629,7 @@ class Parameters(object):
     """
 
     def __eq__(self, other):
-        """Equality based on the paramter values and the command"""
+        """Equality based on the parameter values and the command"""
         if type(self) != type(other):
             return False
         if self.command != other.command:
@@ -642,8 +642,8 @@ class Parameters(object):
     def load(cls, command, json_str=None, values_dict=None):
         """Load the parameters set form a json str or from a dict of values
 
-        Paramters
-        ---------
+        Parameters
+        ----------
         command : qiita_db.software.Command
             The command to which the parameter set belongs to
         json_str : str, optional
@@ -745,7 +745,7 @@ class Parameters(object):
                 raise qdb.exceptions.QiitaDBError(
                     "Provided required parameters not expected.\n"
                     "Missing required parameters: %s\n"
-                    "Extra required paramters: %s\n"
+                    "Extra required parameters: %s\n"
                     % (', '.join(missing_reqd), ', '.join(extra_reqd)))
 
             if opt_params:
--- a/qiita_db/software.py
+++ b/qiita_db/software.py
@@ -8,6 +8,7 @@
 
 from json import dumps, loads
 from copy import deepcopy
+import inspect
 
 import qiita_db as qdb
 
@@ -93,11 +94,6 @@ class Command(qdb.base.QiitaObject):
             in which json-dump-of-list is the JSON dump of a list containing
             the acceptable values
 
-        If the default value for a parameter is NULL, then the parameter will
-        be required. On the other hand, if it is provided, the parameter will
-        be optional and the default value will be used when the user doesn't
-        overwrite it.
-
         Parameters
         ----------
         software : qiita_db.software.Software
@@ -126,6 +122,13 @@ class Command(qdb.base.QiitaObject):
             the available choices
         QiitaDBDuplicateError
             - If the command already exists
+
+        Notes
+        -----
+        If the default value for a parameter is NULL, then the parameter will
+        be required. On the other hand, if it is provided, the parameter will
+        be optional and the default value will be used when the user doesn't
+        overwrite it.
         """
         # Perform some sanity checks in the parameters dictionary
         sql_param_values = []
@@ -608,9 +611,12 @@ class DefaultParameters(qdb.base.QiitaObject):
 class Parameters(object):
     """Represents an instance of parameters used to process an artifact
 
-    This class should never be isntantiated using the __init__ method. The
-    recommended procedure is using of the classmethods 'load' or
-    'from_default_params'
+    Raises
+    ------
+    qiita_db.exceptions.QiitaDBOperationNotPermittedError
+        If trying to instantiate this class directly. In order to instantiate
+        this class, the classmethods `load` or `from_default_params` should
+        be used.
     """
 
     def __eq__(self, other):
@@ -626,10 +632,6 @@ class Parameters(object):
     @classmethod
     def load(cls, command, json_str=None, values_dict=None):
         """Load the parameters set form a json str or from a dict of values
-
-        The parameters `json_str` and `values_dict` are mutually exclusive,
-        only one of them should be provided at a time. However, one of them
-        should always be provided.
 
         Paramters
         ---------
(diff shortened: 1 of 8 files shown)


qiita_db/software.py
```diff
@@ -70,14 +73,31 @@ def exists(cls, software, cli_cmd):
...
+            - float: the parameter is a float
+            - artifact: the parameter is an artifact instance, the artifact id
+            will be stored
+            - reference: the parameter is a reference instance, the reference
+            id will be stored
+            - choice: the format of this should be `choice:<json-dump-of-list>`
+            in which json-dump-of-list is the JSON dump of a list containing
+            the acceptable values
+
+        If the default value for a parameter is NULL, then the parameter will
```
antgonza: Should this be a note?

qiita_db/software.py
```diff
@@ -515,18 +578,15 @@ def values(self):
...
+            sql = """SELECT parameter_set
+                     FROM qiita.default_parameter_set
+                     WHERE default_parameter_set_id = %s"""
             qdb.sql_connection.TRN.add(sql, [self.id])
-            # There should be only one row
-            result = dict(qdb.sql_connection.TRN.execute_fetchindex()[0])
-            del result["parameters_id"]
-            del result["param_set_name"]
-            return result
+            return qdb.sql_connection.TRN.execute_fetchlast()
```
antgonza: Should this be fetchall?

qiita_db/software.py
```diff
@@ -537,29 +597,187 @@ def command(self):
...
+            qdb.sql_connection.TRN.add(sql, [self.id])
+            return Command(qdb.sql_connection.TRN.execute_fetchlast())
 
-    def to_str(self):
-        """Generates a string with the parameter values
+
+class Parameters(object):
+    """Represents an instance of parameters used to process an artifact
+
+    This class should never be isntantiated using the __init__ method. The
```
antgonza: instantiated

antgonza: could you add an error/warning for the __init__?

qiita_db/software.py
```diff
@@ -537,29 +597,187 @@ def command(self):
...
+            return False
+        if self.values != other.values:
+            return False
+        return True
+
+    @classmethod
+    def load(cls, command, json_str=None, values_dict=None):
+        """Load the parameters set form a json str or from a dict of values
+
+        The parameters `json_str` and `values_dict` are mutually exclusive,
```
antgonza: This should be in the Notes, right?

qiita_db/support_files/patches/33.sql
```diff
@@ -38,13 +45,39 @@ CREATE TABLE qiita.software_command (
...
+	parameter_type       varchar   NOT NULL,
+	required             bool      NOT NULL,
+	default_value        varchar  ,
+	CONSTRAINT idx_command_parameter_0 PRIMARY KEY ( command_id, parameter_name )
+ ) ;
+CREATE INDEX idx_command_parameter ON qiita.command_parameter ( command_id ) ;
+ALTER TABLE qiita.command_parameter ADD CONSTRAINT fk_command_parameter FOREIGN KEY ( command_id ) REFERENCES qiita.software_command( command_id )    ;
+
+-- default_parameter_set tables - holds the default parameter sets defined by
+-- the system. If no arbitrary parameters is allowed in the system only the
```
antgonza: is -> are

qiita_db/software.py
antgonza: ... and a test?

qiita_db/software.py
antgonza: ... and add a test?

antgonza:
Some minor comments but it will be good to modify qiita and qiita-env so the tests can run.

antgonza:
Got it, https://github.com/biocore/qiita/pull/1545 makes the tests pass.

qiita_db/software.py
josenavas: Moved to the `Notes` section.

qiita_db/software.py
josenavas: Nope, it is only retrieving a single value from the database, `parameter_set` which is of JSON type and psycopg2 is so awesome that transforms it to a dict directly :boom: Also there is already a test for it.

antgonza: Got it, thanks!

qiita_db/software.py
josenavas: I don't know how to do it. The main issue is that the `classmethods` are actually instantiating the object at the end, so those will raise the warning, although the user will be using the correct methods. Do you have any idea on how to do it?

qiita_db/software.py

qiita_db/support_files/patches/33.sql
josenavas: Done

qiita_db/software.py
josenavas: Thanks for the link to the `inspect` usage - we could do it using that route. For others that want to know more about it, here is the link that @antgonza sent me http://blog.lerner.co.il/making-init-methods-magical-with-autoinit/ and here is the documentation https://docs.python.org/2/library/inspect.html

antgonza:
Thanks! :+1:

qiita_db/software.py
```diff
@@ -86,32 +102,91 @@ def create(cls, software, name, description, cli_cmd, parameters_table):
...
-        cli_cmd : str
-            The CLI used to call this command
-        parameters_table : str
-            The name of the table in which the parameters of the commands are
-            stored
+        parameters : dict
+            The description of the parameters that this command received. The
+            format is: {parameter_name: (paramter_type, default)},
+            where parameter_name, paramter_type and default are strings. If
+            default is None,
```
ElDeveloper: Can you change the comma to a period?

qiita_db/software.py
```diff
@@ -86,32 +102,91 @@ def create(cls, software, name, description, cli_cmd, parameters_table):
...
+        for pname, vals in parameters.items():
+            if len(vals) != 2:
+                raise qdb.exceptions.QiitaDBError(
+                    "Malformed parameters dictionary, the format should be "
+                    "{param_name: [paramter_type, default]}. Found: "
+                    "%s for parameter name %s" % (vals, pname))
+
+            ptype, dflt = vals
+            # Check that the type is one of the supported types
+            supported_types = ['string', 'integer', 'float', 'artifact',
```
ElDeveloper: Do we really need to have a distinction between integer, float and string? For all terms an purposes that I can foresee we won't be using the type information in any way. Out of curiosity [I googled](http://stackoverflow.com/a/12722107/379593), and it seems like bash does not even support floats.

qiita_db/software.py
```diff
@@ -86,32 +102,91 @@ def create(cls, software, name, description, cli_cmd, parameters_table):
...
             c_id = qdb.sql_connection.TRN.execute_fetchlast()
 
+            # Add the parameters to the DB
+            sql = """INSERT INTO qiita.command_parameter
+                        (command_id, parameter_name, parameter_type, required,
+                         default_value)
+                     VALUES (%s, %s, %s, %s, %s)"""
+            sql_params = [[c_id, pname, p_type, reqd, default]
+                          for pname, p_type, reqd, default in sql_param_values]
+            qdb.sql_connection.TRN.add(sql, sql_params, many=True)
```
ElDeveloper: Just to make sure I understand, if the parameters dictionary is empty, then this line is not going to add any parameters associated with the newly created command, but that's ok, right? Or should we disallow this?

qiita_db/software.py
```diff
@@ -147,36 +222,70 @@ def description(self):
...
-    def cli(self):
-        """The CLI used to call the command
+    def parameters(self):
+        """Returns the parameters that the command accepts
 
         Returns
         -------
-        str
-            The CLI used to call the command
+        dict of {parameter_name: [ptype, dflt]}
```
ElDeveloper: I believe it should be something like:

```python
dict
    Dictionary of the {parameter_name: [ptype, dflt]}.
```

qiita_db/software.py
```diff
@@ -147,36 +222,70 @@ def description(self):
...
-    def parameters_table(self):
-        """The table in which the parameters of th e command are stored
+    def required_parameters(self):
+        """Returns the required parameters that the command accepts
 
         Returns
         -------
-        str
-            The name of the table
+        dict of {parameter_name: ptype}
```
ElDeveloper: Something similar as above.

qiita_db/software.py
```diff
@@ -147,36 +222,70 @@ def description(self):
...
+            res = qdb.sql_connection.TRN.execute_fetchindex()
+            return {pname: ptype for pname, ptype in res}
+
+    @property
+    def optional_parameters(self):
+        """Returns the optional parameters that the command accepts
+
+        Returns
+        -------
+        dict of {parameter_name: [ptype, default]}
```
ElDeveloper: Same, and same for the next method.

qiita_db/software.py
```diff
@@ -413,31 +484,37 @@ def exists(cls, command, **kwargs):
         -------
         bool
             Whether if the parameter set exists in the given command
+
+        Raises
+        ------
+        qiita_db.exceptions.QiitaDBError
+            - If there are missing parameters for the given command
+            - If there are extra parameters in `kwargs` than for the given
```
ElDeveloper: This means that there's an error raised if `kwargs` contains parameters not originally defined in command?

qiita_db/software.py
```diff
@@ -413,31 +484,37 @@ def exists(cls, command, **kwargs):
...
-            if extra:
-                raise qdb.exceptions.QiitaDBError(
-                    "Extra parameters for command %s: %s"
-                    % (command.name, ', '.join(extra)))
+            sql = """SELECT parameter_set
+                     FROM qiita.default_parameter_set
+                     WHERE command_id = %s"""
+            qdb.sql_connection.TRN.add(sql, [command.id])
+            for p_set in qdb.sql_connection.TRN.execute_fetchflatten():
+                if p_set == kwargs:
```
ElDeveloper: Is `p_set` a dictionary of the same form as kwargs? Maybe I'm not following this piece of code very well.

qiita_db/software.py
```diff
@@ -537,29 +600,214 @@ def command(self):
...
+    def load(cls, command, json_str=None, values_dict=None):
+        """Load the parameters set form a json str or from a dict of values
+
+        Paramters
+        ---------
+        command : qiita_db.software.Command
+            The command to which the parameter set belongs to
+        json_str : str, optional
+            The json string encoding the parameters
+        values_dict : dict of {str: object}, optinal
```
ElDeveloper: optinal -> optional

qiita_db/software.py
```diff
@@ -537,29 +600,214 @@ def command(self):
...
+            The dictionary witht the parameter values
```
ElDeveloper: witht -> with

qiita_db/software.py
```diff
@@ -537,29 +600,214 @@ def command(self):
...
-            The string with all the parameters
+        qiita_db.software.Parameters
+            The loaded parameter set
+
+        Raises
+        ------
+        qiita_db.exceptions.QiitaDBError
+            - If `json_str` and `values` are both provided
+            - If neither `json_str` or `values` are provided
+            - If the provided JSON or values do not encode a parameter set of
```
ElDeveloper: For consistency, I think you mean JSON -> `json_str` and `values`.

qiita_db/software.py
```diff
@@ -537,29 +600,214 @@ def command(self):
...
+            if parameters:
+                raise qdb.exceptions.QiitaDBError(
+                    "%s. Extra parameters: %s"
+                    % (error_msg, ', '.join(parameters.keys())))
+
+            return cls(values, command)
+
+    @classmethod
+    def from_default_params(cls, dflt_params, req_params, opt_params=None):
+        """Creates the parameter set from a dflt_parameter set
```
ElDeveloper: dflt_parameter -> dflt_params

qiita_db/software.py
ElDeveloper: Or maybe not, I just noticed these are properties ... hmmm, up to you.

qiita_db/software.py
```diff
@@ -537,29 +600,214 @@ def command(self):
...
+
+        Parameters
+        ----------
+        dflt_params : qiita_db.software.DefaultParameters
+            The DefaultParameters object in which this instance is based on
+        req_params : dict of {str: object}
+            The required parameters values, keyed by parameter name
+        opt_params : dict of {str: object}, optional
+            The optional parameters to change from the default set, keyed by
+            parameter name. Default: None, user the values in `dflt_params`
```
ElDeveloper: user -> use

qiita_db/software.py
```diff
@@ -537,29 +600,214 @@ def command(self):
...
+
+            if opt_params:
+                values.update(opt_params)
+
+            return cls(values, command)
+
+    def __init__(self, values, command):
+        # Time for some python magic! The __init__ function should not be used
+        # outside of this module, users should always be using one of the above
+        # classmethods to instantiate the object. Less test that it is the case
```
ElDeveloper: less -> lets

qiita_db/software.py
```diff
@@ -537,29 +600,214 @@ def command(self):
...
+        # First, we are going to get the current frame (i.e. this __init__)
+        # function and the caller to the __init__
+        current_frame = inspect.currentframe()
+        caller_frame = current_frame.f_back
+        # The file names where the function is defines is stored in the
```
ElDeveloper: defines -> defined

qiita_db/software.py
```diff
@@ -537,29 +600,214 @@ def command(self):
...
+        # f_code.co_filename attribute,and in this case it has to be the same
```
ElDeveloper: attribute,and -> attribute, and

qiita_db/software.py
```diff
@@ -537,29 +600,214 @@ def command(self):
...
-            result = []
-            for p_name, p_type in sorted(table_cols):
-                if p_type == 'boolean':
-                    if values[p_name]:
-                        result.append("--%s" % p_name)
-                else:
-                    result.append("--%s '%s'" % (p_name, values[p_name]))
+    @property
+    def command(self):
+        """The command to which this parametr set belongs to
```
ElDeveloper: parametr -> parameter

qiita_db/software.py
```diff
@@ -537,29 +600,214 @@ def command(self):
...
 
-            return " ".join(result)
+        Returns
+        -------
+        qiita_db.software.Command
+            The command to which this parametr set belongs to
```
ElDeveloper: parametr -> parameter

qiita_db/software.py
```diff
@@ -537,29 +600,214 @@ def command(self):
...
+        """
+        return self._values
+
+    def dump(self):
+        """Return the values in the parameter as JSON
+
+        Returns
+        -------
+        str
+            The parameter values as a json string
```
ElDeveloper: json -> JSON

qiita_db/test/test_software.py
```diff
@@ -16,27 +16,57 @@
...
+                self.parameters)
+
+    def test_create_error_unsupported_parameter_type(self):
+        self.parameters['opt_int_param'][0] = 'unsupported_type'
+        with self.assertRaises(qdb.exceptions.QiitaDBError):
+            qdb.software.Command.create(
+                self.software, "Test command", "Testing command",
+                self.parameters)
+
+    def test_create_error_bad_defalut_choice(self):
```
ElDeveloper: defalut -> default

ElDeveloper:
Done reviewing, looks good overall, just a few questions! Thanks!

qiita_db/software.py

qiita_db/software.py
josenavas: The reason why I make the difference is to be able to generate the interface automatically if we want, so we can limit the user input.

qiita_db/software.py
josenavas: That is actually a really good point - the commands should at least have always one parameter, which is the artifact that they operate on - I will add a check/test that we at least have one parameter. Good catch!

qiita_db/software.py
josenavas: Changed

qiita_db/software.py

qiita_db/software.py

qiita_db/software.py
josenavas: Yup, reworded

qiita_db/software.py
josenavas: Yes, they should both be a dict of `{parameter_name: value}`, and this is checking if the given combination of values already exists in the system.

qiita_db/software.py

qiita_db/software.py

qiita_db/software.py

qiita_db/software.py

qiita_db/software.py
josenavas: Ugh so many typos, sorry about that! the result of coding at 4 am...

qiita_db/software.py

qiita_db/software.py

qiita_db/software.py

qiita_db/software.py

qiita_db/software.py
josenavas: copy-paste!

qiita_db/software.py

qiita_db/test/test_software.py
josenavas: Done!

josenavas:
Thanks for the detailed review @ElDeveloper !!

qiita_db/software.py
```diff
@@ -537,29 +609,214 @@ def command(self):
...
+            return False
+        if self.values != other.values:
+            return False
+        return True
+
+    @classmethod
+    def load(cls, command, json_str=None, values_dict=None):
+        """Load the parameters set form a json str or from a dict of values
+
+        Paramters
```
mortonjt: Paramters -> Parameters

qiita_db/software.py
```diff
@@ -86,32 +102,96 @@ def create(cls, software, name, description, cli_cmd, parameters_table):
...
         description : str
             The description of the command
-        cli_cmd : str
-            The CLI used to call this command
-        parameters_table : str
-            The name of the table in which the parameters of the commands are
-            stored
+        parameters : dict
+            The description of the parameters that this command received. The
+            format is: {parameter_name: (paramter_type, default)},
```
mortonjt: paramter_type -> parameter_type ?

qiita_db/software.py
```diff
@@ -537,29 +609,214 @@ def command(self):
...
+            cmd_req_params = command.required_parameters
+            cmd_opt_params = command.optional_parameters
+
+            missing_reqd = set(cmd_req_params) - set(req_params)
+            extra_reqd = set(req_params) - set(cmd_req_params)
+            if missing_reqd or extra_reqd:
+                raise qdb.exceptions.QiitaDBError(
+                    "Provided required parameters not expected.\n"
+                    "Missing required parameters: %s\n"
+                    "Extra required paramters: %s\n"
```
mortonjt: There are about 7 of these typos altogether.

qiita_db/software.py
```diff
@@ -537,29 +609,214 @@ def command(self):
...
+
+            values = dflt_params.values
+            values.update(req_params)
+
+            if opt_params:
+                values.update(opt_params)
+
+            return cls(values, command)
+
+    def __init__(self, values, command):
```
mortonjt: Not sure if you'd like to add a numpy docstring here or not.

mortonjt:
First pass through with minor comments.
This is looking really awesome!

qiita_db/software.py

qiita_db/software.py

qiita_db/software.py
josenavas: I really apologize for this typos - it just needs to happen once and the editor autocomplete breaks everything... did a search/repleace

qiita_db/software.py
josenavas: Usually the numpy docstring of the `__init__` method goes to the class level documentation, which I intentionally left it out except for the Raises section since I don't want users to use the __init__ method (they should go through the classmethods)

josenavas:
Thanks @mortonjt

qiita_db/software.py
ElDeveloper: You know, Vim has built-in spell checker ... just saying :P

qiita_db/software.py
ElDeveloper: Oh, ok, would it then make sense to somehow specify the allowed ranges
for these values? i.e. replicating click's Range object.

josenavas: I think that is an _awesome_ idea. However, I don't think it should be blocking ATM and it is something that we can extend later, as we are not going to immediately use this right now (i.e. can you open an issue about that?)

ElDeveloper: Sure, should it be its own issue or part of the PR that's integrating
this branch in master?

josenavas: I would suggest it's own issue, since once we have the basic plugin system it will be easier to extend :D

josenavas:
Is there any remaining comment in here?

PR #1545, opened 2015-11-24
Plugin: parameters revamping delete files

josenavas:
Built on top of https://github.com/biocore/qiita/pull/1544, so review/merge that one first.

This PR actually deletes the qiita_db.parameter.py file and its tests. It also fixes the commands.py, and the CLI. Finally it removes the imports from qiita_ware and qiita_pet of this module so we can start the webserver.


antgonza:
This should be at least running the tests but is not ...

scripts/qiita
```diff
@@ -130,24 +130,31 @@ def load_study(owner, title, info):
               help="The parent artifacts from which the new artifact has been "
                    "processed",
               type=click.IntRange(1))
-@click.option('--command', required=False,
-              help="The command used to process the new artifact",
-              type=click.IntRange(1))
-@click.option('--parameter_set', required=False,
-              help="The parameter set used to process the new artifact",
+@click.option('--dflt_params', required=False,
```
antgonza: It will be good to add a command to show the available parameters and their defaults. Fine to open as a new issue if too time demanding.

antgonza:
one comment, after that :+1:

INSTALL.md
```diff
@@ -6,7 +6,7 @@ Qiita is pip installable, but depends on some non-python packages that must be i
 Install the non-python dependencies
 -----------------------------------
 
-* [PostgreSQL](http://www.postgresql.org/download/) (we have tested most extensively with 9.3)
+* [PostgreSQL](http://www.postgresql.org/download/) (minimum required version 9.3.0, we have tested most extensively with 9.3)
```
ElDeveloper: What's the motivation for this? Also, we should make a note in the ChangeLog.md and update `qiita-test-install`.

qiita_db/commands.py
```diff
@@ -189,20 +193,20 @@ def load_prep_template_from_cmd(prep_temp_path, study_id, data_type):
...
+def load_parameters_from_cmd(name, fp, cmd_id):
     """Add a new parameters entry on `table`
 
     Parameters
     ----------
     fp : str
         The filepath to the parameters file
-    table : str
-        The name of the table to add the parameters
+    cmd_id : int
```
ElDeveloper: I thought we were going to be using objects directly? Why are we using the id instead?

qiita_db/test/test_commands.py
```diff
@@ -134,8 +136,8 @@ def test_load_artifact_from_cmd_biom(self):
             with open(fp, 'w') as f:
                 f.write("\n")
         obs = qdb.commands.load_artifact_from_cmd(
-            fps, ftypes, 'BIOM', parents=[3], processing_command_id=3,
-            processing_parameters_id=1)
+            fps, ftypes, 'BIOM', parents=[3], dflt_params_id=10,
+            required_params='{"input_data": 3}')
```
mortonjt: Worthwhile adding a comment specifying where these magic numbers come from?

mortonjt: Just realized that this was the test suite :P.  Nvm.

INSTALL.md
ElDeveloper: Just noticed that qiita-test-install was updated.

scripts/qiita
```diff
@@ -221,16 +228,16 @@ def load_reference_db(name, version, seq_fp, tax_fp, tree_fp):
...
               help='The name of the parameters set')
-def load_parameters(fp, table, name):
+def load_parameters(fp, cmd_id, name):
     """Loads a parameter set to the database"""
-    param = qdb.commands.load_parameters_from_cmd(name, fp, table)
-    click.echo("Parameters successfully added to table %s with id %s"
-               % (table, param.id))
+    param = qdb.commands.load_parameters_from_cmd(name, fp, cmd_id)
+    click.echo("Parameters successfully added to the command %s with id %s"
+               % (cmd_id, param.id))
```
mortonjt: Would this even work since `cmd_id` is an int?

scripts/qiita
ElDeveloper: Hmmm that's true, would it make sense to have this as a "Choice" type then? Or are we gonna have so many different parameters that this is going to become unmanageable?

scripts/qiita
```diff
@@ -130,24 +130,31 @@ def load_study(owner, title, info):
...
+              help="The default parameter set used to process the new "
+                   "artifact",
               type=click.IntRange(1))
+@click.option('--required_params', required=False,
+              help="JSON string containing the required parameters used to "
+                   "processed the new artifact",
+              type=str)
+@click.option('--optional_params', required=False,
+              help="JSON string containing the optional parameters used to "
+                   "processed the new artifact, if different from the "
```
ElDeveloper: processed -> process

scripts/qiita
```diff
@@ -130,24 +130,31 @@ def load_study(owner, title, info):
...
-              type=click.IntRange(1))
-@click.option('--parameter_set', required=False,
-              help="The parameter set used to process the new artifact",
+@click.option('--dflt_params', required=False,
+              help="The default parameter set used to process the new "
+                   "artifact",
               type=click.IntRange(1))
+@click.option('--required_params', required=False,
+              help="JSON string containing the required parameters used to "
+                   "processed the new artifact",
```
ElDeveloper: processed -> process

scripts/qiita
ElDeveloper: Thinking more about it, it doesn't really make sense because the options would be just a bunch of numbers, if we listed these by an identifier that was guaranteed to be unique that would make more sense. Also if we are going to be doing this with this command, then we should probably do the same with the rest of the commands.

antgonza: Yes, that's the reason of a new issue. Note that is possible to list all available commands/params with their unique identifier (int).

scripts/qiita-test-install
```diff
@@ -238,7 +238,7 @@ class QiitaConfig(TestCase):
         if not self.psql_version:
             self.assertTrue(False, 'PostgreSQL not running or configured')
 
-        acceptable_version = (9, 1, 14)
+        acceptable_version = (9, 3, 0)
```
ElDeveloper: This is fine, however line 245 needs to be changed as according to the ChangeLog we now have a "minimum acceptable" meaning that anything greater or equal to that should be considered valid, right?

ElDeveloper:
Looks good, just a few comments.

qiita_db/commands.py
```diff
@@ -189,20 +193,20 @@ def load_prep_template_from_cmd(prep_temp_path, study_id, data_type):
...
+        The command to add the new default parameter set
 
     Returns
     -------
-    qiita_db.BaseParameters
-        The newly `qiita_db.BaseParameters` object
+    qiita_db.software.DefaultParameters
+        The newly parameter set object cretaed
```
mortonjt: cretaed -> created

scripts/qiita
ElDeveloper: Maybe I just have not used this feature enough, but is that really more
useful? (not a rhetorical question).

qiita_db/commands.py
mortonjt: If I'm understanding this correctly, it is used to load the default parameters from the default.

Is this correct @josenavas ?

mortonjt:
Looking good!

scripts/qiita
antgonza: The benefit is that you don't need to have 2 windows open, one to check the commands/params (ipython or DB) and other one to write your command. Again, a nice to have.

ElDeveloper: Got it.

josenavas: I think a new command is the way to go, as @antgonza points out. Is it possible to open an issue? I think it is a nice to have, but not necessarily a blocker right now, and it will take more time than the one that I've right now.

INSTALL.md
josenavas: Added a note to CHANGELOG.md.
The reason is that we are now using the SQL type JSON, which was introduced in postgres 9.3.0

qiita_db/commands.py
josenavas: @ElDeveloper you're correct - we are going to be using objects. However, this is commands.py, which is targetting the CLI, so the only way to provide the objects are by the id.

qiita_db/test/test_commands.py
josenavas: Yeah, that's emulating what would happen when a user loads an artifact from the CLI :)

scripts/qiita
josenavas: Yeah, python is awesome:
```
In [1]: "foo: %s" % 1
Out[1]: 'foo: 1'
```

scripts/qiita

scripts/qiita

scripts/qiita-test-install
josenavas: I changed it for the postgres case. However I think the wording is incorrect for all of them as all of them are checking using `<=` rather than `==`, meaning that is a minimum required version.

scripts/qiita
antgonza: https://github.com/biocore/qiita/issues/1547

qiita_db/commands.py

scripts/qiita-test-install
josenavas: I opened an issue: https://github.com/biocore/qiita/issues/1548

scripts/qiita
mortonjt: Sweet.

PR #1549, opened 2015-11-25
Linking artifacts and processing jobs

josenavas:
This PR adds links the artifacts with those processing jobs that are using the artifact as input.

Since we removed the 'processing_status' attribute from the artifacts, you can know check if it being processed by doing:
```python
is_processing = len(artifact.jobs(status='running')) > 0
```
Although it may seem complex, the main advantage is that `artifact.jobs(status='running')` actually returns the jobs, so you can ask in which step are they or which parameters are being used.


ElDeveloper:
:+1:

mortonjt:
:+1:

PR #1550, opened 2015-11-26
Plugin processing job api

josenavas:
Creates the API that the plugins will use to communicate job information with Qiita.


qiita_db/handlers/processing_job.py
```diff
@@ -0,0 +1,157 @@
...
+        job = qdb.processing_job.ProcessingJob(job_id)
+    except qdb.exceptions.QiitaDBError as e:
+        return None, False, 'Error instantiating the job: %s' % str(e)
+
+    return job, True, ''
+
+
+class JobHandler(RequestHandler):
+    def get(self, job_id):
+        """Retrieves the work that need to be done for the job
```
antgonza: could you change work for something clearer?

qiita_db/handlers/processing_job.py
```diff
@@ -0,0 +1,157 @@
...
+
+        Parameters
+        ----------
+        job_id : str
+            The job that the information needs to be returned
```
antgonza: Rewrite

qiita_db/handlers/processing_job.py
```diff
@@ -0,0 +1,157 @@
...
+    try:
+        job = qdb.processing_job.ProcessingJob(job_id)
+    except qdb.exceptions.QiitaDBError as e:
+        return None, False, 'Error instantiating the job: %s' % str(e)
+
+    return job, True, ''
+
+
+class JobHandler(RequestHandler):
+    def get(self, job_id):
```
antgonza: Perhaps get_status?

qiita_db/handlers/processing_job.py
```diff
@@ -0,0 +1,157 @@
...
+
+        response = {'success': success, 'error': error_msg,
+                    'command': cmd, 'parameters': params,
+                    'status': status}
+        self.write(response)
+
+
+class HeartbeatHandler(RequestHandler):
+    def post(self, job_id):
+        """Perform a heartbeat against a job
```
antgonza: Checks heartbeat?

qiita_db/handlers/processing_job.py
```diff
@@ -0,0 +1,157 @@
...
+
+        Parameters
+        ----------
+        job_id : str
+            The job to perform the heartbeat
```
antgonza: The job id to retrieve ...

qiita_pet/webserver.py
```diff
@@ -106,6 +108,11 @@ def __init__(self):
             (r"/stats/", StatsHandler),
             (r"/download/(.*)", DownloadHandler),
             (r"/vamps/(.*)", VAMPSHandler),
+            # Plugin handlers - the order matters here so do not change
```
antgonza: Why does it matter? Can you expand the comment?

antgonza:
A few comments but after that! :+1:

qiita_db/handlers/processing_job.py
ElDeveloper: need -> needs

qiita_db/handlers/processing_job.py
```diff
@@ -0,0 +1,157 @@
...
+
+class JobHandler(RequestHandler):
+    def get(self, job_id):
+        """Retrieves the work that need to be done for the job
+
+        Parameters
+        ----------
+        job_id : str
+            The job that the information needs to be returned
+        """
```
ElDeveloper: It seems that for these handlers it is particularly useful to know what the responses are, as these are a proper API and should be used as such, so I would recommend adding the returns section to the numpy doc.

ElDeveloper: Consider doing this for all the other handlers.

qiita_db/handlers/processing_job.py
```diff
@@ -0,0 +1,157 @@
...
+        ----------
+        job_id : str
+            The job to perform the heartbeat
+        """
+        with qdb.sql_connection.TRN:
+            job, success, error_msg = _get_job(job_id)
+            if success:
+                job_status = job.status
+                if job_status == 'queued':
+                    job.status = 'running'
```
ElDeveloper: Why does the status change from queued to running? :runner: :running_shirt_with_sash:

qiita_db/handlers/processing_job.py
```diff
@@ -0,0 +1,157 @@
...
+                    job.heartbeat = datetime.now()
+                elif job_status == 'running':
+                    job.heartbeat = datetime.now()
+                else:
+                    success = False
```
ElDeveloper: If this somehow represents a failure, should this be returned as a different response code? Seems odd that the this is is not a successful call yet the response returns just fine. I would even consider getting rid of the error message.

qiita_db/handlers/processing_job.py
```diff
@@ -0,0 +1,157 @@
...
+                    error_msg = 'Job already finished. Status: %s' % job_status
+
+        response = {'success': success, 'error': error_msg}
+        self.write(response)
+
+
+class StepHandler(RequestHandler):
+    def post(self, job_id):
+        """Changes the step of the given job
```
ElDeveloper: I didn't think about this before, but what happens when there are multiple jobs running at the moment i.e. you've started a meta analysis and the analysis creation began, now you have three parallel jobs summarize taxa, beta diversity and alpha, would this handler report that only one of them is the "current step" or would it handle this correctly? It seems like it would only report one, right?

qiita_db/handlers/processing_job.py
```diff
@@ -0,0 +1,157 @@
...
+        ----------
+        job_id : str
+            The job to change the step
+        """
+        with qdb.sql_connection.TRN:
+            job, success, error_msg = _get_job(job_id)
+            if success:
+                job_status = job.status
+                if job_status != 'running':
+                    success = False
```
ElDeveloper: Same thing about errors, should probably be an exception.

qiita_db/handlers/processing_job.py
```diff
@@ -0,0 +1,157 @@
...
+                    job.heartbeat = datetime.now()
+                else:
+                    success = False
+                    error_msg = 'Job already finished. Status: %s' % job_status
+
+        response = {'success': success, 'error': error_msg}
+        self.write(response)
+
+
+class StepHandler(RequestHandler):
```
ElDeveloper: StepHandler sounds a bit odd, would it make more sense as "ActiveStepHandler" or something else, not sure what others think.

qiita_db/handlers/processing_job.py
```diff
@@ -0,0 +1,157 @@
...
+                    step = payload['step']
+                    job.step = step
+
+        response = {'success': success, 'error': error_msg}
+        self.write(response)
+
+
+class CompleteHandler(RequestHandler):
+    def post(self, job_id):
+        """Updates the job to one of the completed status ('success', 'error')
```
ElDeveloper: statuts -> statuses

qiita_db/handlers/processing_job.py
```diff
@@ -0,0 +1,157 @@
...
+                            vamps = artifact_data['can_be_submitted_to_vamps']
+                            qdb.artifact.Artifact.create(
+                                filepaths, atype, parents=parents,
+                                processing_parameters=params,
+                                can_be_submitted_to_ebi=ebi,
+                                can_be_submitted_to_vamps=vamps)
+                        job.status = 'success'
+                    else:
+                        log = qdb.logger.LogEntry.create(
+                            'Runtime', payload['error'])
```
ElDeveloper: Not sure about this, but the paylod will include the information about why something failed, correct?

qiita_db/handlers/test/__init__.py
```diff
@@ -0,0 +1,7 @@
+# -----------------------------------------------------------------------------
```
ElDeveloper: I know we haven't been very diligent about this but the folder's name should probably be tests instead of test ... just sayin'

qiita_pet/webserver.py
```diff
@@ -106,6 +108,11 @@ def __init__(self):
...
+            (r"/qiita_db/jobs/(.*)/heartbeat/", HeartbeatHandler),
```
ElDeveloper: I thought your initial intent was to keep these handlers separate from the main server, just wanted to check with you.

qiita_db/handlers/processing_job.py
josenavas: I can't change the name of these functions - there are defining the GET and POST operations of the handlers.

qiita_db/handlers/processing_job.py
josenavas: Done

qiita_db/handlers/processing_job.py
josenavas: Yeah, I was thinking about that - I was unsure on how to document the actual results of the handler given that they're not "returning" anything in the usual python way, but they're actually returning data through the HTTP requests. Given that, I think it is ok to document them in the "Returns" section. Will do that!

qiita_db/handlers/processing_job.py
josenavas: No, it's a post function and it is actually performing the heartbeat. This is how Qiita knows that the plugin is still alive.

qiita_db/handlers/processing_job.py
josenavas: The first heartbeat is the one that tells you that the job is running - the plugin will start the heartbeat when it starts.

qiita_db/handlers/processing_job.py
josenavas: Change all of them to just `The job id`. Everything else is just redundant.

qiita_db/handlers/processing_job.py
josenavas: I was thinking about that - the main reason why I decided to have a (success, error) tuple in all the calls is that this will allow us to differentiate between communication error (i.e. http error) and some processing error. I think this will make debugging easier, as the plugin can write down the reason of the failure, rather than a useless '500'...

qiita_db/handlers/processing_job.py
josenavas: +1 changed

qiita_db/handlers/processing_job.py
josenavas: What I was thinking in that situation is that we actually have multiple jobs, i.e. summarize taxa is one job, beta is one job, etc... Does that make sense?

qiita_db/handlers/processing_job.py
josenavas: If you raise an exception here, the caller will only receive a "500 Internal Sever Error", I think that returning success False and the reason why can be more useful.

qiita_db/handlers/processing_job.py
josenavas: Done

qiita_db/handlers/processing_job.py
josenavas: Yes, `payload['error']` is the string containing the error string

qiita_db/handlers/test/__init__.py
josenavas: Done

qiita_pet/webserver.py
josenavas: Done

qiita_pet/webserver.py
josenavas: Yes, the handlers are in a different location, but we decided in the last meeting that we're gonna be using the same webserver in interests of time.

qiita_db/handlers/processing_job.py
antgonza: Got it, thanks.

antgonza:
Looks good and all comments have been addressed. Going to wait for tests then, merge ...

qiita_db/handlers/processing_job.py
ElDeveloper: This may be way to late, but it seems like there's a motivation to use
error codes whenever possible: http://stackoverflow.com/a/943011

antgonza: Sounds like a good idea but this should be propagated to the full code. What about adding an issue so we don't forget.

ElDeveloper: Sounds good! https://github.com/biocore/qiita/issues/1555

PR #1551, opened 2015-11-26
Plugin artifact api

josenavas:
Creates a minimal API that the plugins will use to request artifact information from Qiita.
It is build on top of #1550 so review/merge that one first to remove those commits from here.


antgonza:
:+1:

qiita_db/handlers/artifact.py
```diff
@@ -0,0 +1,69 @@
...
+
+
+class ArtifactFilepathsHandler(RequestHandler):
+    def get(self, artifact_id):
+        """Retrieves the filepath information of the given artifact
+
+        Parameters
+        ----------
+        artifact_id : str
+            the id of the artifact that the information needs to be returned
```
mortonjt: This is a bit confusing to read. Could you reword this to something like

`the id of the artifact whose file path information is being retrieved`.

mortonjt:
One nit-picky comment.  Otherwise, I think it looks good.

PR #1552, opened 2015-11-27
Plugin qiita-pet

antgonza:
This fixes qiita-pet with artifacts. The changes are minimal but had to create a new method to return the pmid ids from the doi. Additionally, I realized that I had to change the populate so the log display works. The next things to do are fix the 'TODO plugin' sections and make sure that jobs work: split libs, analysis, etc.


qiita_db/support_files/populate_test_db.sql
```diff
@@ -488,7 +488,7 @@ INSERT INTO qiita.message_user (message_id, email) VALUES (1, 'test@foo.bar'),(1
 
 -- Create a loggin entry
 INSERT INTO qiita.logging (time, severity_id, msg, information)
-    VALUES ('Sun Nov 22 21:29:30 2015', 2, 'Error message', NULL);
+    VALUES ('Sun Nov 22 21:29:30 2015', 2, 'Error message', '{}');
```
josenavas: If information can't be NULL there should be an error somewhere (either the interface is assuming that is always there, or either the database should have a NOT NULL restriction and update the python code accordingly).

josenavas: This shouldn't be blocking so I added an issue: https://github.com/biocore/qiita/issues/1553

qiita_db/util.py
```diff
@@ -1057,6 +1058,30 @@ def get_timeseries_types():
...
+        Return dict of doi and pubmed ids
+
+    Notes
+    -----
+    If doi doesn't exist it will not return that {key: value} pair
+    """
+    with qdb.sql_connection.TRN:
+        sql = "SELECT doi, pubmed_id FROM qiita.publication WHERE doi IN %s"
+        qdb.sql_connection.TRN.add(sql, [tuple(doi_ids)])
+        return {row[0]: join(*row[1:])
```
josenavas: Shouldn't this just be `{row[0]: row[1] for row in qdb.sql_connection.TRN.fetchindex()}`?

qiita_pet/handlers/analysis_handlers.py
```diff
@@ -307,13 +307,21 @@ def get(self):
...
-        sel_samps = Analysis(self.current_user.default_analysis).samples
+        sel_samps = self.current_user.default_analysis.samples
         for pid, samps in viewitems(sel_samps):
-            proc_data = ProcessedData(pid)
+            proc_data = Artifact(pid)
             sel_data[proc_data.study][pid] = samps
             # Also get processed data info
-            proc_data_info[pid] = proc_data.processing_info
-            proc_data_info[pid]['data_type'] = proc_data.data_type()
+            # TODO plugin:
```
josenavas: Just a note - we will be able to duplicate this information with the artifacts, however, this is highly tied up with target gene and is not gonna work for other datatypes - so we should rethink this information.

qiita_db/util.py
antgonza: I copied from filepath_ids_to_rel_paths so not sure ... should I change?

qiita_pet/uimodules/preprocessed_data_tab.py
```diff
@@ -57,10 +56,14 @@ def render(self, study_id, preprocessed_data):
         # New Type is for users to add a new user-defined investigation type
         user_defined_terms = ontology.user_defined_terms + ['New Type']
 
-        if PrepTemplate.exists(preprocessed_data.prep_template):
-            prep_template_id = preprocessed_data.prep_template
-            prep_template = PrepTemplate(prep_template_id)
-            raw_data_id = prep_template.raw_data
+        # ppd can only have 1 prep template
+        prep_template = preprocessed_data.prep_templates[0]
+        # this block might seem wrong but is here due to a possible
```
josenavas: This is not gonna work as it is right now. `prep_template` is already a PrepTemplate so if we have that pathological case, it will break in the call above `preprocessed_data.prep_templates[0]`, as it could not instantiate any prep template, so you'll receive an IndexError. If this pathological case still exists (which it shouldn't) the correct solution is:

```python
prep_templates = preprocessed_data.prep_templates
if len(prep_templates == 1):
    prep_template_id = prep_template.id
    raw_data_id = prep_template.artifact.id
    inv_type = prep_template.investigation_type or "None selected"
else:
    prep_template_id = None
```

qiita_pet/handlers/study_handlers/listing_handlers.py
```diff
@@ -112,11 +121,12 @@ def _build_single_proc_data_info(proc_data_id, data_type, samples):
     dict
         The information for the processed data, in the form {info: value, ...}
     """
-    proc_data = ProcessedData(proc_data_id)
-    proc_info = proc_data.processing_info
+    proc_data = Artifact(proc_data_id)
+    # TODO plugin:
```
mortonjt: What is this TODO for?  Is this worth raising an issue?

qiita_db/util.py
josenavas: Yeah, the construction in here is weird

qiita_pet/handlers/study_handlers/listing_handlers.py
josenavas: No, the issue is that some of this changes depends on my changes for the plugin, but the general qiita_pet update should not be blocked due to this. This way, it is easier that we have most of the things changed and I can just search for the TODO and connect the things together in a subsequent PR.

josenavas:
Couple of comments, otherwise :+1:

mortonjt:
:+1:

qiita_db/util.py
antgonza: Done

qiita_pet/uimodules/preprocessed_data_tab.py
antgonza: I know we fixed but prefer to leave it ... changing for your block.

PR #1554, opened 2015-11-30
Plugin default params

josenavas:
Adds default parameters if there isn't any present in the DB.


antgonza:
This are basic DB changes that are needed so my next PR works, which it does now. Thus, merging.

PR #1557, opened 2015-11-30
Plugin: target gene plugin

josenavas:
This adds the target gene plugin (split librares [fastq] and OTU picking).

I'm currently adding tests.
This is heavily based in the code already present in `qiita_ware/processing_pipeline.py` but adapting it to the plugin structure and simplifying some (redundant/complex) code.

PR #1558, opened 2015-12-02
WIP: Plugin connect everything

josenavas:
This is built on top of #1557 so review/merge that one first.

This PR connects all the plugin bits so the system should be all functional again.

Currently, there are still some reference to the data objects (RawData, PreprocessedData and ProcessData) done in qiita_pet (which I thought were already removed); and some parts of the interface is acting up.

Tomorrow I will need to shift for some time to work in some of my class homeworks, so doing a WIP in case that somebody else can also take a look. The vast majority of the work is already done, and the only missing bits in here is to fix the interface to be compliant with the underlying system.


qiita_ware/executor.py
```diff
@@ -0,0 +1,73 @@
...
+from os.path import join
+
+from qiita_core.qiita_settings import qiita_config
+from qiita_db.processing_job import ProcessingJob
+from qiita_db.util import get_work_base_dir
+from qiita_db.logger import LogEntry
+
+
+def system_call(cmd):
+    """Call command and return (stdout, stderr, return_value)
```
antgonza: Call cmd and return ...

antgonza:
One small comment.

qiita_db/software.py
```diff
@@ -331,15 +351,22 @@ def create(cls, name, version, description, publications=None):
             The version of the software
         description : str
             The description of the software
+        environment_script : str
+            The script used to start the environemnt in which the plugin runs
```
mortonjt: environemnt -> environment

qiita_db/software.py
```diff
@@ -455,6 +482,38 @@ def add_publications(self, publications):
...
             qdb.sql_connection.TRN.execute()
 
+    @property
+    def environment_script(self):
+        """The script used to start the plugin environment
+
+        Returns
+        -------
+        str
+            The script used to start the environemnt
```
mortonjt: environemnt -> environment

qiita_db/test/test_software.py
```diff
@@ -233,6 +245,14 @@ def test_add_publications(self):
                ['10.1000/nmeth.f.101', '12345678']]
         self.assertItemsEqual(tester.publications, exp)
 
+    def test_environment_script(self):
+        tester = qdb.software.Software(1)
+        self.assertEqual(tester.environment_script, 'source activate qiita')
```
mortonjt: This is assuming a conda environment?

qiita_pet/handlers/preprocessing_handlers.py
```diff
@@ -13,19 +13,12 @@ class PreprocessHandler(BaseHandler):
...
-            param_constructor = PreprocessedIlluminaParams
-        elif raw_data.filetype in ('FASTA', 'SFF'):
-            param_constructor = Preprocessed454Params
-        else:
-            raise ValueError('Unknown filetype')
+        user_id = self.current_user.id
 
-        job_id = submit(self.current_user.id, preprocessor, study_id,
-                        prep_template_id, param_id, param_constructor)
+        job_id = submit(user_id, preprocessor, user_id, raw_data.id, param_id)
```
mortonjt: This `submit` function is a bit confusing.  Why does `user_id` have to be passed in twice?

Also the function signature for [`submit`](https://github.com/josenavas/QiiTa/blob/plugin-connect-tgp/qiita_ware/context.py) is really confusing.  If you think it is a good idea, I can open up an issue/PR to add numpy docs to fix this.

mortonjt:
A few comments.

qiita_db/software.py

qiita_db/software.py

qiita_db/test/test_software.py
josenavas: That is added in the database - In the last meeting that I attended we agreed that we are going to use conda environments to manage environment encapsulation. However, I found that I can actually leave it up to the user (so it can even put custom scripts). The test that we have in the database however, assumes a conda environment, as that is what we are going to recommend.

qiita_pet/handlers/preprocessing_handlers.py
josenavas: That is part of the execution environment that we have in place, and it is using moi. The first user_id is used in moi to identify which user is launching the job, so we can use the websockets to communicate back to the interface. The second user_id is actually a parameter of the function `preprocessor`.

In short, the parameters of the `submit` function is a user identifier to know who is launching the job, the function to be executed and the `args` and `kwargs` of such function.

qiita_core/configuration_manager.py
```diff
@@ -107,6 +107,8 @@ class ConfigurationManager(object):
         The filepath for the configuration file that is loaded
     portal_fp : str
         The filepath to the portal styling config file
+    plugin_launcher : str
```
ElDeveloper: Should we have some user-level documentation on what this script looks like, maybe in the Admin Tutorials section? Alternatively if we don't expect this to change or be modified, then we probably don't need that.

qiita_db/software.py
```diff
@@ -455,6 +482,38 @@ def add_publications(self, publications):
...
+        with qdb.sql_connection.TRN:
+            sql = """SELECT environment_script
+                     FROM qiita.software
+                     WHERE software_id = %s"""
+            qdb.sql_connection.TRN.add(sql, [self.id])
+            return qdb.sql_connection.TRN.execute_fetchlast()
+
+    @property
+    def start_script(self):
+        """The start_script used to start the plugin
```
ElDeveloper: start_script -> script

qiita_db/software.py
```diff
@@ -455,6 +482,38 @@ def add_publications(self, publications):
...
             qdb.sql_connection.TRN.execute()
 
+    @property
+    def environment_script(self):
+        """The script used to start the plugin environment
+
+        Returns
+        -------
+        str
+            The script used to start the environment
```
ElDeveloper: Hmmm, is this the script itself or is it a path to the script? Does it have to be a full path?

qiita_db/software.py
```diff
@@ -455,6 +482,38 @@ def add_publications(self, publications):
...
+            return qdb.sql_connection.TRN.execute_fetchlast()
+
+    @property
+    def start_script(self):
+        """The start_script used to start the plugin
+
+        Returns
+        -------
+        str
+            The plugin's start script
```
ElDeveloper: Same question as above?

ElDeveloper: nix that question mark :P

qiita_pet/handlers/study_handlers/listing_handlers.py
```diff
@@ -122,9 +122,7 @@ def _build_single_proc_data_info(proc_data_id, data_type, samples):
         The information for the processed data, in the form {info: value, ...}
     """
     proc_data = Artifact(proc_data_id)
-    # TODO plugin:
-    # proc_info = proc_data.processing_info
-    proc_info = {'processed_date': '10/10/1981'}
+    proc_info = {'processed_date': str(proc_data.timestamp)}
```
ElDeveloper: Is line 128 necessary at all?

qiita_pet/uimodules/prep_template_tab.py
```diff
@@ -218,7 +219,19 @@ def render(self, study, prep_template, full_access, ena_terms,
                         no_preprocess_msg = (
                             "Preprocessing disabled due to missing columns in "
                             "the prep template: %s" % ', '.join(missing_cols))
-        preprocessing_status = 'TODO: plugin'
+
+            # Check the processing status
+            jobs = raw_data.jobs(status='running')
```
ElDeveloper: I almost feel like this block of logic should be in the artifact class, and not in here, what do you think?

qiita_ware/dispatchable.py
```diff
@@ -5,48 +5,29 @@
...
+def processor(user_id, preprocessed_data_id, param_id):
     """Dispatch the processor work"""
-    preprocessed_data = PreprocessedData(preprocessed_data_id)
-    params = param_constructor(param_id)
-
-    sp = StudyProcessor()
-    try:
-        process_out = sp(preprocessed_data, params)
-    except Exception as e:
-        error_msg = ''.join(format_exception_only(e, exc_info()))
```
ElDeveloper: I think I just don't remember anymore, but I wanted to check with you that the exception is being caught and formatted properly like it was being done here?

qiita_ware/executor.py
```diff
@@ -0,0 +1,73 @@
...
+    plugin_start_script = software.start_script
+    plugin_env_name = software.environment_name
+
+    # Get the command to start the plugin
+    cmd = [qiita_config.plugin_launcher, plugin_env_name, plugin_start_script,
+           qiita_config.base_url, job.id, job_dir]
+
+    # Start the plugin
+    std_out, std_err, return_value = system_call(cmd)
+    if return_value != 0:
```
ElDeveloper: Should we be raising an error or something, this looks a bit too dangerous as errors can go unnoticed.

ElDeveloper:
shoooot, my browser didn't update the fact that this PR was merged ... I guess all comments are irrelevant now :(

scripts/qiita
```diff
@@ -425,18 +424,5 @@ def start(port):
...
 
-# #############################################################################
-# WARE COMMANDS
-# #############################################################################
-
-@ware.command()
-@click.argument('sl_out_dir', required=True,
-                type=click.Path(resolve_path=True, readable=True, exists=True,
-                                file_okay=False))
-def generate_demux(sl_out_dir):
```
ElDeveloper: Will this functionality be replaced somehow?

scripts/qiita-plugin-launcher
```diff
@@ -0,0 +1,32 @@
...
+@click.argument('url', required=True)
+@click.argument('job_id', required=True)
+@click.argument('job_dir', required=True)
+def start(env_script, start_script, url, job_id, job_dir):
+    """Starts the plugin environment"""
+    cmd = [start_script, url, job_id, job_dir]
+    # When Popen executes, the shell is not in interactive mode,
+    # so it is not sourcing any of the bash configuration files
+    # We need to source it so the env_script are available
+    cmd = "source ~/.bash_profile; %s; %s" % (env_script, ' '.join(cmd))
```
ElDeveloper: Hmmm ... I don't think hardcoding this is a good idea.

qiita_core/configuration_manager.py
josenavas: I've added an item in the main WIP PR so I can add the docs later.

qiita_db/software.py
josenavas: Path to the scrip - not necessarily full as it will be executed inside the environment in which the plugin lives, so it should be added in the path already.

josenavas: So that actually means that is not the path but just the name of the script

josenavas: Sorry, I got confused here. This is the call used to start the environment. So it can be either a script, or whatever. For example, in my system this looks like `workon qiita`; in the test environment this looks like `source activate qiita` (using conda) or I can have a `start_plugin_X_env.sh` which is a bash script. At the end is a CLI call that I can issue to start the environment

qiita_db/software.py
josenavas: Changed (this and all other changes will go in with my PR)

qiita_db/software.py
josenavas: This is the one that I wanted to answer first :D
This one is executed inside the plugin environment (ie. after it has been already activated) so is the script name and we are assuming that is already in the path.

qiita_pet/handlers/study_handlers/listing_handlers.py
josenavas: Yeah, that value is needed by the interface, we can probably initialize all the dict values at once, but I wanted to minimize changes.

qiita_pet/uimodules/prep_template_tab.py
josenavas: The reason why it is here is because the artifact (and qiita db) supports multiple processing of an artifact in parallel. However, the interface was designed to be processing only once at a time. I have the feeling that this is a limitation that is going to go away once we re-implement the study interface, so this is just a hack to maintain the same interface.

qiita_ware/dispatchable.py
josenavas: Yeah, execute should not raise any exception and it is caching the error and storing it in the job accoridingly.

qiita_ware/executor.py
josenavas: I don't understand this concern, we are storing the error in the job itself, and it will be shown to the user in the interface

scripts/qiita
josenavas: generate demux is part of the plugin. I'm unsure if we want to add CLI to the plugins... I would suggest adding an issue to see what we can do. Note that this functionality is available anyways through ipython, by importing the generate_demux_file func from the plugin

scripts/qiita-plugin-launcher
josenavas: The script `qiita-plugin-launcher` is just an example on how they have to look and it is related to the comment that you made above about adding admin documentation. For example, this script will be different in barnacle.

qiita_db/software.py
ElDeveloper: Got it, would be useful to have that info!

ElDeveloper: I see, makes sense, thanks! Would be useful to note that this refers to
a command available within a non-interactive session.

qiita_pet/uimodules/prep_template_tab.py
ElDeveloper: Got it thanks for the explanation!

qiita_ware/executor.py
ElDeveloper: Thanks, just wanted to make sure we were not losing this information!

scripts/qiita
ElDeveloper: I feel like the CLIs have proved useful to add/process studies in bulk,
and given that I recently used this tool for one study myself, the new
issue is here:
https://github.com/biocore/qiita/issues/1563

scripts/qiita-plugin-launcher
ElDeveloper: Got it, would be useful to note this somewhere in the documentation of
this script.

PR #1561, opened 2015-12-05
Plugin gui

josenavas:
I was able to create a new study and go from raw data to a biom table using the interface.

I will keep looking at the interface tomorrow to see if there is any other bug due to the changes, but I think I caught all the issues in the study page.


antgonza:
Thanks @josenavas. Notice how there are still failures, I guess fixing those will help to minimize the possible errors.

josenavas:
We should be back on green! I've been able to go form the raw data to an analysis.

qiita_db/handlers/reference.py
```diff
@@ -0,0 +1,71 @@
...
+
+from tornado.web import RequestHandler
+
+import qiita_db as qdb
+
+
+def _get_reference(r_id):
+    """Returns the reference with the given id if exists
+
+    Parameters
```
antgonza: There are Returns and needs Parameters.

qiita_db/handlers/reference.py
```diff
@@ -0,0 +1,71 @@
...
+            retrieved
+
+        Returns
+        -------
+        dict
+            Format:
+            {'success': bool,
+             'error': str,
+             'filepaths': list of (str, str)}
+            - success: whether the request is successful or not
```
antgonza: Should this be Notes? Not 100% sure.

qiita_pet/handlers/analysis_handlers.py
```diff
@@ -269,9 +266,14 @@ def validate_absolute_path(self, root, absolute_path):
         """Overrides StaticFileHandler's method to include authentication
         """
         # Get the filename (or the base directory) of the result
+        if root[-1] != '/':
+            root = "%s/" % root
         len_prefix = len(commonprefix([root, absolute_path]))
         base_requested_fp = absolute_path[len_prefix:].split(sep, 1)[0]
 
+        print len_prefix, base_requested_fp, absolute_path
```
antgonza: Remove these prints?

qiita_pet/handlers/study_handlers/listing_handlers.py
```diff
@@ -126,6 +128,18 @@ def _build_single_proc_data_info(proc_data_id, data_type, samples):
     proc_info['pid'] = proc_data_id
     proc_info['data_type'] = data_type
     proc_info['processed_date'] = str(proc_info['processed_date'])
+    params = proc_data.processing_parameters.values
+    del params['input_data']
```
antgonza: Why do we need to remove this field? Can you document?

qiita_pet/handlers/study_handlers/listing_handlers.py
```diff
@@ -177,10 +191,15 @@ def _build_study_info(user, study_proc=None, proc_samples=None):
     cols = ['study_id', 'email', 'principal_investigator_id',
             'publication_doi', 'study_title', 'metadata_complete',
             'number_samples_collected', 'study_abstract']
+    study_ids = [s.id for s in study_set]
```
antgonza: Can these 2 lines be merged? Just asking ...

qiita_pet/util.py
```diff
@@ -116,3 +116,45 @@ def is_localhost(host):
...
+            preprocessing_status_msg.append(
+                "<b>Job %s</b>: %s" % (job.id, job_status))
+
+    if not preprocessing_status_msg:
+        preprocessing_status_msg = 'Not processed'
+    else:
+        preprocessing_status_msg = convert_text_html(
+            '<br/>'.join(preprocessing_status_msg))
+
+    return preprocessing_status, preprocessing_status_msg
```
antgonza: Is this really a tuple or two separate vars?

qiita_plugins/target_gene/tgp/split_libraries/util.py
```diff
@@ -142,5 +142,25 @@ def generate_demux_file(sl_out):
...
+    """Creates the artifact information to attach to the payload
+
+    Parameters
+    ----------
+    sl_out : str
+        Path to the split libraries output directory
+
+    Returns
+    -------
+    list
```
antgonza: It returns a list of list, can you expand documentation?

antgonza:
A few comments, nothing major. Gonna check the interface ...

qiita_db/handlers/reference.py
josenavas: Done

qiita_db/handlers/reference.py
josenavas: I think for consistency we should leave it here (this is how there are in the other handlers documentation).
Since there is an issue that @ElDeveloper opened to change the behavior of handlers (i.e. do not return the success bool, just return error codes) I suggest to leave as it is and fix when fixing that other issue.

qiita_pet/handlers/analysis_handlers.py
josenavas: Ooops, debugging stuff - removed.

qiita_pet/handlers/study_handlers/listing_handlers.py
josenavas: Here I'm trying to replicate whatever was in the interface before. Before we were not showing which was the origin artifact, so that's why I'm just removing it here. Note that a lot of the changes here are kind of hackish because the system is way more powerful than the interface allows, so we need these kind of things until we change the interface.

qiita_pet/handlers/study_handlers/listing_handlers.py
josenavas: Done

qiita_pet/util.py
josenavas: I'm assuming that you're asking due to the documentation? I just fixed it, because they're 2 separate vars. Note, however, that python automatically creates a tuple when you have a return like this.

qiita_plugins/target_gene/tgp/split_libraries/util.py
josenavas: Done

josenavas:
Thanks @antgonza 

While I was checking the interface yesterday, I had the feeling that split libraries was running faster in my machine that usual - did you had the same impression?

qiita_db/study.py
```diff
@@ -652,8 +652,13 @@ def publications(self, values):
 
             # Set the new ones
             sql = """INSERT INTO qiita.publication (doi, pubmed_id)
-                     VALUES (%s, %s)"""
-            qdb.sql_connection.TRN.add(sql, values, many=True)
+                     SELECT %s, %s
+                     WHERE NOT EXISTS(
+                        SELECT doi FROM qiita.publication WHERE doi = %s)"""
+            # sql = """INSERT INTO qiita.publication (doi, pubmed_id)
```
mortonjt: Want to delete this comment?

qiita_db/artifact.py
```diff
@@ -45,25 +45,36 @@ class Artifact(qdb.base.QiitaObject):
     _table = "artifact"
 
     @classmethod
-    def iter_public(cls):
-        r"""Iterator over the public artifacts available in the system
+    def iter_by_visibility(cls, visibility):
+        r"""Iterator over the artifacts with the given visibility
```
mortonjt: Would you like to add a Parameter section here?

qiita_db/study.py

qiita_db/artifact.py

mortonjt:
:+1:

qiita_db/handlers/reference.py
antgonza: K

qiita_pet/handlers/study_handlers/listing_handlers.py
antgonza: K

qiita_pet/util.py
antgonza: Thanks

qiita_pet/util.py
```diff
@@ -116,3 +116,45 @@ def is_localhost(host):
...
+    for job in artifact.jobs():
+        job_status = job.status
+        if job_status == 'error':
+            if preprocessing_status != 'success':
+                preprocessing_status = 'failed'
+            preprocessing_status_msg.append(
+                "<b>Job %s</b>: failed - %s"
+                % (job.id, job.log.msg))
+        elif job_status == 'success':
+            preprocessing_status = 'success'
```
mortonjt: Just to sanity check

Wouldn't it be possible to have an artifact that has multiple jobs, the first job fails, but the last job succeeds, resulting in `get_artifact_status` to return `'sucess'`.

Is this behavior wanted?

josenavas: Yes - unfortunately, the current interface is designed in a way that only one job can be executed, so whenever we have a successful job we disallow launching any other job. This is just an interface limitation, since the underlying system supports way more. However, since we're not changing the interface in this PR, we need to emulate that behavior like that in here.

PR #1562, opened 2015-12-05
Oauth2 addition

squirrelo:
This pull request adds Oauth2-style authentication to the API. Every plugin now requires a client_id and client_secret to ping the API with to get an access token. This access token allows connection for one hour, and then the token will need to be re-created and the new one used. Users can also get access by giving a client ID and their username and password. This allows us to have a secure login system for the API that can also be secured on a per-user basis as well as a per-client (plugin) basis.
Two of the four access types have been implemented: [Client Credentials Grant](http://tools.ietf.org/html/rfc6749#section-4.4) for the plugins and [Resource Owner Password Credentials Grant](http://tools.ietf.org/html/rfc6749#section-4.3) for qiita-pet.


josenavas:
Do not merge this changes until I can test this locally with a working system.

antgonza:
@squirrelo could you pull from artifact? Thanks!

squirrelo:
pulled from artifact. Also added daily request limiting.

squirrelo:
The auth check is now a decorator, so this should be in a final state and ready for review.

mortonjt:
Hmm.  Looks like there are some test files that can't be found ...

squirrelo:
I can't reproduce the test error on my system. Anyone have ideas?

josenavas:
@squirrelo will investigate shortly - I'm planning to pull this down, and test locally, so I can also update the plugins accordingly

qiita_db/handlers/oauth2.py
```diff
@@ -0,0 +1,280 @@
...
+    -------
+    bool
+        Whether client_id exists and, if client_secret given, the secret
+        matches the id
+    """
+    with qdb.sql_connection.TRN:
+        sql = """SELECT EXISTS(
+                    SELECT *
+                    FROM qiita.oauth_identifiers
+                    WHERE client_id = %s{0})"""
```
josenavas: Missing a space after %s

qiita_db/handlers/oauth2.py
```diff
@@ -0,0 +1,280 @@
...
+        else:
+            sql = sql.format("AND client_secret IS NULL")
+        qdb.sql_connection.TRN.add(sql, sql_info)
+        return qdb.sql_connection.TRN.execute_fetchlast()
+
+
+def authenticate_oauth(f):
+    """Decorate methods with this to require valid Oauth2 Authorization header
+
+    If a valid header is given, teh handof is odne and the page is rendered. If
```
josenavas: Multiple typos in here.

qiita_db/handlers/oauth2.py
```diff
@@ -0,0 +1,280 @@
...
+    an invalid header is given, the json error message is automatically sent.
```
josenavas: Can you also comment that a 400 is returned?

qiita_db/handlers/oauth2.py
```diff
@@ -0,0 +1,280 @@
...
+    """
+    @functools.wraps(f)
+    def wrapper(self, *args, **kwargs):
+        def oauth_error(self, error_msg, error):
```
josenavas: Why are you re-defining `oauth_error` here? Can't you use the same already defined in the class?

qiita_db/handlers/oauth2.py
```diff
@@ -0,0 +1,280 @@
...
+        qdb.sql_connection.TRN.add(sql, sql_info)
+        return qdb.sql_connection.TRN.execute_fetchlast()
+
+
+def authenticate_oauth(f):
+    """Decorate methods with this to require valid Oauth2 Authorization header
+
+    If a valid header is given, teh handof is odne and the page is rendered. If
+    an invalid header is given, the json error message is automatically sent.
+    """
```
josenavas: A link to the actual oauth2 documentation would be useful here

qiita_db/handlers/oauth2.py
```diff
@@ -0,0 +1,280 @@
...
+                            'invalid_grant')
+                return
+
+        return f(self, *args, **kwargs)
+    return wrapper
+
+
+class OauthBaseHandler(RequestHandler):
+    def write_error(self, status_code, **kwargs):
+        """Overriding the default write error in tornado RequestHandler"""
```
josenavas: Can you add full numpy documentation, as well as why it is needed to rewrite the one from tornado?

qiita_db/handlers/oauth2.py
```diff
@@ -0,0 +1,280 @@
...
+        if status_code in {403, 404, 405}:
+            # We don't need to log these failues in the logging table
+            return
+        # log the error
+        from traceback import format_exception
```
josenavas: Why is this import done here rather than at the top of the file?

qiita_db/handlers/oauth2.py
```diff
@@ -0,0 +1,280 @@
...
+    def oauth_error(self, error_msg, error='invalid_request'):
+        self.set_status(400)
+        self.write({'error': error,
+                    'error_description': error_msg})
+        self.finish()
+
+    def check_rate_limit(self, client_id, user):
+        limit_key = '%s_%s_daily_limit' % (client_id, user)
+        limiter = r_client.get(limit_key)
+        if limiter is None:
```
josenavas: Why this limitation of 5000 requests per day? I'm unsure we want to limit the number of requests that a user does a day - and this is totally breaking the plugins...

qiita_db/handlers/oauth2.py
```diff
@@ -0,0 +1,280 @@
...
+
+
+class TokenAuthHandler(OauthBaseHandler):
+    def generate_access_token(self):
+        """Creates the random alphanumeric token
+
+        Returns
+        -------
+        str
+            55 character random alphanumeric string
```
josenavas: Where does the 55 come from? It is required by oauth2, please link the docs.

qiita_db/handlers/oauth2.py
```diff
@@ -0,0 +1,280 @@
...
+        """
+        pool = ascii_letters + digits
+        return ''.join((SystemRandom().choice(pool) for _ in range(55)))
```
josenavas: Why `SystemRandom()` instead of simply `choice`?

qiita_db/handlers/oauth2.py
```diff
@@ -0,0 +1,280 @@
...
+
+        Parameters
+        ----------
+        token: str
+            Random token string for authorization
+        client_id : str
+            Client that requested the token
+        user : str, optional
+            If password grant type requested, the user requesting the key.
+        timeout : int
```
josenavas: Missing `optional`. Also, suggest increasing timeout to 1 day - better for plugins.

qiita_db/handlers/oauth2.py
```diff
@@ -0,0 +1,280 @@
...
+        r_client.hset(token, 'timestamp', datetime.datetime.now())
+        r_client.hset(token, 'client_id', client_id)
+        r_client.expire(token, timeout)
+        if user:
+            r_client.hset(token, 'user', user)
+            # Check if client has access limit key, and if not, create it
+            limit_key = '%s_%s_daily_limit' % (client_id, user)
+            limiter = r_client.get(limit_key)
+            if limiter is None:
+                # Set limit to 5,000 requests per day
```
josenavas: This seems redundant? Isn't it already set up above?

qiita_db/handlers/oauth2.py
```diff
@@ -0,0 +1,280 @@
...
+    def validate_client(self, client_id, client_secret):
+        """Make sure client exists, then set the token and send it
+
+        Parameters
+        ----------
+        client_id : str
+            The client making the request
+        client_secret : str
+            The secret key for the client
+        """
```
josenavas: Although this doesn't have a `return`, we have been documenting the API returns in the `returns` sections of the numpy documentation so we know how the API looks like.

qiita_db/handlers/oauth2.py
```diff
@@ -0,0 +1,280 @@
...
+
+        Parameters
+        ----------
+        username : str
+            The username to validate
+        password : str
+            The password for the username
+        client_id : str
+            The client making the request
+        """
```
josenavas: Same here re documenting returns

qiita_db/handlers/oauth2.py
```diff
@@ -0,0 +1,280 @@
...
+        try:
+            qdb.user.User.login(username, password)
+        except (IncorrectEmailError, IncorrectPasswordError,
+                UnverifiedEmailError):
+            self.oauth_error('Oauth2 error: invalid user information',
+                             'invalid_client')
+            return
+
+        if login_client(client_id):
```
josenavas: This code and the one in `validate_client` are the same - can you decompose in another function so we avoid code duplication?

qiita_db/handlers/oauth2.py
```diff
@@ -0,0 +1,280 @@
...
+            self.write({'access_token': token,
+                        'token_type': 'Bearer',
+                        'expires_in': '3600'})
+            self.finish()
+            self.set_token(token, client_id, user=username)
+        else:
+            self.oauth_error('Oauth2 error: invalid client information',
+                             'invalid_client')
+
+    def post(self):
```
josenavas: Just to double check on how oauth2 works. You get a new token by posting? It looks counter-intuitive.

squirrelo: Yes, that is how it works.

qiita_db/handlers/tests/test_oauth2.py
```diff
@@ -0,0 +1,294 @@
...
+        self.assertEqual(loads(obs.body), exp)
+
+    def test_authenticate_header_username(self):
+        obs = self.get('/qiita_db/artifacts/100/mapping/', headers={
+            'Authorization': 'Bearer ' + self.user_token})
+        self.assertEqual(obs.code, 200)
+        exp = {'success': False, 'error': 'Artifact does not exist',
+               'mapping': None}
+        self.assertEqual(loads(obs.body), exp)
+        # Check rate limiting works
```
josenavas: Split this in another test.

qiita_db/handlers/tests/test_oauth2.py
```diff
@@ -0,0 +1,294 @@
...
+
+    def test_authenticate_header_bad_info(self):
+        obs = self.get('/qiita_db/artifacts/100/mapping/', headers={
+            'Authorization': 'Bearer BADTOKEN'})
+        self.assertEqual(obs.code, 400)
+        exp = {'error': 'invalid_grant',
+               'error_description': 'Oauth2 error: token has timed out'}
+        self.assertEqual(loads(obs.body), exp)
+
+        obs = self.get('/qiita_db/artifacts/100/mapping/', headers={
```
josenavas: Split this in another test

qiita_db/handlers/tests/test_oauth2.py
```diff
@@ -0,0 +1,294 @@
...
+        self.assertEqual(len(obs_body['access_token']), 55)
+        self.assertEqual(type(obs_body['access_token']), unicode)
+
+        # Make sure token in system with proper ttl
+        token = r_client.hgetall(obs_body['access_token'])
+        self.assertNotEqual(token, {})
+        self.assertItemsEqual(token.keys(), ['timestamp', 'client_id'])
+        self.assertEqual(r_client.ttl(obs_body['access_token']), 3600)
+
+        # Authenticate using post only
```
josenavas: Split in another test

qiita_db/handlers/tests/test_oauth2.py
```diff
@@ -0,0 +1,294 @@
...
+        self.assertEqual(len(obs_body['access_token']), 55)
+        self.assertEqual(type(obs_body['access_token']), unicode)
+
+        # Make sure token in system with proper ttl
+        token = r_client.hgetall(obs_body['access_token'])
+        self.assertNotEqual(token, {})
+        self.assertItemsEqual(token.keys(), ['timestamp', 'client_id'])
+        self.assertEqual(r_client.ttl(obs_body['access_token']), 3600)
+
+    def test_authenticate_client_bad_info(self):
```
josenavas: Split this in multiple tests - please see my email from Oct 20th about using AAA for tests.

qiita_db/handlers/tests/test_oauth2.py
```diff
@@ -0,0 +1,294 @@
...
+                'grant_type': 'client',
+                'client_id': '19ndkO3oMKsoChjVVWluF7QkxHRfYhTKSFbAVt8IhK7gZgDa'
+                             'O4'})
+        self.assertEqual(obs.code, 400)
+        obs_body = loads(obs.body)
+        exp = {'error': 'invalid_request',
+               'error_description': 'Oauth2 error: missing client information'}
+        self.assertEqual(obs_body, exp)
+
+    def test_authenticate_password(self):
```
josenavas: Same here - last time I note this, applies to all.

qiita_db/handlers/tests/test_processing_job.py
```diff
@@ -35,15 +36,23 @@ def test_get_job(self):
 
 
 class JobHandlerTests(TestHandlerBase):
+    def setUp(self):
```
josenavas: I've seen this construction quite a few times already - I think it is worth to move it to a base class setUp so we avoid code duplication.

qiita_db/handlers/oauth2.py
squirrelo: This is meant ot be used on anything that extends the RequestHandler, not just the OauthBaseHandler, so I need the function available outside the class. I'll rework to allow the class and this decorator to use the same function.

qiita_db/handlers/tests/test_processing_job.py
```diff
@@ -192,7 +240,7 @@ def test_post_job_success(self):
              ]})
         obs = self.post(
             '/qiita_db/jobs/bcc7ebcd-39c1-43e4-af2d-822e3589f14d/complete/',
-            payload)
+            payload, headers={'Authorization': 'Bearer ' + self.token})
```
josenavas: To avoid more code duplication - in the base class you can add `self.headers = {'Authorization': 'Bearer ' + self.token}`, so we only need to do headers=headers in these calls. Either that or add a custom post that will help us to reduce all these calls...

qiita_db/handlers/oauth2.py
squirrelo: this was copy pasted from the original base handler, so I don't recall why it was done this way. I can update this one and the other base handler too I guess.

qiita_db/handlers/oauth2.py
squirrelo: Seemed like a sane limit. Google gives you 2500 requests per day, so on a per-user basis 5,000 seemed appropriate. Note that this limit only applies to users, not client authenticated models, so plugins still have unlimited calls.

qiita_db/handlers/oauth2.py
josenavas: If this is copy-paste, shouldn't this point out that we are doing something wrong? i.e. we are duplicating code...

qiita_db/handlers/oauth2.py
josenavas: ok

josenavas:
Few comments

qiita_db/handlers/oauth2.py
josenavas: It looks like the code is written so if you need oauth2 you need to extend from OauthBaseHandler, e.g. you're calling check_rate_limit below.

qiita_db/handlers/oauth2.py
squirrelo: It's the same functionality, but just changed enough that we can't make it universal as a parent class.

qiita_db/handlers/oauth2.py
squirrelo: Yes, just cauth that. Removing that will be part of the rework, as that is the only place it is used.

qiita_db/handlers/oauth2.py
josenavas: Please, do not put any kind of such hard limitation unless discussed with the group.

qiita_db/handlers/oauth2.py
squirrelo: `SystemRandom` is cryptographically secure as it changes per system, `choice` is not.

qiita_db/handlers/oauth2.py
squirrelo: 1 day is too long for a registration token. The idea is that tokens are very short lived, so if they are compromised their usefulness is in a very small window. With plugins, which have full access to everything and an unlimited amount of calls, this is especially important.

qiita_db/handlers/oauth2.py
squirrelo: Technically this is the first set up here. The one above is in case the access token's validity spans when the rate limit expires, and we need to reset the rate limit key during the lookup of token validity instead of the initial creation of the token.

qiita_db/handlers/oauth2.py
squirrelo: I'm unsure what this means. Could you elaborate?

qiita_db/handlers/oauth2.py
josenavas: ok, thanks!

qiita_db/handlers/oauth2.py
squirrelo: Again, a cryptographically secure limit that needs to be hard coded so we can set the same varchar length for the postgres column storing the key.

qiita_db/handlers/oauth2.py
josenavas: Do you have links that support that 1 day is too long for registration token? I would like to have at lest more than a single our, to avoid the plugins requesting it multiple times... Most of our jobs are longer than an hour.

qiita_db/handlers/oauth2.py
josenavas: Check the documentation of the other handlers. In the returns section we are including the explanation of the JSON that you're returning.

qiita_db/handlers/oauth2.py
josenavas: Again, magic numbers are not useful w/o some context for other devs. You may know what they're for but not other devs, which makes this code hard to maintain. If that is something that you decided, at least add a comment, specially because 55 looks random...

qiita_db/handlers/oauth2.py
squirrelo: That does not fit with the standard numpy documentation methods. If a change like this happened, it should have been discussed at the meetings.

qiita_db/handlers/oauth2.py
squirrelo: I don't have specific links to what the expiration should be, as that is not part of the oauth2 spec, but the consensus on stack overflow and others is  "Very short lived". Facebook, for example, has also taken this to mean 1 hour. However, from [their own documentation](https://developers.facebook.com/docs/facebook-login/access-tokens)
`You should not depend on these lifetimes remaining the same - the lifetime may change without warning or expire early. See more under handling errors.` So client systems need to be robust enough to handle changes in expiration times anyway, and we should be able to handle that. Also, if the client plugins can take care of a multi-day process key timing out, it is no different then a 1 hour key timing out in terms of what the client needs to do.

josenavas:
@squirrelo do you have an ETA for the changes comments being addressed here?

squirrelo:
Working on it now, should be done by end of day.

qiita_db/handlers/oauth2.py
squirrelo: done

qiita_db/handlers/tests/test_oauth2.py
squirrelo: done

qiita_db/handlers/tests/test_oauth2.py
squirrelo: done

squirrelo:
Changes made, save ones we need to discuss as a group.

qiita_db/handlers/oauth2.py
```diff
@@ -0,0 +1,306 @@
...
+                               (k, req_dict[k]) for k in
+                                req_dict.keys() if k != 'files'])
+        error = exc_info[1]
+        qdb.logger.LogEntry.create(
+            'Runtime',
+            'ERROR:\n%s\nTRACE:\n%s\nHTTP INFO:\n%s\n' %
+            (error, trace_info, request_info))
+
+    # Allow call to oauth_error function as part of the class
+    oauth_error = _oauth_error
```
josenavas: ?! Define the function here - you can then call the function in the decorator, as you are calling other methods/attributes of the class.

squirrelo: As stated previously, the decorator function is for anything extending tornado's `RequestHandler`, not specifically the  oauth handler, and the changes just pushed reflect that. Calling the function in the decorator breaks this functionality.

josenavas: I'm getting really confused here. As you stated here, the `authenticate_oauth` is written so it works with any `RequestHandler`. On the other hand, you created the base class `OauthBaseHandler` that all the handlers that require oauth2 need to subclass. The question then is, is the OauthBaseHandler class needed or the authenticate_oauth really need to be as general for any of the Handlers type?

squirrelo: The `OauthBaseHandler` is for anything that needs to actually set and manipulate the authorization keys in the system, or authenticate user/client information, so not every oauth2 using page needs to subclass it. I did it for convenience, as there are things like 404 and error handling that are better used in base classes, but it's not strictly needed. Leaving the decorator open like this allows extensibility in the future with no code change, as all you need to do is import the `authenticate_oauth` decorator and extend `RequestHandler` again, exactly the same as the tornado library already works.

qiita_db/handlers/oauth2.py
```diff
@@ -0,0 +1,306 @@
...
+        """Creates the random alphanumeric token
+
+        Returns
+        -------
+        str
+            55 character random alphanumeric string
+
+        Notes
+        -----
+        55 was chosen as a cryptographically secure limit that needs to be
```
josenavas: I kind of understand this explanation. However I still don't understand why 55. Why we don't use 100? wouldn't it be more secure?

qiita_db/handlers/oauth2.py
```diff
@@ -0,0 +1,306 @@
...
+        if client_secret is not None:
+            sql = sql.format("AND client_secret = %s")
+            sql_info.append(client_secret)
+        else:
+            sql = sql.format("AND client_secret IS NULL")
+        qdb.sql_connection.TRN.add(sql, sql_info)
+        return qdb.sql_connection.TRN.execute_fetchlast()
+
+
+def _oauth_error(self, error_msg, error):
```
josenavas: Can you change `self` here with something different? Like `handler`. Self doesn't make sense in a function like this.

josenavas: Actually, thinking a bit more about this, shouldn't this just simply raise an HTTPError, so we don't need our own function? Based on the [docs](http://www.tornadoweb.org/en/stable/web.html#tornado.web.HTTPError) it looks like it will do everything that we need.

josenavas: Looking a bit more around - if this is authentication, should this behave similarly as the authenticate from tornado, reference code [here](http://www.tornadoweb.org/en/stable/_modules/tornado/web.html#authenticated) the error code should be 403 right?

squirrelo: This is built with ides from the tornado decorator, but pursuant to the Oauth2 framework linked:
```
 The authorization server responds with an HTTP 400 (Bad Request)
   status code (unless specified otherwise) and includes the following
   parameters with the response:
```
So this needs to return 400 error code and the controlled vocab error json in the body.

qiita_db/handlers/oauth2.py
squirrelo: 62^55 = 3.815425e+98 which is probably secure enough, given that we also have request and rate limits.

josenavas: I think you did not understand the point that I was trying to make. It looks like 55 is secure enough but 54 isn't? That is the point that I want to make, if you decided 55 based on some documentation that you read online, yo should link it, if it is something that you decided, you should have some other founding that 'looks like secure enough'. 55 is not a multiple of 2 and looks like strange number - this is too arbitrary and the same questions that I have are the same questions that another dev in 2 years can have, and we may or may not be around to answer those questions. So please, include the needed documentation so a dev that looks at the system for the first time understands what is going on.

squirrelo: I can update the database column to store 255 characters (max you can safely put in an HTTP variable), but as stated there is no documentation to cite for length. The number is not specified as part of the Oauth2 spec, so it's up to the implementer to decide.

antgonza:
Can we close this one? I think it has been superseded by https://github.com/biocore/qiita/pull/1584 ..

squirrelo:
It would be much easier to review in two parts, as #1584 becomes monolithic with this PR merged in as well.

antgonza:
OK, reviewing now.

qiita_db/handlers/oauth2.py
```diff
@@ -0,0 +1,306 @@
...
+    If an invalid header is given, a 400 error code is returned and the json
+    error message is automatically sent.
+
+    References
+    ---------
+    [1] The OAuth 2.0 Authorization Framework.
+    http://tools.ietf.org/html/rfc6749
+    """
+    @functools.wraps(f)
+    def wrapper(self, *args, **kwargs):
```
antgonza: Should this have a doc string?

qiita_db/handlers/oauth2.py
```diff
@@ -0,0 +1,306 @@
...
+        header = self.request.headers.get('Authorization', None)
+        if header is None:
```
squirrelo: It's the inner wrapper of the decorator, so no. The docstring of the containing`authenticate_oauth`  function is what's going to be seen.

qiita_db/handlers/oauth2.py
squirrelo: It's the inner wrapper of the decorator, so no. The docstring of the containingauthenticate_oauth function is what's going to be seen.

qiita_db/handlers/oauth2.py
```diff
@@ -0,0 +1,306 @@
...
+            _oauth_error(self, 'Oauth2 error: invalid access token',
+                         'invalid_request')
+            return
+        token_info = header.split()
+        if len(token_info) != 2 or token_info[0] != 'Bearer':
```
antgonza: Why are you looking for 'Bearer', could you clarify via documentation? Perhaps this is something obvious but someone without previous experience, me, it's not.

qiita_db/handlers/oauth2.py
josenavas: @squirrelo ping about the first comment in here - change `self` to `handler`. There is no self here.

qiita_db/handlers/oauth2.py
```diff
@@ -0,0 +1,306 @@
...
+
+def authenticate_oauth(f):
+    """Decorate methods to require valid Oauth2 Authorization header[1]
+
+    If a valid header is given, the handoff is done and the page is rendered.
+    If an invalid header is given, a 400 error code is returned and the json
+    error message is automatically sent.
+
+    References
+    ---------
```
josenavas: Missing 1 `-`.

qiita_db/handlers/oauth2.py
squirrelo: Required by the OAuth2 authorization framework, which is explicitly linked in the docstring.

antgonza: I guess these 2 comments show how is not fully clear how this works and expecting that everybody reads a full RFC seems unrealistic, what about adding the main assumptions/cases you are using here so is easier to fix/improve in the future?

qiita_db/handlers/oauth2.py
josenavas: Is there any specific necessity of calling `oauth_error` as `self.oauth_error`? If not, remove this line and and just call the function `_oauth_error` as `_oauth_error(self, params...)`. This is just making maintaining the code harder.

qiita_db/handlers/oauth2.py
josenavas: I still don't understand the decision of the 55 chars. The only answer that I received here is "I have the feeling that is secure enough". What are other systems using?

qiita_db/handlers/oauth2.py
```diff
@@ -0,0 +1,306 @@
...
+        else:
+            self.oauth_error('Oauth2 error: invalid client information',
+                             'invalid_client')
+
+    def post(self):
+        # first check for header version of sending auth, meaning client ID
+        header = self.request.headers.get('Authorization', None)
+        if header is not None:
+            header_info = header.split()
+            if header_info[0] != 'Basic':
```
antgonza: What's Basic and why is invalid ?

qiita_db/handlers/oauth2.py
josenavas: Also, I just noticed that the description here is totally incorrect. As far as I know, the `access_token` is never stored in the postgres DB, cause you are storing it in redis. Please, update the documentation accordingly and do some research on what other systems are using, and link it here.

qiita_db/handlers/oauth2.py
squirrelo: I'd argue you shouldn't be modifying code that must conform to a very specific set of rules laid out in the RFC unless you have read and understand the RFC. This is a web-wide standard we did not write, and it is very explicit on what it requires, which allows for things like the libraries used to do client verification in #1584.

antgonza: I feel like having the code well documented doesn't hurt any one ... do you feel otherwise?

squirrelo: Good documentation is one thing. Overly documenting is another, and frowned upon by pep8. What do you gain from having the documentation lines like `Make sure Auth header passed conforms to what is expected by Oauth2 standard`?

qiita_db/handlers/oauth2.py
josenavas: Please add numpy documentation in the post. Also, as it was agreed over email on documenting the RESTapi handlers (@ElDeveloper, @antgonza and I agreed, and no body complain), add the `Returns` section including the json response of the query. If you don't agree with this solution, please propose an alternative one, but we are not going to merge undocumented code.

qiita_db/handlers/oauth2.py
antgonza: Agree that your example doesn't help - not even sure why will you use it - but I was expecting something more like:
`Based on RFC6750 if reply is not 2 elements in the format of: ['Bearer', token] we assume a wrong reply.`
or something like this. Simple, easy, and no need to read lots of pages.

josenavas: Agree with @antgonza on adding a bit more points of documentation. @squirrelo we are not asking to duplicate the entire RFC, but you can't expect that every single developer will read the RFC in order to understand the code that is written here.

As an example of documentation that I would expect having here can be seen here: https://github.com/idan/oauthlib/blob/master/oauthlib/oauth2/rfc6749/request_validator.py

Also, looking at that - the question that raises is why did you re-implement the RFC if the oauthlib already includes an implementation-independent framework for OAuth2? We shouldn't be reinventing the wheel, specially on topics that we are not experts.

qiita_db/handlers/oauth2.py
```diff
@@ -0,0 +1,306 @@
...
+from base64 import urlsafe_b64decode
+from string import ascii_letters, digits
+import datetime
+from random import SystemRandom
+import functools
+from traceback import format_exception
+
+from tornado.web import RequestHandler
+
+from moi import r_client
```
josenavas: Group the import from moi with tornado rather than with qiita (3rd party lib group)

qiita_db/handlers/reference.py
```diff
@@ -6,7 +6,7 @@
 # The full license is in the file LICENSE, distributed with this software.
 # -----------------------------------------------------------------------------
 
-from tornado.web import RequestHandler
+from .oauth2 import OauthBaseHandler, authenticate_oauth
```
josenavas: remove extra blank line

qiita_db/handlers/tests/oauthbase.py
```diff
@@ -0,0 +1,12 @@
+from moi import r_client
+from qiita_pet.test.tornado_test_base import TestHandlerBase
```
josenavas: Add blank line

qiita_db/handlers/tests/test_oauth2.py
```diff
@@ -0,0 +1,300 @@
+# -----------------------------------------------------------------------------
+# Copyright (c) 2014--, The Qiita Development Team.
+#
+# Distributed under the terms of the BSD 3-clause License.
+#
+# The full license is in the file LICENSE, distributed with this software.
+# -----------------------------------------------------------------------------
+from unittest import main
+from json import loads
+from moi import r_client
```
josenavas: Add blank line above and below of the `moi` import

qiita_db/handlers/tests/test_oauth2.py
josenavas: This doesn't look it has been split.

qiita_db/handlers/tests/test_oauth2.py
```diff
@@ -0,0 +1,300 @@
...
+        obs = self.get('/qiita_db/artifacts/100/mapping/', headers={
+            'Authorization': 'WRONG ' + self.client_token})
+        self.assertEqual(obs.code, 400)
+        exp = {'error': 'invalid_grant',
+               'error_description': 'Oauth2 error: invalid access token'}
+        self.assertEqual(loads(obs.body), exp)
+
+
+class OAuth2HandlerTests(TestHandlerBase):
+    def test_authenticate_client(self):
```
josenavas: Split tests.

qiita_db/handlers/tests/test_oauth2.py
josenavas: Ignore this, the make me think it was a different test.

qiita_db/handlers/oauth2.py
rob-knight: I agree with @antgonza and @josenavas. Documentation is essential and it doesn't make sense to require every reader of the code to be an expert on topics they use infrequently. @squirrelo, please work with the rest of the group on this productively in this and similar instances -- part of the reason why we're so productive is precisely that the good documentation means people don't have to waste a lot of additional time looking up arcane topics.

qiita_db/handlers/oauth2.py
```diff
@@ -0,0 +1,306 @@
...
+
+
+def login_client(client_id, client_secret=None):
+    """Login client_id, optionally with client_secret
+
+    Parameters
+    ----------
+    client_id : str
+        The client ID to validate
+    client_secret : str, optional
```
josenavas: I don't think this is actually correct, neither if we want to support it. According to the RFC6749:
```
client_secret
         REQUIRED.  The client secret.  The client MAY omit the
         parameter if the client secret is an empty string.
```
According to this description, the only time that the parameter can be omitted is that if the `client_secret` of the client is an empty string. Here, it will validate the client correctly as long as the client_id is the same. That means, the only thing that I need to impersonate somebody else is using his client_id, I don't need this `client_secret` to gain access to the system. All our clients are going to have `client_id` and `client_secret` and `client_secret` will never be an empty string, given the way they're generated.

qiita_db/handlers/oauth2.py
squirrelo: The explanation is still there is no set size for the key, and it must arbitrarily be decided by the implementer because there is no standard. [Twitter](https://dev.twitter.com/oauth/reference/post/oauth2/token) seems to be using 21 characters although this is an example so it may be a different length, [oauthlib](https://github.com/idan/oauthlib/blob/master/oauthlib/common.py#L220) seems to be using 30 characters by default, sometimes more, etc. It is completely arbitrary, and most systems don't even say how long the code is because it may change length at any time. The only fixture is that it can not be more than 255 characters for HTTP transfer reasons.

qiita_db/handlers/oauth2.py
squirrelo: `client_secret` as optional is so it can also be used to check client_id for the `Resource Owner Password Credentials` login type, which only requires client ID and is in the database as a client ID with a NULL client secret.

qiita_db/handlers/oauth2.py
josenavas: Ok, can you actually add the comment that the exact length of the token can change, but it should never be longer than 255? That is a technology limit and it will be good to have it exposed here in the documentation.

I don't see any check in here but, what happens if the token that is generated already exists in the system? The RFC does not say anything specific, but I think we should double check that each new access_token is unique, or am I missing something?

squirrelo: Done

qiita_db/handlers/oauth2.py
josenavas: I had to read the entire code to understand why this is correct. My take on this is that this is really dangerous and it is really easy to introduce bugs in the code.

Now, from the code structure correctness, this function is used in 2 places, the function `validate_resource_owner` and the function `validate_client`. In this function, you have an if statement, which the only thing that is doing is differentiating the 2 cases that is calling this function. This exposes that there is no functionality in common and this function should've not been decomposed in a single function - I'm not even sure that is saving lines of code...

Can you remove this function and just write the call in the given functions?

qiita_db/handlers/reference.py
squirrelo: breaks pep8, can't.

squirrelo: sorry, misread. Done.

squirrelo:
Comments addressed

qiita_db/handlers/reference.py
josenavas: It is really confusing that it breaks pep8 - as that is blank line is
specific defined by pep8. is it because you are using relative imports? If
so change to global.

2015-12-21 10:33 GMT-08:00 Joshua Shorenstein <notifications@github.com>:

> In qiita_db/handlers/reference.py
> <https://github.com/biocore/qiita/pull/1562#discussion_r48175040>:
>
> > @@ -6,7 +6,7 @@
> >  # The full license is in the file LICENSE, distributed with this software.
> >  # -----------------------------------------------------------------------------
> >
> > -from tornado.web import RequestHandler
> > +from .oauth2 import OauthBaseHandler, authenticate_oauth
> >
>
> breaks pep8, can't.
>
> —
> <https://github.com/biocore/qiita/pull/1562/files#r48175040>.
>



-- 
Jose Navas

josenavas: Oh - thanks!

2015-12-21 10:33 GMT-08:00 Joshua Shorenstein <notifications@github.com>:

> In qiita_db/handlers/reference.py
> <https://github.com/biocore/qiita/pull/1562#discussion_r48175070>:
>
> > @@ -6,7 +6,7 @@
> >  # The full license is in the file LICENSE, distributed with this software.
> >  # -----------------------------------------------------------------------------
> >
> > -from tornado.web import RequestHandler
> > +from .oauth2 import OauthBaseHandler, authenticate_oauth
> >
>
> sorry, misread. Done.
>
> —
> <https://github.com/biocore/qiita/pull/1562/files#r48175070>.
>



-- 
Jose Navas

qiita_db/handlers/oauth2.py
```diff
@@ -0,0 +1,394 @@
...
+        ----------
+        username : str
+            The username to validate
+        password : str
+            The password for the username
+        client_id : str
+            The client making the request
+
+        Returns
+        -------
```
mortonjt: Maybe Returns is a misnomer here, since it doesn't actually return anything.

qiita_db/handlers/oauth2.py
```diff
@@ -0,0 +1,394 @@
...
+            Username to authenticate
+        password : str
+            Password for the username
+
+        If client grant type:
+        client_secret : str
+            client authentication secret
+
+        Returns
+        -------
```
mortonjt: Same here.

qiita_db/handlers/oauth2.py
squirrelo: That's what I was saying.

josenavas: @mortonjt in a previous discussion over email we agreed on documenting the RESTapi using the numpy returns.

We are happy to change to a different standard to document the RESTapi in the same code, but please provide alternatives - it is better to document like this than leave undocumented.

squirrelo: Using the examples section would be easier, especially since these all respond differently depending on errors, warnings, etc and that could be documented better.

josenavas:
:+1: @antgonza @mortonjt available for reviewing this?

PR #1564, opened 2015-12-07
WIP - Study main page

squirrelo:
This adds the first steps for the new study interface, and demos how it will be done, so please feel free to comment right now on the general idea of ajax handled pages. 

The idea is that the page is modularized into divs for specific information, and only the divs that need it are updated through ajax calls back to the main system. This allows a modularized page design like the original tabs information, but we don't have to build the entire page information at once and only have to build and send the changes needed for the page,  again instead of building the entire page with all information upfront. 

This also allows expensive things like the prep template artifact graph to be generated only the first time the prep template link is clicked, and does not change until then next prep template is clicked (essentially lazy load and then leave things in place). This allows us to save a lot of overhead by not recomputing this every time a different file in the prep template is clicked.


qiita_db/study.py
```diff
@@ -152,6 +152,19 @@ def _lock_non_sandbox(self):
...
+        list of str
+        """
+        with qdb.sql_connection.TRN:
+            sql = "SELECT DISTINCT data_type FROM qiita.data_type"
+            qdb.sql_connection.TRN.add(sql)
+            return qdb.sql_connection.TRN.execute_fetchflatten()
+
     @property
     def status(self):
         r"""The status is inferred by the status of its artifacts"""
```
mortonjt: What do you say about updating the docstring (considering that the return type as completely changed)?

qiita_db/study.py
```diff
@@ -263,7 +276,7 @@ def get_info(cls, study_ids=None, info_cols=None):
             if study_ids is not None and len(res) != len(study_ids):
                 raise qdb.exceptions.QiitaDBError(
                     'Non-portal-accessible studies asked for!')
-            return res
+            return [dict(r) for r in res]
```
mortonjt: Sorry, that comment meant to go here.

I think the docstring should be updated here.

qiita_pet/handlers/study_handlers/description_handlers.py
```diff
@@ -83,865 +60,81 @@ def _to_int(value):
...
-            pd_id = None
-        except Exception as e:
-            msg = ("Couldn't remove processed data %d: %s" %
-                   (pd_id, str(e)))
-            msg_level = "danger"
-
-        callback((msg, msg_level, 'processed_data_tab', pd_id, None))
+class StudyIndexHandler(BaseHandler):
+    def study_prep_proxy(self, study):
+        """Proxies expected json from the API for existing prep templates"""
```
mortonjt: Can we bring back the standard numpy doc strings?  It makes it a bit easier to understand what's going on here.

qiita_pet/handlers/study_handlers/description_handlers.py
```diff
@@ -83,865 +60,81 @@ def _to_int(value):
...
+        return data_types if data_types else []
+
+    def study_info_proxy(self, study):
+        """Proxies expected json from the API for base study info"""
+        study_info = Study.get_info([study.id])[0]
+
+        pi = StudyPerson(study_info["principal_investigator_id"])
+        study_info["principal_investigator"] = '%s (%s)' % (
+            pi.name, pi.affiliation)
+        del study_info["principal_investigator_id"]
```
mortonjt: I'm a bit confused. Why is "principal_investigator" information being filled in, then deleted?

squirrelo: Because just the ID comes from the database, not the actual studyperson object. Need to get the real info and replace the ID with the real info.

qiita_db/study.py
squirrelo: done

antgonza:
I will like to start testing this but there is conflicts and I'm worry about all the code that is being deleted, what's the plan?

PR #1571, opened 2015-12-09
Base for new study page

squirrelo:
This adds in the base HTML and index handler for the study pages, and a single example of the AJAX-call-based updates done on the page. The idea is that we load the minimal information to show the first page, and then lazy load each bit of information needed as it is called for, so we save a lot of time and database calls that would be duplicated if we just reloaded the page for every link click. It also means we no longer need to load all information for all files when the page is accessed.

### DIFF
--- a/qiita_db/study.py
+++ b/qiita_db/study.py
@@ -169,9 +169,11 @@ class Study(qdb.base.QiitaObject):
     @staticmethod
     def all_data_types():
         """Returns list of all the data types available in the system
+
         Returns
         -------
         list of str
+            All the data types available in the system
         """
         with qdb.sql_connection.TRN:
             sql = "SELECT DISTINCT data_type FROM qiita.data_type"
--- a/qiita_db/study.py
+++ b/qiita_db/study.py
@@ -512,8 +512,7 @@ class Study(qdb.base.QiitaObject):
             # remove non-info items from info
             for item in self._non_info:
                 info.pop(item)
-            # This is an optional column, but should not be considered part
-            # of the info
+            # removed because redundant to the id already stored in the object
             info.pop('study_id')
 
             if info['principal_investigator_id']:
--- a/qiita_db/study.py
+++ b/qiita_db/study.py
@@ -166,6 +166,18 @@ class Study(qdb.base.QiitaObject):
             return qdb.util.infer_status(
                 qdb.sql_connection.TRN.execute_fetchindex())
 
+    @staticmethod
+    def all_data_types():
+        """Returns list of all the data types available in the system
+        Returns
+        -------
+        list of str
+        """
+        with qdb.sql_connection.TRN:
+            sql = "SELECT DISTINCT data_type FROM qiita.data_type"
+            qdb.sql_connection.TRN.add(sql)
+            return qdb.sql_connection.TRN.execute_fetchflatten()
+
     @classmethod
     def get_by_status(cls, status):
         """Returns study id for all Studies with given status
@@ -503,6 +515,21 @@ class Study(qdb.base.QiitaObject):
             # This is an optional column, but should not be considered part
             # of the info
             info.pop('study_id')
+
+            if info['principal_investigator_id']:
+                info['principal_investigator'] = qdb.study.StudyPerson(
+                    info["principal_investigator_id"])
+            else:
+                info['principal_investigator'] = None
+            del info['principal_investigator_id']
+
+            if info['lab_person_id']:
+                info['lab_person'] = qdb.study.StudyPerson(
+                    info["lab_person_id"])
+            else:
+                info['lab_person'] = None
+            del info['lab_person_id']
+
             return info
 
     @info.setter
--- a/qiita_pet/handlers/api_proxy.py
+++ b/qiita_pet/handlers/api_proxy.py
@@ -9,7 +9,7 @@ from __future__ import division
 
 # This is the only file in qiita_pet that should import from outside qiita_pet
 # The idea is that this proxies the call and response dicts we expect from the
-# QIITA API once we build it. This will be removed and replaced with API calls
+# Qiita API once we build it. This will be removed and replaced with API calls
 # when the API is complete.
 from qiita_core.qiita_settings import qiita_config
 from qiita_db.study import Study
@@ -37,7 +37,7 @@ def _approve(level):
 
 
 class StudyAPIProxy(BaseHandler):
-    """Adds API proxy functions to the handler. Can be removed once the restful
+    """Adds API proxy functions to the handler. Can be removed once the RESTful
        API is in place."""
     def study_prep_proxy(self, study_id):
         """Proxies expected json from the API for existing prep templates
@@ -47,10 +47,16 @@ class StudyAPIProxy(BaseHandler):
         study_id : int
             Study id to get prep template info for
 
-        Returns:
+        Returns
+        -------
         dict of list of dict
             prep template information seperated by data type, in the form
             {data_type: [{prep 1 info dict}, ....], ...}
+
+        Raises
+        ------
+        HTTPError
+            Raises code 403 if user does not have access to the study
         """
         # Can only pass ids over API, so need to instantiate object
         study = Study(study_id)
@@ -90,7 +96,8 @@ class StudyAPIProxy(BaseHandler):
         study_id : int
             Study id to get prep template info for
 
-        Returns:
+        Returns
+        -------
         dict of list of dict
             prep template information seperated by data type, in the form
             {data_type: [{prep 1 info dict}, ....], ...}
--- a/qiita_pet/handlers/api_proxy.py
+++ b/qiita_pet/handlers/api_proxy.py
@@ -86,7 +86,7 @@ class StudyAPIProxy(BaseHandler):
             Data types available on the system
         """
         data_types = Study.all_data_types()
-        return data_types if data_types else []
+        return data_types
 
     def study_info_proxy(self, study_id):
         """Proxies expected json from the API for base study info
@@ -101,6 +101,11 @@ class StudyAPIProxy(BaseHandler):
         dict of list of dict
             prep template information seperated by data type, in the form
             {data_type: [{prep 1 info dict}, ....], ...}
+
+        Raises
+        ------
+        HTTPError
+            Raises code 403 if user does not have access to the study
         """
         # Can only pass ids over API, so need to instantiate object
         study = Study(study_id)
@@ -110,6 +115,8 @@ class StudyAPIProxy(BaseHandler):
         study_info['publications'] = study.publications
         study_info['study_id'] = study.id
         study_info['study_title'] = study.title
+        study_info['shared_with'] = [s.id for s in study.shared_with]
+        study_info['status'] = study.status
 
         # Clean up StudyPerson objects to string for display
         pi = study_info["principal_investigator"]
--- a/qiita_pet/handlers/api_proxy.py
+++ b/qiita_pet/handlers/api_proxy.py
@@ -9,7 +9,8 @@ from __future__ import division
 
 # This is the only file in qiita_pet that should import from outside qiita_pet
 # The idea is that this proxies the call and response dicts we expect from the
-# QIITA API once we build it.
+# QIITA API once we build it. This will be removed and replaced with API calls
+# when the API is complete.
 from qiita_core.qiita_settings import qiita_config
 from qiita_db.study import Study
 
@@ -65,7 +66,7 @@ class StudyAPIProxy(BaseHandler):
                     'status': prep.status,
                     'start_artifact': start_artifact.artifact_type,
                     'start_artifact_id': start_artifact.id,
-                    'last_artifact': 'TODO'
+                    'last_artifact': 'TODO new gui'
                 }
                 prep_info[dtype].append(info)
         return prep_info
@@ -99,7 +100,7 @@ class StudyAPIProxy(BaseHandler):
         check_access(self.current_user, study, raise_error=True)
         study_info = study.info
         # Add needed info that is not part of the initial info pull
-        study_info['publication_doi'] = [p[0] for p in study.publications]
+        study_info['publications'] = study.publications
         study_info['study_id'] = study.id
         study_info['study_title'] = study.title
 
--- a/qiita_pet/handlers/study_handlers/api_proxy.py
+++ b/qiita_pet/handlers/study_handlers/api_proxy.py
@@ -7,6 +7,9 @@
 # -----------------------------------------------------------------------------
 from __future__ import division
 
+# This is the only file in qiita_pet that should import from outside qiita_pet
+# The idea is that this proxies the call and response dicts we expect from the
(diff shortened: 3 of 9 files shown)


qiita_db/study.py
```diff
@@ -503,6 +515,21 @@ def info(self):
             # This is an optional column, but should not be considered part
             # of the info
             info.pop('study_id')
```
antgonza: If optional and not present this is going to fail.

qiita_pet/handlers/study_handlers/api_proxy.py
```diff
@@ -0,0 +1,116 @@
...
+# Copyright (c) 2014--, The Qiita Development Team.
+#
+# Distributed under the terms of the BSD 3-clause License.
+#
+# The full license is in the file LICENSE, distributed with this software.
+# -----------------------------------------------------------------------------
+from __future__ import division
+
+# This is the only file in qiita_pet that should import from outside qiita_pet
+# The idea is that this proxies the call and response dicts we expect from the
```
antgonza: and response dicts -> and responses the dicts

qiita_pet/handlers/study_handlers/api_proxy.py
```diff
@@ -0,0 +1,116 @@
...
+# QIITA API once we build it.
```
antgonza: remove: once we build it

qiita_pet/handlers/study_handlers/api_proxy.py
```diff
@@ -0,0 +1,116 @@
...
+            prep_info[dtype] = []
+            for prep in study.prep_templates(dtype):
+                start_artifact = prep.artifact
+                info = {
+                    'name': 'PREP %d NAME' % prep.id,
+                    'id': prep.id,
+                    'status': prep.status,
+                    'start_artifact': start_artifact.artifact_type,
+                    'start_artifact_id': start_artifact.id,
+                    'last_artifact': 'TODO'
```
antgonza: Suggest to add `TODO new gui` or something similar and keeping it consistent so we can search at the end.

qiita_pet/handlers/util.py
```diff
@@ -64,3 +64,28 @@ def download_link_or_path(is_local_request, filepath, fp_id, label):
...
+
+    Raises
+    ------
+    HTTPError
+        If `value` cannot be transformed to an integer
+    """
+    try:
+        res = int(value)
+    except ValueError:
+        raise HTTPError(500, "%s cannot be converted to an integer" % value)
```
antgonza: Not sure if this is something we want to be specific about or simply say: "Not valid study identifier"

qiita_pet/templates/study_ajax/base_info.html
```diff
@@ -0,0 +1,7 @@
+<p style="font-weight:bold;">Abstract</p>
+<p>{{study_info['study_abstract']}}</p>
+Study ID:  {{study_info['study_id']}}<br />
+Publications: {% for pub in study_info['publication_doi'] %}<a href="http://dx.doi.org/{{pub}}" target="_blank">{{pub}}</a> {% end %}<br />
```
antgonza: This formatter already exist in other sections of the code, not sure if it will be better to have that for loop and the formatter outside the template and simply print the value here, what do you think?

antgonza:
Some comments but good start. I guess the idea is to merge this once comments are addressed + other reviewer and start building from there, right?

qiita_db/study.py
squirrelo: The wording is bad, but I think it's supposed to mean it could be part of the main info, but we thought no. I'll reword, since really it's removed because it's redundant to the study id already stored in the object.

qiita_pet/handlers/study_handlers/api_proxy.py
squirrelo: That doesn't make sense grammatically.

qiita_pet/handlers/study_handlers/api_proxy.py
squirrelo: Yeah, that makes sense. The TODO is there because we don't have the functionality needed in the prep template object yet.

qiita_pet/handlers/util.py
squirrelo: I don't think so, since this is now in handlers.util, so it's generic.

qiita_pet/templates/study_ajax/base_info.html
squirrelo: Forgot about that, I'll take a look and change if it works.

squirrelo:
Exactly. This framework allows the rest of things to be done in parallel. The only thing that may get merge conflicts is adding the ajax functions to the top of the main study html page.

antgonza:
:+1:

squirrelo:
Tests are expected to fail since the study description page has been removed and replaced by this skeleton, which is missing most of the functionality at this moment.

qiita_db/study.py
```diff
@@ -166,6 +166,18 @@ def status(self):
             return qdb.util.infer_status(
                 qdb.sql_connection.TRN.execute_fetchindex())
 
+    @staticmethod
+    def all_data_types():
+        """Returns list of all the data types available in the system
```
ElDeveloper: This is missing a space between the description and the `Returns` header.

qiita_db/study.py
```diff
@@ -166,6 +166,18 @@ def status(self):
...
+        Returns
+        -------
+        list of str
```
ElDeveloper: Mind describing what the strings represent? I know its repetitive, but that's the suggested way.

qiita_pet/handlers/api_proxy.py
```diff
@@ -0,0 +1,117 @@
...
+#
+# Distributed under the terms of the BSD 3-clause License.
+#
+# The full license is in the file LICENSE, distributed with this software.
+# -----------------------------------------------------------------------------
+from __future__ import division
+
+# This is the only file in qiita_pet that should import from outside qiita_pet
+# The idea is that this proxies the call and response dicts we expect from the
+# QIITA API once we build it. This will be removed and replaced with API calls
```
ElDeveloper: QIITA -> Qiita.

qiita_pet/handlers/api_proxy.py
```diff
@@ -0,0 +1,117 @@
+# -----------------------------------------------------------------------------
+# Copyright (c) 2014--, The Qiita Development Team.
+#
+# Distributed under the terms of the BSD 3-clause License.
+#
+# The full license is in the file LICENSE, distributed with this software.
+# -----------------------------------------------------------------------------
+from __future__ import division
+
+# This is the only file in qiita_pet that should import from outside qiita_pet
```
ElDeveloper: Are we sure we want to have this? Regrettably we have a few pieces of code that are "to be removed when X happens" and this compromises the usability, and the stability of those pieces of code and everything that gets built on top of them. If nobody sees this is as a possible problem, then lets leave as is.

qiita_pet/handlers/api_proxy.py
```diff
@@ -0,0 +1,117 @@
...
+    Returns
+    -------
+    bool
+        Whether the study can be approved or not
+    """
+    return True if not qiita_config.require_approval else level == 'admin'
+
+
+class StudyAPIProxy(BaseHandler):
+    """Adds API proxy functions to the handler. Can be removed once the restful
```
ElDeveloper: restful -> RESTful

qiita_pet/handlers/api_proxy.py
```diff
@@ -0,0 +1,117 @@
...
+       API is in place."""
+    def study_prep_proxy(self, study_id):
+        """Proxies expected json from the API for existing prep templates
+
+        Parameters
+        ----------
+        study_id : int
+            Study id to get prep template info for
+
+        Returns:
```
ElDeveloper: Can you format this with underscores and remove the semicolon?

qiita_pet/handlers/api_proxy.py
```diff
@@ -0,0 +1,117 @@
...
+        dict of list of dict
+            prep template information seperated by data type, in the form
+            {data_type: [{prep 1 info dict}, ....], ...}
+        """
+        # Can only pass ids over API, so need to instantiate object
+        study = Study(study_id)
+        check_access(self.current_user, study, raise_error=True)
```
ElDeveloper: Can you add whatever error this raises to a `Raises` section in the docstring?

qiita_pet/handlers/api_proxy.py
squirrelo: The idea is qiita_pet will be it's own self contained client for the qiita RESTful API. Since we are developing the UI before developing the RESTful interface, this proxy is needed so I don't have to completely redevelop the UI again once the API goes into place.

qiita_db/study.py
```diff
@@ -166,6 +166,18 @@ def status(self):
...
+        """
+        with qdb.sql_connection.TRN:
+            sql = "SELECT DISTINCT data_type FROM qiita.data_type"
+            qdb.sql_connection.TRN.add(sql)
+            return qdb.sql_connection.TRN.execute_fetchflatten()
```
ElDeveloper: Seems like this could return None, should this be checking and instead returning an empty list? Otherwise it would probably be worth updating the docstring. My preference would be to return an empty list, not sure what others think about that.

qiita_pet/handlers/api_proxy.py
```diff
@@ -0,0 +1,117 @@
...
+
+    def study_info_proxy(self, study_id):
+        """Proxies expected json from the API for base study info
+
+        Parameters
+        ----------
+        study_id : int
+            Study id to get prep template info for
+
+        Returns:
```
ElDeveloper: Can you fix this header.

qiita_pet/handlers/api_proxy.py
```diff
@@ -0,0 +1,117 @@
...
+        dict of list of dict
+            prep template information seperated by data type, in the form
+            {data_type: [{prep 1 info dict}, ....], ...}
+        """
+        # Can only pass ids over API, so need to instantiate object
+        study = Study(study_id)
+        check_access(self.current_user, study, raise_error=True)
```
ElDeveloper: Can you add this error to the `Raises` section of the docstring?

qiita_db/study.py
squirrelo: Done

qiita_db/study.py
squirrelo: This will return an empty list if nothing is in the SQL.

qiita_pet/handlers/util.py
ElDeveloper: I think this is actually a 4xx error, the best candidates from what I've read so far are: 417, 406 and 400. http://www.w3.org/Protocols/rfc2616/rfc2616-sec10.html

qiita_pet/templates/study_base.html
```diff
@@ -0,0 +1,72 @@
+{% extends sitebase.html %}
+{% block head %}
+<script type="text/javascript">
+  function fill_base(sid) {
+  $.get( "/study/description/baseinfo/", { study_id: sid} )
```
ElDeveloper: This block is under-indented, should be two spaces in.

qiita_pet/templates/study_base.html
```diff
@@ -0,0 +1,72 @@
...
+    <div class="panel-group" id="accordion" role="tablist" aria-multiselectable="true">
+      <div class="panel panel-default">
+      {% for dt in prep_info %}
+      {% set cleaned_dt = dt.replace(' ', '_') %}
+        <div class="panel-heading" role="tab" id="heading{{cleaned_dt}}">
+          <h4 class="panel-title">
+            <a role="button" data-toggle="collapse" data-parent="#accordion" href="#collapse{{cleaned_dt}}" aria-expanded="true" aria-controls="collapse{{cleaned_dt}}">{{dt}}</a>
+          </h4>
+        </div>
+        {% end %}
```
ElDeveloper: I believe this statement is over-indented by two spaces, same thing for the statement below.

qiita_pet/templates/study_base.html
```diff
@@ -0,0 +1,72 @@
...
+  $(document).ready(function() {
+    fill_base({{study_info['study_id']}});
+  });
+</script>
+{% end %}
+{% block content %}
+<div class="row">
+  <div class="col-md-3">
+  <h3>Study Information</h3>
+    <a href="#">Sample Template</a><br />
```
ElDeveloper: Some of these options should be showed if you are not the owner of this study, right?

qiita_pet/templates/study_base.html
```diff
@@ -0,0 +1,72 @@
...
+      <div class="panel panel-default">
+      {% for dt in prep_info %}
+      {% set cleaned_dt = dt.replace(' ', '_') %}
+        <div class="panel-heading" role="tab" id="heading{{cleaned_dt}}">
+          <h4 class="panel-title">
+            <a role="button" data-toggle="collapse" data-parent="#accordion" href="#collapse{{cleaned_dt}}" aria-expanded="true" aria-controls="collapse{{cleaned_dt}}">{{dt}}</a>
+          </h4>
+        </div>
+        {% end %}
+        {% for prep in prep_info[dt] %}
```
ElDeveloper: Only owners of the study should see "all" the information, right, this should be filtered for things the user can see, which are presumably only "public" artifacts right?

qiita_db/study.py
ElDeveloper: Got it, then I'm confused what you are checking for in line 89 of api_proxy.py, maybe that's unnecessary then?

squirrelo: ah, yes that's a redundant check.

qiita_pet/handlers/api_proxy.py
squirrelo: Done

qiita_pet/handlers/util.py
squirrelo: Best candidate is probably 400 then, since it is a bad study/prep/whatever ID being passed in the URL.

qiita_pet/templates/study_base.html
squirrelo: I'm leveraging the API on this one, since this design keeps with the current assumption that the API call object only returns what information is viewable by the user.

ElDeveloper: OK, got it, that makes perfect sense, we should really keep an eye
opened for this as otherwise it will lead to a rather strange-looking
UI.

qiita_pet/templates/study_base.html
squirrelo: I agree. I've added an if statement to show certain things only if study is yours or shared with you.

squirrelo:
Comments addressed

qiita_pet/handlers/api_proxy.py
ElDeveloper: OK, sounds good then!

ElDeveloper:
Thanks @squirrelo

PR #1572, opened 2015-12-11
Fixing raw data sharing

josenavas:
This PR allows to "share" the raw data objects across multiple studies correctly.

The main issue is that artifacts are designed to not be shared across studies. Instead, the artifact is copied.

This is currently copying the files. I think this is fine for now, but at some point we need to modify the interface and some parts of the code so we can share files across artifacts. It will be complicated to do this right now w/o major changes to the interface, which I don't think is worth given the on-going refactoring.

### DIFF
--- a/qiita_db/artifact.py
+++ b/qiita_db/artifact.py
@@ -96,7 +96,7 @@ class Artifact(qdb.base.QiitaObject):
             A new instance of Artifact
         """
         with qdb.sql_connection.TRN:
-            visibility = qdb.util.convert_to_id("sandbox", "visibility")
+            visibility_id = qdb.util.convert_to_id("sandbox", "visibility")
             atype = artifact.artifact_type
             atype_id = qdb.util.convert_to_id(atype, "artifact_type")
             dtype_id = qdb.util.convert_to_id(
@@ -107,7 +107,7 @@ class Artifact(qdb.base.QiitaObject):
                         can_be_submitted_to_vamps, submitted_to_vamps)
                      VALUES (%s, %s, %s, %s, %s, %s, %s)
                      RETURNING artifact_id"""
-            sql_args = [datetime.now(), visibility, atype_id, dtype_id,
+            sql_args = [datetime.now(), visibility_id, atype_id, dtype_id,
                         artifact.can_be_submitted_to_ebi,
                         artifact.can_be_submitted_to_vamps, False]
             qdb.sql_connection.TRN.add(sql, sql_args)
--- a/qiita_db/artifact.py
+++ b/qiita_db/artifact.py
@@ -80,6 +80,63 @@ class Artifact(qdb.base.QiitaObject):
         return cls.iter_by_visibility('public')
 
     @classmethod
+    def copy(cls, artifact, prep_template):
+        """Creates a copy of `artifact` and attaches it to `prep_template`
+
+        Parameters
+        ----------
+        artifact : qiita_db.artifact.Artifact
+            Artifact to copy from
+        prep_template : qiita_db.metadata_template.prep_template.PrepTemplate
+            The prep template to attach the new artifact to
+
+        Returns
+        -------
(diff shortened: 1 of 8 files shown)


josenavas:
@mortonjt @ElDeveloper @antgonza can I get a review here?

qiita_db/test/test_artifact.py
```diff
@@ -107,6 +107,42 @@ def test_iter_public(self):
         exp = [a1, a4]
         self.assertEqual(obs, exp)
 
+    def test_copy(self):
+        src = qdb.artifact.Artifact(1)
+        # Craete the files to the first artifact
```
mortonjt: Craete -> Create

qiita_pet/handlers/compute.py
```diff
@@ -64,8 +58,27 @@ def _split(x):
...
+    def post(self):
+        pt_id = self.get_argument('prep_template_id')
+        pt = PrepTemplate(pt_id)
+        study_id = pt.study_id
+
+        artifact_id = self.get_argument('artifact_id', default=None)
+        print artifact_id
+
+        if artifact_id is not None:
+            job_id = self.create_from_artifact(pt, artifact_id)
```
mortonjt: Do we really need `create_from_artifact`? Or is that for readability?

qiita_db/test/test_util.py
```diff
@@ -164,6 +164,34 @@ def test_insert_filepaths(self):
...
+        self.assertEqual(obs, exp)
+
+    def test_insert_filepaths_copy(self):
+        fd, fp = mkstemp()
+        close(fd)
+        with open(fp, "w") as f:
+            f.write("\n")
+        self.files_to_remove.append(fp)
+
+        exp_new_id = 1 + self.conn_handler.execute_fetchone(
```
mortonjt: I'm assuming that `1 + ` is being done to assign a unique id correct?

Worthwhile adding a comment about this?

qiita_pet/handlers/compute.py
```diff
@@ -64,8 +58,27 @@ def _split(x):
...
+
+    @authenticated
+    @execute_as_transaction
+    def post(self):
+        pt_id = self.get_argument('prep_template_id')
+        pt = PrepTemplate(pt_id)
+        study_id = pt.study_id
+
+        artifact_id = self.get_argument('artifact_id', default=None)
+        print artifact_id
```
mortonjt: Don't forget to remove the print :)

qiita_db/artifact.py
```diff
@@ -80,6 +80,63 @@ def iter_public(cls):
...
+        prep_template : qiita_db.metadata_template.prep_template.PrepTemplate
+            The prep template to attach the new artifact to
+
+        Returns
+        -------
+        qiita_db.artifact.Artifact
+            A new instance of Artifact
+        """
+        with qdb.sql_connection.TRN:
+            visibility = qdb.util.convert_to_id("sandbox", "visibility")
```
antgonza: I guess this should be visibility_id, right?

antgonza:
After my and @mortonjt comments are addressed :+1:

qiita_db/artifact.py

qiita_db/test/test_artifact.py
josenavas: Done

qiita_db/test/test_util.py
josenavas: We are using bigserials in the DB, i.e. they get autoincremented by the database, so that's why I can predict the id. I added a comment anyways.

qiita_pet/handlers/compute.py
josenavas: Ooops! Done :)

qiita_pet/handlers/compute.py
josenavas: I don't understand your question here - what is the alternative?

PR #1577, opened 2015-12-15
Artifact lineage

josenavas:
This PR adds methods to retrieve all the ancestors and descendants of a given artifact represented in a networkx directed graph in which the nodes are the artifacts themselves.

This is needed for the interface changes that @squirrelo is working on as well as for being able to propagate the status of the artifacts.

### DIFF
--- a/qiita_db/artifact.py
+++ b/qiita_db/artifact.py
@@ -8,7 +8,10 @@
 
 from __future__ import division
 from future.utils import viewitems
+from itertools import chain
 from datetime import datetime
+
+import networkx as nx
 
 import qiita_db as qdb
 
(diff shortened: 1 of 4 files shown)


qiita_db/artifact.py
```diff
@@ -646,6 +649,62 @@ def parents(self):
             return [Artifact(p_id)
                     for p_id in qdb.sql_connection.TRN.execute_fetchflatten()]
 
+    def _create_lineage_graph_from_edge_list(self, edge_list):
+        """Generates an artifact graph from the given `edge_list`
+
+        Parameters
+        ----------
+        edge_list : list of (int, int)
```
squirrelo: From the if statement below, edge_list here should be optional and set to None by default in the function definition.

qiita_db/test/test_artifact.py
```diff
@@ -496,6 +497,87 @@ def test_parents(self):
         exp_parents = [qdb.artifact.Artifact(2)]
         self.assertEqual(qdb.artifact.Artifact(4).parents, exp_parents)
```
squirrelo: Would be good to add a test in for `_create_lineage_graph_from_edge_list` explicitly.

squirrelo:
Couple comments

squirrelo:
One more observation: has this been tested for raw data or others that can possibly be shared with multiple prep templates? What happens if I have a raw data containing artifact that is shared with 3 prep templates, and then call `descendants`? Looking at this, it will give the graphs for all files made as part of the three prep templates, which may not be what we want depending on the use.

qiita_db/artifact.py
josenavas: Nope, the if statement is specifically for the case in which the list is empty, but it is not an optional parameter.

qiita_db/test/test_artifact.py
josenavas: Adding

josenavas:
Artifacts can't be shared between studies.

qiita_db/artifact.py
squirrelo: Ah, can you add a note to that effect? That was not clear from the code.

squirrelo:
Ah, the ability to use the same Raw Data between multiple studies is gone? Cool.

josenavas:
Yeah, after seeing how the people work with the system, we realize that sharing "raw data" is not useful, as usually you have to modify it by either adding more lanes or removing some existing lanes.

In the new interface, however, we should allow users to import the files from the other raw data objects, but not the raw data object itself. Does it make sense? I was planning on doing that on my last PR, but the amount of interface changes needed were not worth given the refactor that you're currently doing.

BTW, comments addressed.

qiita_db/artifact.py
```diff
@@ -646,6 +649,62 @@ def parents(self):
...
+            List of (artifact_id, parent_artifact_id)
+
+        Returns
+        -------
+        networkx.DiGraph
+            The graph representing the artifact lineage stored in `edge_list`
+        """
+        lineage = nx.DiGraph()
+        if edge_list:
```
squirrelo: If this is for the case when the list is empty, this should probably be set to check that. Right now anything can be passed as the edge_list and it will just return the single node graph.

qiita_db/artifact.py

qiita_db/artifact.py
josenavas: Can you elaborate? this is the pythonic way of testing if a list is empty or not.

squirrelo:
:+1: assuming tests pass.

josenavas:
@mortonjt @ElDeveloper @antgonza available for a review in here?

antgonza:
:+1: waiting for tests ...

squirrelo:
tests are good, merging.

PR #1581, opened 2015-12-18
HTTPS

josenavas:
This PR enables HTTPS in the system.

This is adding default certificates and keys to the system. Since these are public, I highly recommend to not use these in a public system open to everybody. However, I think they should be ok to use in personal laptops for development and/or when users install Qiita in their systems.

If anybody sees a potential security problem by using these ones, please let me know and I will add code to generate those certificates.

I will add admin documentation in how to create these certificates and use them in a PR with the other documentation.

Please, test this in your machine using the certificates that are included. If they do not work, let me know so I can add the code rather than include them in the repository.


qiita_core/support_files/config_test.cfg
```diff
@@ -45,6 +45,10 @@ PORTAL = QIITA
 #Script used for launching plugins
 PLUGIN_LAUNCHER = qiita-plugin-launcher
 
+#Webserver certificate files
```
antgonza: Space between # W

qiita_plugins/target_gene/tgp/qiita_client.py
```diff
@@ -0,0 +1,79 @@
...
+            # The server certificate is not provided, use standard certificate
+            # verification methods
+            self._verify = True
+        else:
+            # The server certificate is provided, use it to verify the identity
+            # of the server
+            self._verify = server_cert
+
+    def _request_retry(self, req, url, **kwargs):
+        """Executes a request retrying it 3 times in case of failure
```
antgonza: Should 3 be a parameter in the config file?

qiita_plugins/target_gene/tgp/qiita_client.py
```diff
@@ -0,0 +1,79 @@
...
+        Returns
+        -------
+        dict
+            The JSON information in the request reply
+        """
+        url = self._server_url + url
+        success = False
+        retries = 3
+        json_reply = None
+        while not success and retries > 0:
```
antgonza: Where does success changes its status? In other words, it's always False.

qiita_plugins/target_gene/tgp/qiita_client.py
```diff
@@ -0,0 +1,79 @@
...
+            retries -= 1
+            r = req(url, verify=self._verify, **kwargs)
+            r.close()
+            if r.status_code == 200:
+                json_reply = r.json()
+                break
```
antgonza: This break does the trick for success ...

qiita_core/support_files/config_test.cfg
josenavas: Done, also fixed a couple of similar errors in this file

qiita_plugins/target_gene/tgp/qiita_client.py
josenavas: Not sure. This functionality is hard to test. In case that the communication fails I'm just re-trying such communication up to 3 times, in case that something weird happened. Do you have strong feelings on moving this to a config file? (note that this was like that in the previous version but in a different function)

qiita_plugins/target_gene/tgp/qiita_client.py
josenavas: Good point, I guess I no longer need the success boolean. I think this was a bit of leftover code when I ported from the previous function.

qiita_plugins/target_gene/tgp/qiita_client.py
antgonza: No strong feelings, just that trying 3 times seems kind of random when you can try once or twice. If nobody else has an issue with this it is fine to leave as is.

qiita_core/support_files/server.key
```diff
@@ -0,0 +1,27 @@
+-----BEGIN RSA PRIVATE KEY-----
```
squirrelo: Have these been tested when serving from other systems? I want to make sure they don't trigger HTTPS error pages.

qiita_plugins/target_gene/tgp/qiita_client.py
squirrelo: Add documentation in the notes section why 3 specifically.

qiita_core/support_files/config_test.cfg
```diff
@@ -39,12 +39,16 @@ BASE_DATA_DIR =
...
 
-#Portal the site is working under
+# Portal the site is working under
 PORTAL = QIITA
 
-#Script used for launching plugins
+# Script used for launching plugins
 PLUGIN_LAUNCHER = qiita-plugin-launcher
 
+# Webserver certificate files
```
squirrelo: Can you state these must be filepaths and not the certificate text itself?

qiita_plugins/target_gene/tgp/qiita_client.py
```diff
@@ -0,0 +1,116 @@
...
+
+        config = ConfigParser()
+        with open(conf_fp, 'U') as conf_file:
+            config.readfp(conf_file)
+
+        server_cert = config.get('main', 'SERVER_CERT')
+        if not server_cert:
+            # The server certificate is not provided, use standard certificate
+            # verification methods
+            self._verify = True
```
squirrelo: From the looks of this, _verify can be bool or string type and hold True or a filepath. This is confusing and will be hard to maintain, especially since both are "truthy" for if statements. Please decompose this better so we are not mixing types in the same variable.

qiita_plugins/target_gene/tgp/tests/test_qiita_client.py
```diff
@@ -0,0 +1,108 @@
...
+    def test_init(self):
+        obs = QiitaClient("https://test_server.com")
+        self.assertEqual(obs._server_url, "https://test_server.com")
+        self.assertTrue(obs._verify)
+
+    def test_init_cert(self):
+        fd, conf_fp = mkstemp()
+        close(fd)
+        with open(conf_fp, 'w') as f:
+            f.write(CONF_FP)
```
squirrelo: Why are you doing it this way instead of using `qiita_plugins/target_gene/tgp/support_files/config_file.cfg` like in all other tests? This way doesn't verify the config file given is correct, and adds two points to change if a config variable is added or removed.

qiita_plugins/target_gene/tgp/util.py
```diff
@@ -47,95 +46,65 @@ def system_call(cmd):
...
-    server_url : str
-        The ust of the server
+    qclient : tgp.qiita_client.QiitaClient
+        The Qiita server client
     job_id : str
         The job id
     """
-    url = "%s/qiita_db/jobs/%s/heartbeat/" % (server_url, job_id)
-    heartbeat_thread = threading.Thread(target=heartbeat, args=(url,))
+    url = "/qiita_db/jobs/%s/heartbeat/" % job_id
```
squirrelo: Should this be hardcoded here? Probably better as a config file variable in case API change happen.

qiita_plugins/target_gene/tgp/util.py
```diff
@@ -145,9 +114,8 @@ def complete_job(server_url, job_id, payload):
     global JOB_COMPLETED
     JOB_COMPLETED = True
     # Create the URL where we have to post the results
-    url = "%s/qiita_db/jobs/%s/complete/" % (server_url, job_id)
     json_payload = dumps(payload)
-    execute_request_retry(requests.post, url, data=json_payload)
+    qclient.post("/qiita_db/jobs/%s/complete/" % job_id, data=json_payload)
```
squirrelo: Same as above about hard coding URLs.

qiita_plugins/target_gene/tgp/util.py
```diff
@@ -47,95 +46,65 @@ def system_call(cmd):
...
+        The Qiita server client
     jon_id : str
         The job id
     new_step : str
         The new step
     """
-    url = "%s/qiita_db/jobs/%s/step/" % (server_url, job_id)
     json_payload = dumps({'step': new_step})
-    execute_request_retry(requests.post, url, data=json_payload)
+    qclient.post("/qiita_db/jobs/%s/step/" % job_id, data=json_payload)
```
squirrelo: Same as above about hardcoding URLs.

squirrelo:
Some comments.

squirrelo:
Running tests on my system for the plugins, I get that `httpretty` isn't installed. I also don't see it in the setup.py.

squirrelo:
Also, when I try to acc

qiita_core/support_files/server.key
squirrelo: When I try and access any page on the system, I get hte following error: 
```python 
WARNING:tornado.general:SSL Error on 61 ('::1', 58512, 0, 0): [SSL: HTTP_REQUEST] http request (_ssl.c:590)
```
So these may not be good universally.

squirrelo:
Sorry, cat on keyboard.

qiita_core/support_files/config_test.cfg
josenavas: Done

qiita_core/support_files/server.key
josenavas: I couldn't test this in another system - that is what I exposed in the PR comments.

The error that you're seeing is because you are using http rather than https. Make sure to type https - in the live system the redirect is done through nginx, but in your local machine you need to type the https. Also make sure that your config file includes the https in the base_url, so when redirects happen in the system it works as expected.

qiita_plugins/target_gene/tgp/qiita_client.py
josenavas: Yeah, I hate this - however this is a limitation from `requests` (see documentation here: `http://docs.python-requests.org/en/latest/user/advanced/#ssl-cert-verification`). By setting up the `verify` parameter here, I'm avoiding the execution of the if statement every time a request is done.

qiita_plugins/target_gene/tgp/qiita_client.py
josenavas: Changed to 2 and added documentation on why.

qiita_core/support_files/server.key
squirrelo: Done, and I now get the error I was expecting from chrome:
```
Your connection is not private

Attackers might be trying to steal your information from localhost (for example, passwords, messages, or credit cards). NET::ERR_CERT_AUTHORITY_INVALID
```

josenavas: Can you post a screenshot of the error rather than copy paste? I would like to see better the error from chrome.

Thanks,

squirrelo: screen shot 2015-12-21 at 12 34 25

josenavas: @squirrelo That is expected - you should go to the `Advanced` button and move forward. Note that the certificates that we are using here (even if you generate it in your machine) is not validated by any certificate authority, so chrome will complain.

Note that if we want this to go away, we have to require to our users to request a certificate to one of these authorities, which I don't think we want to do.

Also, we cannot add a certificate approved by any authority in the repository, otherwise anyone can use such certificate and appear safe, while in reality is stealing information.

I'm going to add documentation on how to create your certificate, but even if you create it using those steps, this warning is going to appear.

squirrelo: I think this is a bad idea if we are targeting this for single user laptop use as well, as I can foresee a lot of people freaking out when they first start the system and see that error.

josenavas: @biocore/qiita-dev can I have some comments here? The only alternative then is to ask each of our users to ask for a certificate to an authority - which can take some time and I don't think this is what we need to do. Without this, we can't use https neither oauth2.

I think that we good documentation this is not a problem - and we let the user know that is not a problem.

@squirrelo If you don't agree with this, please provide an alternative and constructive feedback.

squirrelo: We should really default back to HTTP if we don't have a certificate provided, possibly with a warning when you start the webserver. This allows users to use standard HTTP when it's on their laptop and safe, and HTTPS in larger deploys.

ElDeveloper: This issue is a complicated one, on the one hand we want to maintain the installation process as simple as possible, but on the other hand we also want to have a consistent system that isn't raising flags and worrying our users. It almost feels like these two are incompatible, so these are the possible partial solutions I can think of:

- Document this situation as part of the installation procedure, something along the lines of " ... you'll need to provide authorized and validated certificates, otherwise you may .... XXXXX".
- Add a permanent warning to the system i.e. a bootstrap alert that never leaves the screen saying " your certificates are not xxxxx, to fix this xxxxx, or read a discussion on why this xxxxxxx".
- Check the system before starting the webserver and halting execution if valid certificates are not found, this could be bypassed with a flag maybe something like `--i-dont-care-about-security-` or `--local-system`.
- Add an installation manager that requires valid certificates to be present at installation time, so that the situation shown above cannot be reached.

The solution should probably be a combination of some of the points above.

josenavas: Just discussed some of this information with Jeff. Few points:
 - Valid certificates are not free: you have to buy them and their prices can be around ~$50. Even if it was $1 I don't think it is ok to ask our users running Qiita in their local machine to buy some certificate just because chrome complains about it.
 - The goal of a certificate is to let the user that the server is trusted - if we have documentation explaining that the server is your own machine, you can trust it as much as you trust your own machine.
 - There is no way of generating self-signed certificates that are recognized by the entire world unless you buy them. However, you can add an exception to your browser so it accepts such certificate.

About the comment of defaulting to HTTP: the entire work to use HTTPS has beed done so the RESTapi/plugins work as expected communicating through OAuth2 - if we default to HTTP, we can't use any of the plugins as they're authenticating through OAuth2, which requires HTTPS.

My vote goes top add detailed documentation in which we specify that if you are running locally in your machine, you will receive that warning the first time that you log in to the server (per @squirrelo 's observation, maybe every time that you start the server - I did not have that problem). Then, also add documentation that if you want to make your own Qiita server, you should get your own valid certificates from a Certificate Authority.

antgonza: I think the solution is to improve documentation and/or try to do some of these suggestions: http://superuser.com/questions/772762/how-can-i-disable-security-checks-for-localhost

qiita_plugins/target_gene/tgp/tests/test_qiita_client.py
josenavas: Cause it will brake the point of this test.

What I'm specifically testing here is that it is using the values of a different config file, not the one as default - if I use the default one as example, I can't tell the difference.

qiita_plugins/target_gene/tgp/util.py
josenavas: I don't agree with that - in order to be consistent it will require to put every single URL in the config file just in case it changes.

qiita_plugins/target_gene/tgp/util.py
josenavas: Same as above

qiita_plugins/target_gene/tgp/util.py
josenavas: Same as above - these are three examples, but there are a lot more calls that it will make the entire code just messy - and the API should not be extremely dynamic, otherwise it looses its power..

josenavas:
I am hesitant to add httpretty to the setup.py
httpretty is required by the plugin, not by the main Qiita - the only reason why the plugin is in the same repository is for convenience in early development stages of the plugin, but it should be branched of to its own repository soon, once we can make the interfaces a bit more stable. An alternative that I was managing was to add it's own setup.py in the qiita_plugins/target_gene folder, what do others think about this? Looks like a good solution?

qiita_plugins/target_gene/tgp/qiita_client.py
squirrelo: That is pretty weird. If we do want to keep this, this needs a lot more documentation as to what's going on, including a link to that documentation.

qiita_plugins/target_gene/tgp/tests/test_qiita_client.py
squirrelo: you can change options in an already loaded config to something else, as demonstrated in the [analysis object tests](https://github.com/biocore/qiita/blob/master/qiita_db/test/test_analysis.py#L32). This will be more robust.

josenavas: No, I don't want to change the values of the config file specifically, I want to make sure that the `__init__` functionality is parsing the correct config file. Note that in Qiita we have a configuration manager that is parsing such config and providing the information to the rest of the system. Here, the configuration file only affects this class, so the class can be though as the configuration manager itself.

If I use the same strategy that you are proposing, I still don't know if the class is parsing the config file in the environmental variable or the default config file in the repository.

qiita_plugins/target_gene/tgp/tests/test_qiita_client.py
```diff
@@ -0,0 +1,108 @@
...
+
+        self._clean_up_files.append(conf_fp)
+        old_fp = environ.get('QP_TARGET_GENE_CONFIG_FP')
+        environ['QP_TARGET_GENE_CONFIG_FP'] = conf_fp
+        try:
```
squirrelo: As has been stated in the past, try/finally should never be in tests. Please move this into the setUp/tearDown.

josenavas: Good call - done

qiita_plugins/target_gene/tgp/tests/test_qiita_client.py
squirrelo: This brings up a good point: we do not seem to have tests for the qiita config manager. Could you open an issue on this?

josenavas: Done https://github.com/biocore/qiita/issues/1586

josenavas:
Is there anything missing in this PR? The creation of the setup.py for the plugin is done here: https://github.com/biocore/qiita/pull/1587

antgonza:
httpretty on setup.py in qiita_plugins/target_gene sounds like a good idea but open to other suggestions, the only requirement: one single command to install everything. I think it will be fine to have multiple commands once we have separate repos.

qiita_plugins/target_gene/tgp/qiita_client.py
```diff
@@ -0,0 +1,197 @@
...
+          retries already done, as well as the cost of a retry
+          3. Retry just once
+
+        Number 1 could create an infinite loop. Number 2 is too complex and
+        the cost of retrying depends on the actual work that we are currently
+        doing (which is unknown to the current function). We thus decided to
+        implement 3, which is simple and allows to overcome simple
+        communication problems.
+        """
+        url = self._server_url + url
```
antgonza: should this use path join?

antgonza:
2 comments then :+1:

qiita_plugins/target_gene/tgp/qiita_client.py
josenavas: I disagree as those are not paths and, although we don't support it, on windows it will not work.

josenavas:
Can we address the setup.py comment in the other PR?

antgonza:
Just to keep moving, lets do the setup.py in other PR but please open it ASAP.

josenavas:
@antgonza that PR is already open: https://github.com/biocore/qiita/pull/1587

antgonza:
Thanks for the clarification.

PR #1582, opened 2015-12-21
API refactor and addition of sample template page

squirrelo:
This rebuilds the API proxy to be a separate folder of files that can easily be moved to be the new API once it is established. This also allows tests on the API. Finally, it adds the sample information page and actions on the page.


squirrelo:
The errors are expected, as files and functions have been moved around for the API and the original UI handlers have not been deleted.

qiita_pet/handlers/api_proxy/prep_template.py
```diff
@@ -0,0 +1,53 @@
+# -----------------------------------------------------------------------------
+# Copyright (c) 2014--, The Qiita Development Team.
+#
+# Distributed under the terms of the BSD 3-clause License.
+#
+# The full license is in the file LICENSE, distributed with this software.
+# -----------------------------------------------------------------------------
+from __future__ import division
+
+# This is the only folder in qiita_pet that should import outside qiita_pet
```
antgonza: Not sure if we need this warning in every file of just in __init__.py, what do other think?

qiita_pet/handlers/api_proxy/sample_template.py
```diff
@@ -0,0 +1,242 @@
...
+                                 QiitaDBDuplicateError, QiitaDBError,
+                                 QiitaDBDuplicateHeaderError)
+from qiita_ware.metadata_pipeline import (
+    create_templates_from_qiime_mapping_file)
+from qiita_ware.exceptions import QiitaWareError
+from qiita_pet.util import convert_text_html
+from qiita_pet.handlers.api_proxy.util import check_access
+
+
+def sample_template_info(samp_id, user_id):
```
antgonza: Just to make it easy to search, what about changing the name to `sample_template_get_request` or `st_get_request`? Same for the following function: `sample_template_post_request`

qiita_pet/handlers/api_proxy/sample_template.py
```diff
@@ -0,0 +1,242 @@
...
+    data_type : str
+        Data type for the sample template
+    sample_template : str
+        filepath to use for creation
+
+    Raises
+    ------
+    HTTPError
+        If the sample template file does not exists
+    """
```
antgonza: Missing Returns, note the different types, not sure if this is desirable; and I couldn't find the raise ...

qiita_pet/handlers/api_proxy/sample_template.py
```diff
@@ -0,0 +1,242 @@
...
+        status = 'error'
+        msg = str(e)
+        status = "error"
+        return {'status': status,
+                'message': msg,
+                'file': sample_template}
+
+
+@execute_as_transaction
+def update_sample_template(study_id, user_id, sample_template):
```
antgonza: Same that previous suggestion: `sample_template_post_request_update`

qiita_pet/handlers/api_proxy/sample_template.py
```diff
@@ -0,0 +1,242 @@
...
+    Parameters
+    ----------
+    study_id : int
+        The current study object id
+    user_id : str
+        The current user object id
+    sample_template : str
+        filepath to use for updating
+
+    Raises
```
antgonza: Same comments about Raises and Returns than the previous function.

qiita_pet/handlers/api_proxy/sample_template.py
```diff
@@ -0,0 +1,242 @@
...
+            status = 'error'
+            msg = format_exc(e)  # str(e)
+    return {'status': status,
+            'message': msg,
+            'file': sample_template}
+
+
+@execute_as_transaction
+def delete_sample_template(study_id, user_id):
+    """Equivalent to DELETE request to `/study/(ID)/sample_template'
```
antgonza: Note how you have ` and '

qiita_pet/handlers/api_proxy/sample_template.py
```diff
@@ -0,0 +1,242 @@
...
+
+    Parameters
```
antgonza: Missing Return

qiita_pet/handlers/api_proxy/sample_template.py
```diff
@@ -0,0 +1,242 @@
...
+        return access_error
+    try:
+        SampleTemplate.delete(int(study_id))
+    except Exception as e:
+        return {'status': 'error', 'message': str(e)}
+    return {'status': 'success'}
+
+
+@execute_as_transaction
+def get_sample_template_filepaths(study_id, user_id):
```
antgonza: Same comments than previous func

qiita_pet/handlers/api_proxy/studies.py
```diff
@@ -0,0 +1,74 @@
...
+    user_id : str
+        User requesting the info
+
+    Returns
+    -------
+    dict of info
+        Study information seperated by data type, in the form
+        {col_name: value, ...}
+
+    Raises
```
antgonza: No raises ...

qiita_pet/handlers/api_proxy/studies.py
```diff
@@ -0,0 +1,74 @@
...
+    # Clean up StudyPerson objects to string for display
+    pi = study_info["principal_investigator"]
+    study_info["principal_investigator"] = '%s (%s)' % (pi.name,
+                                                        pi.affiliation)
+    lab_person = study_info["lab_person"]
+    study_info["lab_person"] = '%s (%s)' % (lab_person.name,
+                                            lab_person.affiliation)
+
+    samples = study.sample_template.keys()
+    study_info['num_samples'] = 0 if samples is None else len(set(samples))
```
antgonza: Do we really need to make it a set? My guess is no ...

qiita_pet/handlers/api_proxy/tests/test_sample_template.py
```diff
@@ -0,0 +1,154 @@
...
+
+    def test_delete_sample_template_no_access(self):
+        obs = delete_sample_template(1, 'demo@microbio.me')
+        exp = {'status': 'error',
+               'message': 'User does not have access to study'}
+        self.assertEqual(obs, exp)
+
+    def test_get_sample_template_filepaths(self):
+        obs = get_sample_template_filepaths(1, 'test@foo.bar')
+        exp = [(14, '/Users/jshorens/Repositories/qiita/qiita_db/support_'
```
antgonza: This is hardcoded to your computer ...

qiita_pet/handlers/study_handlers/sample_template.py
```diff
@@ -0,0 +1,51 @@
...
+            is_local, download[0], download[1], "Download sample information")
+
+        stats = sample_template_info(study_id, self.current_user.id)
+        self.render('study_ajax/sample_summary.html', stats=stats['summary'],
+                    num_samples=stats['num_samples'], dl_path=dl_path,
+                    files=files, study_id=study_id, data_types=data_types)
+
+    @authenticated
+    def post(self):
+        """Edit/delete/recreate sample template"""
```
antgonza: recreate?

antgonza:
Some comments.

qiita_pet/handlers/api_proxy/prep_template.py
```diff
@@ -0,0 +1,53 @@
...
+    ----------
+    study_id : int
+        Study id to get prep template info for
+    user_id : str
+        User id requesting the prep templates
+
+    Returns
+    -------
+    dict of list of dict
+        prep template information seperated by data type, in the form
```
mortonjt: Coming really late to the party.

What do you guys think about substituting out this dict of list of dict data types with pandas?

qiita_pet/handlers/api_proxy/sample_template.py
```diff
@@ -0,0 +1,242 @@
...
+    Parameters
+    ----------
+    samp_id : int
+        SampleTemplate id to get info for
+    user_id : str
+        User requesting the sample template info
+
+    Returns
+    -------
+    dict of list of tuples
```
mortonjt: Same comment about pandas here.

qiita_pet/handlers/api_proxy/prep_template.py
squirrelo: Can't use pandas since this is just proxying the JSON that will come from the RESTful API. Goes for everything in the api_proxy folder.

qiita_pet/handlers/api_proxy/sample_template.py
squirrelo: Same as above

qiita_pet/handlers/api_proxy/sample_template.py
squirrelo: Could you expand on what you mean by it makes it easier to search? Find works regardless of the string being searched for.

qiita_pet/handlers/api_proxy/sample_template.py
squirrelo: Could you expand on what you mean by different types? Also, added the returns.

qiita_pet/handlers/api_proxy/sample_template.py
antgonza: Basically, IMOO sample_template_info is too general and if you search for it in the code it could return a lot of hits. However, I we use get_request or post_request on the names it might simplify this. For example, you could quickly get all the funcs that do a post_request and see which one you need to modify. I feel like this is a good idea after spending a lot of time making modifications to the code and trying to find the correct function to change and hoping to have something like this to make it easier.

qiita_pet/handlers/api_proxy/sample_template.py
antgonza: Different types: 
- return access_error
- return {'status': 'error',
              'error': 'filepath does not exist',
              'filepath': sample_template}
-             return {'status': 'error',
                          'msg': 'Please, choose a data type if uploading a '
                          'QIIME mapping file',
                          'file': sample_template
                }
-         return {'status': status,
                      'message': msg,
                      'file': sample_template}

qiita_pet/handlers/api_proxy/sample_template.py
```diff
@@ -0,0 +1,242 @@
...
+                                 for w in warns])
+                status = 'warning'
+
+    except (TypeError, QiitaDBColumnError, QiitaDBExecutionError,
+            QiitaDBDuplicateError, IOError, ValueError, KeyError,
+            CParserError, QiitaDBDuplicateHeaderError,
+            QiitaDBError, QiitaWareError) as e:
+        # Some error occurred while processing the sample template
+        # Show the error to the user so they can fix the template
+        status = 'error'
```
antgonza: A couple of things here:
- There are 2 times status = 'error'
- AFIK The return is within the except so if warning if success there is no return

qiita_pet/handlers/api_proxy/sample_template.py
squirrelo: Ah, that was fixed but not pushed

qiita_pet/handlers/api_proxy/studies.py
squirrelo: Yes, it errors if not:
```python
TypeError: object of type 'setiterator' has no len()
```

antgonza: Got it, it returns an iterator, what about making it a list? AFAIK is a bit faster.

qiita_pet/handlers/study_handlers/sample_template.py
squirrelo: If the sample template already exists and we want to wipe it out and recreate it.

antgonza: OK, is that possible for the sample template? I remember it was but we remove it cause to do this we need to verify that there is no prep template and no artifacts associated ...

squirrelo: Makes my life easier, removed

antgonza: Thanks. Just to be clear we still need to have something that calls delete but not the delete/add in one single step ...

qiita_pet/handlers/api_proxy/prep_template.py
```diff
@@ -0,0 +1,53 @@
...
+    if access_error:
+        return access_error
+    # Can only pass ids over API, so need to instantiate object
+    study = Study(int(study_id))
+    prep_info = {}
+    for dtype in study.data_types:
+        prep_info[dtype] = []
+        for prep in study.prep_templates(dtype):
+            start_artifact = prep.artifact
+            info = {
```
wasade: Do `id` and  `start_artifact.id` refer to other endpoints in the inferface? If so, they should be formed URIs

qiita_pet/handlers/api_proxy/util.py
```diff
@@ -0,0 +1,15 @@
...
+
+def check_access(study_id, user_id):
+    try:
+        study = Study(int(study_id))
+    except QiitaDBUnknownIDError:
+        return {'status': 'error',
+                'message': 'Study does not exist'}
+    if not study.has_access(User(user_id)):
+        return {'status': 'error',
+                'message': 'User does not have access to study'}
```
wasade: if you look at the github api, it doesn't return a `status` on error but just has a `message`. In addition, includes a key `documentation_url` so that the api documentation is discoverable

qiita_pet/handlers/api_proxy/sample_template.py
```diff
@@ -0,0 +1,242 @@
...
+
+    # Define here the message and message level in case of success
+    msg = ''
+    status = 'success'
+    is_mapping_file = looks_like_qiime_mapping_file(fp_rsp)
+    study = Study(int(study_id))
+    try:
+        if is_mapping_file and not data_type:
+            return {'status': 'error',
+                    'msg': 'Please, choose a data type if uploading a '
```
wasade: `check_access` uses "message"

qiita_pet/handlers/api_proxy/sample_template.py
```diff
@@ -0,0 +1,242 @@
...
+                           'QIIME mapping file',
+                    'file': sample_template
```
wasade: is this filepath relative to the server and would it leak internal details about the filesystem?

qiita_pet/handlers/api_proxy/sample_template.py
```diff
@@ -0,0 +1,242 @@
...
+            remove(fp_rsp)
+
+            # join all the warning messages into one. Note that this
+            # info will be ignored if an exception is raised
+            if warns:
+                msg = '; '.join([convert_text_html(str(w.message))
+                                 for w in warns])
+                status = 'warning'
+
+    except (TypeError, QiitaDBColumnError, QiitaDBExecutionError,
```
wasade: why not just `except:`?

qiita_pet/handlers/api_proxy/sample_template.py
squirrelo: This is the filename, not a filepath, so we're good there.

qiita_pet/handlers/api_proxy/sample_template.py
```diff
@@ -0,0 +1,242 @@
...
+        return {'status': 'error',
+                'error': 'filepath does not exist',
+                'filepath': sample_template}
+
+    # Define here the message and message level in case of success
+    msg = ''
+    status = 'success'
+    is_mapping_file = looks_like_qiime_mapping_file(fp_rsp)
+    study = Study(int(study_id))
+    try:
```
wasade: there is a lot done in this try/except block. it appears as well that this method can return either a `dict` or `None`

squirrelo: All the code in here to update, create, and delete was copied over from the existing study description page. I can refactor it if we want, but for now I thought we just needed the new UI working.

qiita_pet/handlers/api_proxy/sample_template.py
```diff
@@ -0,0 +1,242 @@
...
+    user_id : int
+        The current user object id
+    """
+    access_error = check_access(int(study_id), user_id)
+    if access_error:
+        return access_error
+    try:
+        SampleTemplate.delete(int(study_id))
+    except Exception as e:
+        return {'status': 'error', 'message': str(e)}
```
wasade: Should `DELETE` be idempotent?

qiita_pet/handlers/api_proxy/sample_template.py
```diff
@@ -0,0 +1,242 @@
...
+    ----------
+    study_id : int
+        The current study object id
+    user_id : int
+        The current user object id
+    """
+    access_error = check_access(study_id, user_id)
+    if access_error:
+        return access_error
+    return SampleTemplate(int(study_id)).get_filepaths()
```
wasade: This returns server side details that are irrelevant for a client

qiita_pet/handlers/api_proxy/tests/test_sample_template.py
wasade: Yes. More specifically, the interface should not leak server side details: the client should not care about the nature of how templates are stored especially since the client doesn't have access to the filesystem.

qiita_pet/handlers/api_proxy/util.py
squirrelo: We can have a success message though, so that's why I have status. Google API also has status for geocoding and it makes it super easy to check results.

qiita_pet/handlers/api_proxy/sample_template.py
antgonza: Improving old code is good practice ...

qiita_pet/handlers/study_handlers/sample_template.py
```diff
@@ -0,0 +1,51 @@
...
+        action = self.get_argument('action')
```
wasade: Isn't this a hybrid SOAP + REST?

squirrelo: Yes. The quote UI (which is this handler) calls the rest interface to do everything. Sine we can only get and post through web forms, this is a needed check.

qiita_pet/webserver.py
```diff
@@ -106,6 +104,7 @@ def __init__(self):
             (r"/study/sharing/", ShareStudyAJAX),
             # ORDER FOR /study/description/ SUBPAGES HERE MATTERS.
             # Same reasoning as below. /study/description/(.*) should be last.
+            (r"/study/description/sample_template/", SampleTemplateAJAX),
```
wasade: Shouldn't the endpoints and collections described in the docstrings be expressed here?

squirrelo: These are not the RESTful API endpoints but qiita_pet AJAX call endpoints. These allow lazy loading the needed elements on the study info page.

wasade: Sorry, what I was expecting to see where handler entries for the API but
they do not appear to exist

squirrelo: They do not. The issue wis we are essentially now maintaining three separate projects in this single repo: qiita core functionality, qiita_pet UI, and the target gene plugin. This is why things are getting a bit odd. The api_proxy folder is what will become the API for qiita core once we separate these projects.

qiita_pet/handlers/study_handlers/sample_template.py
wasade: The nature of REST is to encode the actions in the URI as opposed to the arguments, so for instance, `create` is `POST`, `update` is `PUT`, etc. Why not just use the handlers to resolve the action?

squirrelo: This is not part of the RESTful interface though. This is the handler for the actual rendering and actions from the qiita_pet UI. This will then use the restful interface (currently just function calls) once the projects are separated.

qiita_pet/handlers/api_proxy/studies.py
squirrelo: Timings?

qiita_pet/webserver.py
wasade: ah

qiita_pet/handlers/api_proxy/studies.py
wasade: isn't it just O(n) vs O(n + n), which is O(n)?

antgonza: http://stackoverflow.com/questions/7717011/which-is-faster-and-why-set-or-list

antgonza: http://stackoverflow.com/questions/2831212/python-sets-vs-lists

wasade: Why are we discussing optimizing something without evidence that it needs to be optimized? 

@squirrelo, the motive to cast to `set` is to reduce duplicate elements here but I don't think the expectation is that `samples` contains duplicated elements. Using `set` obscures the motive which is to unroll the iterator. Given the motive, `list` is more apt.

qiita_pet/handlers/api_proxy/sample_template.py
squirrelo: Probably not. If it doesn't exist we should let the user know it doesn't exist.

qiita_pet/handlers/api_proxy/tests/test_sample_template.py
squirrelo: Now returns a URL for download instead of a file path.

squirrelo:
Comments addressed.

qiita_pet/handlers/api_proxy/prep_template.py
squirrelo: Not sure yet. The API design is still evolving.

qiita_pet/handlers/api_proxy/sample_template.py
wasade: Should it? The client can determine it's existence with a `GET`. Without the error, the operation is idempotent which simplifies clientside handling

squirrelo: True, but that’s also two hits against the API instead of one and logic to check that first response instead of the error on delete. Kind of a tradeoff either way.

wasade: Right, but this is complicating the API under the guise of optimization without evidence that the optimization is needed. The internals can always be reworked and sped up independent of clients, but changing the API later will be a nightmare.

squirrelo: This code will still be the same, just a pass here instead of the current return, since we don't want the server erroring if it can't delete. The dictionary layout is consistent with every other one in the API, so behavior isn't any different for this call specifically. I'm not sure what you mean by it is complicating the API.

qiita_pet/handlers/api_proxy/util.py
josenavas: @wasade @squirrelo just a comment on this
@ElDeveloper made the same comment in one of my PR and it actually goes against the design of RESTapi. However, we agreed that in interest of time we are going to leave as it was and we are going to update later. The RESTapi that is currently used by the plugins is in a very early development stage and it can still change a bit.

I'm planning to do those changes while I'm away, so we have a more standard RESTapi.

qiita_pet/handlers/api_proxy/tests/test_sample_template.py
josenavas: @wasade - just a quick comment on this.

In order to simplify the early development of the RESTapi for the plugins
(since they're clients of the RESTapi) was to assume that they have access
to the same filepaths, so we avoid unnecessary file transfers. It was
discussed that we will improve this later, using just URIs to represent the
files, so the plugins will decide if they have to download the file or just
access directly to the filesystem. Does this make sense?

2015-12-22 9:29 GMT-08:00 Daniel McDonald <notifications@github.com>:

> In qiita_pet/handlers/api_proxy/tests/test_sample_template.py
> <https://github.com/biocore/qiita/pull/1582#discussion_r48276601>:
>
> > +    def test_delete_sample_template(self):
> > +        obs = delete_sample_template(1, 'test@foo.bar')
> > +        exp = {'status': 'error',
> > +               'message': 'Sample template can not be erased because there are'
> > +                          ' prep templates associated.'}
> > +        self.assertEqual(obs, exp)
> > +
> > +    def test_delete_sample_template_no_access(self):
> > +        obs = delete_sample_template(1, 'demo@microbio.me')
> > +        exp = {'status': 'error',
> > +               'message': 'User does not have access to study'}
> > +        self.assertEqual(obs, exp)
> > +
> > +    def test_get_sample_template_filepaths(self):
> > +        obs = get_sample_template_filepaths(1, 'test@foo.bar')
> > +        exp = [(14, '/Users/jshorens/Repositories/qiita/qiita_db/support_'
>
> Yes. More specifically, the interface should not leak server side details:
> the client should not care about the nature of how templates are stored
> especially since the client doesn't have access to the filesystem.
>
> —
> <https://github.com/biocore/qiita/pull/1582/files#r48276601>.
>



-- 
Jose Navas

qiita_pet/handlers/api_proxy/sample_template.py
wasade: Sorry, had a slight confusion here. The method is actually idempotent as server side state does not change if the same request is made repeatedly. But in thinking about this further, shouldn't this set HTTP status to 404? 

In other words, I retract the initial comment as it was made in error but now wondering about HTTP status codes...

squirrelo: There are no status codes to worry about right now as these are not actually being served anywhere, just called directly in the qiita_pet handlers. I agree that once the API is actually in place and being served, status codes are important.

josenavas:
I just want to raise one concern here, which I already shared with @squirrelo in private.

Although we can say that this is proxying the RESTapi, I disagree that whatever decision is made here is actually the final RESTapi. A RESTapi cannot be designed by @squirrelo alone, neither by @squirrelo and I as he tried to do last Friday, and I specifically proposed to move forward using 2 directions: (1) using the document that @wasade and @ElDeveloper put together a while ago; and (2) group discussions in how the interface should look like.

My fear is that if we allow the RESTapi be defined here, it's design is going to be driven by the interface needs, rather than by common practices in RESTapi design and (specially) simplicity of the RESTapi.

If the interface needs to perform multiple hits to the RESTapi to obtain all the information is fine, the interface is a specific case, while the RESTapi will be used by a lot more users (i.e. users, interface, plugins, and who knows? people can develop their own clients!).

That being said, I don't think it is worth discussing the specifics of the RESTapi in this PR, since the proxy can be re-work later to provide the information to the interface as it is expecting - thus the proxy is a nice abstraction layer between what the interface needs and what the RESTapi provides.

Just wanted to comment this here, as I don't want my concerns to go unheard as it happened at the moment that this PR has been created.

qiita_pet/handlers/api_proxy/tests/test_sample_template.py
wasade: Can't the file transfers be avoided by a translation? f(uri) -> filepath.
Or are you saying that the development of that translation will be done
later?

josenavas: Yup, the development of that translation will be done later

2015-12-22 11:31 GMT-08:00 Daniel McDonald <notifications@github.com>:

> In qiita_pet/handlers/api_proxy/tests/test_sample_template.py
> <https://github.com/biocore/qiita/pull/1582#discussion_r48288874>:
>
> > +    def test_delete_sample_template(self):
> > +        obs = delete_sample_template(1, 'test@foo.bar')
> > +        exp = {'status': 'error',
> > +               'message': 'Sample template can not be erased because there are'
> > +                          ' prep templates associated.'}
> > +        self.assertEqual(obs, exp)
> > +
> > +    def test_delete_sample_template_no_access(self):
> > +        obs = delete_sample_template(1, 'demo@microbio.me')
> > +        exp = {'status': 'error',
> > +               'message': 'User does not have access to study'}
> > +        self.assertEqual(obs, exp)
> > +
> > +    def test_get_sample_template_filepaths(self):
> > +        obs = get_sample_template_filepaths(1, 'test@foo.bar')
> > +        exp = [(14, '/Users/jshorens/Repositories/qiita/qiita_db/support_'
>
> Can't the file transfers be avoided by a translation? f(uri) -> filepath.
> Or are you saying that the development of that translation will be done
> later?
> … <#151cb2c9c8e9077d_>
> On Tue, Dec 22, 2015 at 11:12 AM, Jose Navas <notifications@github.com>
> wrote: In qiita_pet/handlers/api_proxy/tests/test_sample_template.py <#1582
> (comment)
> <https://github.com/biocore/qiita/pull/1582#discussion_r48286866>>: > +
> def test_delete_sample_template(self): > + obs = delete_sample_template(1,
> 'test@foo.bar') > + exp = {'status': 'error', > + 'message': 'Sample
> template can not be erased because there are' > + ' prep templates
> associated.'} > + self.assertEqual(obs, exp) > + > + def
> test_delete_sample_template_no_access(self): > + obs =
> delete_sample_template(1, 'demo@microbio.me') > + exp = {'status':
> 'error', > + 'message': 'User does not have access to study'} > +
> self.assertEqual(obs, exp) > + > + def
> test_get_sample_template_filepaths(self): > + obs =
> get_sample_template_filepaths(1, 'test@foo.bar') > + exp = [(14,
> '/Users/jshorens/Repositories/qiita/qiita_db/support_' @wasade
> <https://github.com/wasade> <https://github.com/wasade> - just a quick
> comment on this. In order to simplify the early development of the RESTapi
> for the plugins (since they're clients of the RESTapi) was to assume that
> they have access to the same filepaths, so we avoid unnecessary file
> transfers. It was discussed that we will improve this later, using just
> URIs to represent the files, so the plugins will decide if they have to
> download the file or just access directly to the filesystem. Does this make
> sense? 2015-12-22 9:29 GMT-08:00 Daniel McDonald <notifications@github.com>:
> … <#151cb1bb6bffcca4_> In
> qiita_pet/handlers/api_proxy/tests/test_sample_template.py <#1582
> <https://github.com/biocore/qiita/pull/1582> (comment) <#1582 (comment)
> <https://github.com/biocore/qiita/pull/1582#discussion_r48276601>>>: > +
> def test_delete_sample_template(self): > + obs = delete_sample_template(1,
> 'test@foo.bar') > + exp = {'status': 'error', > + 'message': 'Sample
> template can not be erased because there are' > + ' prep templates
> associated.'} > + self.assertEqual(obs, exp) > + > + def
> test_delete_sample_template_no_access(self): > + obs =
> delete_sample_template(1, 'demo@microbio.me') > + exp = {'status':
> 'error', > + 'message': 'User does not have access to study'} > +
> self.assertEqual(obs, exp) > + > + def
> test_get_sample_template_filepaths(self): > + obs =
> get_sample_template_filepaths(1, 'test@foo.bar') > + exp = [(14,
> '/Users/jshorens/Repositories/qiita/qiita_db/support_' Yes. More
> specifically, the interface should not leak server side details: the client
> should not care about the nature of how templates are stored especially
> since the client doesn't have access to the filesystem. — Reply to this
> email directly or view it on GitHub <
> https://github.com/biocore/qiita/pull/1582/files#r48276601>. -- Jose
> Navas — Reply to this email directly or view it on GitHub <
> https://github.com/biocore/qiita/pull/1582/files#r48286866>.
>
> —
> <https://github.com/biocore/qiita/pull/1582/files#r48288874>.
>



-- 
Jose Navas

squirrelo:
At this point in development the only difference in opinion over the API is the addition of possible API URLs to the docstrings. In an effort to get this merged, those have been removed.

qiita_pet/handlers/api_proxy/sample_template.py
```diff
@@ -0,0 +1,266 @@
...
+        filepath to use for updating
+
+    Returns
+    -------
+    dict
+        results dictonary in the format
+        {'status': status,
+         'message': msg,
+         'file': sample_template}
+
```
mortonjt: Can you add a Raises section and specify the exceptions you expect to catch?

qiita_pet/handlers/api_proxy/sample_template.py
```diff
@@ -0,0 +1,266 @@
...
+    Parameters
+    ----------
+    study_id : int
+        The current study object id
+    user_id : str
+        The current user object id
+
+    Returns
+    -------
+    list of tuple of int and str
```
mortonjt: This needs to be updated, because it also returns an access error.

Would this be a good time to throw an exception/warning?

qiita_pet/handlers/api_proxy/sample_template.py
```diff
@@ -0,0 +1,266 @@
...
+    user_id : str
+        The current user object id
+
+    Returns
+    -------
+    dict
+        results dictonary in the format
+        {'status': status,
+         'message': msg}
+
```
mortonjt: Same here, the Returns clause needs to be updated at the very least

qiita_pet/handlers/api_proxy/studies.py
```diff
@@ -0,0 +1,69 @@
...
+    ----------
+    study_id : int
+        Study id to get prep template info for
+    user_id : str
+        User requesting the info
+
+    Returns
+    -------
+    dict of info
+        Study information seperated by data type, in the form
```
mortonjt: Same here.

qiita_pet/handlers/api_proxy/sample_template.py
```diff
@@ -0,0 +1,266 @@
...
+        {'status': status,
+         'message': msg,
+         'file': sample_template}
+
+    status can be success, warning, or error depending on result
+    message has the warnings or errors
+    file has the file name
+    """
+    access_error = check_access(int(study_id), user_id)
+    if access_error:
```
mortonjt: This is the last one that I will mark.  There are a few more incidents of this

qiita_pet/handlers/api_proxy/sample_template.py
squirrelo: No, no errors should be raised in the entirety of the api proxy since it is a proxy. That is the point of having the status: error and status: warning in the system.

qiita_pet/handlers/api_proxy/sample_template.py
squirrelo: No, this is correct, as the access_error is a dictionary in the exact same format. This goes for all the other incidents pointed out.

qiita_pet/handlers/api_proxy/sample_template.py
squirrelo: I do not understand what you are asking for here.

mortonjt: You are trying to catch all exceptions, regardless of the exception is.

Why are you doing exception handling?  Can you spell that out?  And can you add to a Notes section?

qiita_pet/handlers/api_proxy/sample_template.py
mortonjt: That is not correct.  Please take a look at the function signature [here](https://github.com/biocore/qiita/blob/master/qiita_pet/handlers/util.py#L18)

qiita_pet/handlers/api_proxy/sample_template.py
squirrelo: The whole point of this proxy is that all errors are caught and passed out as part of the return dict, hence catching all errors. They need to be shown to the user, not raised in the system and taking down the webserver.

squirrelo: Also, this is explained in the exceptions comments.

qiita_pet/handlers/api_proxy/sample_template.py
squirrelo: Please look at what the code is calling, as it is not that function. It is [this one](https://github.com/squirrelo/qiita/blob/artifact-study-api/qiita_pet/handlers/api_proxy/util.py#L6).

mortonjt: Wow. That's confusing :S

I'll eat all my words

qiita_pet/handlers/api_proxy/prep_template.py
josenavas: :+1:

qiita_pet/handlers/api_proxy/prep_template.py
josenavas: Agree with @squirrelo .

qiita_pet/handlers/api_proxy/sample_template.py
```diff
@@ -0,0 +1,269 @@
...
+        return access_error
+    template = SampleTemplate(int(samp_id))
+    df = template.to_dataframe()
+    out = {'num_samples': df.shape[0],
+           'summary': {}}
+
+    # drop the samp_id column if it exists
+    if 'study_id' in df.columns:
+        df.drop('study_id', axis=1, inplace=True)
+    cols = list(df.columns)
```
josenavas: Why is this needed? Can't you just simply do `for column in df.columns`?

qiita_pet/handlers/api_proxy/sample_template.py
```diff
@@ -0,0 +1,269 @@
...
+    for column in cols:
+        counts = df[column].value_counts()
+        out['summary'][str(column)] = [(str(key), counts[key])
+                                       for key in natsorted(counts.index)]
+
+    return out
+
+
+@execute_as_transaction
```
josenavas: Why is this executed as a transaction but not the other ones?

qiita_pet/handlers/api_proxy/sample_template.py
```diff
@@ -0,0 +1,269 @@
...
+                msg = '; '.join([convert_text_html(str(w.message))
+                                 for w in warns])
+                status = 'warning'
+
+    except Exception as e:
+        # Some error occurred while processing the sample template
+        # Show the error to the user so they can fix the template
+        status = 'error'
+        msg = str(e)
+    return {'status': status,
```
josenavas: Why is this returning the file itself? Shouldn't it only be status and message?

qiita_pet/handlers/api_proxy/sample_template.py
```diff
@@ -0,0 +1,269 @@
...
+    status can be success, warning, or error depending on result
+    message has the warnings or errors
+    file has the file name
+    """
+    access_error = check_access(study_id, user_id)
+    if access_error:
+        return access_error
+    # Define here the message and message level in case of success
+    status = "success"
+    # Get the uploads folder
```
josenavas: Code duplication with the above function.

qiita_pet/handlers/api_proxy/sample_template.py
```diff
@@ -0,0 +1,269 @@
...
+            if warns:
+                msg = '\n'.join(set(str(w.message) for w in warns))
+                status = 'warning'
+
+    except Exception as e:
+            status = 'error'
+            msg = str(e)
+    return {'status': status,
+            'message': msg,
+            'file': sample_template}
```
josenavas: Same here re returning the file.

qiita_pet/handlers/api_proxy/sample_template.py
```diff
@@ -0,0 +1,269 @@
...
+    user_id : str
+        The current user object id
+
+    Returns
+    -------
+    dict
+        Filepaths in the form
+        {'status': status, message: msg, filepaths: filepaths}
+        Where filepaths is a list of tuple of int and str
+        All files in the sample template, as [(id, URL), ...]
```
josenavas: Does the interface care about filepaths ids?

qiita_pet/handlers/api_proxy/studies.py
```diff
@@ -0,0 +1,69 @@
...
+    Parameters
+    ----------
+    study_id : int
+        Study id to get prep template info for
+    user_id : str
+        User requesting the info
+
+    Returns
+    -------
+    dict of info
```
josenavas: Here you should describe the types correctly: `dict of {str: str}`?

qiita_pet/handlers/api_proxy/tests/test_prep_template.py
```diff
@@ -0,0 +1,26 @@
+from unittest import TestCase, main
```
josenavas: Missing copyright notice, here and in all the other test files.

qiita_pet/handlers/api_proxy/tests/test_prep_template.py
```diff
@@ -0,0 +1,26 @@
...
+
+from qiita_pet.handlers.api_proxy.prep_template import prep_template_get_req
+
+
+class TestPrepAPI(TestCase):
+    def test_prep_template_get_req(self):
+        obs = prep_template_get_req(1, 'test@foo.bar')
+        exp = {'18S': [
+            {'id': 1,
+             'name': 'PREP 1 NAME',
```
josenavas: Is there an issue to add the name property to the prep template? I remember that we talked about adding the name property to the artifacts but not sure if there is an issue for the prep template, and I agree that is a good idea.

qiita_pet/handlers/api_proxy/tests/test_prep_template.py
```diff
@@ -0,0 +1,26 @@
...
+             'status': 'private',
+             'start_artifact_id': 1,
+             'start_artifact': 'FASTQ',
+             'last_artifact': 'TODO new gui'
+             }]}
+        self.assertEqual(obs, exp)
+
+    def test_prep_template_get_req_no_access(self):
+        obs = prep_template_get_req(1, 'demo@microbio.me')
+        exp = {'status': 'error',
```
josenavas: If you are using the status + message strategy that has been used in other points, the previous test case reflects that the returns are inconsistent. The return in the previous case (success) should be:
`{'status': 'success', 'message': None, 'info': {<previous_result>}}`

qiita_pet/handlers/api_proxy/tests/test_sample_template.py
```diff
@@ -0,0 +1,161 @@
...
+
+from qiita_core.qiita_settings import qiita_config
+from qiita_pet.handlers.api_proxy.sample_template import (
+    sample_template_get_req, sample_template_post_req,
+    sample_template_put_req, sample_template_delete_req,
+    sample_template_filepaths_get_req)
+
+
+class TestSampleAPI(TestCase):
+    def test_sample_template_get_req(self):
```
josenavas: Same here regarding status + message strategy, and similarly to all the other ones.

qiita_pet/handlers/util.py
```diff
@@ -50,7 +50,7 @@ def download_link_or_path(is_local_request, filepath, fp_id, label):
         resp = "<b>%s:</b> %s" % (label, filepath)
     else:
         resp = ('<a class="btn btn-default glyphicon glyphicon-download-alt" '
-                'href="/download/%s" style="word-spacing: -10px;"> %s</a>'
+                'href="%s" style="word-spacing: -10px;"> %s</a>'
```
josenavas: I'm confused about this change. Why is this needed? Isn't this doing the download? In the live system I think this is the mechanism used to offload the download to nginx.

qiita_pet/templates/study_ajax/sample_summary.html
```diff
@@ -0,0 +1,78 @@
+{% from future.utils import viewitems %}
```
josenavas: Can we avoid as much as python code here? I don't think is strictly necessary to use viewitems here.

qiita_ware/util.py
```diff
@@ -97,38 +96,6 @@ def per_sample_sequences(iter_, max_seqs, min_seqs=1, random_buf_size=100000):
             yield (sequence_id, sequence)
 
 
-def stats_from_df(df):
```
josenavas: Should the test for this function also be removed?

josenavas:
Few comments.

qiita_pet/handlers/api_proxy/sample_template.py
squirrelo: It's returning the filepath, not the file itself. This allows better messaging to users and anything else that acts on the system, since you know not only if it was successful, but what the function is acting on.

qiita_pet/handlers/api_proxy/sample_template.py
squirrelo: Sane answer

qiita_pet/handlers/api_proxy/sample_template.py
squirrelo: No, not currently

qiita_pet/handlers/api_proxy/studies.py
squirrelo: it's not always a dict of strings. Some are list of strings as the value, others are list of ints.

qiita_pet/handlers/api_proxy/tests/test_prep_template.py
squirrelo: Good catch, this was made before settling on that design. Fixed.

qiita_pet/handlers/util.py
squirrelo: This was changed because the back and forth over whether to return the URL or the filepath. This can get reverted since we decided filepath.

qiita_pet/handlers/api_proxy/tests/test_sample_template.py
squirrelo: Fixed

qiita_pet/templates/study_ajax/sample_summary.html
squirrelo: Are we still trying to be 2/3 compatible? If so, we nee this here.

qiita_pet/handlers/api_proxy/sample_template.py
squirrelo: Sorry, not filepath, the filename.

qiita_pet/handlers/api_proxy/sample_template.py
squirrelo: fixed

qiita_pet/handlers/api_proxy/sample_template.py
squirrelo: The ones that make change to the DB and filesystem are. The only one that isn't is the get request, since it only does select statements.

qiita_pet/handlers/api_proxy/sample_template.py
squirrelo: Wait, yes. Needed for the download since it downloads using the filepath ID in the url.

qiita_pet/handlers/api_proxy/tests/test_prep_template.py
squirrelo: Added #1598

qiita_ware/util.py
squirrelo: I'm dealing with removals at the end, since tests are failing anyway. That way everything is gutted in one fell swoop.

squirrelo:
Comments addressed.

qiita_pet/handlers/api_proxy/prep_template.py
josenavas: @squirrelo you missed this one

qiita_pet/handlers/api_proxy/sample_template.py
josenavas: From a RESTapi point of view, this doesn't make any sense. However, since this is not the RESTapi, it is ok.

qiita_pet/handlers/api_proxy/studies.py
josenavas: in such case `dict of {str: object}`

qiita_pet/handlers/api_proxy/prep_template.py
squirrelo: done

qiita_pet/handlers/api_proxy/studies.py
squirrelo: done

squirrelo:
comments addressed

qiita_pet/handlers/api_proxy/sample_template.py
```diff
@@ -0,0 +1,284 @@
...
+import warnings
+
+from os import remove
+from os.path import exists, join
+
+from natsort import natsorted
+from qiita_db.metadata_template.sample_template import SampleTemplate
+from qiita_db.study import Study
+from qiita_core.util import execute_as_transaction
+
```
josenavas: Remove space

qiita_pet/handlers/api_proxy/studies.py
```diff
@@ -0,0 +1,71 @@
...
+    """Returns data types available in the system
+
+    Returns
+    -------
+    list of str
+        Data types available on the system
+    """
+    return {'status': 'success',
+            'message': '',
+            'data_types': Study.all_data_types()
```
josenavas: I just saw this and I think this is not correct. Why should we go through the study to retrieve all the data types? Shouldn't it be a util function? This function has been merged in the artifact-study-pages branch previously but I highly recommend removing it from the study object and create a new util function.

qiita_pet/handlers/api_proxy/studies.py
```diff
@@ -0,0 +1,71 @@
...
+    Parameters
+    ----------
+    study_id : int
+        Study id to get prep template info for
+    user_id : str
+        User requesting the info
+
+    Returns
+    -------
+    dict of {str: object}
```
josenavas: This documentation is actually incorrect. It returns `dict of {'status': str, 'message': str, 'info': dict of {str: object}`

The same applies for the above function.

qiita_pet/handlers/api_proxy/tests/test_sample_template.py
```diff
@@ -0,0 +1,170 @@
...
+from qiita_pet.handlers.api_proxy.sample_template import (
+    sample_template_summary_get_req, sample_template_post_req,
+    sample_template_put_req, sample_template_delete_req,
+    sample_template_filepaths_get_req)
+
+
+class TestSampleAPI(TestCase):
+    def test_sample_template_summary_get_req(self):
+        obs = sample_template_summary_get_req(1, 'test@foo.bar')
+        exp = {'summary': {
```
josenavas: Here also applies the above comment about returning `{'status': 'success', 'message': None, 'info': {<previous_result>}}`, please fix here and in any other instance.

qiita_pet/templates/study_ajax/sample_summary.html
josenavas: I think the idea was to eventually move to py3 only (we are currently not even py3 compatible). In interest of time, fine with leaving as it is.

qiita_ware/util.py
josenavas: That sentence is actually contradictory, since you are removing the function now. I would suggest either remove both this function and the code or leave both, but not remove one but not the other. Note that this is creating an import error in the test file, so the other tests on that file do not run, which are independent to your changes. There is no need to break more tests than the strictly needed ones.

josenavas:
Few more comments.

qiita_pet/handlers/api_proxy/studies.py
squirrelo: A study tells you what data types it has, so it makes sense that the uninstantiated study can tell you all data types available. Also, as you have stated, this has already been reviewed and merged previously, with no comment before.  If you didn't agree with that, you should raise up your concern then.

qiita_pet/handlers/api_proxy/tests/test_sample_template.py
squirrelo: done

qiita_ware/util.py
squirrelo: done

josenavas:
:+1: Can somebody else review this PR? There has been a few changes since other dev than me reviewed this PR.

antgonza:
Merging!

Issue #1576, opened 2015-12-14
Add HTML storage capability to Artifacts


squirrelo:
Since artifacts generate static files, the summary pages for the artifacts will also be static HTML. We can store this HTML as part of the Artifact, so we do not have to get summary information and re-render the page every time it is called. This allows much faster responses by the UI.

antgonza:
Solved in: https://github.com/biocore/qiita/pull/1738

Issue #1573, opened 2015-12-14
Add last created artifact information


squirrelo:
For the study page UI, we need to easily grab the last Artifact created using the prep template information. This will allow showing the status of that artifact on the UI.

PR #1583, opened 2015-12-21
Study prep

squirrelo:
Requires #1582 

Adds the prep template information and graphs to the study page.

Still needs the `Artifact(id).to_html()` #1576  and `PrepTemplate(id).last_artifact` #1573 to be completed.


antgonza:
@squirrelo could you update ?

squirrelo:
I will update when #1582 is merged, as it is already a complicated conflict to do that update.

squirrelo:
Updated

antgonza:
Could you pull from the branch? What's the plan for #1576 and #1573? Are they going to be implemented in this PR or in another one?

squirrelo:
They will be implemented as a different PR. Right now there are placeholders where that information will go once it is available.

antgonza:
OK, thanks. Do you know who is implementing them or are they part of an open PR?

squirrelo:
They will be implemented by @josenavas, no current PR is in.

josenavas:
@squirrelo see #1599 for how to get the youngest artifact in the lineage. Since that is something that has to do with the lineage of the artifact, I don't think it is correct to add it to the prep template, and I think it should be a call in the artifact. Thus, the call should look something like: `PrepTemplate(id).artifact.youngest_artifact`
Note that the name change is also to be consistent with the other nomenclature used in the Artifact object.

squirrelo:
The `youngest_artifact` and `artifact.name` information are now implemented in this pull request. The to_html can wait and be added in as a micro-PR, so that should not block review of this PR.

qiita_pet/handlers/api_proxy/prep_template.py
```diff
@@ -6,52 +6,358 @@
...
+
+    Returns
+    -------
+    str
+        The investigation type chosen by the user
+    """
+    if inv_type == '':
+        inv_type = None
+    elif inv_type == 'Other' and user_def_type == 'New Type':
+        # This is a nre user defined investigation type so store it
```
antgonza: what's nre?

qiita_pet/handlers/api_proxy/prep_template.py
```diff
@@ -6,52 +6,358 @@
...
+def prep_template_post_req(study_id, user_id, prep_template, data_type,
+                           investigation_type=None,
+                           user_defined_investigation_type=None,
+                           new_investigation_type=None):
+    """Adds a prep template to the system
+
+    Parameters
+    ----------
+    study_id : int
+        Study to attach the pre ptemplate to
```
antgonza: prep template.

qiita_pet/handlers/api_proxy/prep_template.py
rob-knight: typo for “new”?

qiita_pet/handlers/api_proxy/prep_template.py
```diff
@@ -6,52 +6,358 @@
...
+        Existing investigation type to attach to the prep template
+    user_defined_investigation_type: str, optional
+        Existing user added investigation type to attach to the prep template
+    new_investigation_type: str, optional
+        Investigation type to add to the system
     """
     access_error = check_access(study_id, user_id)
     if access_error:
         return access_error
-    # Can only pass ids over API, so need to instantiate object
```
antgonza: This block wasn't use before, right?

squirrelo: This code was moved to the study.py file in the api-proxy folder since it is a summary of what is in the study.

qiita_pet/handlers/api_proxy/prep_template.py
```diff
@@ -6,52 +6,358 @@
...
+    prep_template : str, optional
+        filepath to use for updating
+    investigation_type: str, optional
+        Existing investigation type to attach to the prep template
+    user_defined_investigation_type: str, optional
+        Existing user added investigation type to attach to the prep template
+    new_investigation_type: str, optional
+        Investigation type to add to the system
+    """
+    # Get the uploads folder
```
antgonza: This comment is wrong, right?

qiita_pet/handlers/api_proxy/prep_template.py
antgonza: Cool, thanks for the clarification.

qiita_pet/handlers/api_proxy/artifact.py
```diff
@@ -0,0 +1,53 @@
...
+    -------
+    dict of lists of tuples
+        A dictionary containing the edge list representation of the graph,
+        and the node labels. Formatted as:
+        {'status': status,
+         'message': message,
+         'edge_list': [(0, 1), (0, 2)...],
+         'node_labels': [(0, 'label0'), (1, 'label1'), ...]}
+
+    Raises
```
antgonza: This per se doesn't raise anything, right?

qiita_pet/handlers/api_proxy/prep_template.py
```diff
@@ -6,52 +6,358 @@
...
+    prep_template : str
+        Filepath to the prep template being added
+    data_type : str
+        Data type of the processed samples
+    investigation_type: str, optional
+        Existing investigation type to attach to the prep template
+    user_defined_investigation_type: str, optional
+        Existing user added investigation type to attach to the prep template
+    new_investigation_type: str, optional
+        Investigation type to add to the system
```
antgonza: This is missing returns

qiita_pet/handlers/api_proxy/prep_template.py
```diff
@@ -6,52 +6,358 @@
...
+        The current user object id
+    prep_template : str, optional
+        filepath to use for updating
+    investigation_type: str, optional
+        Existing investigation type to attach to the prep template
+    user_defined_investigation_type: str, optional
+        Existing user added investigation type to attach to the prep template
+    new_investigation_type: str, optional
+        Investigation type to add to the system
+    """
```
antgonza: Missing returns.

qiita_pet/handlers/api_proxy/prep_template.py
```diff
@@ -6,52 +6,358 @@
...
+def prep_template_filepaths_get_req(prep_id, user_id):
+    """Returns all filepaths attached to a prep template
+
+    Parameters
+    ----------
+    prep_id : int
+        The current prep template id
+    user_id : int
+        The current user object id
+    """
```
antgonza: Missing returns

qiita_pet/handlers/api_proxy/prep_template.py
```diff
@@ -6,52 +6,358 @@
...
+
+    Returns
+    -------
+    dict of lists of tuples
+        A dictionary containing the edge list representation of the graph,
+        and the node labels. Formatted as:
+        {'edge_list': [(0, 1), (0, 2)...],
+         'node_labels': [(0, 'label0'), (1, 'label1'), ...]}
+
+    Raises
```
antgonza: This doesn't raise anything, right?

qiita_pet/handlers/api_proxy/prep_template.py
```diff
@@ -6,52 +6,358 @@
...
+            'User': sorted(ontology.user_defined_terms)
+            }
+
+
+def prep_template_graph_get_req(prep_id, user_id):
+    """Returns graph of all artifacts created from the prep base artifact
+
+    Parameters
+    ----------
+    study_id : int
```
antgonza: No longer a parameter.

qiita_pet/handlers/api_proxy/prep_template.py
```diff
@@ -6,52 +6,358 @@
...
+from qiita_db.util import convert_to_id
 from qiita_db.study import Study
+from qiita_db.ontology import Ontology
+from qiita_db.metadata_template.prep_template import PrepTemplate
 
 
-def prep_template_get_req(study_id, user_id):
-    """Gives a summary of each prep template attached to the study
+@execute_as_transaction
+def _process_investigation_type(inv_type, user_def_type, new_type):
```
antgonza: This doesn't have an specific test, right?

qiita_pet/handlers/api_proxy/prep_template.py
```diff
@@ -6,52 +6,358 @@
...
-            start_artifact = prep.artifact
-            info = {
-                'name': 'PREP %d NAME' % prep.id,
-                'id': prep.id,
-                'status': prep.status,
-                'start_artifact': start_artifact.artifact_type,
-                'start_artifact_id': start_artifact.id,
-                'last_artifact': 'TODO new gui'
+    fp_rpt = check_fp(study_id, prep_template)
+    if isinstance(fp_rpt, dict):
```
antgonza: This if is not tested

qiita_pet/handlers/api_proxy/prep_template.py
```diff
@@ -6,52 +6,358 @@
...
+            data_type_id = convert_to_id(data_type, 'data_type')
+            # deleting previous uploads and inserting new one
+            PrepTemplate.create(load_template_to_dataframe(fp_rpt),
+                                Study(int(study_id)), int(data_type_id),
+                                investigation_type=investigation_type)
+            remove(fp_rpt)
+
+            # join all the warning messages into one. Note that this info
+            # will be ignored if an exception is raised
+            if warns:
```
antgonza: This is not tested.

qiita_pet/handlers/api_proxy/prep_template.py
```diff
@@ -6,52 +6,358 @@
...
+        and the values are list of tuples. Each tuple is an observed value in
+        the category and the number of times its seen.
+        Format {status: status,
+                message: message,
+                num_samples: value,
+                category: [(val1, count1), (val2, count2), ...], ...}
+    """
+    prep = PrepTemplate(int(prep_id))
+    access_error = check_access(prep.study_id, user_id)
+    if access_error:
```
antgonza: This is not tested.

antgonza: Deleting, wrong line/if.

qiita_pet/handlers/api_proxy/prep_template.py
```diff
@@ -6,52 +6,358 @@
...
+        return access_error
+    df = prep.to_dataframe()
+    out = {'num_samples': df.shape[0],
+           'summary': {}}
+
+    # drop the prep_id column if it exists
+    if 'study_id' in df.columns:
```
antgonza: This is not tested.

qiita_pet/handlers/api_proxy/prep_template.py
```diff
@@ -6,52 +6,358 @@
...
+        Investigation type to add to the system
+    """
+    # Get the uploads folder
+    prep = PrepTemplate(int(prep_id))
+    study_id = prep.study_id
+    access_error = check_access(study_id, user_id)
+    if access_error:
+        return access_error
+
+    if investigation_type:
```
antgonza: Could you confirm that test_prep_template_put_req covers this section? Thanks.

qiita_pet/handlers/api_proxy/prep_template.py
```diff
@@ -6,52 +6,358 @@
...
+    -------
+    dict of list of str
+        {'ENA': [term1, term2, ...],
+         'User': [userterm1, userterm2, ...]}
+    """
+    # Get all the ENA terms for the investigation type
+    ontology = Ontology(convert_to_id('ENA', 'ontology'))
+    # make "Other" last on the list
+    ena_terms = sorted(ontology.terms)
+    ena_terms.remove('Other')
```
antgonza: Why do you remove it and then add it?

qiita_pet/handlers/api_proxy/util.py
```diff
@@ -37,3 +40,31 @@ def check_access(study_id, user_id):
...
+    Parameters
+    ----------
+    study_id : int
+        Study file uploaded to
+    filename : str
+        name of the uploaded file
+
+    Returns
+    -------
+    dict or str
```
antgonza: To have a consistent return, should this always return a dict and if success return fp_rsp in file?

qiita_pet/handlers/api_proxy/util.py
```diff
@@ -37,3 +40,31 @@ def check_access(study_id, user_id):
         return {'status': 'error',
                 'message': 'User does not have access to study'}
     return {}
+
+
+def check_fp(study_id, filename):
```
antgonza: No test for this function.

qiita_pet/handlers/study_handlers/prep_template.py
```diff
@@ -0,0 +1,79 @@
+# -----------------------------------------------------------------------------
```
antgonza: One of the ideas of these changes was to create tests, right? However, there are no tests for this ...

antgonza:
Done with first pass: tests missing and some minor comments. Also, looking at travis' errors it seems like most of them are solvable as some of the new additions are failing or are simple things. I know we agree that some failures are fine but I think the new additions shouldn't fail.

qiita_pet/handlers/api_proxy/prep_template.py
squirrelo: yup, fixed.

qiita_pet/handlers/api_proxy/prep_template.py
squirrelo: To make sure it is the last item in the list.

antgonza: OK. That means that is always there, right? Cause if not it's gonna fail. So what about always removing and adding in the template?

squirrelo: `Other` is actually part of the ontology, so it is always there. Removing it completely from what is returned would alter the ontology, which we don't want to do.

antgonza: OK

qiita_pet/handlers/api_proxy/prep_template.py
squirrelo: Added a specific test for here.

qiita_pet/handlers/api_proxy/util.py
squirrelo: I suppose, although this means the contents of the `file` key will be different for error and non-error, since the full filepath will be there on success and just the filename on error.

antgonza: Well, both of them are gonna return a string, right? The difference is what you do with that string.

squirrelo:
Comments addressed.

antgonza:
Could you minimize qiita_pet/static/vendor/css/vis.css? After that :+1:

qiita_pet/handlers/api_proxy/prep_template.py
```diff
@@ -6,52 +6,383 @@
...
+        The investigation type chosen by the user
+    """
+    if inv_type == '':
+        inv_type = None
+    elif inv_type == 'Other' and user_def_type == 'New Type':
+        # This is a new user defined investigation type so store it
+        inv_type = new_type
+        ontology = Ontology(convert_to_id('ENA', 'ontology'))
+        ontology.add_user_defined_term(inv_type)
+    elif inv_type == 'Other' and user_def_type != 'New Type':
```
mortonjt: This segment of code doesn't get tested (as far as I know).  Does this need a test?

qiita_pet/handlers/api_proxy/artifact.py
```diff
@@ -0,0 +1,48 @@
+from qiita_db.artifact import Artifact
+from qiita_pet.handlers.api_proxy.util import check_access
+
+
+def artifact_graph_get_req(artifact_id, direction, user_id):
```
mortonjt: This also doesn't have a test.  Are we ok with this?

squirrelo:
tests added

qiita_pet/handlers/api_proxy/artifact.py
```diff
@@ -0,0 +1,48 @@
+from qiita_db.artifact import Artifact
```
josenavas: This file is missing the Copyright information on top.

qiita_pet/handlers/api_proxy/prep_template.py
```diff
@@ -6,52 +6,383 @@
...
+                'message': message,
+                'num_samples': value,
+                'category': [(val1, count1), (val2, count2), ...], ...}
+    """
+    prep = PrepTemplate(int(prep_id))
+    access_error = check_access(prep.study_id, user_id)
+    if access_error:
+        return access_error
+    df = prep.to_dataframe()
+    out = {'num_samples': df.shape[0],
```
josenavas: Missing the 'status' and 'message' keywords.

josenavas: Ooops - just saw that it is added later.

qiita_pet/handlers/api_proxy/prep_template.py
josenavas: I don't think this type of handling should be done outside of the prep template. Why are you checking if the study_id is present? Isn't it showing a knowledge of how that data is stored outside of the DB?

Also, recent discussions showed that we may want to add the study_id as a column, since it will help for meta-analysis - but that is a different discussion.

qiita_pet/handlers/api_proxy/prep_template.py
```diff
@@ -6,52 +6,383 @@
...
+    # drop the prep_id column if it exists
+    if 'study_id' in df.columns:
+        df.drop('study_id', axis=1, inplace=True)
+    cols = list(df.columns)
+    for column in cols:
+        counts = df[column].value_counts()
+        out['summary'][str(column)] = [(str(key), counts[key])
+                                       for key in natsorted(counts.index)]
+    # Add expected messaging and status info
+    out.update({'status': 'success', 'message': ''})
```
josenavas: Why don't you add those keywords/values at the creation time of the dictionary? I don't see the usage of adding them now and it will make the code more readable.

qiita_pet/handlers/api_proxy/prep_template.py
```diff
@@ -6,52 +6,383 @@
...
+    investigation_type = _process_investigation_type(
+        investigation_type, user_defined_investigation_type,
+        new_investigation_type)
+
+    msg = ''
+    status = 'success'
+    try:
+        with warnings.catch_warnings(record=True) as warns:
+            # force all warnings to always be triggered
+            warnings.simplefilter("always")
```
josenavas: I don't understand why you are forcing the warnings here - they should be already forced in the code base when the Warning is defined.

qiita_pet/handlers/api_proxy/prep_template.py
```diff
@@ -6,52 +6,383 @@
...
+    access_error = check_access(prep.study_id, user_id)
+    if access_error:
+        return access_error
+    return {'status': 'success',
+            'message': '',
+            'filepaths': prep.get_filepaths()
+            }
+
+
+def prep_ontology_get_req():
```
josenavas: Why is the ontology information attached to the prep template?

qiita_pet/handlers/api_proxy/tests/test_prep_template.py
```diff
@@ -6,28 +6,112 @@
...
+    def test_process_investigation_type_new_term(self):
+        randstr = ''.join([choice(ascii_letters) for x in range(30)])
+        obs = _process_investigation_type('Other', 'New Type', randstr)
+        self.assertEqual(obs, randstr)
+
+        # Make sure New Type added
+        ontology = Ontology(999999999)
+        self.assertIn(randstr, ontology.user_defined_terms)
+
+    def test_prep_ontology_get_req(self):
```
josenavas: Can we add a test in which the user defined terms are not empty?

qiita_pet/handlers/api_proxy/tests/test_prep_template.py
```diff
@@ -6,28 +6,112 @@
...
+                       'Metagenomics', 'Pooled Clone Sequencing',
+                       'Population Genomics', 'RNASeq', 'Resequencing',
+                       'Synthetic Genomics', 'Transcriptome Analysis',
+                       'Whole Genome Sequencing', 'Other'],
+               'User': [],
+               'status': 'success',
+               'message': ''}
+        self.assertEqual(obs, exp)
+
     def test_prep_template_get_req(self):
```
josenavas: I think this test is incomplete. It is only testing that one of the samples is present. I'm not sure if that is extensive enough.

qiita_pet/handlers/api_proxy/tests/test_prep_template.py
```diff
@@ -35,5 +119,184 @@ def test_prep_template_get_req_no_access(self):
                'message': 'User does not have access to study'}
         self.assertEqual(obs, exp)
 
+    def test_prep_template_post_req(self):
```
josenavas: It doesn't look like there is a test in which prep_template_post_req works correctly (i.e. adds a prep template to the system).

qiita_pet/handlers/api_proxy/prep_template.py
josenavas: @squirrelo ping on this.

qiita_pet/handlers/api_proxy/tests/test_prep_template.py
```diff
@@ -35,5 +119,184 @@ def test_prep_template_get_req_no_access(self):
...
+        self.assertEqual(obs, exp)
+
+    def test_prep_template_post_req_bad_filepath(self):
+        obs = prep_template_post_req(1, 'test@foo.bar', 'badfilepath', '16S')
+        exp = {'status': 'error',
+               'message': 'file does not exist',
+               'file': 'badfilepath'}
+        self.assertEqual(obs, exp)
+
+    def test_prep_template_put_req(self):
```
josenavas: This is also missing a test in which the function works with no issues.

qiita_pet/handlers/api_proxy/tests/test_prep_template.py
```diff
@@ -35,5 +119,184 @@ def test_prep_template_get_req_no_access(self):
...
+        self.assertEqual(obs, exp)
+
+    def test_prep_template_put_req_bad_filepath(self):
+        obs = prep_template_put_req(1, 'test@foo.bar', 'badfilepath')
+        exp = {'status': 'error',
+               'message': 'file does not exist',
+               'file': 'badfilepath'}
+        self.assertEqual(obs, exp)
+
+    def test_prep_template_delete_req(self):
```
josenavas: There is no test in which the delete works.

qiita_pet/handlers/api_proxy/util.py
josenavas: This always returns a str, what is different is the contents of the 'file' key. @squirrelo please update

qiita_pet/handlers/api_proxy/prep_template.py
squirrelo: This is so every warning is caught and recorded properly, so they can be displayed to the user.

qiita_pet/handlers/api_proxy/prep_template.py
squirrelo: Because this is the ENA ontology that is already attached to the prep template and prep template page currently.

qiita_pet/handlers/study_handlers/prep_template.py
josenavas: @squirrelo I don't think the tests for this have been added.

One thing that I found useful when I'm addressing comments from a PR is to answer "done" to the comment, so I make sure that I don't forget to address any of the comments.

qiita_pet/handlers/study_handlers/prep_template.py
```diff
@@ -0,0 +1,79 @@
...
+class PrepTemplateGraphAJAX(BaseHandler):
+    @authenticated
+    def get(self):
+        prep = to_int(self.get_argument('prep_id'))
+        self.write(prep_template_graph_get_req(prep, self.current_user.id))
+
+
+class PrepTemplateAJAX(BaseHandler):
+    @authenticated
+    def get(self):
```
josenavas: I still think there is too much work done in the handlers. Since we are already doing a rework of these handlers, I think it would be useful to start addressing #1145

squirrelo: I don't understand what you mean by too much work. All this is doing is requesting data to display in the HTML page. There is no processing of any information actually done here.

qiita_pet/handlers/study_handlers/sample_template.py
```diff
@@ -39,10 +40,15 @@ def get(self):
 
     @authenticated
     def post(self):
-        """Edit/delete sample template"""
+        """Edit/delete/create sample template"""
```
josenavas: I'm not seeing delete in here?

qiita_pet/handlers/study_handlers/sample_template.py
```diff
@@ -12,7 +12,8 @@
 from qiita_pet.handlers.util import download_link_or_path
 from qiita_db.util import get_files_from_uploads_folders
 from qiita_pet.handlers.api_proxy import (
-    sample_template_summary_get_req, sample_template_put_req,
+    sample_template_summary_get_req,
```
josenavas: I think this file is also missing tests?

qiita_pet/handlers/api_proxy/prep_template.py
josenavas: I don't think this needs to be here. Note that below you have a similar function and you are not calling `warnings.simplefilter`. The reason for that is because the system is already set up so by default the warnings are always triggered. Please remove as it is not needed.

qiita_pet/handlers/api_proxy/prep_template.py
josenavas: change name to `ena_ontology_get_req` as it is independent of the prep template.

qiita_pet/handlers/study_handlers/prep_template.py
```diff
@@ -0,0 +1,79 @@
...
+        """Send formatted summary page of prep template"""
+        study_id = self.get_argument('study_id')
+        prep_id = self.get_argument('prep_id')
+        files = [f for _, f in get_files_from_uploads_folders(study_id)
```
josenavas: This should be done somewhere else

qiita_pet/handlers/study_handlers/prep_template.py
josenavas: The problem is that you are not testing anywhere that the formatting that is done in this handlers is correct or not. What this means is that the interface can be broken in some cases without you noticing it because the tests are not exhaustive enough.

As I exposed in yesterday's meeting - ideally we want to be able to list all the possible ways to get to this handler and provide that as tests inputs. If that is done in a different function is way easier to test.

qiita_pet/handlers/api_proxy/prep_template.py
squirrelo: This is code from the original implementation that was part of the study description handler, so I can not speak to the original intention of the code. I can remove this if wanted.

qiita_pet/handlers/api_proxy/prep_template.py
squirrelo: done

qiita_pet/handlers/api_proxy/prep_template.py
squirrelo: done

qiita_pet/handlers/api_proxy/tests/test_prep_template.py
squirrelo: Done

qiita_pet/handlers/study_handlers/prep_template.py
squirrelo: The tests exist in `qiita_pet/test/test_study_handlers.py`

qiita_pet/handlers/api_proxy/prep_template.py
squirrelo: test was added previously

qiita_pet/handlers/api_proxy/tests/test_prep_template.py
squirrelo: Now tests all keys in template to make sure all samples are returned.

qiita_pet/handlers/api_proxy/util.py
squirrelo: done

qiita_pet/handlers/study_handlers/sample_template.py
squirrelo: line 55

qiita_pet/handlers/api_proxy/tests/test_prep_template.py
squirrelo: Done

qiita_pet/handlers/api_proxy/tests/test_prep_template.py
squirrelo: done

qiita_pet/handlers/api_proxy/tests/test_prep_template.py
squirrelo: done

qiita_pet/handlers/study_handlers/sample_template.py
squirrelo: The tests exist in `qiita_pet/test/test_study_handlers.py`

antgonza:
@squirrelo and @josenavas, what's the status of this PR?

qiita_pet/handlers/api_proxy/prep_template.py
```diff
@@ -6,52 +6,379 @@
...
+           'summary': {},
+           'status': 'success',
+           'message': ''}
+
+    cols = list(df.columns)
+    for column in cols:
+        counts = df[column].value_counts()
+        out['summary'][str(column)] = [(str(key), counts[key])
+                                       for key in natsorted(counts.index)]
+    # Add expected messaging and status info
```
josenavas: This comment is outdated.

josenavas:
Just one comment @squirrelo - let me know when it is addressed and I will merge.

squirrelo:
Done, ready to merge.

antgonza:
Thanks!!

PR #1584, opened 2015-12-21
Plugins oauth2

josenavas:
This modifies the plugins so they work with OAuth2.

This includes the commits from @squirrelo in https://github.com/biocore/qiita/pull/1562

### DIFF
--- a/qiita_db/handlers/oauth2.py
+++ b/qiita_db/handlers/oauth2.py
@@ -14,50 +14,37 @@ import functools
 from traceback import format_exception
 
 from tornado.web import RequestHandler
-
 from moi import r_client
+
 from qiita_core.exceptions import (IncorrectPasswordError, IncorrectEmailError,
                                    UnverifiedEmailError)
 import qiita_db as qdb
 
 
-def login_client(client_id, client_secret=None):
-    """Login client_id, optionally with client_secret
+def _oauth_error(handler, error_msg, error):
+    """Set expected status and error formatting for Oauth2 style error
 
-    Parameters
-    ----------
-    client_id : str
-        The client ID to validate
-    client_secret : str, optional
-        The client secret code. If not given, will only validate if client_id
-        exists
+   Parameters
+   ----------
+   error_msg : str
+        Human parsable error message
+    error : str
+        Oauth2 controlled vocab error
 
     Returns
     -------
-    bool
-        Whether client_id exists and, if client_secret given, the secret
-        matches the id
+    Writes out Oauth2 formatted error JSON of
+        {error: error,
+         error_description: error_msg}
+
+    Notes
+    -----
+    Expects handler to be a tornado RequestHandler or subclass
     """
-    with qdb.sql_connection.TRN:
-        sql = """SELECT EXISTS(
-                    SELECT *
-                    FROM qiita.oauth_identifiers
-                    WHERE client_id = %s {0})"""
-        sql_info = [client_id]
-        if client_secret is not None:
-            sql = sql.format("AND client_secret = %s")
-            sql_info.append(client_secret)
-        else:
-            sql = sql.format("AND client_secret IS NULL")
-        qdb.sql_connection.TRN.add(sql, sql_info)
-        return qdb.sql_connection.TRN.execute_fetchlast()
-
-
-def _oauth_error(self, error_msg, error):
-            self.set_status(400)
-            self.write({'error': error,
-                        'error_description': error_msg})
-            self.finish()
+    handler.set_status(400)
+    handler.write({'error': error,
+                   'error_description': error_msg})
+    handler.finish()
 
 
 def authenticate_oauth(f):
@@ -67,21 +54,31 @@ def authenticate_oauth(f):
     If an invalid header is given, a 400 error code is returned and the json
     error message is automatically sent.
 
+    Returns
+    -------
+    Sends oauth2 formatted error JSON if authorizaton fails
+
+    Notes
+    -----
+    Expects handler to be a tornado RequestHandler or subclass
+
     References
-    ---------
+    ----------
     [1] The OAuth 2.0 Authorization Framework.
     http://tools.ietf.org/html/rfc6749
     """
     @functools.wraps(f)
-    def wrapper(self, *args, **kwargs):
-        header = self.request.headers.get('Authorization', None)
+    def wrapper(handler, *args, **kwargs):
+        header = handler.request.headers.get('Authorization', None)
         if header is None:
-            _oauth_error(self, 'Oauth2 error: invalid access token',
+            _oauth_error(handler, 'Oauth2 error: invalid access token',
                          'invalid_request')
             return
         token_info = header.split()
+        # Based on RFC6750 if reply is not 2 elements in the format of:
+        # ['Bearer', token] we assume a wrong reply
         if len(token_info) != 2 or token_info[0] != 'Bearer':
-            _oauth_error(self, 'Oauth2 error: invalid access token',
+            _oauth_error(handler, 'Oauth2 error: invalid access token',
                          'invalid_grant')
             return
 
@@ -89,7 +86,7 @@ def authenticate_oauth(f):
         db_token = r_client.hgetall(token)
         if not db_token:
             # token has timed out or never existed
-            _oauth_error(self, 'Oauth2 error: token has timed out',
+            _oauth_error(handler, 'Oauth2 error: token has timed out',
                          'invalid_grant')
             return
         # Check daily rate limit for key if password style key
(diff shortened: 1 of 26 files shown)


antgonza:
Just to double check, the plan is to close #1562 and fully review this one, right?

qiita_core/support_files/server.key
```diff
@@ -0,0 +1,27 @@
+-----BEGIN RSA PRIVATE KEY-----
```
mortonjt: Maybe I'm missing something here...

Should this be a public key, not a private key?

josenavas: Nope, the server needs the private key to run.

antgonza:
@josenavas could you update your branch? Note that I merged #1562 with the idea that all test should pass on this one.

qiita_db/handlers/oauth2.py
```diff
@@ -0,0 +1,306 @@
...
+            sql = sql.format("AND client_secret IS NULL")
+        qdb.sql_connection.TRN.add(sql, sql_info)
+        return qdb.sql_connection.TRN.execute_fetchlast()
+
+
+def _oauth_error(self, error_msg, error):
+            self.set_status(400)
+            self.write({'error': error,
+                        'error_description': error_msg})
+            self.finish()
```
mortonjt: Over-indent!

squirrelo: @mortonjt this still needs a pull and update before review.

josenavas:
Done

antgonza:
Now this has conflicts ... could you fix? Thanks.

josenavas:
Done

antgonza:
:+1: when tests pass.

antgonza:
Note that this PR is going to have failures, @josenavas can you fix?

qiita_core/support_files/server.key
wasade: Will the server key be automatically regenerated? Otherwise, distributing a
private key is usually not a good idea

qiita_plugins/target_gene/tgp/tests/test_qiita_client.py
```diff
@@ -31,16 +37,32 @@ def tearDown(self):
...
+            "https://test_server.com/qiita_db/authenticate/",
+            body='{"access_token": "token", "token_type": "Bearer", '
+                 '"expires_in": "3600"}')
         obs = QiitaClient("https://test_server.com")
         self.assertEqual(obs._server_url, "https://test_server.com")
         self.assertTrue(obs._verify)
 
+    @httpretty.activate
     def test_init_cert(self):
+        httpretty.register_uri(
```
squirrelo: This is duplicated from setUp

qiita_plugins/target_gene/tgp/split_libraries/tests/test_util.py
```diff
@@ -51,6 +58,11 @@ def test_get_artifact_information(self):
             httpretty.GET,
             "https://test_server.com/qiita_db/artifacts/1/type/",
             body='{"type": "FASTQ", "success": true, "error": ""}')
+        httpretty.register_uri(
```
squirrelo: Duplicated from setUp

qiita_plugins/target_gene/tgp/split_libraries/tests/test_util.py
```diff
@@ -67,6 +79,11 @@ def test_get_artifact_error(self):
             httpretty.GET,
             "https://test_server.com/qiita_db/artifacts/1/filepaths/",
             body='{"filepaths": '', "success": false, "error": "some error"}')
+        httpretty.register_uri(
```
squirrelo: Same here

qiita_plugins/target_gene/tgp/tests/test_qiita_client.py
josenavas: I can't remove this one - tests fail. The other 2, however, have been removed.

josenavas:
Tests should be passing right now.

antgonza:
Thanks, one more question, should the values of the keys be shown in the qiita-test-install output?

josenavas:
That's probably a good point. However, how is that going to scale? Meaning, as the number of plugins increase, how ugly is going to be showing that list?

Do you think that adding to qiita-test-install output is the best one, or should we add a flag, --show-plugin-info that show such information? If so, we can potentially remove the print statements from the patch and just point to the script.

What do you think?

antgonza:
I thought the plan was to have a certificate per system not per plugin, is that inaccurate?

josenavas:
The certificate is for the system but each plugin has its own client id and client secret.

antgonza:
IMOO is fine to display all the plugins ids/secrets in the qiita-test-install by default but fine if you want to add a flag for this.

josenavas:
Ok, thanks! Will try to update later today - busy with classes right now...

squirrelo:
That seems like a bad security issue if we are going to have people post their qiita-test-install outputs anywhere for debugging, like we do with print_qiime_config, as it will expose all the system's client username and passwords. Flag would be safer, but this issue may still happen if someone accidentally posts their output with the flag on.

antgonza:
IMOO this is fine and why we have an email for help and not a forum. At least for the moment.

josenavas:
Ok, I just realized that to do this correctly I need to do some modifications on the DB. The issue is that the system does not know which client_id belongs to each plugin and although @squirrelo suggested that is not needed for the oauth2 workflow, it is actually needed for the system to keep track of ownerships.

I have to work in something else right now, but I will update the PR soon.

josenavas:
PR updated

qiita_db/support_files/qiita-db.dbs
```diff
@@ -1,9 +1,9 @@
 <?xml version="1.0" encoding="UTF-8" ?>
-<project name="qiita" id="Project669804" html="n" template="Default" database="PostgreSQL" >
-	<comment><![CDATA[Qiita-DB layout]]></comment>
+<project name="qiita" id="Project669804" database="PostgreSQL" >
+	<comment>Qiita-DB layout</comment>
 	<schema name="qiita" schemaname="qiita" >
 		<table name="analysis" >
-			<comment><![CDATA[hHolds analysis information]]></comment>
+			<comment>hHolds analysis information</comment>
```
antgonza: This has an extra h

qiita_db/support_files/qiita-db.dbs
```diff
@@ -48,7 +48,7 @@
 			</fk>
 		</table>
 		<table name="analysis_chain" >
-			<comment><![CDATA[Keeps track of the chain of analysis edits. Tracks what previous analysis a given analysis came from.If a given analysis is not in child_id, it is the root of the chain. ]]></comment>
+			<comment>Keeps track of the chain of analysis edits. Tracks what previous analysis a given analysis came from.If a given analysis is not in child_id, it is the root of the chain. </comment>
```
antgonza: Missing a space between "." and "If"

antgonza:
Thanks. One minor comment and then :+1:

qiita_db/support_files/qiita-db.dbs
josenavas: Not sure why this lines changed, as I didn't touch this part. Anyways, I try fixing this line and for some reason dbschema is not fixing it. When I save, it removes the space again.

josenavas:
Comments addressed

squirrelo:
What do you mean ownerships? By oauth2 standards, there are no specific ownerships for keys. All you transmit is the client id and client secret, and no other information. If you want to transmit the plugin id  or other information as well, that is no longer following oauth2 specification, and we need to discuss this.

josenavas:
I'm not transmitting this information anywhere - is a information that the system needs to know to inform the user which client id and client secret needs to provide to the plugin config file - does it make sense?

squirrelo:
Not really. There's no reason we couldn't use a single client id/secret per system for app plugins if we wanted to, and this may actually make sense for single-computer systems like laptop deploys, since it means the user only needs one client login instead of many. The multi-client login does makes sense for larger or distributed setups, however. 
I do not think we need to tie specific logins to specific plugins when displaying, since it creates a tie in the display that does not actually exist in the system. The system does not actually connect a login to a plugin in at all, as you could just as easily use the client id/secret for plugin A on plugin B and still have it work fine. The display should reflect that.

josenavas:
Ok this is getting annoying - in one of my previous comments I stated that each plugin has its own cliend_id/client_secret pair. If you didn't agree with that, you should raise up your concern then, instead of wasting my time. I'm not going to continue working on this until we reach a consensus.

antgonza:
There is nothing in the code that limits the use of the same id/secret if the user chooses to, right? In that case, we are covering both scenarios, right?

josenavas:
Yes - right now we are covering both scenarios.

josenavas:
I'm unsure why the tests are failing here - my guess is that it has to do with the Oauth2 authorization. I've done a small change to see if tests pass now as I can't reproduce in my machine.

josenavas:
Ok, it looks like it was a timing issue

josenavas:
@squirrelo ping

PR #1587, opened 2015-12-23
Artifact plugin setup

josenavas:
Creates a setup.py for the plugin.


josenavas:
This depends on https://github.com/biocore/qiita/pull/1581 and https://github.com/biocore/qiita/pull/1584 so review/merge those ones first.

squirrelo:
@antgonza brought up in #1581 that having a single command to install everything would be good to have, and I agree. Could you add this to the pull request?

josenavas:
I will try to add something hackish to the current setup.py that would work, but it is going to be hard to have a long maintainable setup.py with this option.

For example, I think I can make it to work when installing by doing python setup.py but I don't think I can make it to work when installing through pip. Other repos that have multiple setup.py (like numpy) the extra code is actually part of the repository so it is a bit different.

That being said, I think we need to split the target gene plugin apart before we actually cut the next Qiita release, but we still don't have a target date and I would like to have the plugins more mature in order to cut a release; so I don't think it is a big deal adding some hackish code to the setup.py that we will delete prior release. Does this sounds good to everybody?

antgonza:
It does, thanks. Let's discuss ETA/deadlines during tomorrow's meeting.

josenavas:
After after trying multiple options and an offline discussion with @antgonza, we agreed that it is ok to have to install Qiita using two commands instead of one. Note that this is a temporary solution while the target_gene plugin is still in the same repository that Qiita, but it will be simplified once the target_gene plugin is moved to its own repository.

I've updated the INSTALL.md document to include a section on how to install the development version and added the instructions here. I tested those instructions locally and I successfully installed Qiita and the plugin.

mortonjt:
Looks good to me

PR #1589, opened 2015-12-28
Sample summary page

squirrelo:
Depends on #1583

Adds the sample summary page for tabularizing what samples are in what prep templates, along with additional sample metadata if wanted.


josenavas:
@squirrelo let us know when this branch is ready for review.

squirrelo:
Ready for review.

qiita_pet/handlers/api_proxy/sample_template.py
```diff
@@ -23,6 +23,33 @@
...
+        User requesting the sample template info
+
+    Returns
+    -------
+    dict of dictionaries
+        Dictionary object where the keys are the metadata samples
+        and the values are a dictionary of column and value.
+        Format {sample: {column: value, ...}, ...}
+    """
+    template = SampleTemplate(int(samp_id))
```
antgonza: If the parameter is already int, not sure why we need this cast.

antgonza:
1 minor comment, rest looks good.

qiita_pet/handlers/api_proxy/sample_template.py
```diff
@@ -23,6 +23,33 @@
...
+    Parameters
+    ----------
+    samp_id : int
+        SampleTemplate id to get info for
+    user_id : str
+        User requesting the sample template info
+
+    Returns
+    -------
+    dict of dictionaries
```
josenavas: Can you update the docs so it actually matches with what is returned?

qiita_pet/handlers/api_proxy/sample_template.py
```diff
@@ -23,6 +23,33 @@
 from qiita_pet.handlers.api_proxy.util import check_access, check_fp
 
 
+def sample_template_get_req(samp_id, user_id):
```
josenavas: This method seems unsafe - what happens if the given SampleTemplate does not exist?

qiita_pet/handlers/api_proxy/sample_template.py
squirrelo: There have been a few times in past code where the direct input from a form was passed, so a string was passed instead of an int. The object still works in most cases with the string, so it's really hard to debug the few cases where it breaks. I put the explicit cast just to make absolutely sure it is an int when it goes into the instantiation.

qiita_pet/handlers/api_proxy/sample_template.py
squirrelo: done

qiita_pet/handlers/study_handlers/sample_template.py
```diff
@@ -57,3 +61,50 @@ def post(self):
         else:
             raise HTTPError(400, 'Unknown sample template action: %s' % action)
         self.write(result)
+
+
+def build_sample_summary(study_id, user_id, visible=None):
```
josenavas: I'm concerned about the performance of this call. This is going over the entire sample template and all the prep templates on that study. In cases like the AGP, this can become _really_ expensive.

Why has been all the metadata summary merged in a single page? Before it was per each template. I think that was the correct way of doing it...

qiita_pet/handlers/api_proxy/sample_template.py
```diff
@@ -23,6 +23,37 @@
...
+    dict of objects
+        {'status': status,
+         'message': msg,
+         'template': dict of {str: {str: object, ...}, ...}
+
+        template is dictionary where the keys access_error the metadata samples
+        and the values are a dictionary of column and value.
+        Format {sample: {column: value, ...}, ...}
+    """
+    template = SampleTemplate(int(samp_id))
```
josenavas: I think @antgonza comment here was the fact that the documentation of the function and the code are out of sync. You are describing the input as an int, but you are actually accepting ints and str that can be casted to int.

josenavas:
Few comments

qiita_pet/handlers/study_handlers/sample_template.py
squirrelo: This is the overall sample summary page @rob-knight and @antgonza asked for, where each sample is listed as part of or not part of a prep template's data set. The mockup is [here](https://biocore.mybalsamiq.com/projects/qiitastudiesv3/mouse%20decomp%20-%20sample%20summary).

josenavas: Ok, I see now, can you make the loading of the information dynamic? This looks extremely expensive as it is right now...

qiita_pet/handlers/api_proxy/sample_template.py
squirrelo: Good catch, fixed.

qiita_pet/handlers/study_handlers/sample_template.py
antgonza: That's correct and we believe is going to be super useful and cool. However, perhaps we can precompute that every time the templates are updated (including creation) and just display it here to make it safe. What do you think?

qiita_pet/handlers/api_proxy/sample_template.py
squirrelo: updated

qiita_pet/handlers/api_proxy/sample_template.py
josenavas: Discussed offline with @squirrelo - this is still not fixed and will be fixed soon

qiita_pet/handlers/study_handlers/sample_template.py
wasade: Memoize?

josenavas: I don't think memoization will buy us that much - and I don't think that precomputing is the way to go in here. The main concern that I have from here is that this is loading all the templates, but in the mockup it doesn't look like you need all that information. If I understand correctly, the only parts that you need is know in which prep template each sample is present and all the metadata categories that are available.

Based on that, you can load that information first and then just load the metadata columns that the user want to show once it adds them to the table.

It is also unclear from the mockup how columns that are present in one prep template but not in another prep template are handled.

antgonza: Good point, you only need which samples are in which prep ...

wasade: A generator + websocket would be pretty cool

qiita_pet/handlers/api_proxy/sample_template.py
squirrelo: now fixed.

qiita_pet/handlers/study_handlers/sample_template.py
squirrelo: The prep template is a bit of an issue right now, since multiple prep templates can have the same column name and not all samples will be on each column. Because of this issue, I'm only showing sample template columns for now.

qiita_pet/handlers/api_proxy/sample_template.py
```diff
@@ -23,6 +23,53 @@
 from qiita_pet.handlers.api_proxy.util import check_access, check_fp
 
 
+def _check_sample_template_exists(samp_id):
```
josenavas: Missing tests for this function.

josenavas: Also documentation.

qiita_pet/handlers/study_handlers/sample_template.py
josenavas: Missing documentation.

josenavas: Also, i'm unsure what did you changed in this function, but this is still asking for all the prep templates info - i.e. you are still transferring all the information. If you want, we can meet later today to find a way to improve this function.

qiita_pet/handlers/study_handlers/sample_template.py
```diff
@@ -57,3 +61,50 @@ def post(self):
...
+    # Add one column per prep template highlighting what samples exist
+    prep_cols = []
+    preps = study_prep_get_req(study_id, user_id)['info']
+    for dt in preps:
+        for prep in preps[dt]:
+            prep_samples = prep_template_get_req(
+                prep['id'], user_id)['template'].keys()
+            prep_df = pd.Series(['X'] * len(prep_samples),
+                                index=prep_samples, dtype=str)
+            col_name = ' - '.join([prep['name'], str(prep['id'])])
```
josenavas: col_name = '%s - %d' % (prep['name'], prep['id'])

qiita_pet/handlers/study_handlers/sample_template.py
```diff
@@ -57,3 +61,50 @@ def post(self):
...
+    return table, meta_available
+
+
+class SampleAJAX(BaseHandler):
+    def get(self):
+        """Show the sample summary page"""
+        study_id = self.get_argument('study_id')
+        meta_col = self.get_argument('meta_col', None)
+        visible = self.get_argument('meta_visible', [])
+        # Build list of metadta columns to keep
```
josenavas: metadta -> metadata

qiita_pet/handlers/study_handlers/sample_template.py
```diff
@@ -57,3 +61,50 @@ def post(self):
...
+
+    # Format the dataframe to html table
+    meta_available = meta_available.difference(visible)
+    table = df.to_html(classes='table table-striped', na_rep='',
+                       columns=prep_cols + visible)
+    return table, meta_available
+
+
+class SampleAJAX(BaseHandler):
+    def get(self):
```
josenavas: I'm having problems understanding what the parameters of this Handler means/do. Can you add more comments to the code?

qiita_pet/handlers/study_handlers/sample_template.py
```diff
@@ -57,3 +61,50 @@ def post(self):
...
+        """Show the sample summary page"""
+        study_id = self.get_argument('study_id')
+        meta_col = self.get_argument('meta_col', None)
+        visible = self.get_argument('meta_visible', [])
+        # Build list of metadta columns to keep
+        if visible:
+            visible = loads(visible)
```
josenavas: If `visible` is not provided, thus `visible = []`, this breaks:

```python
In [2]: loads([])
TypeError                                 Traceback (most recent call last)
<ipython-input-2-3e5b14ee389a> in <module>()
----> 1 loads([])

/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/json/__init__.pyc in loads(s, encoding, cls, object_hook, parse_float, parse_int, parse_constant, object_pairs_hook, **kw)
    336             parse_int is None and parse_float is None and
    337             parse_constant is None and object_pairs_hook is None and not kw):
--> 338         return _default_decoder.decode(s)
    339     if cls is None:
    340         cls = JSONDecoder

/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/json/decoder.pyc in decode(self, s, _w)
    364 
    365         """
--> 366         obj, end = self.raw_decode(s, idx=_w(s, 0).end())
    367         end = _w(s, end).end()
    368         if end != len(s):

TypeError: expected string or buffer
```

josenavas:
Few more comments.

qiita_pet/handlers/study_handlers/sample_template.py
squirrelo: I have not changed anything, just making the comment. Still working on the new implementation.

josenavas: Oops! The latest pushes made me think that you addressed the comments.

qiita_pet/handlers/api_proxy/sample_template.py
squirrelo: Added

squirrelo:
In the interest of not growing this PR any larger, the prep template information/sample will be added in another pull request. It has also been added to the list of things to complete for this branch PR.

qiita_db/metadata_template/base_metadata_template.py
```diff
@@ -1320,6 +1320,30 @@ def update_category(self, category, samples_and_values):
...
+            Metadata category to get information for
+
+        Returns
+        -------
+        dict
+            Sample metadata for the category in the form {sample_id: value}
+
+        Raises
+        ------
+        ValueError
```
josenavas: This error is not raised here. At the same time, I would argue that the error raised here should be a subtype of QiitaDB. I would prefer to follow the structure that is done in other places, in which we first check that the column actually exists, and if it exists, we retrieve such information. This will also remove the usage of "scrub_data" which I don't think it is safe in terms that we may suffer from SQL injection.

qiita_pet/handlers/api_proxy/sample_template.py
josenavas: @squirrelo this is still missing the documentation

qiita_pet/handlers/study_handlers/sample_template.py
```diff
@@ -57,3 +62,71 @@ def post(self):
         else:
             raise HTTPError(400, 'Unknown sample template action: %s' % action)
         self.write(result)
+
+
+def _build_sample_summary(study_id, user_id):
```
josenavas: I would like to hear from others @antgonza @ElDeveloper 

In my opinion, using pandas here is an overkill. The arguments of using pandas was that it handles the merge of the series for us and it can generate the HTML for us. I think this can be simplified by a simple set of dictionaries and just generate the HTML ourselves.

Also, here we are generating the HTML, then parsing it with xml.etree and then writing back the string. This just looks like patches to fix something that doesn't work in the way that we actually want.

However, these are my ideas and I leave it open to other devs to decide if this is ok as it is.

qiita_pet/static/vendor/js/tableaddcolumn.js
```diff
@@ -0,0 +1,24 @@
+// 2006-08-21 - Created
```
josenavas: This file should be minimized.

qiita_pet/templates/study_ajax/sample_prep_summary.html
```diff
@@ -0,0 +1,29 @@
...
+      .done( function ( data ) {
+        if(data.status == "success"){
+          addColumn("samples-table", meta_val, data.values);
+        } else {
+          bootstrapAlert(data.message.replace("\n", "<br/>"), "danger", 4000);
+        }
+      });
+  }
+
+  function delete_meta() {}
```
josenavas: What is this function doing?

josenavas:
Few more comments.

josenavas:
Also, I just realized, you've been adding a couple of javascript libraries to the repository. Which is the license of those libraries and should be include their license in our repo as we did with other libraries (I think?). cc @ElDeveloper and @antgonza as I think they've more experience with JavaScript libraries usage.

squirrelo:
The licence on the javascript is public domain, so we don't need to include a licence file.

rob-knight:
We should include the statement that its public domain to reduce future
iterations of this conversation, however...

On Jan 22, 2016, at 1:10 PM, Joshua Shorenstein <notifications@github.com>
wrote:

The licence on the javascript is public domain, so we don't need to include
a licence file.

—
<https://github.com/biocore/qiita/pull/1589#issuecomment-174049842>.

squirrelo:
All comments addressed, now waiting for DataFrames discussion.

qiita_pet/handlers/study_handlers/sample_template.py
squirrelo: Remember that HTML tables are built row by row, not column by column, so we would essentially still need a dictionary with sample as key and, at best, the `<td>yes/no</td>` for each prep template. This is effectively what the DataFrame gives us, with optimizations to the column merging for each prep template and HTML generation.

antgonza:
JS license: We like to keep track of the name of the library, the version and actual license the library was released under, for example: https://github.com/biocore/emperor/blob/master/LICENSE.md or the newer version: https://github.com/biocore/emperor/blob/new-api/LICENSE.md

qiita_pet/handlers/study_handlers/sample_template.py
squirrelo: After some testing, this is the fastest way to build the table. It still takes ~1.5 seconds to build and render on my computer, but it beats the minutes the current one takes (`df.to_html` is really slow) and the javascript version, which is fast on the server but takes a long time to render on the page.

qiita_pet/handlers/study_handlers/sample_template.py
```diff
@@ -57,3 +66,80 @@ def post(self):
...
+            # Empty cell for samples not in the prep template
+            for s in all_samps.difference(prep_samples):
+                samps_table[s].append('<td></td>')
+            # X in cell for samples in the prep template
+            for s in all_samps.intersection(prep_samples):
+                samps_table[s].append('<td>X</td>')
+            col_names.append('%s - %d' % (prep['name'], prep['id']))
+
+    # build the table header
+    table = ('<table id="samples-table" class="table table-striped"><thead>'
```
ElDeveloper: I just want to bring up the fact that this is not going to work very well and that rendering a table of this size all at once is excessive, there are other ways of rendering the data, for example using clusterize http://www.plugingeek.com/repos/NeXTs/Clusterize.js

squirrelo:
This has now been updated to use slickgrid, and is super fast. Ready for review again.

qiita_pet/handlers/api_proxy/prep_template.py
```diff
@@ -221,6 +281,9 @@ def prep_template_put_req(prep_id, user_id, prep_template=None,
     access_error = check_access(study_id, user_id)
     if access_error:
         return access_error
+    exists = _check_prep_template_exists(int(prep_id))
```
antgonza: Couldn't find a test for this, is there one?

qiita_pet/handlers/api_proxy/prep_template.py
```diff
@@ -280,6 +343,9 @@ def prep_template_delete_req(prep_id, user_id):
     access_error = check_access(prep.study_id, user_id)
     if access_error:
         return access_error
+    exists = _check_prep_template_exists(int(prep_id))
```
antgonza: Couldn't find a test for this, is there one?

qiita_pet/handlers/api_proxy/prep_template.py
```diff
@@ -315,6 +381,9 @@ def prep_template_filepaths_get_req(prep_id, user_id):
...
```
antgonza: Couldn't find a test for this, is there one?

qiita_pet/handlers/api_proxy/prep_template.py
```diff
@@ -374,6 +443,9 @@ def prep_template_graph_get_req(prep_id, user_id):
...
```
antgonza: Couldn't find a test for this, is there one?

qiita_pet/handlers/api_proxy/sample_template.py
```diff
@@ -23,6 +24,167 @@
...
+        {'status': str,
+         'message': str,
+         'samples': list of str}
+         samples is list of samples in the template
+    """
+    access_error = check_access(samp_id, user_id)
+    if access_error:
+        return access_error
+    exists = _check_sample_template_exists(int(samp_id))
+    if exists['status'] != 'success':
```
antgonza: This if is missing a test.

qiita_pet/handlers/api_proxy/sample_template.py
```diff
@@ -23,6 +24,167 @@
...
+        return exists
+    return {'status': 'success',
+            'message': '',
+            'samples': sorted(x for x in SampleTemplate(int(samp_id)))
+            }
+
+
+def sample_template_meta_cats_get_req(samp_id, user_id):
```
antgonza: Couldn't find a test for this, is there one?

qiita_pet/handlers/api_proxy/sample_template.py
```diff
@@ -23,6 +24,167 @@
...
+    if access_error:
+        return access_error
+    exists = _check_sample_template_exists(int(samp_id))
+    if exists['status'] != 'success':
+        return exists
+
+    st = SampleTemplate(int(samp_id))
+    try:
+        values = st.get_category(category)
+    except QiitaDBColumnError:
```
antgonza: This except doesn't have a test.

qiita_pet/handlers/api_proxy/sample_template.py
```diff
@@ -23,6 +24,167 @@
...
+        Returns information in the form
+        {'status': str,
+         'message': str,
+         'values': dict of {str: object}}
+    """
+    access_error = check_access(samp_id, user_id)
+    if access_error:
+        return access_error
+    exists = _check_sample_template_exists(int(samp_id))
+    if exists['status'] != 'success':
```
antgonza: This if doesn't have a test.

qiita_pet/handlers/api_proxy/sample_template.py
```diff
@@ -51,6 +213,9 @@ def sample_template_summary_get_req(samp_id, user_id):
     access_error = check_access(samp_id, user_id)
     if access_error:
         return access_error
+    exists = _check_sample_template_exists(int(samp_id))
+    if exists['status'] != 'success':
```
antgonza: I think this if doesn't have a test.

qiita_pet/handlers/api_proxy/sample_template.py
```diff
@@ -175,6 +335,9 @@ def sample_template_put_req(study_id, user_id, sample_template):
     access_error = check_access(int(study_id), user_id)
     if access_error:
         return access_error
+    exists = _check_sample_template_exists(int(study_id))
```
antgonza: As far as I can tell there is no tests for this if.

qiita_pet/handlers/api_proxy/sample_template.py
```diff
@@ -229,6 +392,9 @@ def sample_template_delete_req(study_id, user_id):
...
```
antgonza: As far as I can tell there is no tests for this if.

qiita_pet/handlers/api_proxy/sample_template.py
```diff
@@ -262,6 +428,9 @@ def sample_template_filepaths_get_req(study_id, user_id):
     access_error = check_access(study_id, user_id)
     if access_error:
         return access_error
+    exists = _check_sample_template_exists(int(study_id))
```
antgonza: As far as I can tell there is no tests for this if.

qiita_pet/handlers/api_proxy/studies.py
```diff
@@ -113,16 +113,15 @@ def study_prep_get_req(study_id, user_id):
     for dtype in study.data_types:
         prep_info[dtype] = []
         for prep in study.prep_templates(dtype):
-            start_artifact = prep.artifact
-            youngest_artifact = prep.artifact.youngest_artifact
+            # start_artifact = prep.artifact
```
antgonza: Are these lines required?

qiita_pet/handlers/study_handlers/prep_template.py
```diff
@@ -41,7 +41,7 @@ def get(self):
         download = prep_template_filepaths_get_req(
             study_id, self.current_user.id)['filepaths'][-1]
         dl_path = download_link_or_path(
-            is_local, download[0], download[1], "Download prep information")
+            is_local, download[1], download[0], "Download prep information")
```
antgonza: Out of curiosity, why this change?

qiita_pet/handlers/study_handlers/sample_template.py
antgonza: Thanks for checking!

qiita_pet/static/vendor/css/slick.grid.css
```diff
@@ -0,0 +1,158 @@
+/*
```
antgonza: Minimize this file.

qiita_pet/static/vendor/js/jquery.event.drag-2.2.js
```diff
@@ -0,0 +1,402 @@
+/*!
```
antgonza: Minimize this file.

qiita_pet/static/vendor/js/slick.core.js
```diff
@@ -0,0 +1,484 @@
+/***
```
antgonza: Minimize this file.

antgonza:
A few comments. Note that most of them are due to missing tests :cry:. Also, could you minimize qiita_pet/static/vendor/js/slick.grid.js, couldn't add a comment inline.

qiita_pet/handlers/api_proxy/studies.py
squirrelo: ah, that was testing that got through. Fixed.

qiita_pet/handlers/study_handlers/prep_template.py
squirrelo: It was backwards.

squirrelo:
Comments addressed

josenavas:
:+1: @antgonza wanna do the final pass/merge?

PR #1590, opened 2015-12-29
Delete study functionality

squirrelo:
Depends on #1589 

This adds the deletion of studies to the new layout.

PR #1593, opened 2015-12-30
Admin actions

squirrelo:
Requires #1590 

Adds the admin approvals buttons to the interface. Can now change the visibility of all the artifacts individually. Also adds the submit to EBI and VAMPS buttons when applicable.


antgonza:
@squirrelo which should we review next, this one or https://github.com/biocore/qiita/pull/1602 ? Thanks!

squirrelo:
This one.

antgonza:
Thanks, please let us know when should we proceed.

squirrelo:
ready for review.

qiita_pet/handlers/api_proxy/artifact.py
```diff
@@ -53,3 +57,126 @@ def artifact_graph_get_req(artifact_id, direction, user_id):
...
+    -------
+    dict
+        information about the artifact
+    """
+    pd = Artifact(int(artifact_id))
+    access_error = check_access(pd.study.id, user_id)
+    if access_error:
+        return access_error
+
+    try:
```
antgonza: These 3 try/except don't make sense, can you change? Or perhaps I'm missing something?

antgonza: BTW they are also missing tests.

qiita_pet/handlers/api_proxy/artifact.py
```diff
@@ -53,3 +57,126 @@ def artifact_graph_get_req(artifact_id, direction, user_id):
...
+        can_be_submitted_to_vamps = pd.can_be_submitted_to_vamps
+    except QiitaDBOperationNotPermittedError:
+        can_be_submitted_to_vamps = False
+
+    try:
+        is_submitted_to_vamps = pd.is_submitted_to_vamps
+    except QiitaDBOperationNotPermittedError:
+        is_submitted_to_vamps = False
+
+    try:
```
antgonza: I guess this is only valid if can_be_submitted_to_ebi is True so to avoid the try/except you could simply test that first.

qiita_pet/handlers/api_proxy/artifact.py
```diff
@@ -53,3 +57,126 @@ def artifact_graph_get_req(artifact_id, direction, user_id):
...
+        'can_be_submitted_to_vamps': can_be_submitted_to_vamps,
+        'is_submitted_to_vamps': is_submitted_to_vamps,
+        'filepaths': pd.filepaths,
+        'parents': [a.id for a in pd.parents],
+        'prep_templates': [p.id for p in pd.prep_templates],
+        'ebi_run_accessions': ebi_run_accessions,
+        'study': pd.study.id}
+
+
+def artifact_delete_req(artifact_id, user_id):
```
antgonza: Can you add a test for this?

antgonza: Can you add doc strings?

qiita_pet/handlers/api_proxy/artifact.py
```diff
@@ -53,3 +57,126 @@ def artifact_graph_get_req(artifact_id, direction, user_id):
...
+        What to change the visibility to
+
+    Returns
+    -------
+    dict
+        Status of action, in the form {'status': status, 'message': msg}
+        status: status of the action, either success or error
+        message: Human readable message for status
+    """
+    if visibility not in {'sandbox', 'awaiting_approval', 'private', 'public'}:
```
antgonza: This if is not tested.

qiita_pet/handlers/api_proxy/artifact.py
squirrelo: Looks like there were more merge issues then I though, since they were both there before. Let me dig on those.

qiita_pet/handlers/study_handlers/artifact.py
```diff
@@ -19,5 +21,73 @@ class ArtifactGraphAJAX(BaseHandler):
     def get(self):
         direction = self.get_argument('direction')
         artifact = to_int(self.get_argument('artifact_id'))
-        self.write(artifact_graph_get_req(artifact, direction,
-                                          self.current_user.id))
+        self.write(self.artifact_graph_proxy(artifact, direction))
+
+
+class ArtifactAJAX(BaseHandler):
```
antgonza: Can you add tests to this and the next class?

squirrelo:
Merge issues taken care of, so missing tests and docstrings should be in place now.

qiita_pet/handlers/api_proxy/artifact.py
squirrelo: condensed

qiita_pet/handlers/api_proxy/artifact.py
```diff
@@ -53,3 +57,136 @@ def artifact_graph_get_req(artifact_id, direction, user_id):
...
+    """
+    pd = Artifact(int(artifact_id))
+    access_error = check_access(pd.study.id, user_id)
+    if access_error:
+        return access_error
+
+    try:
+        can_be_submitted_to_ebi = pd.can_be_submitted_to_ebi
+        ebi_run_accessions = pd.ebi_run_accessions
+    except QiitaDBOperationNotPermittedError:
```
antgonza: Can you add tests for the except?

qiita_pet/handlers/api_proxy/artifact.py
```diff
@@ -53,3 +57,136 @@ def artifact_graph_get_req(artifact_id, direction, user_id):
...
+        can_be_submitted_to_ebi = False
+        ebi_run_accessions = None
+
+    try:
```
antgonza: Can you add tests for the except?

qiita_pet/handlers/api_proxy/artifact.py
```diff
@@ -53,3 +57,136 @@ def artifact_graph_get_req(artifact_id, direction, user_id):
...
+        status: status of the action, either success or error
+        message: Human readable message for status
+    """
+    pd = Artifact(int(artifact_id))
+    access_error = check_access(pd.study.id, user_id)
+    if access_error:
+        return access_error
+    try:
+        Artifact.delete(int(artifact_id))
+    except QiitaDBArtifactDeletionError as e:
```
antgonza: Can you add a test for the except?

antgonza:
A few more comments.

qiita_pet/handlers/api_proxy/artifact.py
squirrelo: Already in.

qiita_pet/handlers/api_proxy/artifact.py
squirrelo: Already in

qiita_pet/handlers/api_proxy/artifact.py
squirrelo: Done

qiita_pet/handlers/api_proxy/artifact.py
josenavas: `can_be_submitted_to_ebi` is a boolean that tells you if you can query the next attribute or not. Please, use an if statement rather than an exception. try/except should be used when something unexpected may happen, not as a substitute of an if:
```python
can_be_submitted_to_ebi = pd.can_be_submitted_to_ebi
ebi_run_accessions = pd.ebi_run_accessions if can_be_submitted_to_ebi else None
```

qiita_pet/handlers/api_proxy/artifact.py
josenavas: Same comment here.

qiita_pet/handlers/api_proxy/artifact.py
josenavas: Do not hardcode this values. Please, extract from DB if you want to check here.

qiita_pet/handlers/api_proxy/artifact.py
```diff
@@ -53,3 +57,136 @@ def artifact_graph_get_req(artifact_id, direction, user_id):
...
+        return {'status': 'error',
+                'message': 'Unknown visiblity value: %s' % visibility}
+
+    pd = Artifact(int(artifact_id))
+    access_error = check_access(pd.study.id, user_id)
+    if access_error:
+        return access_error
+    user = User(str(user_id))
+    # Set the approval to private if needs approval and admin
+    if all([qiita_config.require_approval, visibility == 'private',
```
josenavas: The first three if statements have common checks. Please extract common checks and use nested ifs to avoid evaluating multiple times the same expression.

qiita_pet/handlers/api_proxy/artifact.py
josenavas: Also I don't see tests in which can_be_submitted_to_XXX gets true/false (only false are tested).

qiita_pet/handlers/api_proxy/tests/test_artifact.py
```diff
@@ -23,25 +115,24 @@ def test_artifact_graph_get_req_descendants(self):
...
-        obs = artifact_graph_get_req(1, 'UNKNOWN', 'test@foo.bar')
-        exp = {'status': 'error',
-               'message': 'Unknown directon UNKNOWN'}
+                               (4, 'BIOM - BIOM')],
+               'edge_list': [(1, 3), (1, 2), (2, 4)]}
         self.assertEqual(obs, exp)
 
     def test_artifact_graph_get_req_no_access(self):
-        obs = artifact_graph_get_req(1, 'descendants', 'demo@microbio.me')
+        obs = artifact_graph_get_req(1, 'ancestors', 'demo@microbio.me')
```
josenavas: Why this change?

qiita_pet/handlers/api_proxy/tests/test_sample_template.py
```diff
@@ -21,34 +20,34 @@
...
-        info)
+        self.new_study = qdb.study.Study.create(
+            qdb.user.User('test@foo.bar'), "Some New Study", [1],
+            info)
 
-    def setUp(self):
-        fp = join(qiita_config.base_data_dir, 'uploads',
-                  str(self.new_study.id), 'uploaded_file.txt')
+        base_dir = qdb.util.get_mountpoint('uploads')[0][1]
+        fp = join(base_dir, str(self.new_study.id), 'uploaded_file.txt')
```
josenavas: I think it will simplify the handling of the tests if you don't use this test (note code duplication here and in tearDown).
My recommendation would be that you actually create a new file using mkstemp (note that you can indicate the directory in which you want it) and just remove it in the tearDown.

qiita_pet/handlers/study_handlers/artifact.py
```diff
@@ -21,3 +25,72 @@ def get(self):
         artifact = to_int(self.get_argument('artifact_id'))
         self.write(artifact_graph_get_req(artifact, direction,
                                           self.current_user.id))
+
+
+class ArtifactAJAX(BaseHandler):
+    def get(self):
+        artifact_id = to_int(self.get_argument('artifact_id'))
+        self.write('TITLE FOR ARTIFACT %d' % artifact_id)
```
josenavas: I don't understand this - why is this hardcoded to 'TITLE FOR ARTIFACT %s'?

qiita_pet/handlers/study_handlers/artifact.py
```diff
@@ -21,3 +25,72 @@ def get(self):
...
+    def get(self):
+        artifact_id = to_int(self.get_argument('artifact_id'))
+        info = artifact_get_req(artifact_id, self.current_user.id)
+        status = info['visibility']
+        buttons = []
+
+        btn_base = ('<button onclick="set_admin_visibility(\'%s\', {0})" '
+                    'class="btn btn-primary">%s</button>').format(artifact_id)
+
+        if all([status == 'sandbox', qiita_config.require_approval]):
```
josenavas: Same comment as above regarding common expression evaluation.

qiita_pet/handlers/study_handlers/artifact.py
```diff
@@ -21,3 +25,72 @@ def get(self):
...
+        artifact_id = to_int(self.get_argument('artifact_id'))
+        self.write('TITLE FOR ARTIFACT %d' % artifact_id)
+
+    def post(self):
+        artifact_id = to_int(self.get_argument('artifact_id'))
+        self.write(artifact_delete_req(artifact_id, self.current_user.id))
+
+
+class ArtifactAdminAJAX(BaseHandler):
+    def get(self):
```
josenavas: How do others feel about creating all those HTML buttons in here?
I think it is better the previous approach that was used. The buttons are already defined in the interface and we just pass booleans to the template to decide if a buttons is shown or not. The approach taken here, splits the definition of the template between the HTML and the handler which, in my opinion, makes it harder to maintain.

qiita_pet/handlers/study_handlers/metadata_summary_handlers.py
```diff
@@ -1,95 +0,0 @@
-# -----------------------------------------------------------------------------
```
josenavas: Why removing this file here? Seems unrelated with the rest of the PR.

qiita_pet/handlers/study_handlers/artifact.py
antgonza: That's a good point and that's the idea of templates. However, there might be a reason to leave them here, @squirrelo?

qiita_pet/templates/study_base.html
```diff
@@ -3,6 +3,7 @@
 <link rel="stylesheet" src="/static/vendor/css/vis.css" type="text/css">
 <script type="text/javascript" src="/static/vendor/js/vis.js"></script>
 <script type="text/javascript">
+  var curr_prep = -1;
```
josenavas: Ugh... global variables...

qiita_pet/handlers/api_proxy/artifact.py
squirrelo: Done

qiita_pet/templates/study_base.html
```diff
@@ -15,12 +16,26 @@
 
   function fill_artifact(aid) {
     $("#study-main").html("ARTIFACT " + aid + " HTML STUFF");
+    $.get("/admin/artifact/", {artifact_id: +aid })
```
josenavas: Remove the '+'

qiita_pet/templates/study_base.html
```diff
@@ -15,12 +16,26 @@
...
+      .done(function ( data ) {
+        $("#study-admin").html(data);
+      });
+  }
+
+  function delete_artifact(aid, pid) {
+    $.post("/artifact/", {artifact_id: +aid })
```
josenavas: Same as above

qiita_pet/templates/study_base.html
josenavas: If you are using global variables, please add a comment on what are they doing and why they're needed. Easier to maintain the code

qiita_pet/templates/study_base.html
```diff
@@ -121,6 +142,22 @@
     }
   }
 
+  function set_admin_visibility(visibility, aid) {
+    $.post('/admin/artifact/', {artifact_id: +aid, visibility: visibility})
```
josenavas: same as above re '+'

qiita_pet/templates/study_base.html
```diff
@@ -121,6 +142,22 @@
...
+      .done(function ( data ) {
+        if(data["status"] == "error") {
+          bootstrapAlert(data.message.replace("\n", "<br/>"), "danger", 4000);
```
josenavas: Code duplication and magic number 4000. Explain number and put "danger"/"info" in a variable and use a single call to bootstrapAlert.

qiita_pet/handlers/api_proxy/artifact.py
squirrelo: done

qiita_pet/templates/study_base.html
squirrelo: This is how you explicitly convert to int in Javascript. It's needed.

qiita_pet/templates/study_base.html
squirrelo: Same answer

qiita_pet/templates/study_base.html
squirrelo: Same answer

josenavas:
done with the review.

qiita_pet/templates/study_base.html
antgonza: Does it have to be converted?

qiita_pet/handlers/api_proxy/artifact.py
squirrelo: Although I don't see why that list cant be hard coded, as the rest is hard coded below. It makes sense to have it hardcoded here because of that.

qiita_pet/templates/study_base.html
josenavas: I thought you were taking care of at the python level...

qiita_pet/handlers/api_proxy/tests/test_sample_template.py
squirrelo: There is no code duplication. One creates a file for the new study upload folder, the other replaces the file for the existing study.

qiita_pet/handlers/api_proxy/artifact.py
josenavas: AFAIK, below you're only checking for the 'private' status - no need for all of them hardocded.

qiita_pet/handlers/study_handlers/artifact.py
squirrelo: missed replacement

qiita_pet/handlers/study_handlers/metadata_summary_handlers.py
squirrelo: was causing a bunch of tests to not run

qiita_pet/templates/study_base.html
squirrelo: yes, and this is pure javascript, no python interaction at this point.

qiita_pet/handlers/api_proxy/tests/test_artifact.py
squirrelo: probably something weird in the merge conflict resolution.

qiita_pet/handlers/api_proxy/artifact.py
squirrelo: done

qiita_pet/templates/study_base.html
squirrelo: done

qiita_pet/handlers/study_handlers/artifact.py
squirrelo: Needed for the logic to work

squirrelo:
Comments addressed

qiita_pet/handlers/study_handlers/artifact.py
squirrelo: Since it's an AJAX call it's a bit easier to build the code here and use write.

qiita_db/util.py
```diff
@@ -979,6 +980,19 @@ def get_environmental_packages():
         return qdb.sql_connection.TRN.execute_fetchindex()
 
 
+def get_visibilities():
```
josenavas: Missing tests

qiita_pet/handlers/api_proxy/artifact.py
```diff
@@ -53,3 +57,128 @@ def artifact_graph_get_req(artifact_id, direction, user_id):
...
+                'message': 'Unknown visiblity value: %s' % visibility}
+
+    pd = Artifact(int(artifact_id))
+    access_error = check_access(pd.study.id, user_id)
+    if access_error:
+        return access_error
+    user = User(str(user_id))
+    # Set the approval to private if needs approval and admin
+    if visibility == 'private':
+        if all([qiita_config.require_approval, user.level == 'admin']):
```
josenavas: There is no need to create a list and use `all`. Also, if you rearrange the ifs, you can save some expression checks:
```python
if not qiita_config.require_approval:
    # Set the visibility to private b/c approval is not required
elif user.level == 'admin':
    # Set the visibility to private b/c approval is required and level is admin
else:
    # Error b/c there are no privileges
```

Also, the msg variable and status are set tu success multiple times. Move that code before the main if statement and only change in case of error.

qiita_pet/handlers/api_proxy/tests/test_sample_template.py
josenavas: ok

josenavas: Does it means that the new file created for the new study is not deleted in the tearDown?

qiita_pet/handlers/study_handlers/artifact.py
josenavas: ? See code below, same functionality but reducing number of checks:
```python
if qiita_config.require_approval:
    if status == 'sandbox':
        # Request approval button
    elif status == 'awaiting approval' and self.current_user.level == 'admin':
        # Approve button
elif status == 'private':
    # Make public button
```

qiita_pet/handlers/study_handlers/artifact.py
```diff
@@ -21,3 +25,73 @@ def get(self):
...
+            # The make public button only appears if the status is private
+            buttons.append(btn_base % ('public', 'Make public'))
+
+        # The revert to sandbox button only appears if the processed data is
+        # not sandboxed or public
+        if status not in {'sandbox', 'public'}:
+            buttons.append(btn_base % ('sandbox', 'Revert to sandbox'))
+
+        # Add EBI and VAMPS submission buttons if allowed
+        if all([not info['ebi_run_accessions'],
```
josenavas: Why are you using `all` and creating lists? Please, use the `and` keyword. Here and the next one.

qiita_pet/templates/study_base.html
josenavas: That's incorrect, you're actually also casting in the python code: https://github.com/biocore/qiita/pull/1593/files#diff-ea0af58e24e01961bafc83e4c0124a05R43

josenavas:
Few comments

qiita_pet/templates/study_base.html
squirrelo: The aid passed in is done using pure javascript on the client side before the get call, so yes it is correct. It is to make sure that what comes from the user side is a valid integer before the call.

qiita_pet/handlers/study_handlers/artifact.py
squirrelo: ease of multi-line if statements. They do the same thing.

qiita_pet/templates/study_base.html
josenavas: We are performing 2 checks, once at python level on other at JS level - is it needed? This goes back to the question above from @antgonza. My opinion is that if you are already making sure that at JS level you're passing an integer, you don't need to perform the check in the python code.

squirrelo: everything is still passed as a string to the server, so yes, both checks are needed. Client side to make sure an integer is passed, and server side to make sure it is cast to integer after it is received.

qiita_pet/handlers/study_handlers/artifact.py
josenavas: They act the same, but they don't do the same:
```
In [7]: def f1(a, b):
    return all([a, b])
   ...: 

In [8]: def f2(a, b):
    return a and b
   ...: 

In [9]: %timeit f1(True, True)
1000000 loops, best of 3: 300 ns per loop

In [10]: %timeit f2(True, True)
10000000 loops, best of 3: 124 ns per loop
```

qiita_pet/templates/study_base.html
josenavas: In which case will not an integer be passed here? If this code is only used in our HTML and no-body else can call this function, I don't think it is needed to have this check. Which type of error is returned through the interface in case that this is not an integer? Is this just to check if a programmer messed up?

josenavas:
@squirrelo ETA on addressing the comments here?

squirrelo:
All comments have been addressed

qiita_pet/templates/study_base.html
squirrelo: It's because javascript is very weakly typed, so it's better to make it explicit that an integer must be here.

qiita_pet/handlers/api_proxy/artifact.py
```diff
@@ -53,3 +57,126 @@ def artifact_graph_get_req(artifact_id, direction, user_id):
...
+    """
+    if visibility not in get_visibilities():
+        return {'status': 'error',
+                'message': 'Unknown visiblity value: %s' % visibility}
+
+    pd = Artifact(int(artifact_id))
+    access_error = check_access(pd.study.id, user_id)
+    if access_error:
+        return access_error
+    user = User(str(user_id))
```
josenavas: out of curiosity - why the specific cast to str in here? it looks unnecessary as you've already said before that everything is passed as a string.

qiita_pet/handlers/api_proxy/artifact.py
```diff
@@ -53,3 +57,126 @@ def artifact_graph_get_req(artifact_id, direction, user_id):
...
+        # Set the approval to private if approval not required
+        elif user.level == 'admin':
+            pd.visibility = 'private'
+        # Trying to set approval without admin privileges
+        else:
+            status = 'error'
+            msg = 'User does not have permissions to approve change'
+    else:
+        pd.visibility = visibility
+        status = 'success'
```
josenavas: This is redundant code form lines 165/166 (as I noticed in my previous comment).

Move the setting of `status` and `msg` outside the if in line 164 and remove lines 165 and 166. The output is the same and if we ever need to change the message there is a single line in the code that we need to change.

qiita_pet/handlers/api_proxy/tests/test_sample_template.py
josenavas: @squirrelo ping about this one

josenavas:
Still some unresolved comments here.

qiita_pet/handlers/api_proxy/artifact.py
squirrelo: Was casting explicitly to int for the other objects so I just followed the same pattern and did an explicit cast here as well.

josenavas: Well, the reasoning doesn't apply here but it is ok - no need to change.

qiita_pet/handlers/api_proxy/artifact.py
josenavas: @squirrelo ETA on this? Having your PR in will reduce my developing time b/c I will be able to connect the interface and my work in the same PR.

squirrelo: Done.

PR #1594, opened 2015-12-30
Clean handlers

squirrelo:
Requires #1593 
This removes the unneeded handlers, uimodules, and tests that were for the old study UI pages. Tests should pass on this PR.


squirrelo:
Closed because now out of sync with what it builds on.

PR #1599, opened 2016-01-07
Artifact names and artifact youngest in the lineage

josenavas:
Adds the name property to the artifacts.

This has been build on top of #1587 so review/merge that one first.

### DIFF
--- a/qiita_db/artifact.py
+++ b/qiita_db/artifact.py
@@ -828,6 +828,31 @@ class Artifact(qdb.base.QiitaObject):
                     for c_id in qdb.sql_connection.TRN.execute_fetchflatten()]
 
     @property
+    def youngest_artifact(self):
+        """Returns the youngest artifact of the artifact's lineage
+
+        Returns
+        -------
+        qiita_db.artifact.Artifact
+            The youngest descendant of the artifact's lineage
+        """
+        with qdb.sql_connection.TRN:
+            sql = """SELECT artifact_id
+                     FROM qiita.artifact_descendants(%s)
+                        JOIN qiita.artifact USING (artifact_id)
+                     ORDER BY generated_timestamp DESC
+                     LIMIT 1"""
+            qdb.sql_connection.TRN.add(sql, [self.id])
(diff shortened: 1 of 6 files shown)


josenavas:
Merge conflict solved and now it also fixes #1573 (I've updated the title to reflect this change).

josenavas:
After a discussion with @antgonza I'm going to also fix #1573 here - so hold on review.

josenavas:
Talking again with @antgonza, including #1573 in this PR will make this PR to grow too much, so we decided to finally include it in another PR.

qiita_db/artifact.py
```diff
@@ -255,8 +257,8 @@ def create(cls, filepaths, artifact_type, prep_template=None,
 
                 # Create the artifact
                 sql = """INSERT INTO qiita.artifact
-                            (generated_timestamp, command_id, data_type_id,
-                             command_parameters, visibility_id,
+                            (generated_timestamp, command_id,
```
ElDeveloper: Why was this changed?

qiita_db/artifact.py
```diff
@@ -405,6 +410,46 @@ def delete(cls, artifact_id):
...
+            The new artifact's name
+
+        Raises
+        ------
+        ValueError
+            If `value` contains more than 35 chars
+        """
+        if len(value) > 35:
+            raise ValueError("The name of an artifact cannot exceed 35 chars. "
+                             "Current lengths: %d" % len(value))
```
ElDeveloper: lengths -> length

ElDeveloper:
:+1:

qiita_db/artifact.py
josenavas: For no good reason - reverting.

qiita_db/artifact.py

josenavas:
Thanks @ElDeveloper! comments addressed

qiita_db/artifact.py
ElDeveloper: Thanks!

antgonza:
Merging as the previous commit tests passed and the latest changes will not change that ...

PR #1603, opened 2016-01-15
Fix artifact delete

josenavas:
There is currently a bug in the system in which if there is an artifact with a processing job attached, it can't be deleted.

This PR allows to delete an artifact that has successful/failed processing jobs and raises a meaningful error if the artifact has queued/running jobs.

@antgonza @ElDeveloper @mortonjt a quick review is appreciated, since this bug is in master and in the live system.


antgonza:
Closing in favor of: https://github.com/biocore/qiita/pull/1604

squirrelo:
:+1:

PR #1604, opened 2016-01-15
Fix artifact delete agp

antgonza:
A non error version of https://github.com/biocore/qiita/pull/1603

### DIFF
--- a/qiita_db/artifact.py
+++ b/qiita_db/artifact.py
@@ -377,6 +377,22 @@ class Artifact(qdb.base.QiitaObject):
                 raise qdb.exceptions.QiitaDBArtifactDeletionError(
                     artifact_id, "it has been submitted to VAMPS")
 
+            # Check if there is a job queued that will use the artifact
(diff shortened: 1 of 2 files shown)


qiita_db/artifact.py
```diff
@@ -377,10 +377,43 @@ def delete(cls, artifact_id):
...
+                        SELECT *
+                        FROM qiita.artifact_processing_job
+                            JOIN qiita.processing_job USING (processing_job_id)
+                            JOIN qiita.processing_job_status
+                                USING (processing_job_status_id)
+                        WHERE artifact_id = %s
+                            AND processing_job_status IN ('queued', 'running'))
+                  """
+            qdb.sql_connection.TRN.add(sql, [artifact_id])
+            if qdb.sql_connection.TRN.execute_fetchlast():
```
squirrelo: This case is not tested.

squirrelo: Sorry, misread test.

squirrelo:
:+1:

PR #1605, opened 2016-01-15
Fixing template for prep deletion

josenavas:
A bug in the interface when deleting a prep template that doesn't have any artifact.


antgonza:
This is not gonna modify the result of the tests and we need this patch. Merging.

PR #1606, opened 2016-01-15
fixing bug with stats page

antgonza:
This doesn't have a test as is part of qiita-pet so is fine to merge if you agree with the change.


josenavas:
:+1:

PR #1612, opened 2016-01-20
Fixes #1611

### LINKED ISSUE #1611: Should any user with access to a study be able to share the study?
Currently, the only one who can share the study is the owner of the study. Do we want to keep it like that or whoever who has access to the study can share the study? As an example, Bob owns a study and she shares it with Alice. Now Alice want to give me access to such study. Should Alice be able to give access or should Alice contact Bob so he can share the study with me?

josenavas:
Now every user with access to the study can share the study.

That handler was missing tests, so I've added them.

This issue is currently in the deployed system, @antgonza @ElDeveloper @mortonjt any of you available for review?


antgonza:
Looks good but what about a test for when the user has no access?

ElDeveloper:
:+1:

mortonjt:
I agree with @antgonza 's comment.  Otherwise it looks good

josenavas:
Just added a test in which the user has no access to the study.

antgonza:
Thanks, looks good. I'm gonna wait for the test to finish.

PR #1618, opened 2016-01-26
Add artifact ui

squirrelo:
This adds the page to associate a prep template and files with the study. It creates the prep template object and initial artifact to attach files to that object.

This is still waiting on the PR for programmatically getting the file_types associated with each of the artifact types.


qiita_db/artifact.py
```diff
@@ -82,6 +82,23 @@ def iter_public(cls):
         """
         return cls.iter_by_visibility('public')
 
+    @staticmethod
+    def types():
+        """Returns list of all artifact types available in the system
```
antgonza: ... and their descriptions?

qiita_pet/handlers/api_proxy/artifact.py
```diff
@@ -5,8 +5,78 @@
...
+            cleaned_filepaths.append((fp, fp_id))
+
+    artifact = Artifact.create(cleaned_filepaths, artifact_type, name=name,
+                               prep_template=prep_template_id)
+    return {'status': 'success',
+            'message': '',
+            'artifact': artifact.id}
+
+
+def artifact_types_get_req():
```
antgonza: Is this a helper function needed by other objects? If so what about `general_reply_get_req(Artifact.types())`?

qiita_pet/handlers/api_proxy/prep_template.py
```diff
@@ -182,9 +184,12 @@ def prep_template_post_req(study_id, user_id, prep_template, data_type,
         # Show the error to the user so he can fix the template
         status = 'error'
         msg = str(e)
-    return {'status': status,
+    info = {'status': status,
             'message': msg,
             'file': prep_template}
+    if prep:
```
antgonza: What will be the value of info['id'] if not prep?

qiita_pet/handlers/api_proxy/artifact.py
```diff
@@ -5,8 +5,78 @@
...
 from qiita_db.artifact import Artifact
-from qiita_pet.handlers.api_proxy.util import check_access
+from qiita_db.metadata_template.prep_template import PrepTemplate
+from qiita_db.util import convert_to_id
+
+
+@execute_as_transaction
+def artifact_post_req(user_id, filepaths, artifact_type, name,
+                      prep_template_id):
+    """Creates the prep template and initial artifact for the prep template
```
josenavas: There will be the option to upload the prep template without the initial artifact? There may be cases where this is useful (i.e. I'm adding the prep template while the sequencing is being done, so I can reduce the time to process once the sequences are available)

josenavas: Actually, reading the code this comment is off. The prep template is already created, so it is creating only the artifact. Please, update the documentation to be sync'd with the code.

qiita_pet/handlers/api_proxy/artifact.py
```diff
@@ -5,8 +5,78 @@
...
+    user_id : str
+        User adding the atrifact
+    filepaths : dict of {str: [str, ...], ...}
+        List of files to attach to the artifact, keyed to the file type
+    }
+    artifact_type : str
+        The type of the artifact
+    name : str
+        Name to give the artifact
+    prep_template_id : int
```
josenavas: Same comment as in a previous PR regarding `int` and `str castable to int`

qiita_pet/handlers/api_proxy/artifact.py
```diff
@@ -5,8 +5,78 @@
...
+    """
+    prep = PrepTemplate(int(prep_template_id))
+    access_error = check_access(prep.study_id, user_id)
+    if access_error:
+        return access_error
+    study_id = PrepTemplate.study_id
+    cleaned_filepaths = []
+    for ftype in filepaths:
+        # Convert filepath type to the database ID for adding artifact
+        fp_id = convert_to_id(ftype, 'filepath_type')
```
josenavas: It looks like the documentation for the artifact creation is a bit off (as a result of the documentation of the function `insert_filepaths` being also off). This line is not needed. If you pass the string, it internally changes it to the id, so it is cleaner here if you don't change to the id here (so the interface code is decouples from qiita_db.

qiita_pet/handlers/api_proxy/artifact.py
josenavas: I don't think that applies here. The caller does not have access to Artifact.types() so the actual goal here is being able to return the result of artifact.types() to the GUI.

qiita_pet/handlers/api_proxy/prep_template.py
josenavas: Suggest doing:

```python
return  {'status': status,
        'message': msg,
        'file': prep_template,
        'id': prep.id if prep else None}
```

qiita_pet/handlers/study_handlers/artifact.py
```diff
@@ -21,3 +24,47 @@ def get(self):
...
+                      if f.endswith(('txt', 'tsv'))]
+        artifact_types = artifact_types_get_req()['types']
+        data_types = sorted(data_types_get_req()['data_types'])
+        ontology = ena_ontology_get_req()
+        self.render("study_ajax/add_prep_artifact.html", prep_files=prep_files,
+                    artifact_types=artifact_types, data_types=data_types,
+                    ontology=ontology, study_id=study_id)
+
+    @authenticated
+    def post(self, study_id):
```
josenavas: I'm unsure on the goal of this post function. Can you explain, in plain words here, what is this post supposed to do?

qiita_pet/handlers/study_handlers/prep_template.py
```diff
@@ -77,3 +79,42 @@ def post(self):
...
+        # TODO: Get file types for the artifact type
+        # FILE TYPE IN POSTION 0 MUST BE DEFAULT FOR SELECTED
+        file_types = ['raw_fwd', 'raw_rev']
+
+        selected = []
+        not_selected = []
+        _, base = get_mountpoint("uploads")[0]
+        uploaded = get_files_from_uploads_folders(study_id)
+        prep = pd.read_table(join(base, study_id, prep_file), sep='\t')
+        if 'run_prefix' in prep.columns:
```
josenavas: the name of the the variables on this piece of code is misleading. If you have run prefix, you don't necessarily has a per sample artifact.

qiita_pet/handlers/study_handlers/prep_template.py
```diff
@@ -77,3 +79,42 @@ def post(self):
...
+            # Use run_prefix column of prep template to auto-select
+            # per-sample uploaded files if available.
+            per_sample = True
+            prep_prefixes = set(prep['run_prefix'])
+            for _, filename in uploaded:
+                if splitext(filename)[0] in prep_prefixes:
```
josenavas: Checking the run_prefix in this way is not correct. Counter examples:
 - `sampleA_FWD.fastq` and `sampleB_REV.fastq`, with run prexix = sampleA <- this is the actual run prefix
 - `run_prefix1.fastq.gz`<- splitext will return `run_prefix1`.

It needs to be done using `startswith` which checks for prefixes.

qiita_pet/handlers/study_handlers/sample_template.py
```diff
@@ -29,13 +29,24 @@ def get(self):
...
-        dl_path = download_link_or_path(
-            is_local, download[1], download[0], "Download sample information")
+            study_id, self.current_user.id)
+        if download['status'] == 'success':
+            download = download['filepaths'][-1]
+            dl_path = download_link_or_path(
+                is_local, download[1], download[0],
+                "Download sample information")
+        else:
+            dl_path = 'No sample template added'
```
josenavas: `...sample template...` -> `...sample information...`

qiita_pet/templates/study_ajax/add_prep_artifact.html
```diff
@@ -0,0 +1,155 @@
...
+
+    $('#form').submit(function( event ){
+    event.preventDefault();
+    get_files();
+    $.ajax({
+      url: $('#form').attr('action'),
+      type: "POST",
+      data : $('#form').serialize(),
+      success: function( data ){
+        bootstrapAlert("<p>Error creating preperation</p>" + data.message, "danger");
```
josenavas: `Error creation preparation` -> `Error adding preparation information`

josenavas:
Few comments added

qiita_pet/handlers/api_proxy/artifact.py
```diff
@@ -5,8 +5,78 @@
...
+from qiita_core.util import execute_as_transaction
+from qiita_pet.handlers.api_proxy.util import check_access, check_fp
 from qiita_db.artifact import Artifact
-from qiita_pet.handlers.api_proxy.util import check_access
+from qiita_db.metadata_template.prep_template import PrepTemplate
+from qiita_db.util import convert_to_id
+
+
+@execute_as_transaction
+def artifact_post_req(user_id, filepaths, artifact_type, name,
```
antgonza: Are there tests for this?

antgonza:
A few comments.

qiita_pet/handlers/api_proxy/artifact.py
squirrelo: Correct.

qiita_pet/handlers/api_proxy/prep_template.py
squirrelo: Done

qiita_pet/handlers/api_proxy/artifact.py
squirrelo: Done

qiita_pet/handlers/api_proxy/artifact.py
squirrelo: Done

qiita_pet/handlers/api_proxy/artifact.py
squirrelo: Ok, done. Need that documentation updated to be in sync with code.

qiita_pet/handlers/study_handlers/artifact.py
squirrelo: It takes in the information for the prep template and artifact files, then creates the prep template and creates+attaches the initial artifact to the prep template.

qiita_pet/handlers/study_handlers/prep_template.py
squirrelo: @ackermag explained that in the old qiimedb the run_prefix column was used to select files. if it's all the same here, you could theoretically auto-select the single file for the run as well.

qiita_pet/handlers/study_handlers/prep_template.py
squirrelo: The issue there is then all the FWD and REV would be put into the forward seqs or whatever the default bin is. This default bin problem will need to be discussed during the meeting today.

qiita_pet/templates/study_ajax/add_prep_artifact.html
squirrelo: Done

squirrelo:
Removed the WIP since it does all work now. Couple things still need to be discussed however.

josenavas:
@squirrelo can you pull from upstream/artifact-study-pages and fix the merge conflicts? I will review afterwards. Let me know when it is done.

squirrelo:
Done, ready for review.

qiita_pet/handlers/api_proxy/prep_template.py
```diff
@@ -193,13 +194,14 @@ def prep_template_post_req(study_id, user_id, prep_template, data_type,
 
     msg = ''
     status = 'success'
+    prep = False
```
josenavas: Is this supposed to be `prep = None`? otherwise prep is getting different types (bool and PrepTemplate)

qiita_pet/handlers/study_handlers/prep_template.py
josenavas: This comment is still unresolved. The name of the variables here makes the illusion that this code is only used one we have per_sample info which is not the case. Per sample files, is just a special case when we have the column run_prefix. That will make code maintainability harder.

Use a name such `per_prefix`.

qiita_pet/handlers/study_handlers/prep_template.py
```diff
@@ -77,3 +80,43 @@ def post(self):
...
+                    else:
+                        not_selected.append(filename)
+        else:
+            per_sample = False
+            not_selected = [f for _, f in uploaded]
+
+        # Write out if this prep template supports per-sample files, and the
+        # as well as pre-selected and remaining files
+        self.write({
+            'per_sample': per_sample,
```
josenavas: The same applies here to be passed through JSON (which means changing the template)

qiita_pet/handlers/study_handlers/sample_template.py
```diff
@@ -33,12 +33,22 @@ def get(self):
...
+
+        if download['status'] == 'success':
             download = download['filepaths'][-1]
+            dl_path = download_link_or_path(
+                is_local, download[1], download[0],
+                "Download sample information")
+        else:
+            dl_path = 'No sample information added'
+
+        stats = sample_template_summary_get_req(study_id, self.current_user.id)
```
josenavas: I think you need to remove the code in lines 53 and 54? Given that they're overwriting what you are doing in here.
That makes me wonder how this code has been tested.

qiita_pet/templates/study_ajax/process_study.html
```diff
@@ -0,0 +1,13 @@
+<script type="text/javascript">
```
josenavas: Where is this file used? And, it only includes javascript, why is not in a javascript file rather than in an HTML?

qiita_pet/templates/study_base.html
```diff
@@ -143,6 +151,23 @@
     }
   }
 
+  function delete_study() {
```
josenavas: This function is defined twice

josenavas:
Why is the file `qiita_db/support_files/test_data/uploads/1/uploaded_file.txt` being removed from the repo? I think there are tests that depend on it, so add it back.

Done with the current review.

qiita_pet/handlers/study_handlers/prep_template.py
squirrelo: I can do that, but it will need to be discussed with @ackermag before deployment.

qiita_pet/handlers/study_handlers/prep_template.py
squirrelo: I don't understand what you mean here.

qiita_pet/handlers/study_handlers/sample_template.py
squirrelo: This code is still in because of a merge error. Lines 53 and 54 are correct, hence why it still renders properly and tests pass.

qiita_pet/templates/study_ajax/process_study.html
squirrelo: another merge error, removed

qiita_pet/templates/study_base.html
squirrelo: another auto-merge error, removed

qiita_pet/handlers/api_proxy/prep_template.py
squirrelo: It's false here for line 220 if statement

qiita_pet/handlers/study_handlers/prep_template.py
josenavas: Why? The functionality is the same, it does work, but the problem is that the name of the variable is misleading.

run_prefix is not useful only when there are per sample files, is also useful when in a single prep template there are more than one lane. In such case, the "run_prefix" column helps to differentiate which set of samples belong to the given lane.

That is way I'm suggesting that change here, because as the code currently is makes the illusion that is only used when per_sample files are used which is not correct.

qiita_pet/handlers/study_handlers/prep_template.py
josenavas: regarding the name of the variable, and the name of the keyword in the output dictionary.

squirrelo: I still don't understand what you mean. Its a boolean variable that says whether it is expecting per-sample files or not, so why would it need to change?

qiita_pet/handlers/study_handlers/sample_template.py
josenavas: the thing is that this code looks better than the one below, and it looks like this code has been added here (since it is not complaining about lines 53 and 54)

qiita_pet/handlers/api_proxy/prep_template.py
josenavas: Please, define prep as None as it is done in common python practices and across the code base. This way there is no doubt if prep is a boolean or a PrepTemplate object. Then change the if to, `prep.id if prep is not None else None

qiita_pet/handlers/study_handlers/prep_template.py
josenavas: the problem is that having the column "run_prefix" in the mapping file does't mean that the data is per_sample, hence the misleading sample name

squirrelo: With the change above to `per_prefix` that you wanted, it now does work that way.

squirrelo:
Comments addressed

qiita_pet/handlers/study_handlers/prep_template.py
```diff
@@ -77,3 +80,43 @@ def post(self):
...
+        # TODO: Get file types for the artifact type
+        # FILE TYPE IN POSTION 0 MUST BE DEFAULT FOR SELECTED
+        file_types = supported_filepath_types(prep_type)
+
+        selected = []
+        not_selected = []
+        _, base = get_mountpoint("uploads")[0]
+        uploaded = get_files_from_uploads_folders(study_id)
+        prep = pd.read_table(join(base, study_id, prep_file), sep='\t')
+        if 'per_prefix' in prep.columns:
```
josenavas: @squirrelo the name of the column was correct "run_prefix"
What you needed to change is the name of the variable below, `per_sample` -> `per_prefix`

Please make that change.

qiita_pet/handlers/study_handlers/prep_template.py
josenavas: And this one

qiita_pet/templates/study_base.html
josenavas: Did you modify this function in some way? or it just says that has been moved due to the merge conflict?

Please be careful solving merge conflicts, otherwise is really hard to review.

squirrelo: No changes, just moved.

qiita_pet/handlers/study_handlers/prep_template.py
squirrelo: Change made. Please comment on the line you have issue with in the future. Commenting on this if statement and talking about changing run_prefix means you have an issue with the column name.

qiita_pet/handlers/api_proxy/artifact.py
```diff
@@ -5,12 +5,84 @@
...
+                      prep_template_id):
+    """Creates the initial artifact for the prep template
+
+    Parameters
+    ----------
+    user_id : str
+        User adding the atrifact
+    filepaths : dict of {str: [str, ...], ...}
+        List of files to attach to the artifact, keyed to the file type
+    }
```
josenavas: Probably another merge conflict issue? I think this '}' doesn't belong here

qiita_pet/handlers/study_handlers/prep_template.py
```diff
@@ -77,3 +80,43 @@ def post(self):
...
+        file_types = supported_filepath_types(prep_type)
+
+        selected = []
+        not_selected = []
+        _, base = get_mountpoint("uploads")[0]
+        uploaded = get_files_from_uploads_folders(study_id)
+        prep = pd.read_table(join(base, study_id, prep_file), sep='\t')
+        if 'run_prefix' in prep.columns:
+            # Use run_prefix column of prep template to auto-select
+            # per-sample uploaded files if available.
```
josenavas: Please update comments accordingly

qiita_pet/handlers/study_handlers/prep_template.py
```diff
@@ -77,3 +80,43 @@ def post(self):
...
+                for prefix in prep_prefixes:
+                    if filename.startswith(prefix):
+                        selected.append(filename)
+                    else:
+                        not_selected.append(filename)
+        else:
+            per_prefix = False
+            not_selected = [f for _, f in uploaded]
+
+        # Write out if this prep template supports per-sample files, and the
```
josenavas: Same here

josenavas:
More comments.

squirrelo:
All addressed

PR #1622, opened 2016-01-27
artifact_type <-> filepath_type relationship

josenavas:
Modifies the DB so each artifact_type is linked with the filepath_type that can be added to the given artifact_type. Also adds a util function to retrieve such information.

This is needed by the GUI in order to reduce the options provided to the user when uploading data. By storing them in the DB we avoid the necessity to hardcode them in the GUI, as it was before.

@squirrelo @ElDeveloper @mortonjt available to review?

### DIFF
--- a/qiita_db/support_files/patches/37.sql
+++ b/qiita_db/support_files/patches/37.sql
@@ -1,0 +1,30 @@
+-- Jan 26, 2016
+-- Relate the artifact types with the filepath types that they support
+
+CREATE TABLE qiita.artifact_type_filepath_type (
+	artifact_type_id     bigint  NOT NULL,
+	filepath_type_id     bigint  NOT NULL,
+	required             bool DEFAULT 'TRUE' NOT NULL,
+	CONSTRAINT idx_artifact_type_filepath_type PRIMARY KEY ( artifact_type_id, filepath_type_id )
+ ) ;
+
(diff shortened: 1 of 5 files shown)


antgonza:
The code seems fine but my concern is what happens when we allow a new file type. For example, what are the steps to allow fna/qual as FASTA and tsv as mapping files in the system

josenavas:
Note that this is not looking at the file themselves, so the extensions do not apply.
This only affects artifacts, so adding tsv as mapping file is not affected here (i.e. it still needs to be done through the config file).

Now, the important part :smile: 
There is currently no "easy" way of changing those relationships. But as an extension I would also note that there is no easy way of adding new filepath_types or artifact_types. So far we have been adding them in patches when the processing pipeline supported those new types.

Working with the target_gene_plugin was easy because the system already "depends" on it. However, this also has some drawbacks. For example, there is no way of "installing" a plugin in the system ATM. All this information (new filepath_types, artifact_types, and filepath_types <-> artifact_types relationships) should be instantiated at plugin installation time. So I would add the mechanisms to add these relationships at that point, but that is not critical right now, IMO.

antgonza:
Makes sense, do we have issues open for these missing functionalities?

josenavas:
Now we do :smile: https://github.com/biocore/qiita/issues/1625

antgonza:
@squirrelo could you review?

squirrelo:
:+1:

PR #1624, opened 2016-01-27
URGENT: Fixes #1621

### LINKED ISSUE #1621: Studies awaiting approval page does not show user email
It instead shows the user object:
screen shot 2016-01-26 at 1 31 46 pm

josenavas:
I'm unsure on how to test this part programmatically, since the page was rendering.

It would be good if we also have this PR merged by tomorrow's meeting. It shouldn't take more than 5 seconds to review, any takers?


antgonza:
:+1: What about adding a test that the email has an @?

josenavas:
Actually, I may be able to test if the response itself contains the email of the user. Thanks for the tip @antgonza , will add now.

josenavas:
Test added - should be ready!

Issue #1616, opened 2016-01-22
Portals admin page shows 500


ElDeveloper:
Seems like the /admin/portals/studies page is showing a 500.

![screen shot 2016-01-21 at 10 28 52 am](https://cloud.githubusercontent.com/assets/375307/12522843/ea8c65dc-c107-11e5-8b69-118e3c67d847.png)

antgonza:
fixed in #1631

PR #1631, opened 2016-02-01
fixing error with portals

antgonza:
An error was identified when looking at the portals list. The error is simple to solve but no idea how to test it as the error was due to the changes we made (now returning objects vs. ids). BTW I think we had a discussion about the get_info method before and agreed that it was fine to just work with ids.

### DIFF
--- a/qiita_pet/test/test_portal.py
+++ b/qiita_pet/test/test_portal.py
@@ -13,41 +13,39 @@ class TestPortal(TestHandlerBase):
         response = self.get('/admin/portals/studies/')
         self.assertEqual(response.code, 200)
 
-    def test_post(self):
+    def test_post_add(self):
         BaseHandler.get_current_user = Mock(return_value=User("admin@foo.bar"))
         response = self.post('/admin/portals/studies/', {'portal': 'EMP',
                                                          'selected': [1],
                                                          'action': 'Add'})
         self.assertEqual(response.code, 200)
 
+    def test_post_remove(self):
+        BaseHandler.get_current_user = Mock(return_value=User("admin@foo.bar"))
         response = self.post('/admin/portals/studies/', {'portal': 'EMP',
                                                          'selected': [1],
                                                          'action': 'Remove'})
         self.assertEqual(response.code, 200)
 
-    def test_get_errors(self):
-        # not valid user
+    def test_get_not_valid_user(self):
         response = self.get('/admin/portals/studies/')
         self.assertEqual(response.code, 403)
 
-    def test_post_errors(self):
-        # not valid user
+    def test_post_not_valid_user(self):
         response = self.post('/admin/portals/studies/', {'portal': 'EMP',
                                                          'selected': [1],
                                                          'action': 'Add'})
         self.assertEqual(response.code, 403)
 
-        # making an admin the valid user so the next tests actually test
-        # what they should
+    def test_post_not_valid_portal(self):
         BaseHandler.get_current_user = Mock(return_value=User("admin@foo.bar"))
-
-        # not valid portal
         response = self.post('/admin/portals/studies/', {'portal': 'not-valid',
                                                          'selected': [1],
                                                          'action': 'Add'})
         self.assertEqual(response.code, 400)
 
-        # not a valid action
+    def test_post_not_valid_action(self):
+        BaseHandler.get_current_user = Mock(return_value=User("admin@foo.bar"))
         response = self.post('/admin/portals/studies/', {'portal': 'EMP',
                                                          'selected': [1],
                                                          'action': 'Error'})
--- a/qiita_pet/test/test_portal.py
+++ b/qiita_pet/test/test_portal.py
@@ -12,7 +12,7 @@ class TestPortal(TestHandlerBase):
         BaseHandler.get_current_user = Mock(return_value=User("admin@foo.bar"))
         response = self.get('/admin/portals/studies/')
         self.assertEqual(response.code, 200)
-        self.assertNotEqual(response.code, "")
+        self.assertNotEqual(response.body, "")
 
     def test_post_add(self):
         BaseHandler.get_current_user = Mock(return_value=User("admin@foo.bar"))
@@ -20,7 +20,7 @@ class TestPortal(TestHandlerBase):
                                                          'selected': [1],
                                                          'action': 'Add'})
         self.assertEqual(response.code, 200)
-        self.assertNotEqual(response.code, "")
+        self.assertNotEqual(response.body, "")
 
     def test_post_remove(self):
         BaseHandler.get_current_user = Mock(return_value=User("admin@foo.bar"))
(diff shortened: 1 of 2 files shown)


squirrelo:
You can test by using mock to log in as admin@foo.bar, then doing the standard code and body tests like [here](https://github.com/biocore/qiita/blob/master/qiita_pet/test/test_study_handlers.py#L211)

antgonza:
Thanks. It turns out that portals had no tests in qiita_pet. Added all (AFAIK). Ready for review.

qiita_pet/test/test_portal.py
```diff
@@ -0,0 +1,58 @@
...
+
+from qiita_db.user import User
+from qiita_pet.handlers.base_handlers import BaseHandler
+
+
+class TestPortal(TestHandlerBase):
+    def test_get(self):
+        BaseHandler.get_current_user = Mock(return_value=User("admin@foo.bar"))
+        response = self.get('/admin/portals/studies/')
+        self.assertEqual(response.code, 200)
```
squirrelo: needs test of `response.body`

qiita_pet/test/test_portal.py
```diff
@@ -0,0 +1,58 @@
...
+
+    def test_post(self):
+        BaseHandler.get_current_user = Mock(return_value=User("admin@foo.bar"))
+        response = self.post('/admin/portals/studies/', {'portal': 'EMP',
+                                                         'selected': [1],
+                                                         'action': 'Add'})
+        self.assertEqual(response.code, 200)
+
+        response = self.post('/admin/portals/studies/', {'portal': 'EMP',
```
squirrelo: Break this into two tests.

qiita_pet/test/test_portal.py
```diff
@@ -0,0 +1,58 @@
...
+        BaseHandler.get_current_user = Mock(return_value=User("admin@foo.bar"))
+        response = self.get('/admin/portals/studies/')
+        self.assertEqual(response.code, 200)
+
+    def test_post(self):
+        BaseHandler.get_current_user = Mock(return_value=User("admin@foo.bar"))
+        response = self.post('/admin/portals/studies/', {'portal': 'EMP',
+                                                         'selected': [1],
+                                                         'action': 'Add'})
+        self.assertEqual(response.code, 200)
```
squirrelo: needs test of `response.body`

qiita_pet/test/test_portal.py
```diff
@@ -0,0 +1,58 @@
...
+
+        response = self.post('/admin/portals/studies/', {'portal': 'EMP',
+                                                         'selected': [1],
+                                                         'action': 'Remove'})
+        self.assertEqual(response.code, 200)
```
squirrelo: needs test of `response.body`

squirrelo:
Some comments.

qiita_pet/test/test_portal.py
antgonza: This applies to all response.body comments. What should I tests? That the display is accurate all html (too much IMOO, as any minimal change will make the test fail), that the response has the study id/title, other?

qiita_pet/test/test_portal.py
antgonza: Out of curiosity, why do you think is needed? Note that these are really simple tests and seem like too much to have a test for each case, for example take a look at test_*_errors

squirrelo: We agreed before that we were going to do atomic tests, with one action tested per test function.

qiita_pet/test/test_portal.py
squirrelo: Some line of the HTML that is dynamically generated from data is proper, like in the example I posted. We have had cases of 200 code with an empty html because of rendering issues.

antgonza: "Funny" enough there is nothing dynamic created in this page as it's retrieved via AJAX, and JS is not executed. Everything is created in /admin/portals/studiesAJAX/, for example: /admin/portals/studiesAJAX/?sEcho=1001&view-portal=QIITA ...

josenavas:
@antgonza I think this is fixing https://github.com/biocore/qiita/issues/1616 correct?

antgonza:
Yup, missed that one. Is there one about adding tests to portals? :yum:

qiita_pet/test/test_portal.py
squirrelo: The JSON it returns is dynamic. Please test that it comes out correctly.

josenavas: Agree with @squirrelo . The easiest way is to grab some static text from the template and check it in the response. In this case, it will suffice to do: `self.assertIn("View from portal:", response.body)`

qiita_pet/test/test_portal.py
josenavas: Which you can use the same as above.

qiita_pet/test/test_portal.py
```diff
@@ -0,0 +1,56 @@
...
+                                                         'selected': [1],
+                                                         'action': 'Add'})
+        self.assertEqual(response.code, 200)
+
+    def test_post_remove(self):
+        BaseHandler.get_current_user = Mock(return_value=User("admin@foo.bar"))
+        response = self.post('/admin/portals/studies/', {'portal': 'EMP',
+                                                         'selected': [1],
+                                                         'action': 'Remove'})
+        self.assertEqual(response.code, 200)
```
josenavas: And here.

josenavas:
@antgonza I don't think so, my guess is that we englobed all missing tests in the interface in the general issue.

Just a small comment on the tests.

Thanks!

qiita_pet/test/test_portal.py
antgonza: @squirrelo, you are right, the JSON is dynamic. However, as stated in my previous comment the AJAX call is not executed as it is in JS. Thus, the page never "renders" the study list == no way to test in here. The only option is to add tests to /admin/portals/studiesAJAX/, which is way beyond the issue this PR is solving but I could do it.

antgonza: Just realized that adding a test for the AJAX page is not that difficult. Can do it here.

antgonza: After chatting offline with @josenavas we agreed on adding a test for not empty body when reply is 200. Adding now but I want to discuss this in the next meeting (already in the list) because we should be doing something smarter than adding tests that test that there are specific words in the rendering.

josenavas:
:+1: once tests pass

qiita_pet/test/test_portal.py
```diff
@@ -0,0 +1,73 @@
...
+        self.assertNotEqual(response.code, "")
+
+    def test_post_add(self):
+        BaseHandler.get_current_user = Mock(return_value=User("admin@foo.bar"))
+        response = self.post('/admin/portals/studies/', {'portal': 'EMP',
+                                                         'selected': [1],
+                                                         'action': 'Add'})
+        self.assertEqual(response.code, 200)
+        self.assertNotEqual(response.code, "")
```
squirrelo: Shouldn't this be `response.body`? Same for the other 2.

antgonza: Good catch!

antgonza:
I think this is ready for merge.

PR #1632, opened 2016-02-02
Fix 1610

### LINKED ISSUE #1610: Errors get hidden in step 3 of split_libraries_fastq.py
If step 3 of `split_libraries_fastq` in `qiita_plugins/target_gene/tgp/split_libraries/split_libraries_fastq.py` fails, we don't get the actual error but: ` The split libraries output directory does not contain the demultiplexed fastq file.`, which is accurate cause the previous step failed. However it will be good to know why it failed, for example:
```
skbio.parse.sequences._exception.FastqParseError: Failed qual conversion for seq id: M00365:7:000000000-A44J0:1:1101:15990:1748 1:N:0:0. This may be because you passed an incorrect value for phred_offset.
```

antgonza:
This fixes #1610. Note that there are no tests for this section as they should be added once the plugin is in its own repo. This also means that we don't need to wait for tests.


josenavas:
Thanks @antgonza :+1: 
@ElDeveloper @mortonjt able to review/merge?

PR #1638, opened 2016-02-04
small fix for ebi

antgonza:
The current EBI GUI submission is broken. Obviously, we didn't catch it cause there are no test for this section, for which we have an open issue: https://github.com/biocore/qiita/issues/406, which points to the importance of having this ASAP.


josenavas:
:+1:

josenavas:
@antgonza do you mind changing the issue number in the tests? I think they're not pointing to the correct one.

antgonza:
#406 is the correct one about testing EBI, without that one we can't test. I know there is another one about general testing ....

PR #1640, opened 2016-02-07
Adding version to GUI

antgonza:
I have also added some initial testing for qiita_core

### DIFF
--- a/qiita_core/tests/test_util.py
+++ b/qiita_core/tests/test_util.py
@@ -48,7 +48,7 @@ class UtilTests(TestCase):
             # retrieve transaction id
             with qdb.sql_connection.TRN:
                 sql = "SELECT txid_current();"
-                qdb.sql_connection.TRN.add(sql, [])
+                qdb.sql_connection.TRN.add(sql)
                 return qdb.sql_connection.TRN.execute_fetchlast()
 
         f1 = function()
--- a/qiita_core/tests/test_util.py
+++ b/qiita_core/tests/test_util.py
@@ -1,5 +1,3 @@
-#!/usr/bin/env python
-
 from __future__ import division
 
 # -----------------------------------------------------------------------------
@@ -58,9 +56,8 @@ class UtilTests(TestCase):
         self.assertNotEqual(f1, f2)
 
     def test_get_qiita_version(self):
-        version = get_qiita_version()
+        exp_version, exp_sha = get_qiita_version()
         # testing just the version
-        exp_version = version.split(' ')[0]
         self.assertEqual(exp_version, qdb.__version__)
 
 if __name__ == '__main__':
--- a/qiita_core/tests/test_util.py
+++ b/qiita_core/tests/test_util.py
@@ -1,0 +1,67 @@
+#!/usr/bin/env python
+
+from __future__ import division
+
+# -----------------------------------------------------------------------------
+# Copyright (c) 2014--, The Qiita Development Team.
+#
+# Distributed under the terms of the BSD 3-clause License.
+#
+# The full license is in the file LICENSE, distributed with this software.
+# -----------------------------------------------------------------------------
+
+from unittest import TestCase, main
+
+from qiita_core.util import (
+    send_email, qiita_test_checker, execute_as_transaction, get_qiita_version)
+import qiita_db as qdb
+
+
+class UtilTests(TestCase):
+    def test_send_email(self):
+        # TODO: issue 1639
+        # figure out how to test sending emails
+        pass
+
+    def test_send_email_fail(self):
+        """testing send email functionality"""
+        # the default configuration is not correct and should fail
+        with self.assertRaises(IOError):
+            send_email("antgonza@gmail.com", "This is a test",
+                       "This is the body of the test")
+
+    def test_qiita_test_checker(self):
+        """testing qiita test checker"""
+        @qiita_test_checker()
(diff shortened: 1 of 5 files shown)


qiita_core/tests/test_util.py
```diff
@@ -0,0 +1,67 @@
+#!/usr/bin/env python
```
ElDeveloper: Can you remove this line and the blank line below?

qiita_core/util.py
```diff
@@ -93,3 +101,21 @@ def wrapper(*args, **kwargs):
         with TRN:
             return func(*args, **kwargs)
     return wrapper
+
+
+def get_qiita_version():
+    """Returns the Qiita version and Git sha if present"""
```
ElDeveloper: Can you add a Returns section?

qiita_core/util.py
```diff
@@ -93,3 +101,21 @@ def wrapper(*args, **kwargs):
...
+    # the actual repo is the abspath of the current file without
+    # qiita_core
+    current_fp = dirname(__file__)
+    len_for_trimming = len('qiita_core/')
+    git_repo_path = current_fp[:-len_for_trimming]
+
+    try:
+        repo = Repo(git_repo_path)
+        sha = '@' + repo.active_branch.commit.hexsha[0:7]
+    except:
```
ElDeveloper: Can we narrow down the "exception space"?

qiita_pet/templates/sitebase.html
```diff
@@ -241,6 +243,11 @@
                 <li>
                   <a href="http://github.com/biocore/qiita">GitHub</a>
                 </li>
+                <li role="separator" class="divider"></li>
+                <li>
+                  <a href="#"><h6>{{qiita_version}}</h6></a>
```
ElDeveloper: The version and SHA makes perfect sense to me, but not sure it will make sense to everyone, do you think it would help adding a "Current version: XXXXXXXX" label or something along those lines?

ElDeveloper:
Looks good!! Just some minor comments. What do you guys think about linking to the commit on GitHub? i.e. when you click on the version number (`16624f7`) you get linked to `https://github.com/biocore/qiita/commit/16624f7`, not sure if it would be useful though.

qiita_core/util.py
antgonza: We could but not sure why will we want to do this, could you comment? Basically, not sure which errors should we catch, I think all-exceptions is fine.

qiita_pet/templates/sitebase.html
antgonza: I tried adding in 1 line but it looks horrible. I can try in 2, what do you think?

antgonza:
Thought about the linking but that will mean that either get_qiita_version will return 2 elements (qiita_version, sha) or the GUI will need to split and then create the link. Any preferences?

qiita_core/util.py
ElDeveloper: The reason why you want to only catch a specific exception is so as to
not mask real errors, does that make sense? It is briefly mentioned in
the PEP8 document, look for "When catching exceptions, mention ..." 
https://www.python.org/dev/peps/pep-0008/#exception-names

ElDeveloper:
I think it makes more sense to go for two elements on return. Note that
we can only do the linking if we have the SHA (i.e. if we are on a
GitHub repo), so that complicates things a little ...

qiita_pet/templates/sitebase.html
ElDeveloper: Hahaha, if it looks horrible then maybe two lines.

qiita_core/tests/test_util.py
```diff
@@ -0,0 +1,64 @@
...
+                pass
+
+    def test_execute_as_transaction(self):
+        """testing that execute as transaction returns 2 different wrappers"""
+        @execute_as_transaction
+        def function():
+            # retrieve transaction id
+            with qdb.sql_connection.TRN:
+                sql = "SELECT txid_current();"
+                qdb.sql_connection.TRN.add(sql, [])
```
josenavas: Minor, but you should be able to just do `qdb.sql_connection.TRN.add(sql)`

qiita_core/util.py
```diff
@@ -93,3 +102,26 @@ def wrapper(*args, **kwargs):
...
+
+def get_qiita_version():
+    """Returns the Qiita version and Git sha if present
+
+    Returns
+    ------
+    tuple (version, sha)
+        The Qiita version and SHA. SHA can be an empty string.
+    """
+    # the actual repo is the abspath of the current file without
```
josenavas: Is this true even if you install it using pip install and is not located in the actual git repo directory?

josenavas:
2 minor comments

qiita_core/util.py
antgonza: If that's the case there is no way of knowing the github SHA and the only thing that will be displayed is the Qiita version.

josenavas: That makes sense.

I didn't catch this last night (sorry) but the code in here can be cleaned up like this:
```python
git_repo_path = dirname(dirname(__file__))
```
which is probably safer in case that `qiita_core` changes.

antgonza: good point.

josenavas:
:+1:

ElDeveloper:
:+1:

PR #1641, opened 2016-02-11
Workflow object

josenavas:
This PR adds the necessary DB structure to support default workflows from plugins and user-generated workflows.

Overview of changes in this PR:
- Structures added to the DB to support default workflows from plugins.
- DefaultWorkflow object (and aux classes) added to access the default workflows.
- Small table added to keep track of the user-generated workflow
- _Currently not implemented (will be in tomorrow):_ ProcessingWorkflow object to access the user defined workflows.
- In order to implement the above class, we need to keep track of which artifacts have been generated from which ProcessingJob. This adds another table to the database and adds some work on job completion. This triggered changes in the ProcessingJob object, so more work is kept internally to the object rather than outside of the object, forcing to make changes in the restAPI handlers, restAPI interface and plugins (to meet the new requirements of the restAPI).


qiita_db/handlers/processing_job.py
```diff
@@ -175,26 +170,19 @@ def post(self, job_id):
...
-                            qdb.artifact.Artifact.create(
-                                filepaths, atype, parents=parents,
-                                processing_parameters=params)
-                        job.status = 'success'
+                payload = loads(self.request.body)
+                try:
+                    payload_success = payload['success']
+                    if payload_success:
+                        artifacts = payload['artifacts']
+                        error = None
```
antgonza: Should error be '' so we always return a string?

antgonza: Ignore this, got confused with error_msg. However, error_msg will not be defined, right?

qiita_db/processing_job.py
```diff
@@ -219,6 +221,60 @@ def status(self, value):
...
             qdb.sql_connection.TRN.execute()
 
+    def complete(self, success, artifacts_data=None, error=None):
+        """Completes the job, either with a success or error status
+
+        Parameters
+        ----------
+        success : bool
+            Whether the job has completed successfully or not
+        artifacts_data : dict of dicts
```
antgonza: This one is also optional, right?

qiita_db/processing_job.py
```diff
@@ -219,6 +221,60 @@ def status(self, value):
...
+        error : str, optional
+            If the job was not successful, the error message
+
+        Raises
+        ------
+        qiita_db.exceptions.QiitaDBOperationNotPermittedError
+            If the job is not in running state
+        """
+        with qdb.sql_connection.TRN:
+            if self.status != 'running':
```
antgonza: missing test

qiita_db/processing_job.py
```diff
@@ -219,6 +221,60 @@ def status(self, value):
...
+                        atype = a_data['artifact_type']
+                        parents = self.input_artifacts
+                        params = self.parameters
+                        a = qdb.artifact.Artifact.create(
+                            filepaths, atype, parents=parents,
+                            processing_parameters=params)
+                        cmd_out_id = qdb.util.convert_to_id(
+                            out_name, "command_output", "name")
+                        artifact_ids.append((cmd_out_id, a.id))
+                    if artifact_ids:
```
antgonza: In which cases will artifact_ids be empty and what should happen? I don't think there is test for the case where artifact_ids doesn't exist.

qiita_db/processing_job.py
```diff
@@ -281,20 +336,28 @@ def heartbeat(self):
             qdb.sql_connection.TRN.add(sql, [self.id])
             return qdb.sql_connection.TRN.execute_fetchlast()
 
-    @heartbeat.setter
-    def heartbeat(self, value):
-        """Sets the timestamp for the last received heartbeat
+    def execute_heartbeat(self):
```
antgonza: This only changes the heartbeat from one state to another, right? If so, this name is confusing and the first line of the docstring is misleading.

qiita_db/software.py
```diff
@@ -514,6 +516,21 @@ def start_script(self):
             qdb.sql_connection.TRN.add(sql, [self.id])
             return qdb.sql_connection.TRN.execute_fetchlast()
 
+    @property
+    def default_workflows(self):
```
antgonza: I think this is missing tests.

qiita_db/software.py
```diff
@@ -879,3 +896,139 @@ def dump(self):
...
+            qdb.sql_connection.TRN.add(sql, [self.id])
+            db_nodes = qdb.sql_connection.TRN.execute_fetchflatten()
+
+            nodes = {}
+            for node_id in db_nodes:
+                n = DefaultWorkflowNode(node_id)
+                nodes[node_id] = n
+                g.add_node(n)
+
+            # Retrieve all grpah edges
```
antgonza: grpah -> graph

qiita_db/support_files/patches/38.sql
```diff
@@ -0,0 +1,241 @@
...
+	VALUES ('in_construction', 'The job is one of the source nodes of a workflow that is in construction'),
+		   ('waiting', 'The job is waiting for a previous job in the workflow to be completed in order to be executed.');
+
+-- Populate the newly created tables
+DO $do$
+DECLARE
+	in_slq_param_id		bigint;
+	in_sl_param_id		bigint;
+	in_po_param_id		bigint;
+	dflt_slq_id			bigint;
```
antgonza: extra tabs?

qiita_db/support_files/patches/38.sql
```diff
@@ -0,0 +1,241 @@
...
+	dflt_per_sample_id	bigint;
+	dflt_po_id			bigint;
+BEGIN
+	-- Add the artifact type information for the input parameters for the commands
+	-- command_id = 1 -> Split libraries FASTQ
+	SELECT command_parameter_id FROM qiita.command_parameter
+		WHERE command_id = 1 AND parameter_name = 'input_data'
+		INTO in_slq_param_id;
+
+	-- Split libraries FASTQ supports FASTQ (3) and per_sample_FASTQ (5)
```
antgonza: I'm not 100% sure that this is true in the live system. Can you check or query by name?

antgonza: This applies to all ids in this file.

qiita_db/support_files/patches/38.sql
```diff
@@ -0,0 +1,241 @@
...
+	-- the command id because they're inserted in patch 33.sql, and that is
+	-- the only way of adding commands at this point.
+
+	-- Insert default workflow
+	INSERT INTO qiita.default_workflow (software_id, name)
+		VALUES (1, 'FASTQ upstream workflow'),
+			   (1, 'FASTA upstream workflow'),
+			   (1, 'Per sample FASTQ upstream workflow');
+
+	-- Retrieve all the ids of the default parameter set that we need
```
antgonza: Can you make sure that these are all the ids we have?

antgonza: I know they should be but makes me a bit worry ...

antgonza:
A few comments.

qiita_db/handlers/processing_job.py
josenavas: error_msg is defined in line 171.

qiita_db/processing_job.py
josenavas: Good catch

qiita_db/processing_job.py
josenavas: Added

qiita_db/processing_job.py
josenavas: Currently there is no command that doesn't generate an artifact. I'm preparing the system for such cases. One example may be EBI submission - they don't generate new artifacts. Since EBI submission is not a plugin command, I can't test.

qiita_db/processing_job.py
josenavas: Can you provide a better name? I didn't like it either, but I couldn't come up with a good name.

qiita_db/software.py
josenavas: Good catch - added

qiita_db/software.py

qiita_db/support_files/patches/38.sql
josenavas: Argh - fixed in the entire file

qiita_db/support_files/patches/38.sql
josenavas: Yes - those are the ones that we care about here. Note that the artifact types is a controlled vocab. That is, all Qiita databases should have the same values ATM, since it is only changed i the patches.

qiita_db/support_files/patches/38.sql
josenavas: Those are "all" that we need here. Note that each query returns a set but I'm getting the minimum of those, specifically because at the patch level there is no control over these.

josenavas:
This is ready for review - Note that the `submit` methods (or any method/branch that ends up doing a submit) can't be tested since those will need to execute the commands (i.e. split libraries). I've tried to decompose those methods in small testable pieces, but that is not always achievable.

qiita_db/processing_job.py
```diff
@@ -7,10 +7,62 @@
...
+from qiita_core.qiita_settings import qiita_config
 import qiita_db as qdb
 
 
+def _system_call(cmd):
+    """Call cmd and return (stdout, stderr, return_value)
+
+    Parameters
+    ----------
+    cmd : str or iterator of str
```
antgonza: Should this have a test for an iterator of str? I can see one for success (str) and error.

qiita_db/processing_job.py
```diff
@@ -7,10 +7,62 @@
...
+    proc = Popen(cmd, universal_newlines=True, shell=True, stdout=PIPE,
+                 stderr=PIPE)
+    # Communicate pulls all stdout/stderr from the PIPEs
+    # This call blocks until the command is done
+    stdout, stderr = proc.communicate()
+    return_value = proc.returncode
+    return stdout, stderr, return_value
+
+
+def _job_submitter(job, cmd):
```
antgonza: Test for this?

qiita_db/processing_job.py
```diff
@@ -186,13 +253,13 @@ def status(self):
             qdb.sql_connection.TRN.add(sql, [self.id])
             return qdb.sql_connection.TRN.execute_fetchlast()
 
-    @status.setter
-    def status(self, value):
+    def _set_status(self, value):
```
antgonza: Test?

qiita_db/processing_job.py
```diff
@@ -219,6 +286,95 @@ def status(self, value):
...
+        job_dir = join(qdb.util.get_work_base_dir(), self.id)
+        software = self.command.software
+        plugin_start_script = software.start_script
+        plugin_env_script = software.environment_script
+        cmd = '%s "%s" "%s" "%s" "%s" "%s"' % (
+            qiita_config.plugin_launcher, plugin_env_script,
+            plugin_start_script, qiita_config.base_url, self.id, job_dir)
+        return cmd
+
+    def submit(self):
```
antgonza: Is there a test for success?

qiita_db/processing_job.py
antgonza: update_heartbeat_state?

qiita_db/processing_job.py
```diff
@@ -337,3 +507,445 @@ def step(self, value):
...
+        """
+        with qdb.sql_connection.TRN:
+            sql = """SELECT child_id
+                     FROM qiita.parent_processing_job
+                     WHERE parent_id = %s"""
+            qdb.sql_connection.TRN.add(sql, [self.id])
+            for jid in qdb.sql_connection.TRN.execute_fetchflatten():
+                yield ProcessingJob(jid)
+
+    def _update_children(self, mapping):
```
antgonza: Test for this?

qiita_db/processing_job.py
```diff
@@ -337,3 +507,445 @@ def step(self, value):
...
+    ----------
+    user : qiita_db.user.User
+        The user that modeled the workflow
+    root : list of qiita_db.processing_job.ProcessingJob
+        The first job in the workflow
+    """
+    _table = "processing_job_workflow"
+
+    @classmethod
+    def _common_creation_steps(cls, user, root_jobs, name=None):
```
antgonza: Test for this?

qiita_db/processing_job.py
```diff
@@ -337,3 +507,445 @@ def step(self, value):
...
+            # multiple times, so caching in here
+            all_nodes = {n: (n.command, n.parameters)
+                         for n in in_degrees}
+            roots = {n: (n.command, n.parameters)
+                     for n, d in viewitems(in_degrees) if d == 0}
+
+            # Check that we have all the required parameters
+            root_cmds = set(c for c, _ in viewvalues(roots))
+            if root_cmds != set(req_params):
+                error_msg = ['Provided required parameters do not match the '
```
antgonza: This needs a test.

antgonza: It seems like all errors need a test.

qiita_db/processing_job.py
```diff
@@ -337,3 +507,445 @@ def step(self, value):
...
+            qdb.sql_connection.TRN.add(sql, [self.id])
+            edges = qdb.sql_connection.TRN.execute_fetchindex()
+            if edges:
+                nodes = {jid: ProcessingJob(jid)
+                         for jid in set(chain.from_iterable(edges))}
+                g.add_nodes_from(nodes.values())
+                edges = [(nodes[s], nodes[d]) for s, d in edges]
+                g.add_edges_from(edges)
+            else:
+                # It is possible that there are no edges because we are still
```
antgonza: One of this is missing a test.

qiita_db/processing_job.py
```diff
@@ -337,3 +507,445 @@ def step(self, value):
...
+                         FROM qiita.processing_job_workflow_root
+                         WHERE processing_job_workflow_id = %s"""
+                qdb.sql_connection.TRN.add(sql, [self.id])
+                nodes = [
+                    ProcessingJob(jid)
+                    for jid in qdb.sql_connection.TRN.execute_fetchflatten()]
+                g.add_nodes_from(nodes)
+        return g
+
+    def _raise_if_not_in_construction(self):
```
antgonza: No tests.

qiita_db/processing_job.py
```diff
@@ -337,3 +507,445 @@ def step(self, value):
...
+            qdb.sql_connection.TRN.add(sql, [job.id])
+
+            # Remove the job
+            sql = """DELETE FROM qiita.processing_job
+                     WHERE processing_job_id = %s"""
+            qdb.sql_connection.TRN.add(sql, [job.id])
+
+            qdb.sql_connection.TRN.execute()
+
+    def submit(self):
```
antgonza: Test for success?

antgonza:
A few more comments.

qiita_db/processing_job.py
josenavas: Good catch - when adding the test, I found that it doesn't actually work with an iterable of str. The only reason why we were accepting an iterable is b/c Popen supports both. However, from the [Popen](https://docs.python.org/2/library/subprocess.html#popen-constructor) docs:
> If shell is True, it is recommended to pass args as a string rather than as a sequence
Thus, I think disabling iterable is the best (we just need to update the docs).

qiita_db/processing_job.py
josenavas: I can of hacked in a test for this.

qiita_db/processing_job.py
josenavas: Note that I changed the this from being a property setter to be an internal function - so the tests for the property setter still applies (I renamed them)

qiita_db/processing_job.py
josenavas: Can't add - it will mean to run split libraries. Added a comment in the test though, for future reference.

qiita_db/processing_job.py
josenavas: Done

qiita_db/processing_job.py
josenavas: I broke up this function in 2 pieces - the first one updates the children and returns the ones that are ready for submission (and added a test), the second one submits those jobs. Can't add a tests for this second one (same reasoning as above) but noted in the tests.

qiita_db/processing_job.py
josenavas: This function is tested through the tests for the 2 next functions. The reason why I'm hesitant to add a specific test for this function is that I will need to test the actual SQL, and in previous discussions we agreed that the tests should be SQL-free.

antgonza: OK, thanks.

qiita_db/processing_job.py
josenavas: Good catch - I thought I wrote those but looking back at the code I don't see them. Totally not my day yesterday...

qiita_db/processing_job.py
josenavas: Added

qiita_db/processing_job.py
josenavas: Good call - it is indirectly checked but better to add a test in this case.

qiita_db/processing_job.py
josenavas: Added a note why this can't be tested ATM.

josenavas:
@ElDeveloper and/or @mortonjt can you review this PR?

qiita_db/handlers/processing_job.py
```diff
@@ -175,26 +170,19 @@ def post(self, job_id):
...
-                            filepaths = artifact_data['filepaths']
-                            atype = artifact_data['artifact_type']
-                            parents = job.input_artifacts
-                            params = job.parameters
-                            qdb.artifact.Artifact.create(
-                                filepaths, atype, parents=parents,
-                                processing_parameters=params)
-                        job.status = 'success'
+                payload = loads(self.request.body)
+                try:
```
ElDeveloper: Would you mind restricting the scope of this try block to just the line that might raise the exception?

qiita_db/processing_job.py
```diff
@@ -7,10 +7,62 @@
...
+
+from future.utils import viewitems, viewvalues
+import networkx as nx
+
+from qiita_core.qiita_settings import qiita_config
 import qiita_db as qdb
 
 
+def _system_call(cmd):
+    """Call cmd and return (stdout, stderr, return_value)
```
ElDeveloper: Can you list the outputs under a Returns section?

qiita_db/processing_job.py
```diff
@@ -174,7 +241,7 @@ def status(self):
 
         Returns
         -------
-        {'queued', 'running', 'success', 'error'}
+        {'queued', 'running', 'success', 'error', 'in_construction', 'waiting'}
```
ElDeveloper: Minor, but the numpy way to do this would be:

```python
str
    The current status of the job, one of .....
```
Or something along those lines right?

qiita_db/processing_job.py
```diff
@@ -186,13 +253,13 @@ def status(self):
...
 
-    @status.setter
-    def status(self, value):
+    def _set_status(self, value):
         """Sets the status of the job
 
         Parameters
         ----------
-        value : str, {'queued', 'running', 'success', 'error'}
+        value : str, {'queued', 'running', 'success', 'error',
```
ElDeveloper: Should we check that the value of `value` is valid? Seems like this could have caught the typo that's being fixed in `test_processing_job.py` unless that's totally unrelated.

qiita_db/processing_job.py
```diff
@@ -219,6 +286,95 @@ def status(self, value):
...
+
+    def complete(self, success, artifacts_data=None, error=None):
+        """Completes the job, either with a success or error status
+
+        Parameters
+        ----------
+        success : bool
+            Whether the job has completed successfully or not
+        artifacts_data : dict of dicts, optional
+            The generated artifact information, if any, keyed by output name.
```
ElDeveloper: the "if any" part is redundant as the argument is already marked as "optional" right?

qiita_db/processing_job.py
```diff
@@ -219,6 +286,95 @@ def status(self, value):
...
+        Raises
+        ------
+        qiita_db.exceptions.QiitaDBOperationNotPermittedError
+            If the job is not in running state
+        """
+        with qdb.sql_connection.TRN:
+            if success:
+                if self.status != 'running':
+                    # If the job is not running, we only allow to complete it
+                    # if it did not success
```
ElDeveloper: success -> succeeded?

ElDeveloper: fixed typo :P

qiita_db/processing_job.py
```diff
@@ -219,6 +286,95 @@ def status(self, value):
...
+        """Submits the job to execution
+
+        Raises
+        ------
+        QiitaDBOperationNotPermittedError
+            If the job is not in 'waiting' or 'in_construction' status
+        """
+        if self.status not in {'in_construction', 'waiting'}:
+            raise qdb.exceptions.QiitaDBOperationNotPermittedError(
+                "Can't submit job, not in 'in_construction' or "
```
ElDeveloper: For debugging purposes with this and the other exceptions, would it make sense to what the status is in the error message?

qiita_db/processing_job.py
```diff
@@ -337,3 +507,466 @@ def step(self, value):
...
+            sql = """INSERT INTO qiita.parent_processing_job
+                        (parent_id, child_id)
+                     VALUES (%s, %s)"""
+
+            # Create the rest of the jobs. These are different form the root
+            # jobs because they depend on other jobs to complete in order to be
+            # submitted
+            for n in nx.topological_sort(dflt_g):
+                if n in node_to_job:
+                    # We have already seend this node
```
ElDeveloper: seend -> sent?

qiita_db/processing_job.py
```diff
@@ -337,3 +507,466 @@ def step(self, value):
...
+                    # because we are visiting the nodes in topological order
+                    source_id = node_to_job[source].id
+                    parent_ids.append(source_id)
+                    # Get the connections between the job and the source
+                    connections = data['connections'].connections
+                    for out, in_param in connections:
+                        # We take advantage of the fact the parameters are
+                        # stored in JSON to encode the name of the output
+                        # artifact from
+                        # the previous job
```
ElDeveloper: this comment can probably be joined with the line above.

qiita_db/processing_job.py
```diff
@@ -337,3 +507,466 @@ def step(self, value):
...
+                        job_req_params[in_param] = [source_id, out]
+
+                # At this point we should have all the requried paramters for
```
ElDeveloper: paramters -> parameters

qiita_db/processing_job.py
```diff
@@ -337,3 +507,466 @@ def step(self, value):
...
+        """
+        g = nx.DiGraph()
+        with qdb.sql_connection.TRN:
+            # Retrieve all graph workflow nodes
+            sql = """SELECT parent_id, child_id
+                     FROM qiita.get_processing_workflow_edges(%s)"""
+            qdb.sql_connection.TRN.add(sql, [self.id])
+            edges = qdb.sql_connection.TRN.execute_fetchindex()
+            if edges:
+                nodes = {jid: ProcessingJob(jid)
```
ElDeveloper: You should be able to only add the edges and not the nodes and then the edges:

```
In [1]: import networkx as nx

In [2]: g = nx.DiGraph()

In [3]: g.add_edges_from([('mickey', 'mouse'), ('minnie', 'mouse')])

In [4]: g.nodes()
Out[4]: ['mickey', 'mouse', 'minnie']
```

qiita_db/processing_job.py
```diff
@@ -337,3 +507,466 @@ def step(self, value):
...
+        ------
+        qiita_db.exceptions.QiitaDBOperationNotPermittedError
+            If the workflow is not in construction
+        """
+        with qdb.sql_connection.TRN:
+            self._raise_if_not_in_construction()
+
+            g = self.graph
+            # In order to avoid potential race conditions, we are going to set
+            # all the childrens in 'waiting' status before submitting
```
ElDeveloper: childrens -> children

qiita_db/software.py
```diff
@@ -514,6 +516,22 @@ def start_script(self):
             qdb.sql_connection.TRN.add(sql, [self.id])
             return qdb.sql_connection.TRN.execute_fetchlast()
 
+    @property
+    def default_workflows(self):
+        """Returns the default workflows attached to the current software
+
+        Returns
+        -------
+        generator of qiita_db.software.DefaultWorkflow"""
```
ElDeveloper: Can you indent this and move the triple-quotes to the line below?

qiita_db/software.py
```diff
@@ -879,3 +897,139 @@ def dump(self):
...
+        g = nx.DiGraph()
+        with qdb.sql_connection.TRN:
+            # Retrieve all graph workflow nodes
+            sql = """SELECT default_workflow_node_id
+                     FROM qiita.default_workflow_node
+                     WHERE default_workflow_id = %s"""
+            qdb.sql_connection.TRN.add(sql, [self.id])
+            db_nodes = qdb.sql_connection.TRN.execute_fetchflatten()
+
+            nodes = {}
```
ElDeveloper: If it applies as well, same comment as above regarding node vs edge creation.

ElDeveloper:
:+1:

qiita_db/handlers/processing_job.py
josenavas: Done

qiita_db/processing_job.py
josenavas: Done

qiita_db/processing_job.py
josenavas: Done

qiita_db/processing_job.py
josenavas: The call `new_status = qdb.util.convert_to_id(value, "processing_job_status")` at line 281 should already raise a meaningful error in case that it doesn't exist.

qiita_db/processing_job.py
josenavas: Done

qiita_db/processing_job.py
josenavas: Done

qiita_db/processing_job.py
josenavas: :dancer:

qiita_db/processing_job.py
josenavas: Changed to "visited" - makes more sense in a graph context :smile:

qiita_db/processing_job.py
josenavas: Done

qiita_db/processing_job.py
josenavas: Done

qiita_db/processing_job.py
josenavas: Done - note that I still need the nodes dictionary anyway to reduce the number of DB calls when creating the ProcessingJob (otherwise it will be generated as many times as the node connectivity)

qiita_db/processing_job.py
josenavas: Done

qiita_db/software.py
josenavas: I don't I need to indent this because that line is exposing the type of the return - I've added another line with the description of the return.

qiita_db/software.py

josenavas:
Thanks! Comments addressed

PR #1644, opened 2016-02-16
fix #1236

### LINKED ISSUE #1236: Ensure unique IDs for metaanalysis
If a study has 2 process data from 2 different prep templates and the prep templates have the same sample name (which they should) the analysis will merge the 2 biom into one. The correct functionality should be to keep them apart.

antgonza:
This fixes #1236 as discussed in there and during the last call. The best way to test this was to add a new biom artifact in a study that already had one => added in populate_test_db.sql (artifact 5). This change will also be helpful as we move to test more than one artifact at the time. However, this change also make some scripts to failed and I'm fixing them. I check all errors in my laptop and flake8 so we should be OK to review ASAP.


qiita_db/analysis.py
```diff
@@ -25,7 +25,7 @@
 from biom import load_table
 from biom.util import biom_open
 import pandas as pd
-from skbio.util import find_duplicates
+from skbio.util import flatten
```
josenavas: `flatten` has been removed in later versions of skbio. In other parts of the code, we have been using the construction below to use flatten:
```python
from itertools import chain
list(chain.from_iterable(ITERABLE))
```

qiita_db/analysis.py
```diff
@@ -802,59 +805,69 @@ def build_files(self, rarefaction_depth=None):
...
-                table.add_metadata(samples_meta, axis='sample')
+                        % aid)
+                biom_table = load_table(biom_table_fp)
+                # filtering samples to keep those selected by the user
+                biom_table_samples = set(biom_table.ids())
+                selected_samples = biom_table_samples.intersection(samps)
+                biom_table.filter(selected_samples, axis='sample',
+                                  inplace=True)
+
+                if not duplicated_samples:
```
josenavas: I'm a bit confused by the name of this variable. If `duplicated_samples == True` then we rename samples? Can this be changed to `rename_duplicated_samples`? I'm just thinking in the case in where we have duplicated samples but we want to merge them - the name of this variable is a bit misleading.

qiita_db/analysis.py
```diff
@@ -802,59 +805,69 @@ def build_files(self, rarefaction_depth=None):
...
+                    ids_map = {_id: "%d.%s" % (aid, _id)
+                               for _id in biom_table.ids()}
+                    biom_table.update_ids(ids_map, 'sample', True, True)
+
+                # add the metadata column for study the samples come from,
+                # this is useful in case the user download the bioms
+                study_md = {'Study': artifact.study.title, 'Processed_id': aid}
```
josenavas: Should this be `artifact_id` instead of `processed_id`?

qiita_db/analysis.py
```diff
@@ -883,47 +896,45 @@ def _build_mapping_file(self, samples):
...
-
-                duplicates = all_sample_ids.intersection(qiime_map.index)
-                if duplicates or len(samples) != len(set(samples)):
-                    # Duplicate samples so raise error
-                    raise qdb.exceptions.QiitaDBError(
-                        "Duplicate sample ids found: %s"
-                        % ', '.join(duplicates))
-                all_sample_ids.update(qiime_map.index)
-                to_concat.append(qiime_map)
+                qm = pd.read_csv(join(fp, qm_fp), sep='\t',
```
josenavas: can you use `load_template_to_dataframe` in `qiita_db.metadata_template.util` instead of `read_csv`? This will help to centralize the parsing of these files, and that function also supports parsing QIIME mapping files (pass the parameter `index='#SampleID'`.

qiita_db/analysis.py
```diff
@@ -883,47 +896,45 @@ def _build_mapping_file(self, samples):
...
+                                 keep_default_na=False, na_values=['unknown'],
+                                 index_col=False,
+                                 converters=defaultdict(lambda: str))
+
+                # if we are not going to merge the duplicated samples
+                # append the pid to the sample name
+                if duplicated_samples:
```
josenavas: Same comment as above regarding the name of the variable. Also it looks like here is working the other way around? If `duplicated_samples == True` they don't get renamed, while above they do get renamed.

qiita_db/analysis.py
```diff
@@ -883,47 +896,45 @@ def _build_mapping_file(self, samples):
...
+                    samps = set(samps) - all_ids
+                    all_ids.update(samps)
+                else:
+                    qm['#SampleID'] = qm['#SampleID'] + ".%d" % pid
```
josenavas: Probably I remember this wrong - but I think that we agreed it was better to prefix the samples with the pid rather than suffix? @ElDeveloper do you remember what we agreed in here?

qiita_db/test/test_analysis.py
```diff
@@ -486,26 +494,18 @@ def test_unshare(self):
...
         self.assertEqual(obs, self.map_fp)
 
-        base_dir = qdb.util.get_mountpoint('analysis')[0][1]
         obs = pd.read_csv(obs, sep='\t', infer_datetime_format=True,
                           parse_dates=True, index_col=False, comment='\t')
-        exp = pd.read_csv(join(base_dir, '1_analysis_mapping_exp.txt'),
-                          sep='\t', infer_datetime_format=True,
-                          parse_dates=True, index_col=False, comment='\t')
-
+        exp = pd.read_csv(self.map_exp_fp, sep='\t',
```
josenavas: Same comment here regarding the parsing of the mapping files.

qiita_db/test/test_analysis.py
```diff
@@ -549,13 +567,25 @@ def test_build_biom_tables(self):
...
         self.assertTrue(obs)
 
         sql = """SELECT * FROM qiita.analysis_filepath
                  WHERE analysis_id=%s ORDER BY filepath_id"""
         obs = self.conn_handler.execute_fetchall(sql, (self.analysis.id,))
-        exp = [[1L, 14L, 2L], [1L, 15L, None], [1L, new_id, None]]
+        exp = [[1L, 12L, 2L], [1L, 13L, None], [1L, new_id, 2L]]
+        self.assertEqual(obs, exp)
+
+    def test_build_biom_tables_duplicated_samples_merge(self):
```
josenavas: Yeah, looking at the tests it is clear that the parameter `duplicated_samples` causes different behavior in these two functions - can we normalize?

qiita_pet/handlers/analysis_handlers.py
```diff
@@ -111,6 +111,10 @@ def get(self, analysis_id):
     def post(self, analysis_id):
         analysis_id = int(analysis_id)
         rarefaction_depth = self.get_argument('rarefaction-depth')
+        mdsi = self.get_argument('merge-duplicated-sample-ids', default=False)
+        if mdsi == 'on':
```
josenavas: Should this be changed to: `mdsi = mdsi == 'on'`; so in case that is not equal it goes to `False`? Right now it will have whatever value it had before (i.e. a string - rather than a boolean)

qiita_pet/templates/select_commands.html
```diff
@@ -24,8 +24,34 @@
...
+          <td style="width:20px;">
+            <input type="checkbox" id="merge-duplicated-sample-ids" name="merge-duplicated-sample-ids" class="form-control">
+          </td>
+          <td>
+            <label style="font-weight:normal;" for="merge-duplicated-sample-ids">Merge same sample ids</label>
+          </td>
+        </tr>
+        <tr>
+          <td colspan="2">
+            <small>Merging sample ids is useful for when you have the same sample in different preparations of the same data, i.e. a sample processed twice in 16S.</small>
```
josenavas: Should we add a comment about what happens to the sample ids if we don't merge them (i.e. they get prefixed/suffixed)?

josenavas:
thanks @antgonza - some comments

qiita_db/analysis.py
ElDeveloper: Yes, that's also what I recall, specifically in support of the argument
that all the information would be grouped together at the beginning of
the sample name. And users would still be able to use the "original
sample id" column just in case.

qiita_db/analysis.py
antgonza: K

qiita_db/analysis.py
antgonza: I will rename to rename_dup_samples

qiita_db/analysis.py
antgonza: Yup, note that this was a left over from when we moved to artifacts and didn't change to not move things around. Changing now.

qiita_db/analysis.py
antgonza: After discussing offline we realized that this is correct but the name is misleading.

qiita_db/analysis.py
antgonza: K, fixing.

qiita_db/test/test_analysis.py
antgonza: The confusion here is the test name, not the actual test of behavior. Renaming.

qiita_pet/handlers/analysis_handlers.py
antgonza: Not sure about this one. By default it's always False, line above. Remember that if is not "on", the argument doesn't exist.

qiita_pet/templates/select_commands.html
antgonza: Fine.

qiita_db/analysis.py
```diff
@@ -802,59 +803,70 @@ def build_files(self, rarefaction_depth=None):
...
-            return dict(qdb.sql_connection.TRN.execute_fetchindex())
-
-    def _build_biom_tables(self, samples, rarefaction_depth):
+            # in practice we could retrieve samples in each of the following
+            # calls but this will mean calling the DB multiple times and will
+            # make testing much harder as we will need to have analyses at
+            # different stages and possible errors.
+            samples = self.samples
+            # figuring out if we are going to have duplicated samples, again
+            # doing it here cause it's computetional cheaper
```
mortonjt: computetional -> computational

qiita_db/analysis.py
```diff
@@ -802,59 +803,70 @@ def build_files(self, rarefaction_depth=None):
...
-            # filter and combine all study BIOM tables needed for
-            # each data type
-            new_tables = {dt: None for dt in self.data_types}
             base_fp = qdb.util.get_work_base_dir()
-            for a_id, samps in viewitems(samples):
-                # one biom table attached to each artifact object
-                artifact = qdb.artifact.Artifact(a_id)
-                table_fp = None
+
+            # this assumes that there is only one refererence/pipeline for each
```
mortonjt: refererence -> reference

mortonjt:
Super minor typos.  :+1: otherwise.

qiita_db/analysis.py
```diff
@@ -883,47 +895,44 @@ def _build_mapping_file(self, samples):
...
-                     ORDER BY filepath_id DESC"""
-            _id, fp = qdb.util.get_mountpoint('templates')[0]
+                     ORDER BY filepath_id DESC LIMIT 1"""
+            _, fp = qdb.util.get_mountpoint('templates')[0]
+
+            all_ids = set()
             to_concat = []
+            for aid, samps in viewitems(samples):
+                qdb.sql_connection.TRN.add(sql, [aid])
+                qm_fp = qdb.sql_connection.TRN.execute_fetchindex()[0][1]
```
josenavas: Just realized - if you are accession only to `[0][1]`, doesn't it mean that the above query is retrieving more info than needed? I think it should only be `SELECT filepath` and then here you can use `execute_fetchlast`. I think it will make the code cleaner and also removes the magic numbers.

qiita_db/analysis.py
```diff
@@ -883,47 +895,44 @@ def _build_mapping_file(self, samples):
                      WHERE filepath_type = 'qiime_map'
                         AND artifact_id IN (SELECT *
                                             FROM qiita.find_artifact_roots(%s))
-                     ORDER BY filepath_id DESC"""
-            _id, fp = qdb.util.get_mountpoint('templates')[0]
+                     ORDER BY filepath_id DESC LIMIT 1"""
+            _, fp = qdb.util.get_mountpoint('templates')[0]
```
josenavas: And this is the same that `fp = qdb.util.get_mountpoint('templates')[0][0]`. Probably adding a comment saying that we are only interested in the filepath but not in the id.

qiita_pet/handlers/analysis_handlers.py
josenavas: Oh - ok, sorry about that.

qiita_ware/analysis_pipeline.py
```diff
@@ -26,7 +27,11 @@ def _build_analysis_files(analysis, r_depth=None, **kwargs):
     analysis : Analysis object
         The analysis to build files for
     r_depth : int, optional
-        Rarefaction depth for biom table creation. Default None
+        Rarefaction depth for biom table creation. Default
```
josenavas: I think this line got cut. `Default: None`. Also, it was not there before but I think this will improve the documentation:
`Rarefaction depth for biom table creation. Default: None, no rarefaction is applied`. If I understand correctly, that is what is happening - if my description of `None` is off, then is even more important to have the docs.

josenavas:
Few minor comments.

qiita_db/analysis.py
antgonza: I think that's clear from the code as we are discarding _. Not sure this is needed. Do you feel strong about it?

qiita_db/analysis.py
antgonza: Good point.

qiita_ware/analysis_pipeline.py
antgonza: Adding.

qiita_db/analysis.py
josenavas: Nope

josenavas:
:+1: given tests pass

PR #1646, opened 2016-02-18
adding a new biom and reference for analysis

antgonza:
Here I'm adding a new reference (silva) and a new biom so we can start working on multiple analysis.

Additionally from adding those lines, fixing tests and a file:
- Had to modify tree_fp in reference so it returns an empty string when the tree doesn't exist.
- Change rename_dup_samples from or to and in build_files so it behaves as expected. This line is going to change once the rest of the code is there.

Note that this doesn't change anything in the front end or current status, thus, doing a PR against master.


josenavas:
@antgonza there are test failures here. I think those failures are related with your changes to populate_test_db.sql, not to your new functionality, so I'm going to review while you fix those tests.

qiita_db/analysis.py
```diff
@@ -812,7 +812,7 @@ def build_files(self,
             # doing it here cause it's computational cheaper
             all_ids = list(chain.from_iterable([
                 samps for _, samps in viewitems(samples)]))
-            rename_dup_samples = ((len(all_ids) != len(set(all_ids))) or
+            rename_dup_samples = ((len(all_ids) != len(set(all_ids))) and
```
josenavas: Shouldn't a test be added to this specific change gets caught? The fact that the code got merged earlier points out that the unit tests are not enough to catch this issue?

qiita_db/analysis.py
```diff
@@ -871,6 +871,8 @@ def _build_biom_tables(self, samples, rarefaction_depth=None,
             # add the new tables to the analysis
             _, base_fp = qdb.util.get_mountpoint(self._table)[0]
             for dt, biom_table in viewitems(new_tables):
+                if biom_table is None:
```
josenavas: Same here

qiita_db/reference.py
```diff
@@ -171,6 +171,10 @@ def tree_fp(self):
                         JOIN qiita.{0} r ON r.tree_filepath=f.filepath_id
                      WHERE r.reference_id=%s""".format(self._table)
             qdb.sql_connection.TRN.add(sql, [self._id])
-            rel_path = qdb.sql_connection.TRN.execute_fetchlast()
+            try:
```
josenavas: Add test

qiita_db/support_files/populate_test_db.sql
```diff
@@ -431,12 +438,17 @@ INSERT INTO qiita.filepath (filepath, filepath_type_id, checksum, checksum_algor
...
 -- Attach filepath to analysis
-INSERT INTO qiita.analysis_filepath (analysis_id, filepath_id, data_type_id) VALUES (1, 12, 2), (1, 13, NULL);
+INSERT INTO qiita.analysis_filepath (analysis_id, filepath_id, data_type_id) VALUES (1, 15, 2), (1, 16, NULL);
 
 -- Attach samples to analysis
-INSERT INTO qiita.analysis_sample (analysis_id, artifact_id, sample_id)
-    VALUES (1, 4, '1.SKB8.640193'), (1, 4, '1.SKD8.640184'), (1, 4, '1.SKB7.640196'), (1, 4, '1.SKM9.640192'), (1, 4, '1.SKM4.640180'),
-           (2, 4, '1.SKB8.640193'), (2, 4, '1.SKD8.640184'), (2, 4, '1.SKB7.640196'), (2, 4, '1.SKM3.640197');
+INSERT INTO qiita.analysis_sample (analysis_id, artifact_id, sample_id) VALUES
+-- first table
```
josenavas: Is this comment incomplete? or is missing another comment below?

josenavas:
Thanks @antgonza - few comments.

We really need to move away from having hard-coded id's in the tests. Ugh.

qiita_db/analysis.py
antgonza: The or made sense when you had only one kind of data. The and makes sense when you have multiple data. Once we allow for multiple references this is gonna change to be tested per data_type/reference. Thus, no more tests are needed. Please see my 2nd point of my initial comment.

qiita_db/analysis.py
antgonza: Same reply.

qiita_db/reference.py
antgonza: There is a test for this: Having the Silva new reference. That why I catch that this didn't work.

qiita_db/support_files/populate_test_db.sql
antgonza: Removing

qiita_db/analysis.py
josenavas: Just to make sure - the reason why it is not a big deal to merge in master ATM is b/c this functionality is not exposed to the GUI, correct?

qiita_db/reference.py
josenavas: I'm confused here. I don't see any new test in the test_reference.py file, and if all the previous unit tests where passing, how is possible that there is a test that it's executing this code?
Also, I don't think the functionality is correct here. If the reference doesn't have a tree, it should return `None`. Currently, it will return the reference mountpoint:
```python
In [6]: import qiita_db as qdb

In [7]: _, basefp = qdb.util.get_mountpoint('reference')[0]

In [8]: rel_path = ''

In [9]: from os.path import join

In [10]: join(basefp, rel_path)
Out[10]: '/Users/jose/qiime_software/QiiTa/qiita_db/support_files/test_data/reference/'
```
Which I think reinforces the idea of having a specific test in which the tree is `Null`.
Furthermore, the approach that we used in cases where the value may not exist is:
```python
rel_path = qdb.sql_connection.TRN.execute_fetchindex()
if rel_path:
    rel_path = rel_path[0][0]
else:
    return None
```

josenavas:
1 comment

qiita_db/analysis.py
antgonza: In short, yes.

qiita_db/reference.py
antgonza: The test exist cause the Silva reference and a biom table that uses it exists. Basically, when we try to call it we are testing that functionality without being explicit. I can see your confusion and that this is what we want to do in the future but not what we are doing now. Adding a test so we can merge and then discuss offline.

Now about the return. I really don't like having 2 different returns in the same function (None, string) as this will require some changes to make it work with other sections of the code. For example, when we create the summary of the artifact we display the value of the tree. Having an empty string make way more sense,

josenavas:
:+1: @ElDeveloper @mortonjt able to review? Should be fast.

PR #1647, opened 2016-02-19
Bringing back the green tick (i.e. pass the tests again)

josenavas:
Depends on #1641, so review/merge that one first.

This PR fixes the tests in this branch so they're passing again.


josenavas:
The changes on this PR seems a lot, however, I have not done that many changes. The changes that I have done are:
1. I split up the file `qiita_pet/test/test_study_handlers.py` in multiple files under the `qiita_pet/handlers/study_handlers/tests` directory, which is where they should be and one test file per library file.
2. Remove unused code due to the refactoring done in this branch
3. Remove duplicated tests due to some previous merge conflict
4. Note which objects need still to be tested
5. Added -v to .travis.yml

qiita_db/artifact.py
```diff
@@ -376,8 +376,8 @@ def delete(cls, artifact_id):
                 raise qdb.exceptions.QiitaDBArtifactDeletionError(
                     artifact_id, "it has been submitted to VAMPS")
 
-            # Check if there is a job queued or running that will use/is using
-            # the artifact
+            # Check if there is a job queued, running. waiting or
```
antgonza: running. -> running,

qiita_pet/handlers/study_handlers/artifact.py
```diff
@@ -110,7 +110,7 @@ def get(self):
                 # an admin, the processed data is waiting to be approved and
                 # the qiita config requires processed data approval
                 buttons.append(btn_base % ('private', 'Approve artifact'))
-        elif status == 'private':
+        if status == 'private':
```
antgonza: Why this change?

qiita_db/test/test_study.py
```diff
@@ -585,9 +586,6 @@ def test_ebi_submission_status_setter(self):
             new.ebi_submission_status = "unknown"
 
     def test_retrieve_info(self):
```
antgonza: Do we really need this test? Basically, is doing the same that the one above but instead of checking that the data that was just inserted is there, you are testing against the one that the populate sql has ....

qiita_pet/handlers/study_handlers/tests/test_listing_handlers.py
```diff
@@ -0,0 +1,439 @@
...
+        obs = _build_single_study_info(Study(1), study_info, study_proc,
+                                       proc_samples)
+        self.assertItemsEqual(obs, self.single_exp)
+
+    def test_build_single_proc_data_info(self):
+        obs = _build_single_proc_data_info(4, '18S',
+                                           self.proc_data_exp[0]['samples'])
+        self.assertItemsEqual(obs, self.proc_data_exp[0])
+
+    def test_build_study_info(self):
```
antgonza: This is basically the same test than the one following with the exception of already created and from a new one, right?

antgonza:
Some minor comments, then :+1:

qiita_db/artifact.py

qiita_db/test/test_study.py
josenavas: Done

qiita_pet/handlers/study_handlers/artifact.py
josenavas: The logic was wrong, (and tests where failing), so I fixed it :D

qiita_pet/handlers/study_handlers/tests/test_listing_handlers.py
josenavas: Yeah, I moved the assertRaises to the test below and removed this tests, as it is executing the same code.

josenavas: I actually decided to do it the other way around - I removed the test below b/c they're doing the same thing, but his one has the assertRaises and does not modify the DB, which allow us to save un DB reset

josenavas:
I also split up the tests in read/only and read/write

josenavas:
Looks like I'm not gonna be able to reduce the running time of the tests here w/o blowing up the size of the PR. @ElDeveloper can you review/merge this one? I have another PR in (https://github.com/biocore/qiita/pull/1667) that is doing some work reducing the running time.

qiita_db/test/test_study.py
```diff
@@ -147,11 +145,11 @@ def setUp(self):
             'metadata_complete': True,
             'reprocess': False,
             'number_samples_promised': 27,
-            'emp_person_id': qdb.study.StudyPerson(2),
+            'emp_person_id': 2,
```
squirrelo: not part of tests, but this should be renamed to emp_person since it's no longer the ID.

qiita_db/test/test_study.py
```diff
@@ -611,8 +606,9 @@ def test_set_info(self):
         self.infoexp["spatial_series"] = None
         self.infoexp["most_recent_contact"] = None
         self.infoexp["reprocess"] = False
-        self.infoexp["lab_person_id"] = 2
         self.infoexp["first_contact"] = datetime(2014, 6, 11)
+        self.infoexp["lab_person"] = qdb.study.StudyPerson(2)
```
squirrelo: Why this two step process instead of just changing the key to lab_person on line 594?

qiita_pet/handlers/api_proxy/tests/test_artifact.py
```diff
@@ -97,8 +100,9 @@ def test_artifact_get_req(self):
     def test_artifact_post_req(self):
         # Create new prep template to attach artifact to
         new_prep_id = get_count('qiita.prep_template') + 1
-        PrepTemplate.create(pd.DataFrame(
-            {'new_col': {'1.SKD6.640190': 1}}), Study(1), '16S')
+        npt.assert_warns(
```
squirrelo: This should also test the warning mesage is correct.

qiita_pet/handlers/api_proxy/tests/test_artifact.py
```diff
@@ -115,8 +119,9 @@ def test_artifact_post_req(self):
     def test_artifact_post_req_bad_file(self):
         # Create new prep template to attach artifact to
         new_prep_id = get_count('qiita.prep_template') + 1
-        PrepTemplate.create(pd.DataFrame(
-            {'new_col': {'1.SKD6.640190': 1}}), Study(1), '16S')
+        npt.assert_warns(
```
squirrelo: same here

qiita_pet/handlers/api_proxy/tests/test_prep_template.py
```diff
@@ -187,7 +189,11 @@ def test_prep_template_post_req(self):
                           'LinkerPrimerSequence, BarcodeSequence',
                'file': 'update.txt',
                'id': new_id}
-        self.assertEqual(obs, exp)
+        self.assertEqual(obs['status'], exp['status'])
+        self.assertItemsEqual(obs['message'].split('\n'),
```
squirrelo: Why the split?

qiita_pet/handlers/api_proxy/tests/test_prep_template.py
```diff
@@ -232,7 +238,10 @@ def test_prep_template_put_req_warning(self):
                           'differences between the data stored in the DB and '
                           'the new data provided',
                'file': 'update.txt'}
-        self.assertEqual(obs, exp)
+        self.assertEqual(obs['status'], exp['status'])
+        self.assertItemsEqual(obs['message'].split('\n'),
```
squirrelo: Same question here.

qiita_pet/handlers/study_handlers/tests/test_artifact.py
```diff
@@ -0,0 +1,142 @@
...
+    def tearDown(self):
+        # Replace file if removed as part of function testing
+        uploads_path = get_mountpoint('uploads')[0][1]
+        fp = join(uploads_path, '1', 'uploaded_file.txt')
+        if not exists(fp):
+            with open(fp, 'w') as f:
+                f.write('')
+
+    def test_post_artifact(self):
+        new_artifact_id = get_count('qiita.artifact') + 1
```
squirrelo: If we are moving towards the rollback test framework instead of the current wipe and rebuild the database, this will no longer work. You need to get the next ID in the series attached to the id column, as the IDs still increase regardless of the rollback.

qiita_pet/handlers/study_handlers/tests/test_description_handlers.py
```diff
@@ -0,0 +1,18 @@
...
+#
+# The full license is in the file LICENSE, distributed with this software.
+# -----------------------------------------------------------------------------
+from unittest import main
+
+from qiita_pet.test.tornado_test_base import TestHandlerBase
+
+
+class PreprocessingSummaryHandlerTests(TestHandlerBase):
+    # TODO: Missing tests
```
squirrelo: Is there an issue number for this?

qiita_pet/handlers/study_handlers/tests/test_processing.py
```diff
@@ -0,0 +1,28 @@
...
+#
+# The full license is in the file LICENSE, distributed with this software.
+# -----------------------------------------------------------------------------
+from unittest import main
+
+from qiita_pet.test.tornado_test_base import TestHandlerBase
+
+
+class ProcessArtifactHandler(TestHandlerBase):
+    # TODO: Missing tests
```
squirrelo: Issues for this and the other TODOs in this file?

qiita_pet/handlers/study_handlers/tests/test_vamps_handlers.py
```diff
@@ -0,0 +1,18 @@
...
+#
+# The full license is in the file LICENSE, distributed with this software.
+# -----------------------------------------------------------------------------
+from unittest import main
+
+from qiita_pet.test.tornado_test_base import TestHandlerBase
+
+
+class VAMPSHandlerTests(TestHandlerBase):
+    # TODO: Missing tests
```
squirrelo: Issue for this?

squirrelo:
Some comments

mortonjt:
Noticing that there are quite a few TODOs.  Those aren't being covered in this PR correct?

I don't see any obvious issues aside from what @antgonza and @squirrelo .  Perhaps @ElDeveloper should take another look over this.

qiita_db/test/test_study.py
josenavas: That's incorrect - note how it is actually the id, the ones that are no longer the id are the principal_investigator and the lab_person

qiita_db/test/test_study.py
josenavas: Note that line 594 and this are not the same variable.

qiita_db/test/test_study.py
squirrelo: Went is this still the id but the rest are not? We should be consistent.

qiita_pet/handlers/api_proxy/tests/test_artifact.py
josenavas: Not correct, we are not interested in the warning message thrown here. the only reason why this is wrapped in the npt.assert_warns is because we don't want the warning message to show up on the test suite.

qiita_pet/handlers/api_proxy/tests/test_artifact.py
josenavas: Same here

qiita_pet/handlers/api_proxy/tests/test_prep_template.py
josenavas: If you pay attention to the code that generates this string, it is a result of joining the strings stored in a set. Since it is in a set, it is unordered, so we cannot ensure the order of the strings here. By doing the split and assertItemsEqual I can compare that the contents are the same.

qiita_pet/handlers/api_proxy/tests/test_prep_template.py
josenavas: Same answer

qiita_pet/handlers/study_handlers/tests/test_artifact.py
josenavas: We will take care whenever we do the refactor, as that will include a lot of other changes.

qiita_pet/handlers/study_handlers/tests/test_description_handlers.py
josenavas: No, I just noted here because when I was splitting the file I realized that there are no tests for this function. Also, I think that there are some of these classes (like this one) that will disappear so I didn't want to spend time writing tests that are going to disappear in the next day.

qiita_pet/handlers/study_handlers/tests/test_processing.py
josenavas: Same answer as above - want first to be able to pass the tests and then will finish later - note that this is going to a branch. Will lis them in the todo list.

qiita_pet/handlers/study_handlers/tests/test_vamps_handlers.py
josenavas: Same

josenavas:
Comments answered.

josenavas:
@mortonjt yes, the TODO's will be addressed in subsequent PR. The problem that I was facing here is that I can't do more development reliable w/o having the tests passing. This and my next PR #1667 are making the tests to pass again.

qiita_db/test/test_study.py
josenavas: @antgonza @ElDeveloper @mortonjt `emp_person` is not used anywhere in the code and I think we should drop it as that is something that we will be keeping track in Jira, not Qiita. Should we spend time fixing that or just agree that we are going to drop it?

josenavas: Added an issue to discuss that: https://github.com/biocore/qiita/issues/1668

antgonza:
@josenavas thanks for working on this. @squirrelo thanks for quick reviews. Merging now.

PR #1648, opened 2016-02-19
WIP: Portals subdirectory

josenavas:
This branch holds the WIP to change the portals to work on subdirectories rather than in subdomain.

Remaining things to check:
 - [x] Uploader - when you upload files, it is trying to do it against the main site, rather than test
 - [x] Analysis - in the analysis pipeline, when moi creates the job group, something is acting up as it only creates the first job in the analysis (the creation one). All the links in moi look correct so I'm unsure where the source of the issue is in here.
 - [x] Once the study is created, the link that shows with "study BLA created" is linking to the main site rather than the portal.


qiita_core/configuration_manager.py
```diff
@@ -162,6 +162,9 @@ def _get_main(self, config):
         self.max_upload_size = config.getint('main', 'MAX_UPLOAD_SIZE')
         self.require_approval = config.getboolean('main', 'REQUIRE_APPROVAL')
         self.portal = config.get('main', 'PORTAL')
```
squirrelo: Both portal and portal_dir are missing from the docstring for the class.

qiita_core/support_files/config_test.cfg
```diff
@@ -42,6 +42,9 @@ VALID_UPLOAD_EXTENSION = fastq,fastq.gz,txt,tsv,sff,fna,qual
 # Portal the site is working under
```
squirrelo: Since these have to do with portals, shouldn't these be under the portal section of the config file?

squirrelo:
:+1:  for the config file updates

antgonza:
:+1:

PR #1653, opened 2016-02-19
Fix metadata file retrieval

josenavas:
While working on the test environment a found a bug on how the sample template files are being retrieved. This PR fixes the issue by re-using the function in util and removing a lot of duplicated code :D


antgonza:
:+1:

qiita_db/util.py
```diff
@@ -656,6 +656,9 @@ def retrieve_filepaths(obj_fp_table, obj_id_column, obj_id):
         The name of the column that represents the object id
     obj_id : int
         The object id
+    sort : {'ascendent', 'descendent'}, optional
+        The direction in which the results are sorted, using the filepath id
+        as sorting key. Default: None, no sorting is applyed
```
mortonjt: applyed -> applied

qiita_db/util.py
```diff
@@ -656,6 +656,9 @@ def retrieve_filepaths(obj_fp_table, obj_id_column, obj_id):
         The name of the column that represents the object id
     obj_id : int
         The object id
+    sort : {'ascendent', 'descendent'}, optional
```
mortonjt: Rather have  `{'ascendent', 'descendent'}` or `{'ascending', 'descending'}`?

[pandas](http://pandas.pydata.org/pandas-docs/stable/generated/pandas.DataFrame.sort.html) uses the later.

qiita_db/metadata_template/prep_template.py
```diff
@@ -610,20 +603,11 @@ def qiime_map_fp(self):
...
-            # We know that the good filepath is the one in the first row
-            # because we sorted them in the SQL query
-            fn = qdb.sql_connection.TRN.execute_fetchindex()[0][1]
-            base_dir = qdb.util.get_mountpoint('templates')[0][1]
-            return join(base_dir, fn)
+        for _, fp, fp_type in qdb.util.retrieve_filepaths(
+                self._filepath_table, self._id_column, self.id,
+                sort='descendent'):
+            if fp_type == 'qiime_map':
+                return fp
```
mortonjt: If there is no `'qiime_map'`, this returns `None`.  Does this warrant a test case?

mortonjt:
Few comments.  :+1: once they are addressed

qiita_db/util.py
squirrelo: This is for retrieving the children graph (descendents) or parents graph (ascendents), not sorting, so not sure that terminology fits here.

squirrelo: sorry, commenting on wrong PR, ignore above.

squirrelo:
N comments outside @mortonjt already commented on. :+1: when fixed.

qiita_db/metadata_template/prep_template.py
josenavas: That case should never happen (i.e. all prep template objects should have the qiime_map).
However, in case that it doesn't exist, I think returning None is the correct approach.

qiita_db/util.py
josenavas: Good call - I knew it sounded weird hahahaha - changing

qiita_db/util.py

josenavas:
Thanks for the reviews! Comments addressed

PR #1659, opened 2016-02-22
moving test to use pure python vs. sql

antgonza:


antgonza:
After qiita meeting discussion, closing for now ...

PR #1660, opened 2016-02-22
WIP: fixing multiple references in an analysis

antgonza:
This should pass all test and could be deployed in the main site. However, I'm adding as WIP as I would like to discuss internally the decisions made here via email with other devs. Email should be out in a few minutes.


antgonza:
closing in favor of https://github.com/biocore/qiita/pull/1678

PR #1662, opened 2016-02-24
WIP: Missing links

josenavas:
Here I'm adding all the remaining links that are missing.
We are discovering this while we are testing the interface in the new test environment - so hold on merging until we discover all the issues.

Remaining things to check:
 - [ ] Uploader - when you upload files, it is trying to do it against the main site, rather than test
 - [ ] Analysis - in the analysis pipeline, when moi creates the job group, something is acting up as it only creates the first job in the analysis (the creation one). All the links in moi look correct so I'm unsure where the source of the issue is in here.
 - [ ] Once the study is created, the link that shows with "study BLA created" is linking to the main site rather than the portal.


ElDeveloper:
:+1:

PR #1663, opened 2016-02-25
WIP: split travis in multiple threads

antgonza:


antgonza:
Closing this test as it's not easy to combine the coverall results: http://technology.indiegogo.com/2015/08/how-we-get-coverage-on-parallelized-test-builds/ and we are not really gaining much speedup.

PR #1666, opened 2016-02-26
working on DB tests

antgonza:


josenavas:
@antgonza coverage decreased here so I guess that there are tests that should not have been removed - I'm going to take a look

antgonza:
After qiita meeting discussion closing for now ...

PR #1667, opened 2016-02-26
Reduce tests running time for qiita_db

josenavas:
Built on top of: https://github.com/biocore/qiita/pull/1647 so review/merge that one first.

Here I split the tests of processing job, software and artifact between read/write and read/only, as well as grouping tests that are testing the same thing. This reduces the amount of times the database needs to be reset. Times:
- Artifact:  From 1m59.500s to 0m32.497s
- Processing job: From 1m34.795s to 0m30.331s
- Software: From 1m34.751s to 0m7.590s

### DIFF
--- a/qiita_db/handlers/tests/test_processing_job.py
+++ b/qiita_db/handlers/tests/test_processing_job.py
@@ -159,6 +159,7 @@ class CompleteHandlerTests(OauthTestingBase):
         super(CompleteHandlerTests, self).setUp()
 
     def tearDown(self):
+        super(CompleteHandlerTests, self).tearDown()
         for fp in self._clean_up_files:
             if exists(fp):
                 remove(fp)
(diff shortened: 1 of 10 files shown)


josenavas:
That's interesting, reducing the running time of the tests in my machine by ~4 min is equivalent to ~30 min in Travis? Wow - I saw that there are a couple of failures, I'm gonna fix them here.

antgonza:
I can confirm this behavior from my other improving-tests tests. However, I can't explain why.

josenavas:
This is ready for review. The main changes here are splitting the tests between read-only and read-write. I've also fixed the last 3 tests that were failing, which where mainly due to the usage of sets which makes the output order undefined.

antgonza:
:+1:

ElDeveloper:
:+1:

PR #1671, opened 2016-03-02
fixing VAMPS submission and fix #756

### LINKED ISSUE #756: after submit to vamps return to study page
Currently after a submission you are returned to the main page but it will be nice to be returned to the page that originated the call. Perhaps we have the same behavior in EBI submissions.

antgonza:
:smile:

### DIFF
--- a/qiita_pet/handlers/study_handlers/vamps_handlers.py
+++ b/qiita_pet/handlers/study_handlers/vamps_handlers.py
@@ -12,9 +12,6 @@ from tornado.web import authenticated, HTTPError
 from qiita_ware.context import submit
 from qiita_ware.demux import stats as demux_stats
 from qiita_ware.dispatchable import submit_to_VAMPS
-from qiita_db.metadata_template.prep_template import PrepTemplate
-from qiita_db.metadata_template.sample_template import SampleTemplate
-from qiita_db.study import Study
 from qiita_db.exceptions import QiitaDBUnknownIDError
 from qiita_db.artifact import Artifact
 from qiita_pet.handlers.base_handlers import BaseHandler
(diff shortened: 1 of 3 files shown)


josenavas:
:+1: While doing some test refactoring in the artifact-study-pages branch I realized that there are no tests for this. Is there because this can't be tested? Not blocker for this PR, but is it possible to talk with the VAMPS developers so we can have a test submission site?

antgonza:
The submission is basically a FTP call and should be really easy to test: submit and check that the file is there ... added an issue for that https://github.com/biocore/qiita/issues/1672

josenavas:
Thanks @antgonza 
@ElDeveloper @mortonjt available for a quick review?

qiita_pet/handlers/study_handlers/vamps_handlers.py
```diff
@@ -36,15 +33,26 @@ def display_template(self, preprocessed_data_id, msg, msg_level):
...
+        if not allow_submission:
+            msg_list.append(
+                "Only artifacts with a single prep template can be submitted")
+        # If allow_submission is already false, we technically don't need to
+        # do the following work. However, there is no clean way to fix this
+        # using the current structure, so we perform the work as we
+        # did so it doesn't fail.
+        # We currently support only one prep template for submission, so
+        # grabbing the first one
+        prep_template = prep_templates[0]
```
ElDeveloper: Could it be that at this point, the number of prep templates is zero and that this would throw an `IndexError`?

qiita_pet/handlers/study_handlers/vamps_handlers.py
```diff
@@ -61,10 +69,21 @@ def display_template(self, preprocessed_data_id, msg, msg_level):
             stats.append(('Number of sequences', demux_file_stats.n))
             msg_level = 'success'
 
+        # In EBI here we check that we have the required field for submission,
+        # however for VAMPS we don't need that
+
```
ElDeveloper: Was this line accidental or meant to be here, highlighting just in case.

qiita_pet/templates/vamps_submission.html
```diff
@@ -1,5 +1,6 @@
 {% extends sitebase.html %}
 {% block content %}
+{% from future.utils import viewitems %}
```
ElDeveloper: I don't think this is being used anywhere.

ElDeveloper:
:+1: looks good, just a few minor comments.

qiita_pet/handlers/study_handlers/vamps_handlers.py
antgonza: Theoretically it could but in practice it can't: the only way to get to this point is by clicking on a button that is only shown when you see the summary of the prep template ...

qiita_pet/handlers/study_handlers/vamps_handlers.py
antgonza: Meant to be here.

qiita_pet/handlers/study_handlers/vamps_handlers.py
ElDeveloper: Noted, thanks!

qiita_pet/handlers/study_handlers/vamps_handlers.py
ElDeveloper: thanks!

PR #1677, opened 2016-03-08
Add to jobs the input file reference id and software command id

antgonza:
These changes are needed so we can continue development on https://github.com/biocore/qiita/pull/1660. The plan is to close that PR and open a new one against the artifact-study-pages branch so we can have these DB changes.

I will submit my next PR: merge and solving https://github.com/biocore/qiita/pull/1660 soon. Obviosly on top of this one.

### DIFF
--- a/qiita_db/job.py
+++ b/qiita_db/job.py
@@ -79,6 +79,7 @@ class Job(qdb.base.QiitaStatusObject):
 
     @classmethod
     def exists(cls, datatype, command, options, analysis,
+               input_file_reference_id, input_file_software_command_id,
                return_existing=False):
         """Checks if the given job already exists
 
@@ -92,6 +93,10 @@ class Job(qdb.base.QiitaStatusObject):
             Options for the command in the format {option: value}
         analysis : Analysis object
             The analysis the job will be attached to on creation
+        input_file_reference_id : int
+            The refernce id used to create the input file
+        input_file_software_command_id : int
+            The software command id used to create the input file
         return_existing : bool, optional
             If True, function will return the instatiated Job object for the
             matching job. Default False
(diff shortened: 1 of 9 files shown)


qiita_db/support_files/patches/39.sql
```diff
@@ -0,0 +1,24 @@
...
+
+ALTER TABLE qiita.job ADD CONSTRAINT fk_job_reference FOREIGN KEY ( input_file_reference_id ) REFERENCES qiita.reference( reference_id );
+
+ALTER TABLE qiita.job ADD CONSTRAINT fk_job_software_command FOREIGN KEY ( input_file_software_command_id ) REFERENCES qiita.software_command( command_id );
+
+-- Change values:
+-- input_file_reference_id can be = 1 as it's only needed for job processing
+-- input_file_software_command_id = 3 as it's close reference picking.
+
+UPDATE qiita.job SET input_file_reference_id = 1;
```
josenavas: Is this 100% safe? In the main deploy we have more than one reference so I just want to make sure that setting this always to 1 is not going to bite us later... (e.g. when we moved this to artifacts and plugins)

josenavas:
Just one question, otherwise :+1:

qiita_db/support_files/patches/39.sql
antgonza: I believe it is cause the jobs already ran and the only way they can be successful is if this is 1.

josenavas: Good point

qiita_db/job.py
```diff
@@ -92,6 +93,10 @@ def exists(cls, datatype, command, options, analysis,
             Options for the command in the format {option: value}
         analysis : Analysis object
             The analysis the job will be attached to on creation
+        input_file_reference_id : int
```
squirrelo: Normally these are the string versions, with ID conversion called in the function (see datatype and command in this function). Why passing the ID directly here?

qiita_db/job.py
```diff
@@ -224,6 +235,10 @@ def create(cls, datatype, command, options, analysis,
             The name of the command executed in this job
         analysis : Analysis object
             The analysis which this job belongs to
+        input_file_reference_id : int
```
squirrelo: same question

qiita_db/job.py
antgonza: Because it's easier and I didn't see any strong evidence that doing string or object arguments is better, at least in this case ... BTW it also reduces SQL calls.

squirrelo: We've already standardized on passing in the string though, so we should stick with previous convention and use a standardized way of calling things instead of a mixed string/id situation as above.

josenavas:
:+1: @squirrelo can you merge if you agree with the changes?

PR #1679, opened 2016-03-09
Refactoring the refactor (I): Fix prep info creation

josenavas:
In the previous implementation, the addition of the prep info and the first artifact was forced to occur at the same point. We used that strategy at the beginning of Qiita and we changed it later because that did not follow the workflow.

At the same time, if the artifact creation fail for whatever reason, the entire page was broke and the user was not able to see the study page again.

Here, the addition of the artifact is removed from the addition of the prep information.
At this point, artifacts can't be added, but that will be added in the next PR, to keep them small.
Thus, the interface is broken at this point BUT the tests should pass.


qiita_pet/handlers/api_proxy/prep_template.py
```diff
@@ -14,12 +14,45 @@
...
+    """
+    Parameters
+    ----------
+    study_id : int
+        The study id
+
+    Returns
+    -------
+    (list of str, list of str, dict of {str: list of str})
+        The list of txt,csv files in the upload dir for the given study
```
antgonza: csv is not in the list below.

antgonza:
One minor comment, then :+1: @squirrelo or @ElDeveloper could you review?

qiita_pet/handlers/api_proxy/prep_template.py
```diff
@@ -14,12 +14,45 @@
...
 from qiita_db.metadata_template.util import load_template_to_dataframe
-from qiita_db.util import convert_to_id
+from qiita_db.util import convert_to_id, get_files_from_uploads_folders
 from qiita_db.study import Study
 from qiita_db.ontology import Ontology
 from qiita_db.metadata_template.prep_template import PrepTemplate
 
 
+def new_prep_template_get_req(study_id):
+    """
```
ElDeveloper: Can you add a description to this function?

qiita_pet/handlers/api_proxy/prep_template.py
ElDeveloper: Yeah, same comment, plus `txt,csv` -> `txt, csv`

qiita_pet/handlers/api_proxy/prep_template.py
```diff
@@ -14,12 +14,45 @@
...
+
+    Returns
+    -------
+    (list of str, list of str, dict of {str: list of str})
+        The list of txt,csv files in the upload dir for the given study
+        The list of available data types
+        The investigation type ontology information
+    """
+    prep_files = [f for _, f in get_files_from_uploads_folders(study_id)
+                  if f.endswith(('txt', 'tsv'))]
```
ElDeveloper: I would suggest "anchoring" the extension to the dot i.e. `'.txt', '.tsv'`, though I imagine it is very unlikely to find an occurrence where the file ends in txt and txt is not an extension?

qiita_pet/templates/study_ajax/add_prep_template.html
```diff
@@ -0,0 +1,119 @@
+<script type="text/javascript" src="/static/vendor/js/jquery.validate.min.js"></script>
+<script type="text/javascript>">
+$(document).ready(function () {
+  // Hide the user and new ontology divs
+  $("#user-ena-info").hide();
+  $("#new-ena-info").hide();
+  // If the ena-ontology select changes, check if we need to show the user list
+  $("#ena-ontology").change(function ( event ) {
+      if( $("#ena-ontology").val() == "Other" ) { $("#user-ena-info").show(); }
```
ElDeveloper: This block is over indented for two spaces. Also, any reason why the whole thing is in a line? Most notably for the else block? same comment applies to the rest of the file, would make more sense to have them in their own line.

qiita_pet/templates/study_base.html
```diff
@@ -192,8 +155,61 @@
...
   }
 
+  /**
+   *
+   * Function to populate the main div of the screen
+   *
+   * @param url string with the URL to perform the AJAX get call
+   * @param parameters the parameters for the AJAX call
+   *
+   * This function ejecutes an AJAX get against the URL provided with the given
```
ElDeveloper: ejecutes -> executes

qiita_pet/templates/study_base.html
```diff
@@ -207,56 +223,34 @@
...
-    <a href="#" onclick="fill_sample({{study_info['study_id']}}); return false;">Sample Template</a><br />
-    <a href="#" onclick="fill_sample_summary({{study_info['study_id']}}); return false;">Sample Summary</a><br />
+  <h3><a href="#" onclick="populate_main_div('/study/description/baseinfo/', { study_id: {{study_info['study_id']}} })">Study Information</a></h3>
+    <a href="#" onclick="populate_main_div('/study/description/sample_template/', { study_id: {{study_info['study_id']}} })">Sample Template</a><br />
+    <a href="#" onclick="populate_main_div('/study/description/sample_summary/', { study_id: {{study_info['study_id']}} })">Sample Summary</a><br />
     {% if editable %}
     <a href="/study/upload/{{study_info['study_id']}}">Upload Files</a><br />
-    <a href="#" onclick="fill_new_prep()">Add New Preparation</a><br />
+    <a href="#" onclick="populate_main_div('/study/new_prep_template/', { study_id: {{study_info['study_id']}} })">Add New Preparation</a><br />
+    <!-- <a href="#" onclick="populate_main_div('/study/add_prep/{{study_info['study_id']}}')">Add New Preparation</a><br /> -->
```
ElDeveloper: Can we remove this line?

ElDeveloper:
:+1: Just a few comments, looking good overall!

qiita_pet/handlers/api_proxy/prep_template.py
josenavas: I think this is actually a typo in the documentation. We are supporting tsv (Tab Separated Values) rather than csv (Comma Separated Values).

qiita_pet/handlers/api_proxy/prep_template.py
josenavas: Done

qiita_pet/handlers/api_proxy/prep_template.py
josenavas: Done

qiita_pet/templates/study_ajax/add_prep_template.html
josenavas: No specific reason, changing - thanks for catching the indentation

qiita_pet/templates/study_base.html
josenavas: Done

qiita_pet/templates/study_base.html
josenavas: Done

josenavas:
Comments addressed

ElDeveloper:
:+1:

PR #1680, opened 2016-03-09
Refactoring the refactor (II): Add artifact creation page

josenavas:
This depends on #1679, so review/merge that one first

This PR adds the page and code needed for adding an artifact to the system. At this point, this page is not integrated with the rest of the system in order to keep the PR small. It will be integrated in the next PR, which I'm almost done.

The tests, however, should pass.

The previous implementation was buggy and it did not allow to upload an artifact with more than a single lane or if it was missing the reverse reads. The user feedback has been also improved by performing the files checks on the fly rather when it clicks the create button.


josenavas:
This is ready for review

qiita_pet/handlers/api_proxy/artifact.py
```diff
@@ -73,8 +76,9 @@ def artifact_post_req(user_id, filepaths, artifact_type, name,
     ----------
     user_id : str
         User adding the atrifact
-    filepaths : dict of {str: [str, ...], ...}
-        List of files to attach to the artifact, keyed to the file type
+    filepaths : dict of {str: str}
```
antgonza: Not sure the description and the example match.

qiita_pet/handlers/api_proxy/artifact.py
```diff
@@ -92,23 +96,46 @@ def artifact_post_req(user_id, filepaths, artifact_type, name,
...
-                return {'status': 'error',
-                        'message': 'File does not exist: %s' % fp}
-            cleaned_filepaths.append((full_fp, ftype))
-
-    artifact = Artifact.create(cleaned_filepaths, artifact_type, name=name,
-                               prep_template=prep)
+    for ftype, file_list in viewitems(filepaths):
+        # JavaScript sends us this list as a comma-separated list
+        for fp in file_list.split(','):
+            # If the list was initially empty, the elemnt fp will the empty
```
antgonza: element

qiita_pet/handlers/api_proxy/artifact.py
```diff
@@ -92,23 +96,46 @@ def artifact_post_req(user_id, filepaths, artifact_type, name,
...
+            # string, which will make the check below for "exists" to check
```
antgonza: can you rewrite? I don't think it makes a lot of sense.

qiita_pet/handlers/api_proxy/studies.py
```diff
@@ -168,3 +171,72 @@ def study_prep_get_req(study_id, user_id):
...
+    study_id : int
+        The study id
+    prep_template_id : int
+        The prep template id
+    artifact_type : str
+        The artifact type
+
+    Returns
+    -------
+    dict of {str: object}
```
antgonza: {str: object} doesn't really match the below description.

qiita_pet/templates/study_ajax/add_artifact.html
```diff
@@ -0,0 +1,112 @@
...
+        data: $("#create-artifact-form").serialize(),
+        success: function(data) {
+          if(data.status == 'error') {
+            bootstrapAlert(data.message, "danger");
+          }
+          else {
+            if(data.status == 'warning') {
+              bootstrapAlert(data.message, "warning");
+            }
+            // This reloads needs to be changed to:
```
antgonza: Just want to confirm when this section is gonna be updated, this or which PR?

antgonza:
Some comments.

qiita_pet/handlers/api_proxy/artifact.py
josenavas: Good catch - done

qiita_pet/handlers/api_proxy/artifact.py
josenavas: Done

qiita_pet/handlers/api_proxy/studies.py
josenavas: The returned dictionary is keyed by strings, but the values to those keys can be either strings, int, or lists. In order to englobe all these types in the type description, I used the common base class `object`. Do you have a better alternative?

qiita_pet/templates/study_ajax/add_artifact.html
josenavas: Following PR

qiita_pet/handlers/api_proxy/artifact.py
```diff
@@ -73,8 +76,9 @@ def artifact_post_req(user_id, filepaths, artifact_type, name,
     ----------
     user_id : str
         User adding the atrifact
-    filepaths : dict of {str: [str, ...], ...}
-        List of files to attach to the artifact, keyed to the file type
+    filepaths : str
```
ElDeveloper: This description doesn't match "my inferred type", it seems like it really is a dictionary, not a string, am I missing something?

qiita_pet/handlers/api_proxy/artifact.py
```diff
@@ -92,23 +96,48 @@ def artifact_post_req(user_id, filepaths, artifact_type, name,
...
+    # a explicit check, in case there is something odd with the JS
+    if not cleaned_filepaths:
+        return {'status': 'error',
+                'message': "Can't create artifact, no files provided."}
+
+    try:
+        artifact = Artifact.create(cleaned_filepaths, artifact_type, name=name,
+                                   prep_template=prep)
+    except Exception as e:
+        # We should hit this exception rarelly (that's why is an exception)
```
ElDeveloper: rarelly -> rarely
why is -> why it is

qiita_pet/handlers/api_proxy/artifact.py
```diff
@@ -92,23 +96,48 @@ def artifact_post_req(user_id, filepaths, artifact_type, name,
...
+        # since at this point we have done multiple checks. However, it can
+        # occur in weird cases, so better let the GUI know that this fail
```
ElDeveloper: fail -> failed

qiita_pet/handlers/api_proxy/artifact.py
```diff
@@ -92,23 +96,48 @@ def artifact_post_req(user_id, filepaths, artifact_type, name,
...
-    artifact = Artifact.create(cleaned_filepaths, artifact_type, name=name,
-                               prep_template=prep)
+    for ftype, file_list in viewitems(filepaths):
+        # JavaScript sends us this list as a comma-separated list
+        for fp in file_list.split(','):
+            # JavaScript will send this value as an empty string if the
+            # list of files was empty. In such case, the split will generate
+            # a single element containing the empty string. Check for that case
+            # here and, if fp is not the empty string, proceed to check if
+            # the file exists
```
ElDeveloper: Thanks for the description here, very useful. Just to check, there's no possible way to get commas in these filenames right? I'm wondering if what we should be getting should be a JSON object, but that I guess gets more in the topic of the REST API etc.

qiita_pet/handlers/api_proxy/studies.py
```diff
@@ -168,3 +171,72 @@ def study_prep_get_req(study_id, user_id):
...
+        for _, filename in uploaded:
+            if filename.startswith(prep_prefixes):
+                selected.append(filename)
+            else:
+                remaining.append(filename)
+    else:
+        num_prefixes = 0
+        remaining = [f for _, f in uploaded]
+
+    # At this point we can't make anything smart about selecting by default
```
ElDeveloper: make -> do

qiita_pet/handlers/api_proxy/studies.py
```diff
@@ -168,3 +171,72 @@ def study_prep_get_req(study_id, user_id):
...
+    supp_file_types = supported_filepath_types(artifact_type)
+    selected = []
+    remaining = []
+
+    uploaded = get_files_from_uploads_folders(study_id)
+    pt = PrepTemplate(prep_template_id).to_dataframe()
+
+    if (any(ft.startswith('raw_') for ft, _ in supp_file_types) and
+            'run_prefix' in pt.columns):
+        prep_prefixes = tuple(set(pt['run_prefix']))
```
ElDeveloper: What about `tuple(pt['run_prefixes'].unique())`? Embracing pandas, not sure it's faster or anything, but the intent is clearer.

qiita_pet/templates/study_ajax/add_artifact.html
```diff
@@ -0,0 +1,112 @@
+<script type="text/javascript">
+  // Global counter used to check if the form can be submitted or not
+  var error_count = 0;
```
ElDeveloper: Is there any way we can try an avoid adding global variables? I know it sometimes hard, but this is a slippery slope and it has the potential of getting out of control.

qiita_pet/templates/study_ajax/artifact_file_selector.html
```diff
@@ -0,0 +1,186 @@
...
+   * Function that validates all the files lists
+   *
+   * This function triggers the sortupdate event in all the elements
+   * that are "checkeable"
+   */
+  function check_files_lists() {
+      // Reset the error counter
+      error_count = 0;
+      // initialize the error count to the number of file types
+      $(".checkeable").trigger("sortupdate");
```
ElDeveloper: I just picked this up, but it seems a bit odd that the name of the class is `checkeable`, are you sure it's not `checkable`?

qiita_pet/templates/study_ajax/artifact_file_selector.html
```diff
@@ -0,0 +1,186 @@
...
+   *
+   * This function checks that the contents of the sortable list is correct,
+   * taking into account if the lsit is required and the number of elements
+   * in the list.
+   *
+   */
+  function on_change_validation() {
+    var num_children = this.children.length;
+    var is_required = $(this).attr("data-file-required") === "True";
+    var req_count = +$(this).attr("data-file-count");
```
ElDeveloper: What's the plus sign here for `+$`?

ElDeveloper:
Just a few comments.

qiita_pet/handlers/api_proxy/artifact.py
josenavas: :beer: you're right

qiita_pet/handlers/api_proxy/artifact.py
josenavas: The serialization of that object happens in the HTML serialization of the form. I'm unsure why javascript (or HTML) is not serializing that list as a JSON object and self.argument is not unwrapping it like that, but that is why I'm receiving in the handler. Agree that we should totally revisit for a REST API

qiita_pet/handlers/api_proxy/artifact.py
josenavas: Done

qiita_pet/handlers/api_proxy/artifact.py
josenavas: Done

qiita_pet/handlers/api_proxy/studies.py
josenavas: Pandas is ~2X slower so leaving as it is:
```
In [18]: %timeit tuple(pt['run_prefix'].unique())
10000 loops, best of 3: 19.3 µs per loop

In [19]: %timeit tuple(set(pt['run_prefix']))
100000 loops, best of 3: 9.15 µs per loop
```

qiita_pet/handlers/api_proxy/studies.py
josenavas: Done

qiita_pet/templates/study_ajax/add_artifact.html
josenavas: donezzo - some JS magic and done :P

qiita_pet/templates/study_ajax/artifact_file_selector.html
josenavas: Is a class that I made up, so yeah, probably the correct work is `checkable`.

qiita_pet/templates/study_ajax/artifact_file_selector.html
josenavas: explicit cast to int

qiita_pet/handlers/api_proxy/studies.py
antgonza: What about: "see below"? No hard feelings so fine to leave it as is.

antgonza:
:+1: @ElDeveloper can you merge?

qiita_pet/templates/study_ajax/artifact_file_selector.html
ElDeveloper: Thanks, I keep forgetting this.

qiita_pet/handlers/api_proxy/studies.py
ElDeveloper: :+1: :hourglass_flowing_sand:

PR #1683, opened 2016-03-11
Refactoring the refactor (III): prep summary

josenavas:
Depends on #1680, so review/merge that one first

This PR refactors the prep information summary page and integrates the artifact creation in this page.


josenavas:
Ready for review!

qiita_pet/templates/study_ajax/prep_summary.html
```diff
@@ -1,115 +1,401 @@
...
+</div>
+
+<!-- Update prep template -->
+<div class="row">
+  <div class="col-md-12">
+    <b>Update information:</b>
+    <select id="file-selector">
+      <option value="">Choose file...</option>
+      {% for fp in files %}
+        <option value="{{fp}}">{{fp}}</option>
```
antgonza: Will files with special chars be a problem?

qiita_pet/templates/study_ajax/prep_summary.html
```diff
@@ -1,115 +1,401 @@
...
+
+<!-- Investigation type info -->
+<div class="row">
+  <!-- Ena ontology selector -->
+  <div class="col-md-4">
+    <b>Select Investigation Type:</b>
+    <select id="ena-ontology" name="ena-ontology" value="ena-ontology">
+      <option value=""></option>
+      {% for o in ontology['ENA'] %}
+        <option value="{{o}}">{{o}}</option>
```
antgonza: What about special chars?

qiita_pet/templates/study_ajax/prep_summary.html
```diff
@@ -1,115 +1,401 @@
...
+      {% end %}
+    </select>
+  </div>
+  <!-- user-defined selector -->
+  <div class="col-md-4" id="user-ena-info" hidden> User defined investigation type:
+    <select id="user-ontology" name="user-ontology" value="user-ontology">
+      <option value=""></option>
+      {% for o in ontology['User'] %}
+        <option value="{{o}}">{{o}}</option>
```
antgonza: Same here.

antgonza:
A few questions/comments.

qiita_pet/templates/study_ajax/prep_summary.html
josenavas: I think if we have files with special chars, it will break the upload page before breaking here, right? If not, do you have an idea on how can that be fixed? And which special character will make this to break?

qiita_pet/templates/study_ajax/prep_summary.html
josenavas: In this specific case there shouldn't be any special characters (is the ENA ontology)

qiita_pet/templates/study_ajax/prep_summary.html
josenavas: I guess we can put some validation on the user input, so only alphanumeric values and the space can be added as a user-defined term for the ENA ontology. Does this sound good to you?

qiita_pet/templates/study_ajax/prep_summary.html
antgonza: I think space will do it, which is allowed by the upload. We could change this behavior in the uploader. Wanna open an issue?

qiita_pet/templates/study_ajax/prep_summary.html
antgonza: K, thanks.

qiita_pet/templates/study_ajax/prep_summary.html
antgonza: Yup. Not sure if this is something we need everywhere.

qiita_pet/templates/study_ajax/prep_summary.html
josenavas: Just tested with a space - not an issue here - it works fine.

qiita_pet/templates/study_ajax/prep_summary.html
josenavas: I've added a check that they only contain alphanumeric and space characters

antgonza:
Thanks! :+1:

qiita_pet/handlers/api_proxy/prep_template.py
```diff
@@ -20,6 +21,24 @@
 from qiita_db.metadata_template.prep_template import PrepTemplate
 
 
+def _get_ENA_ontology():
+    """Returns the information of the ENA ontology
+
+    Returns
+    -------
+    dict of {str: list of strings}
+        A dictionary of the form {'ENA': list of str.], 'User': list of str}
```
ElDeveloper: There's an odd `]` here.

qiita_pet/handlers/api_proxy/prep_template.py
```diff
@@ -40,20 +59,81 @@ def new_prep_template_get_req(study_id):
...
+        if 'qiime' in basename(fp):
+            download_qiime.append(fp_id)
+        else:
+            download_prep.append(fp_id)
+    download_prep = download_prep[0]
+    download_qiime = download_qiime[0]
+
+    ontology = _get_ENA_ontology()
+
+    return {'status': 'success',
```
ElDeveloper: If the status is hard-coded as a return value, then do we need this at all?

qiita_pet/handlers/api_proxy/prep_template.py
```diff
@@ -40,20 +59,81 @@ def new_prep_template_get_req(study_id):
...
+        - artifact_attached: bool, whethe the template has an artifact attached
+        - study_id: int, the study id of the template
+    """
+    # Currently there is no name attribute, but it will be soon
+    name = "Prep information %d" % prep_id
+    pt = PrepTemplate(prep_id)
+    artifact_attached = pt.artifact is not None
+    study_id = pt.study_id
+    files = [f for _, f in get_files_from_uploads_folders(study_id)
+             if f.endswith(('txt', 'tsv'))]
```
ElDeveloper: Same comment as before about anchoring the txt and tsv extensions to .txt and .tsv

qiita_pet/handlers/api_proxy/prep_template.py
```diff
@@ -40,20 +59,81 @@ def new_prep_template_get_req(study_id):
...
+    Parameters
+    ----------
+    prep_id : int
+        The prep template id
+
+    Returns
+    -------
+    dict of {str: object}
+        A dictionary with the following keys:
+        - status: str, whether if the request is successful or not
```
ElDeveloper: Can you remove the `if` from this sentence?

qiita_pet/handlers/api_proxy/prep_template.py
```diff
@@ -40,20 +59,81 @@ def new_prep_template_get_req(study_id):
...
+        - message: str, if the request is unsuccessful, a human readable error
+        - name: str, the name of the prep template
+        - files: list of str, the files available to update the prep template
+        - download_prep: int, the filepath_id of the prep file
+        - download_qiime, int, the filepath_id of the qiime mapping file
+        - num_samples: int, the number of samples present in the template
+        - num_columns: int, the number of columns present in the template
+        - investigation_type: str, the investigation type of the template
+        - ontology: str, dict of {str, list of str} conataining the information
```
ElDeveloper: conataining -> containing

qiita_pet/handlers/api_proxy/prep_template.py
```diff
@@ -40,20 +59,81 @@ def new_prep_template_get_req(study_id):
...
+        of the ENA ontology
+        - artifact_attached: bool, whethe the template has an artifact attached
```
ElDeveloper: whethe -> where

qiita_pet/templates/study_ajax/prep_summary.html
```diff
@@ -1,115 +1,407 @@
...
+      data: {'op': 'replace', 'path': '/{{prep_id}}/data', 'value': fp},
+      success: function(data) {
+        if(data.status == 'error') {
+          bootstrapAlert(data.message, "danger");
+        }
+        else {
+          if(data.status == 'warning') {
+            bootstrapAlert(data.message.replace("\n", "<br/>"), "warning");
+          }
+          $("#file-selector option[value='" + fp +"']").remove();
```
ElDeveloper: Correct me if I'm wrong, but it is highly possible that this fp string will contain characters that are reserved for jQuery? Would it make sense to escape them? See this for an example of what I mean: https://learn.jquery.com/using-jquery-core/faq/how-do-i-select-an-element-by-an-id-that-has-characters-used-in-css-notation/

qiita_pet/templates/study_ajax/prep_summary.html
```diff
@@ -1,115 +1,407 @@
...
+   *
+   */
+  function load_template_graph() {
+    show_loading("graph-or-new-artifact-div");
+    $.get("/prep/graph/", {"prep_id": {{prep_id}} })
+      .done(function(data) {
+        var edges = [];
+        var nodes = [];
+
+        for(i=0; i<data.edge_list.length; i++) {
```
ElDeveloper: can you add var before `i` so `i` is not a global variable?

qiita_pet/templates/study_ajax/prep_summary.html
```diff
@@ -1,115 +1,407 @@
...
+          edges.push({from: data.edge_list[i][0], to: data.edge_list[i][1]});
+        }
+        // format nore depth data
+        for(i=0; i<data.node_labels.length; i++) {
```
ElDeveloper: Same as above.

qiita_pet/templates/study_ajax/prep_summary.html
```diff
@@ -1,115 +1,407 @@
...
+   * Load the artifact summary page
+   *
+   * @param artifact_id integer with the artifact id to load
+   *
+   * This function executes an AJAX call to populate the prep-artifact-summary-div
+   *
+   */
+  function load_artifact_summary(artifact_id) {
+    show_loading("prep-artifact-summary-div");
+    console.log("Load artifact summary " + artifact_id);
```
ElDeveloper: Can you remove this print?

qiita_pet/handlers/api_proxy/prep_template.py
josenavas: Done

qiita_pet/handlers/api_proxy/prep_template.py
josenavas: Done

qiita_pet/handlers/api_proxy/prep_template.py
josenavas: Done

qiita_pet/handlers/api_proxy/prep_template.py
josenavas: `whether` sorry for so many typos!

qiita_pet/handlers/api_proxy/prep_template.py
josenavas: Done

qiita_pet/handlers/api_proxy/prep_template.py
josenavas: For consistency with the other similar calls. I agree, we should remove it and I think it will go away once we start discussing the REST api in depth, but for now this is the best in terms of consistency.

qiita_pet/templates/study_ajax/prep_summary.html
josenavas: Probably is a bit hard to read, but the `fp` is actually contained in a javascript string, so I think it shouldn't affect and I didn't had any problem while I was testing this interface:

```JavaScript
$("#file-selector option[value='some_file_with.reserved.chars']").remove();
```

qiita_pet/templates/study_ajax/prep_summary.html
josenavas: Good call - done

qiita_pet/templates/study_ajax/prep_summary.html
josenavas: Done

qiita_pet/templates/study_ajax/prep_summary.html
josenavas: Done

qiita_pet/handlers/api_proxy/prep_template.py
ElDeveloper: Oh, ok, thanks!

qiita_pet/templates/study_ajax/prep_summary.html
ElDeveloper: Interesting, well, the problem for the most part is that if the string
in a query command in jQuery contains special characters, the query is
likely to fail, but there's no chance we'll see special characters here
then.

PR #1684, opened 2016-03-14
fix patch 37

antgonza:
While working on the current PRs we realized that patch 37 had an error in FASTA input, fixing this and some minor typos.


josenavas:
:+1:

josenavas:
@antgonza tests here time out - is this branch in sync with upstream artifact-study-pages? Your changes shouldn't have affected the running time.

antgonza:
What I have seen that sometime is times out and restarting solves the problem. I guess it depends on the machine we get. Anyway, it should be up to date.

PR #1685, opened 2016-03-14
Creating Qiita client package

josenavas:
This depends on #1683, so review/merge that one first.

This PR extracts the module "qiita_client" from the target gene plugin and sets it up as its own package in the plugins directory. Eventually this package would be it's own package that the plugins developer can use so they don't need to re-implement some of the functionality available here.

I've extended the class to support a few common functionality that all plugins will need to implement, and updated the target gene plugin to make use of this new "package".


josenavas:
Ready for review

antgonza:
looks good to me, :+1: once tests pass, as @josenavas said he is basically moving code around to make it reusable.

josenavas:
@ElDeveloper or @mortonjt able to review?

qiita_plugins/qiita_client/qiita_client/qiita_client.py
```diff
@@ -195,3 +245,68 @@ def post(self, url, **kwargs):
             The JSON response from the server
         """
         return self._request_retry(requests.post, url, **kwargs)
+
+    # The functions are shortcuts for common functionality that all plugins
+    # need to implement.
+
+    def start_heartbeat(self, job_id):
+        """Crete and start a thread that would send the heartbeats to the server
```
mortonjt: Crete -> Create

mortonjt:
:+1:

antgonza:
The tests have already passed and the latests change doesn't change that results: Merging.

PR #1686, opened 2016-03-14
Artifact study pages type plugins

josenavas:
This depends on #1685 , so review/merge that one first.

This PR modifies that database and the software and processing job objects to support different plugin types.

Here, we are defining two types of plugins:
 - "processing plugin": these are the plugins that we had before. They take an artifact and perform some kind of processing on it.
 - "type plugin": new type of plugins. They define new artifact types in the system and they perform mainly 2 operations:
    - Create artifact: when a user uploads a new artifact, instead of being added to the system, it gets sent to the plugin and it performs any required validation and/or fixes to the artifacts (examples in following PRs)
    - Generate HTML summary: when a new artifact gets added to the system (either by uploading it or as a result of a job) Qiita fires off a job that generates the HTML summary of the given artifact.


josenavas:
Ready for review @antgonza

qiita_db/support_files/patches/40.sql
```diff
@@ -0,0 +1,118 @@
...
+CREATE INDEX idx_prep_template_processing_job_job ON qiita.prep_template_processing_job ( processing_job_id ) ;
+ALTER TABLE qiita.prep_template_processing_job ADD CONSTRAINT fk_prep_template_processing_job_pt FOREIGN KEY ( prep_template_id ) REFERENCES qiita.prep_template( prep_template_id )    ;
+ALTER TABLE qiita.prep_template_processing_job ADD CONSTRAINT fk_prep_template_processing_job_job FOREIGN KEY ( processing_job_id ) REFERENCES qiita.processing_job( processing_job_id )    ;
+
+-- Populate the table software_type with the 2 types of plugins
+INSERT INTO qiita.software_type (software_type, description)
+    VALUES ('processing plugin', 'A plugin that performs some kind of processing over an artifact.'),
+           ('type plugin', 'A plugin that defines new artifact types.');
+
+-- All the software presente in the system belong to type 1 (processing plugin)
```
antgonza: presente -> present

qiita_db/support_files/patches/40.sql
```diff
@@ -0,0 +1,118 @@
...
+    CONSTRAINT idx_prep_template_processing_job PRIMARY KEY ( prep_template_id, processing_job_id )
+ ) ;
+CREATE INDEX idx_prep_template_processing_job_pt_id ON qiita.prep_template_processing_job ( prep_template_id ) ;
+CREATE INDEX idx_prep_template_processing_job_job ON qiita.prep_template_processing_job ( processing_job_id ) ;
+ALTER TABLE qiita.prep_template_processing_job ADD CONSTRAINT fk_prep_template_processing_job_pt FOREIGN KEY ( prep_template_id ) REFERENCES qiita.prep_template( prep_template_id )    ;
+ALTER TABLE qiita.prep_template_processing_job ADD CONSTRAINT fk_prep_template_processing_job_job FOREIGN KEY ( processing_job_id ) REFERENCES qiita.processing_job( processing_job_id )    ;
+
+-- Populate the table software_type with the 2 types of plugins
+INSERT INTO qiita.software_type (software_type, description)
+    VALUES ('processing plugin', 'A plugin that performs some kind of processing over an artifact.'),
```
antgonza: What about 'processing plugin' -> 'action' and 'type plugin' -> 'create'?

qiita_db/support_files/patches/40.sql
```diff
@@ -0,0 +1,118 @@
...
+-- Setting up NOT NULL attribute here since the value was null until the previous statement
+ALTER TABLE qiita.software ALTER COLUMN software_type_id SET NOT NULL;
+
+-- We are going to create 2 new type plugins.
+-- The first one will define the type BIOM, while the other one will define
+-- all the types needed for TARGET_GENE. This separation is better since BIOM
+-- is a special type that even Qiita understands since it is the input for
+-- analysis and almost all plugins will need. However, the rest of the types
+-- are specific for the target gene plugin.
+INSERT INTO qiita.software (name, version, description, environment_script, start_script, software_type_id)
```
antgonza: I don't like that they have the plugin in the name, is that really necessary?

qiita_db/support_files/patches/python_patches/40.py
```diff
@@ -0,0 +1,22 @@
+from random import SystemRandom
+from string import ascii_letters, digits
+
+from qiita_db.sql_connection import TRN
+
+pool = ascii_letters + digits
+with TRN:
```
antgonza: Perhaps out of the scope of this PR and more cause this is how we are doing the previous ones. However, this means that every DB clean for the test environment we will get new ids and we will need to reset the configuration files, which is not convenient. Why not just define some for each plugin?

antgonza:
A few comments.

qiita_db/support_files/patches/40.sql
josenavas: Done - using the nomenclature that we discussed offline

qiita_db/support_files/patches/40.sql
josenavas: Done

qiita_db/support_files/patches/40.sql
josenavas: Nope - removed

qiita_db/support_files/patches/python_patches/40.py
josenavas: For life systems, these values have to be different in each deployment. If we hardcode those values here, by just using the values in the repository, the world will have full access to the entire database. However, this is a valid concern. I have fixed it by hardcoding the values in the populate _test_db.sql, so in the test environment, they'll always have the same values.

antgonza: Great, thanks!

antgonza: BTW I think you forgot to push those changes.

antgonza: Scratch that, github wasn't showing those changes.

antgonza:
:+1: @ElDeveloper @mortonjt, could you review? Thanks!

qiita_db/software.py
```diff
@@ -534,6 +537,23 @@ def default_workflows(self):
             for wf_id in qdb.sql_connection.TRN.execute_fetchflatten():
                 yield DefaultWorkflow(wf_id)
 
+    @property
+    def type(self):
+        """Returns the tyoe of the software
```
ElDeveloper: tyoe -> type

qiita_db/support_files/patches/40.sql
```diff
@@ -0,0 +1,118 @@
+-- Mar 12, 2016
+-- Add software_type table. This new table allows us to have different types of
+-- plugins. Here, we will introduce a new type, the "type plugin", and define
+-- the previous type as the "processing plugin" type. The new group "type plugin"
+-- define plugins that do not perform any processing on the artifacts but they
+-- are able to validate that they're correct and generate they summary page.
```
ElDeveloper: they -> their

qiita_db/support_files/patches/40.sql
```diff
@@ -0,0 +1,118 @@
...
+-- These new plugins are special. They are not directly visible by the end Qiita
+-- user by they are useful to plugin developers so they do not need to re-define
```
ElDeveloper: by -> but(?)

qiita_db/support_files/patches/40.sql
```diff
@@ -0,0 +1,118 @@
...
+-- types if they already exist. This way, multiple plugins can share the same
+-- type of artifacts without depending in another "processing" plugin.
+
+-- Create the new table to hold the software types
+CREATE TABLE qiita.software_type (
```
ElDeveloper: If I'm following correctly, `software_type` and `software` are tables that have now become our plugin type tables, right? Meaning that we are now a plugin can be a type or a piece of software, or a plugin that acts upon another plugin right?

qiita_db/support_files/patches/python_patches/40.py
```diff
@@ -0,0 +1,22 @@
...
+    # 2 and 3 are the ids of the 2 new software rows, the BIOM and
+    # target gene type plugins
+    for i in [2, 3]:
+        client_id = ''.join([SystemRandom().choice(pool) for _ in range(50)])
```
ElDeveloper: This is to simulate the fact that there was no oauth key because this functionality was not in place? Why not just make NULL, or would that break something?

ElDeveloper:
Looks good, just a few minor comments. The main thing that tripped me off was that I wasn't really understanding what the software table was storing, would probably make more sense to rename that to plugin and plugin_type, but that's not blocking.

qiita_db/software.py

qiita_db/support_files/patches/40.sql
josenavas: Done

qiita_db/support_files/patches/40.sql
josenavas: Yup - done

qiita_db/support_files/patches/40.sql
josenavas: Correct - the software table holds the actual plugins (in its inception was not like that hence the awful name) - I've added an issue (https://github.com/biocore/qiita/issues/1691) so we change the names.

There are 2 types of plugins, the ones who define new artifact types, and others that apply transformations to those artifact. They don't operate directly upon another plugin, but they require it to be installed in the system in order for such type to be defined in the system (and hence the artifact definition plugin be added to the system).

qiita_db/support_files/patches/python_patches/40.py
josenavas: In theory, these steps should happen when a plugin is "registered" in the system. currently we are "registering" the plugins by adding them directly to the database in the patches, so we need to insert their client id/secret so they can communicate with the DB.

ElDeveloper: Awesome, thanks for the description!

Yoshiki Vázquez-Baeza

antgonza:
Merging as the errors should be solved in the latests commit and the next PR is ready, where the tests are gonna be ran again.

PR #1687, opened 2016-03-14
Creating the Type Plugin: BIOM type

josenavas:
This depends on https://github.com/biocore/qiita/pull/1686 so review/merge that one first.

This adds a new type plugin to the system: the BIOM type plugin.

This plugin defines the type BIOM in the system. Since BIOM is a central type in Qiita, I think it is worth to have it as it's own package, since all the processing plugins will depend on this type but not necessarily in the other target gene types.

Currently, only the "create artifact" functionality is implemented: it checks that the sample ids in the BIOM table match the ones in the prep template and if not, tries to fix them. If it can't the artifact is not added to the system and provides information to the user on how to best solve the issue.

The HTML summary generation functionality is currently stubbed out on how it should work in general, and @antgonza will help to add such functionality.

### DIFF
--- a/qiita_plugins/biom_type/biom_type_plugin/create.py
+++ b/qiita_plugins/biom_type/biom_type_plugin/create.py
@@ -12,6 +12,7 @@ from json import loads
 
 from biom import load_table
 from biom.util import biom_open
+from biom.exception import TableException
 from qiita_client import format_payload
 
 
@@ -91,7 +92,16 @@ def create_artifact(qclient, job_id, parameters, out_dir):
                               'the prep information sample ids.')
 
         # Fix the sample ids
-        table.update_ids(id_map, axis='sample')
+        try:
+            table.update_ids(id_map, axis='sample')
+        except TableException:
+            missing = biom_sample_ids - set(id_map)
+            return format_payload(
+                False,
+                error_msg='Your prep information is missing samples that are '
+                          'present in your BIOM table: %s'
+                          % ', '.join(missing))
+
         new_biom_fp = join(out_dir, basename(biom_fp))
         with biom_open(new_biom_fp, 'w') as f:
             table.to_hdf5(f, "Qiita BIOM type plugin")
--- a/qiita_plugins/biom_type/biom_type_plugin/create.py
+++ b/qiita_plugins/biom_type/biom_type_plugin/create.py
@@ -1,0 +1,100 @@
+# -----------------------------------------------------------------------------
+# Copyright (c) 2014--, The Qiita Development Team.
+#
+# Distributed under the terms of the BSD 3-clause License.
+#
+# The full license is in the file LICENSE, distributed with this software.
+# -----------------------------------------------------------------------------
+
+from os.path import join, basename
+
+from json import loads
+
+from biom import load_table
+from biom.util import biom_open
+from qiita_client import format_payload
+
+
+def create_artifact(qclient, job_id, parameters, out_dir):
+    """Validate and fix a new BIOM artifact
+
+    Parameters
+    ----------
+    qclient : qiita_client.QiitaClient
+        The Qiita server client
+    job_id : str
+        The job id
+    parameters : dict
+        The parameter values to validate and create the artifact
+    out_dir : str
+        The path to the job's output directory
+
+    Returns
+    -------
+    dict
+        The results of the job
+
+    Raises
+    ------
+    ValueError
+        If there is any error gathering the information from the server
+    """
+    prep_id = parameters['template']
+    files = loads(parameters['files'])
+    a_type = parameters['artifact_type']
+
+    if a_type != "BIOM":
+        return format_payload(
+            False, error_msg="Unknown artifact type %s. Supported types: BIOM"
+                             % a_type)
+
+    qclient.update_job_step(job_id, "Step 1: Collecting prep information")
+    prep_info = qclient.get("/qiita_db/prep_template/%s/data" % prep_id)
+    if not prep_info or not prep_info['success']:
+        error_msg = "Could not get prep information: %s"
+        if prep_info:
+            error_msg = error_msg % prep_info['error']
+        else:
+            error_msg = error_msg % "could not connect with the server"
+        raise ValueError(error_msg)
+    prep_info = prep_info['data']
+
+    # Check if the biom table have the same sample ids than the prep info
+    qclient.update_job_step(job_id, "Step 2: Validting BIOM file")
+    new_biom_fp = biom_fp = files['BIOM'][0]
+    table = load_table(biom_fp)
+    pt_sample_ids = set(prep_info.keys())
+    biom_sample_ids = set(table.ids())
+    if biom_sample_ids != pt_sample_ids:
+        # The BIOM sample ids are different from the ones in the prep template
+        qclient.update_job_step(job_id, "Step 3: Fixing BIOM sample ids")
+        # Attempt 1: the user provided the run prefix column - in this case
+        # the run prefix column holds the sample ids present in the BIOM file
+        if 'run_prefix' in prep_info[next(iter(pt_sample_ids))]:
+            id_map = {v['run_prefix']: k for k, v in prep_info.items()}
+        else:
+            # Attemp 2: the sample ids in the BIOM table are the same that in
+            # the prep template but without the prefix
+            prefix = next(iter(pt_sample_ids)).split('.', 1)[0]
+            prefixed = set("%s.%s" % (prefix, s) for s in biom_sample_ids)
+            if prefixed == pt_sample_ids:
+                id_map = {s: "%s.%s" % (prefix, s) for s in biom_sample_ids}
+            else:
+                # There is nothing we can do. The samples in the BIOM table do
+                # not match the ones in the prep template and we can't fix it
+                return format_payload(
+                    False,
+                    error_msg='The sample ids in the BIOM table do not match '
+                              'the ones in the prep information. Please, '
+                              'provide the column "run_prefix" in the prep '
+                              'information to map the existing sample ids to '
+                              'the prep information sample ids.')
+
+        # Fix the sample ids
+        table.update_ids(id_map, axis='sample')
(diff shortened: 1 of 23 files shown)


qiita_plugins/biom_type/biom_type_plugin/create.py
```diff
@@ -0,0 +1,100 @@
...
+                return format_payload(
+                    False,
+                    error_msg='The sample ids in the BIOM table do not match '
+                              'the ones in the prep information. Please, '
+                              'provide the column "run_prefix" in the prep '
+                              'information to map the existing sample ids to '
+                              'the prep information sample ids.')
+
+        # Fix the sample ids
+        table.update_ids(id_map, axis='sample')
```
antgonza: Should this be in a try/except so we can give a good error to the user?

antgonza:
One comment. :+1: once my comment is addressed and the tests pass. @ElDeveloper @mortonjt could you take a look?

qiita_plugins/biom_type/biom_type_plugin/create.py
josenavas: Really good point - added

qiita_plugins/biom_type/biom_type_plugin/create.py
```diff
@@ -0,0 +1,110 @@
...
+from json import loads
+
+from biom import load_table
+from biom.util import biom_open
+from biom.exception import TableException
+from qiita_client import format_payload
+
+
+def create_artifact(qclient, job_id, parameters, out_dir):
+    """Validate and fix a new BIOM artifact
```
wasade: aren't validation, and fixing, conceptually different from creating?

qiita_plugins/biom_type/biom_type_plugin/create.py
```diff
@@ -0,0 +1,110 @@
...
+    a_type = parameters['artifact_type']
+
+    if a_type != "BIOM":
+        return format_payload(
+            False, error_msg="Unknown artifact type %s. Supported types: BIOM"
+                             % a_type)
+
+    qclient.update_job_step(job_id, "Step 1: Collecting prep information")
+    prep_info = qclient.get("/qiita_db/prep_template/%s/data" % prep_id)
+    if not prep_info or not prep_info['success']:
```
wasade: would it make sense for the error handling to happen within `qclient.get`? I suspect this pattern of checking the result will be replicated a reasonable amount?

qiita_plugins/biom_type/biom_type_plugin/create.py
```diff
@@ -0,0 +1,110 @@
...
+        error_msg = "Could not get prep information: %s"
+        if prep_info:
+            error_msg = error_msg % prep_info['error']
+        else:
+            error_msg = error_msg % "could not connect with the server"
+        raise ValueError(error_msg)
+    prep_info = prep_info['data']
+
+    # Check if the biom table have the same sample ids than the prep info
```
wasade: "has the same sample ids as the prep info"

qiita_plugins/biom_type/biom_type_plugin/create.py
```diff
@@ -0,0 +1,110 @@
...
+    qclient.update_job_step(job_id, "Step 2: Validting BIOM file")
+    new_biom_fp = biom_fp = files['BIOM'][0]
+    table = load_table(biom_fp)
+    pt_sample_ids = set(prep_info.keys())
```
wasade: call to `keys` shouldn't be necessary if its a `dict`

qiita_plugins/biom_type/biom_type_plugin/create.py
```diff
@@ -0,0 +1,110 @@
...
+    biom_sample_ids = set(table.ids())
+    if biom_sample_ids != pt_sample_ids:
```
wasade: what about the situation where the prep info is a superset of the biom table, as may happen with failed samples?

qiita_plugins/biom_type/biom_type_plugin/create.py
```diff
@@ -0,0 +1,110 @@
...
+            # the prep template but without the prefix
+            prefix = next(iter(pt_sample_ids)).split('.', 1)[0]
+            prefixed = set("%s.%s" % (prefix, s) for s in biom_sample_ids)
+            if prefixed == pt_sample_ids:
+                id_map = {s: "%s.%s" % (prefix, s) for s in biom_sample_ids}
+            else:
+                # There is nothing we can do. The samples in the BIOM table do
+                # not match the ones in the prep template and we can't fix it
+                return format_payload(
+                    False,
```
wasade: using a kwarg would be helpful as it isn't clear what `True` or `False` mean here. I can guess given the other calls to `format_payload` though...

qiita_plugins/biom_type/biom_type_plugin/plugin.py
```diff
@@ -0,0 +1,79 @@
...
+        The job id
+
+    Raises
+    ------
+    RuntimeError
+        If there is a problem gathering the job information
+    """
+    # Set up the Qiita Client
+    try:
+        conf_fp = environ['QP_BIOM_TYPE_CONFIG_FP']
```
wasade: what about doing a `.get` and not having the try/except?

qiita_plugins/biom_type/biom_type_plugin/plugin.py
```diff
@@ -0,0 +1,79 @@
...
+        if not exists(output_dir):
+            makedirs(output_dir)
+        try:
+            payload = task(qclient, job_id, job_info['parameters'], output_dir)
+        except Exception:
+            exc_str = repr(traceback.format_exception(*sys.exc_info()))
+            error_msg = ("Error executing %s: \n%s" % (task_name, exc_str))
+            payload = format_payload(False, error_msg=error_msg)
+        # The job completed
+        qclient.complete_job(job_id, payload)
```
wasade: would it make sense for this to be in a `finally` block?

qiita_plugins/biom_type/biom_type_plugin/summary.py
```diff
@@ -0,0 +1,40 @@
...
+    ValueError
+        If there is any error gathering the information from the server
+    """
+    # Only one paramater
+    # artifact_id = parameters['artifact']
+    # TODO
+    # Step 1: gather file information from qiita using REST api
+    # Step 2: generate HTML summary
+    # Step 3: add the new file to the artifact using REST api
+    pass
```
wasade: should this stub still conform to the described signature (e.g., raise `ValueError` or return `dict`)?

qiita_plugins/biom_type/setup.py
```diff
@@ -0,0 +1,48 @@
...
+      description='Qiita Type Plugin: BIOM',
+      author="Qiita development team",
+      author_email="qiita.help@gmail.com",
+      url='https://github.com/biocore/qiita',
+      test_suite='nose.collector',
+      packages=['biom_type_plugin'],
+      package_data={'biom_type_plugin': ['support_files/config_file.cfg']},
+      scripts=glob('scripts/*'),
+      extras_require={'test': ["nose >= 0.10.1", "pep8", 'httpretty']},
+      install_requires=['click >= 3.3', 'biom == 2.1.4'],
```
wasade: not biom >= 2.1.4 < 2.2.0?

wasade:
Few light comments, :+1: otherwise

qiita_plugins/biom_type/biom_type_plugin/create.py
josenavas: Yeah, I guess you are right I was trying to show that this are the steps performed before the artifact is actually created in the system (that happens once the job is completed) but I think you're right and makes sense to change it to `validate`.

qiita_plugins/biom_type/biom_type_plugin/create.py
josenavas: I thought that also at the beginning, but different plugins may want to react differently depending on the error message. I don't have a direct use case right now, but I think the current set up is the one that limits less the plugins developers.

qiita_plugins/biom_type/biom_type_plugin/create.py
josenavas: Done

qiita_plugins/biom_type/biom_type_plugin/create.py
josenavas: Done

qiita_plugins/biom_type/biom_type_plugin/create.py
josenavas: Yeah - I thought about that but I was unsure which is the best solution. This method is only executed when the user uploads a biom table through the interface (we are assuming that the artifact transformation plugins generate correct files). However, I think you have a good point and it can get annoying to the users - changing.

qiita_plugins/biom_type/biom_type_plugin/create.py
josenavas: good idea - done

qiita_plugins/biom_type/biom_type_plugin/plugin.py
josenavas: Done

qiita_plugins/biom_type/biom_type_plugin/plugin.py
josenavas: I'm unsure if I'm missing something, but what is the benefit of using the `finally` block here?

qiita_plugins/biom_type/biom_type_plugin/summary.py
josenavas: This was just some guidance for @antgonza , who already did a PR implementing this functionality, so I don't think we should worry about this

qiita_plugins/biom_type/setup.py
josenavas: Good catch

antgonza:
Thanks @josenavas and @wasade

PR #1688, opened 2016-03-15
URGENT: split mapping file fix

josenavas:
The `to_csv` call in the target gene plugin was missing the encoding keyword, which was making it to fail.

This is blocking important processing.


mortonjt:
:+1:

squirrelo:
Please add a test for this case, then :+1:

ElDeveloper:
> Please add a test for this case, then :+1:
:+1:  sounds like a good idea!

josenavas:
Modified current test

josenavas:
:speak_no_evil:

PR #1694, opened 2016-03-17
Add HTML summary to the artifacts

josenavas:
This PR adds the property `html_summary_fp` to the artifact. It also adds the method "patch" to the `ArtifactFilepathsHandler` so plugins can add the HTML summary to the artifacts.

### DIFF
--- a/qiita_db/artifact.py
+++ b/qiita_db/artifact.py
@@ -10,6 +10,7 @@ from __future__ import division
 from future.utils import viewitems
 from itertools import chain
 from datetime import datetime
+from os import remove
 
 import networkx as nx
 
@@ -797,6 +798,43 @@ class Artifact(qdb.base.QiitaObject):
             res = None
 
         return res
+
+    @html_summary_fp.setter
+    def html_summary_fp(self, value):
+        """Sets the HTML summary of the artifact
+
+        Parameters
+        ----------
+        value : str
+            Path to the new HTML summary
+        """
+        with qdb.sql_connection.TRN:
+            current = self.html_summary_fp
+            if current:
+                # Delete the current HTML summary
+                fp_id = current[0]
+                fp = current[1]
+                # From the artifact_filepath table
+                sql = """DELETE FROM qiita.artifact_filepath
+                         WHERE filepath_id = %s"""
+                qdb.sql_connection.TRN.add(sql, [fp_id])
+                # From the filepath table
+                sql = "DELETE FROM qiita.filepath WHERE filepath_id=%s"
+                qdb.sql_connection.TRN.add(sql, [fp_id])
+                # And from the filesystem only after the transaction is
+                # successfully completed (after commit)
+                qdb.sql_connection.TRN.add_post_commit_func(remove, fp)
+
+            # Add the new HTML summary
+            fp_ids = qdb.util.insert_filepaths(
+                [(value, 'html_summary')], self.id, self.artifact_type,
+                "filepath")
+            sql = """INSERT INTO qiita.artifact_filepath
+                        (artifact_id, filepath_id)
+                     VALUES (%s, %s)"""
+            # We only inserted a single filepath, so using index 0
+            qdb.sql_connection.TRN.add(sql, [self.id, fp_ids[0]])
+            qdb.sql_connection.TRN.execute()
 
     @property
     def parents(self):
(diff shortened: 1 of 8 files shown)


qiita_db/handlers/artifact.py
```diff
@@ -69,6 +69,46 @@ def get(self, artifact_id):
 
         self.write(response)
 
+    @authenticate_oauth
```
antgonza: This needs a test, right?

antgonza: Just noticed comment below.

qiita_db/handlers/artifact.py
```diff
@@ -69,6 +69,46 @@ def get(self, artifact_id):
...
+    def patch(self, artifact_id):
+        """Patches the filepaths of the artifact
```
antgonza: filepaths or filepath? AFAIK there is only one, no?

qiita_db/handlers/artifact.py
```diff
@@ -69,6 +69,46 @@ def get(self, artifact_id):
...
+
+        if req_op == 'add':
+            req_path = [v for v in req_path.split('/') if v]
+            if len(req_path) != 1 or req_path[0] != 'html_summary':
+                success = False
+                error_msg = 'Incorrect path parameter value'
+            else:
+                artifact, success, error_msg = _get_artifact(artifact_id)
+                if success:
+                    artifact.html_summary_fp = req_value
```
antgonza: Does this mean that we are going to store the full path?

qiita_db/handlers/artifact.py
josenavas: The "PATCH" request will allow us to update the filepaths attribute of the artifact. Currently, I'm only allowing to add a new filepath, which should be the html-summary, by using `op = add` and `path = /html_summary/". However, this may change if we need more functionality to the patch request.

qiita_db/handlers/artifact.py
josenavas: Nope, the setter of that property calls internally "insert_filepaths" which moves the file to the correct location in the file system and inserts the correct relative path to the DB

qiita_db/artifact.py
```diff
@@ -776,6 +777,66 @@ def filepaths(self):
...
+            res = (fps[0][0], fps[0][1])
+        else:
+            res = None
+
+        return res
+
+    @html_summary_fp.setter
+    def html_summary_fp(self, value):
+        """Sets the HTML summary of the artifact
+
```
ElDeveloper: Can you describe what `value` is?

ElDeveloper: nix that, didn't see the parameter below.

qiita_db/handlers/artifact.py
```diff
@@ -69,6 +69,46 @@ def get(self, artifact_id):
...
         self.write(response)
 
+    @authenticate_oauth
+    def patch(self, artifact_id):
+        """Patches the filepaths of the artifact
+
+        Parameter
+        ---------
+        artifact_id : str
+            The id of the artifact whole filepaths information is being updated
```
ElDeveloper: whole?

qiita_db/handlers/artifact.py
```diff
@@ -69,6 +69,46 @@ def get(self, artifact_id):
...
+
+        Returns
+        -------
+        dict
+            Format:
+            {'success': bool,
```
ElDeveloper: On previous return dictionaries, we were using a slightly different scheme, is there a good motivation to change that? I'm referring to the dictionaries that have a "status" and a "message" key, etc.

ElDeveloper:
:+1:

qiita_db/handlers/artifact.py
josenavas: whose oops!

qiita_db/handlers/artifact.py
josenavas: Good point. The difference between here and the other dictionaries is that this is part of the REST api that we put in place for the plugins, while the other ones are used in the GUI. It turns out that we did not catch those differences when we reviewed the GUI code adding this structure.

If you feel strong about it I can change but I will need to change all the qiita_db handlers plus the plugins consuming that API. I would prefer to not change it right now, given that we should end up changing this entire API to not return the success keyword but rather raising an HTTP error in case the call is not successful. Does it make sense?

ElDeveloper: If we are going to change the API regardless, I'm fine and I think we already have an issue open, I've added a note there to make sure we don't forget about this.

antgonza:
Thanks @ElDeveloper and @josenavas. Merging as the changes will not change the tests.

PR #1696, opened 2016-03-17
Artifact study pages adding biom

antgonza:
This is built on top of https://github.com/biocore/qiita/pull/1687, so review that one first. 

Adding biom summary, it looks like this:
screen shot 2016-03-17 at 11 51 48 am


setup.py
```diff
@@ -102,6 +102,6 @@
                         'scikit-bio >= 0.2.3, < 0.3.0', 'wtforms == 2.0.1',
                         'qiime >= 1.9.0, < 1.10.0', 'moi >= 0.2.0, < 0.3.0',
                         'sphinx-bootstrap-theme', 'Sphinx >= 1.2.2',
-                        'gitpython'],
+                        'gitpython', 'seaborn'],
```
josenavas: Can you remove `seaborn` from here and add it to `qiita_plugins/biom_type/setup.py`? I think you will need to add pandas there too, as it is a new dependency for the plugin.

qiita_plugins/biom_type/biom_type_plugin/summary.py
```diff
@@ -6,8 +6,19 @@
...
+from urllib import quote
+from base64 import b64encode
+from os.path import join, basename
+
+from StringIO import StringIO
+import seaborn as sns
+
+
+def generate_html_summary(qclient, job_id, parameters, out_dir,
+                          return_html=False):
```
josenavas: I think this is not needed for testing, as the file is not moved during the testing procedure and the output filepath can be deterministically known where it is going to be.

qiita_plugins/biom_type/biom_type_plugin/summary.py
```diff
@@ -31,10 +44,79 @@ def generate_html_summary(qclient, job_id, parameters, out_dir):
...
+    ax = sns.distplot(sample_counts)
+    ax.set_xlabel("Number of sequences per sample")
+    ax.set_ylabel("Frequency")
+    plot = ax.get_figure()
+    sc_plot = StringIO()
+    plot.savefig(sc_plot, format='png')
+    sc_plot.seek(0)
+
+    uri = 'data:image/png;base64,' + quote(b64encode(sc_plot.buf))
+    artifact_information = [
```
josenavas: Minor and no need to change if you don't want to, but I usually have the feeling that this kind of HTML are easy to read if we just create a global variable on top of the file with the full HTML string and the formatters already there, and here in the code we just format the code correctly:
```python
HTML_FORMAT = """<b>Number of samples:</b> %d<br/>
<b>Number of features:</b> %d<br/>
...
"""

...

html_str = HTML_FORMAT % (num_samples, num_features, ...)
```

qiita_plugins/biom_type/biom_type_plugin/summary.py
```diff
@@ -31,10 +44,79 @@ def generate_html_summary(qclient, job_id, parameters, out_dir):
...
+        '<img src = "%s"/>' % uri
+    ]
+
+    of_fp = join(out_dir, "%s.html" % basename(fps))
+    of = open(of_fp, 'w')
+    of.write('\n'.join(artifact_information))
+
     # Step 3: add the new file to the artifact using REST api
-    pass
+    reply = qclient.patch(qclient_url, 'add', '/html_summary/', value='of_fp')
```
josenavas: `value='of_fp'` -> `value=of_fp`

qiita_plugins/biom_type/biom_type_plugin/summary.py
antgonza: I tried skipping this step but couldn't find the file, not sure if it's deleted or what. I really don't see this as a blocker, it doesn't take anything away. However, I can spend more time on this if you feel really strong about it.

setup.py
antgonza: In general I would agree with you but seaborn is something that all the plugins showing plots are going to use. Thus, IMOO installing it in the main install is fine.

qiita_plugins/biom_type/biom_type_plugin/summary.py
```diff
@@ -31,10 +44,79 @@ def generate_html_summary(qclient, job_id, parameters, out_dir):
...
+
+    return reply if not return_html else (reply, artifact_information)
+
+
+# Help methods taken from: https://goo.gl/8uebr7
+# Not adding tests as this will be implemented and tested in QIIME2
```
josenavas: We should add tests here - I see your point that this "will" be implemented and tested in QIIME2, but this is production code so we should have tests here. It would be better to actually write the tests here and move it to QIIME2, rather than waiting/assume it will be in QIIME2.

antgonza: Excellent point. Let's discuss with the rest of the QIIME2 team during this afternoon's meeting.

qiita_plugins/biom_type/biom_type_plugin/summary.py
```diff
@@ -31,10 +44,79 @@ def generate_html_summary(qclient, job_id, parameters, out_dir):
...
+    result = {}
+    for count_vector, id_, _ in table.iter(axis=axis):
+        result[id_] = float(count_vector.sum())
+    return pd.Series(result)
+
+
+def count_summary(table, axis='sample'):
+    counts = _counts(table, axis=axis)
+
+    summary = pd.Series([counts.min(), counts.quantile(0.25), counts.median(),
```
josenavas: I know you grabbed the code from `https://goo.gl/8uebr7`. However, I think using pandas here is an overkill. `counts` is used just as a numpy array (I think all the call that are here are already available in numpy arrays), and `summary` is just used as a dictionary (actually, the first thing done once `count_summary` returns is a call to `to_dict`.

I think we can drop the dependency on pandas for this plugin (remember that plugins are going to be independent from Qiita).

qiita_plugins/biom_type/biom_type_plugin/summary.py
```diff
@@ -31,10 +44,79 @@ def generate_html_summary(qclient, job_id, parameters, out_dir):
...
+    artifact_id = parameters['input_data']
+    qclient_url = "/qiita_db/artifacts/%s/filepaths/" % artifact_id
+    fps_info = qclient.get(qclient_url)
+    if not fps_info or not fps_info['success']:
+        error_msg = "Could not get artifact filepath information: %s"
+        if fps_info:
+            error_msg = error_msg % fps_info['error']
+        else:
+            error_msg = error_msg % "could not connect with the server"
+        raise ValueError(error_msg)
```
josenavas: Missing tests

setup.py
josenavas: No, because the environments of Qiita and the environments for the plugins can potentially be different. Also, not all the plugins will use, only those that actually generate a plot.

qiita_plugins/biom_type/biom_type_plugin/summary.py
antgonza: Adding.

qiita_plugins/biom_type/biom_type_plugin/summary.py
antgonza: Good catch, changing.

qiita_plugins/biom_type/biom_type_plugin/summary.py
gregcaporaso: I agree re: tests - note that the functionality from Q2D2 and the initial QIIME 2 plugins that we demoed is specifically advertised as being untested, non-production ready code. So, if moving that into a production system it must be tested.

qiita_plugins/biom_type/biom_type_plugin/summary.py
gregcaporaso: :+1: (I know we discussed on the call, but just adding here for the record.) I was just trying to get the nice HTML repr from the ``pd.Series`` for this, since I was displaying the output in an IPython Notebook for our prototype.

qiita_plugins/biom_type/biom_type_plugin/summary.py
antgonza: Yup, adding.

qiita_plugins/biom_type/biom_type_plugin/summary.py
antgonza: Fixing.

setup.py
antgonza: K

josenavas:
:+1:

josenavas:
@ElDeveloper @mortonjt available for a review?

qiita_plugins/biom_type/biom_type_plugin/summary.py
wasade: Btw, can also use Table.sum(axis) and zip with Table.IDs(axis).

I thought there already was a summarization method used by summarize-table?

qiita_plugins/biom_type/biom_type_plugin/summary.py
```diff
@@ -31,10 +44,79 @@ def generate_html_summary(qclient, job_id, parameters, out_dir):
...
+        ("<b>Median count:</b> %d<br/>" %
+         sample_count_summary['Median count']),
+        ("<b>Mean count:</b> %d<br/>" %
+         sample_count_summary['Mean count']),
+        '<br/><hr/><br/>',
+        '<img src = "%s"/>' % uri
+    ]
+
+    of_fp = join(out_dir, "%s.html" % basename(fps))
+    of = open(of_fp, 'w')
```
ElDeveloper: Can you close this file handle or use `with`?

qiita_plugins/biom_type/biom_type_plugin/summary.py
```diff
@@ -20,6 +31,8 @@ def generate_html_summary(qclient, job_id, parameters, out_dir):
         The parameter values to validate and create the artifact
     out_dir : str
         The path to the job's output directory
+    return_html : bool, optional
+        True will return the html str, useful for testing
```
ElDeveloper: Can you update the description of the return value?

qiita_plugins/biom_type/biom_type_plugin/tests/test_summary.py
```diff
@@ -0,0 +1,603 @@
...
+
+EXT_HTML = [
+    '<b>Number of samples:</b> 3<br/>',
+    '<b>Number of features:</b> 2<br/>',
+    '<b>Minimum count:</b> 1<br/>',
+    '<b>Maximum count:</b> 43<br/>',
+    '<b>Median count:</b> 3<br/>',
+    '<b>Mean count:</b> 15<br/>',
+    '<br/><hr/><br/>',
+    ('<img src = "data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAyAAAAImCAYAA'
```
ElDeveloper: We really don't need to have the image data, if we want to test the correctness of the figure, we should return the figure object itself and check its attributes.

qiita_plugins/biom_type/biom_type_plugin/summary.py
```diff
@@ -31,10 +44,79 @@ def generate_html_summary(qclient, job_id, parameters, out_dir):
...
+    ax.set_xlabel("Number of sequences per sample")
+    ax.set_ylabel("Frequency")
+    plot = ax.get_figure()
+    sc_plot = StringIO()
+    plot.savefig(sc_plot, format='png')
+    sc_plot.seek(0)
+
+    uri = 'data:image/png;base64,' + quote(b64encode(sc_plot.buf))
+    artifact_information = [
+        "<b>Number of samples:</b> %d<br/>" % num_samples,
```
ElDeveloper: This whole section would make more sense as an HTML table, rather than just text, right?

qiita_plugins/biom_type/biom_type_plugin/summary.py
```diff
@@ -31,10 +44,79 @@ def generate_html_summary(qclient, job_id, parameters, out_dir):
...
+# Based on methods from: https://goo.gl/8uebr7
+
+def _counts(table, axis='sample'):
+    result = {}
+    for count_vector, id_, _ in table.iter(axis=axis):
+        result[id_] = float(count_vector.sum())
+    return result
+
+
+def count_summary(table, axis='sample'):
```
ElDeveloper: This function  and _counts have no tests nor documentation, any reason for that?

qiita_plugins/biom_type/biom_type_plugin/summary.py
```diff
@@ -31,10 +44,79 @@ def generate_html_summary(qclient, job_id, parameters, out_dir):
...
+    counts = _counts(table, axis=axis)
+    counts_values = asarray(counts.values())
+
+    summary = {
+        'Minimum count': min(counts_values),
```
ElDeveloper: if this is already a numpy array, I would use the min method instead of the min function:

```
In [4]: %timeit x = np.random.rand(1000); x.min()
The slowest run took 177.35 times longer than the fastest. This could mean that an intermediate result is being cached 
100000 loops, best of 3: 18.4 µs per loop

In [5]: %timeit x = np.random.rand(1000); min(x)
10000 loops, best of 3: 78.2 µs per loop
```

qiita_plugins/biom_type/biom_type_plugin/summary.py
```diff
@@ -7,7 +7,18 @@
 # -----------------------------------------------------------------------------
 
 
-def generate_html_summary(qclient, job_id, parameters, out_dir):
+from biom import load_table
+from urllib import quote
+from base64 import b64encode
+from os.path import join, basename
+from numpy import mean, median, asarray
```
ElDeveloper: Can you use `import numpy as np`?

qiita_plugins/biom_type/biom_type_plugin/summary.py
antgonza: This comment made me realize that we don't need this separation as we can simply have this code within generate_html_summary. The original idea was to use QIIME2 code but at this point that doesn't make sense.

qiita_plugins/biom_type/biom_type_plugin/summary.py
antgonza: Good suggestion.

qiita_plugins/biom_type/biom_type_plugin/tests/test_summary.py
antgonza: For the record, after offline discussion we agreed that this is OK.

qiita_plugins/biom_type/biom_type_plugin/summary.py
antgonza: yup.

qiita_plugins/biom_type/biom_type_plugin/summary.py
antgonza: Not sure, I guess it depends on what you like ... as you can imagine I like it like this.

qiita_plugins/biom_type/biom_type_plugin/summary.py
antgonza: yup

qiita_plugins/biom_type/biom_type_plugin/summary.py
```diff
@@ -31,10 +44,79 @@ def generate_html_summary(qclient, job_id, parameters, out_dir):
...
+    return result
+
+
+def count_summary(table, axis='sample'):
+    counts = _counts(table, axis=axis)
+    counts_values = asarray(counts.values())
+
+    summary = {
+        'Minimum count': min(counts_values),
+        'Maximum count': max(counts_values),
```
ElDeveloper: Can you do the same thing for max?

ElDeveloper: My bad, let me rephrase that, can you do the same thing you did for min for max.

qiita_plugins/biom_type/biom_type_plugin/summary.py
ElDeveloper: I'm not as concerned about the visual part of it, I'm more concerned about the structure, seems like this summary all belongs together and that it probably belongs inside a table, but if I'm the only one who thinks this way yeah ignore this  `¯\_(ツ)_/¯`

PR #1705, opened 2016-03-24
Processing interface

josenavas:
This PR adds the processing interface, already connected with the backend.
Only 2 missing points:
 - [x] Remove jobs from the workflow
 - [x] Add JavaScript documentation

PR #1706, opened 2016-03-24
Artifact summary add processing jobs

josenavas:
~~Depends on #1705 so review/merge that one first.~~

This adds a list of jobs that use the given artifact as input, and shows their id, command name, status and current step

### DIFF
--- a/qiita_pet/handlers/api_proxy/artifact.py
+++ b/qiita_pet/handlers/api_proxy/artifact.py
@@ -54,6 +54,9 @@ def artifact_summary_get_request(user_id, artifact_id):
     summary = artifact.html_summary_fp
     job_info = None
     errored_jobs = []
+    processing_jobs = [
+        [j.id, j.command.name, j.status, j.step] for j in artifact.jobs()
+        if j.command.software.type == "artifact transformation"]
     # Check if the HTML summary exists
     if summary:
         with open(summary[1]) as f:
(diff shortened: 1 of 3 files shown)


josenavas:
Ready for review

ElDeveloper:
:+1: looks good!

qiita_pet/handlers/api_proxy/artifact.py
```diff
@@ -54,6 +54,9 @@ def artifact_summary_get_request(user_id, artifact_id):
     summary = artifact.html_summary_fp
     job_info = None
     errored_jobs = []
+    processing_jobs = [
+        [j.id, j.command.name, j.status, j.step] for j in artifact.jobs()
```
antgonza: How does this work? The braces are off.

josenavas: What do you mean that the braces are off? It's a list comprehension generating a list of 4 elements, but filtering for those jobs that are part of an `artifact transformation` software. See [this line](https://github.com/biocore/qiita/pull/1706/files/22c5743ed31c12f7940eeb41a5c707136a3e8894#diff-4220e7483ec380c13e6819b835190569R136) for an example of the data generated here.

antgonza: Got it, thanks.

PR #1707, opened 2016-03-24
Fix patch 39.sql

josenavas:
setting not null property later so we can apply the patch in the live system (discovered while deploying in the test environment)

Should be fast to review and doesn't depend in any other PR.


ElDeveloper:
:+1:

PR #1708, opened 2016-03-25
Making lab person optional

josenavas:
The interface was assuming that `lab person` was always present, but it is an optional attribute of the study.

### DIFF
--- a/qiita_pet/handlers/api_proxy/tests/test_studies.py
+++ b/qiita_pet/handlers/api_proxy/tests/test_studies.py
@@ -52,27 +52,78 @@ class TestStudyAPI(TestCase):
                     'Future studies will attempt to analyze the soils and '
                     'rhizospheres from the same location at different time '
                     'points in the plant lifecycle.',
-                    'status': 'private',
-                    'spatial_series': False,
-                    'study_description': 'Analysis of the Cannabis Plant '
-                                         'Microbiome',
-                    'shared_with': ['shared@foo.bar'],
-                    'lab_person': {'affiliation': 'knight lab',
-                                   'name': 'LabDude',
-                                   'email': 'lab_dude@foo.bar'},
-                    'principal_investigator': {'affiliation': 'Wash U',
-                                               'name': 'PIDude',
-                                               'email': 'PI_dude@foo.bar'},
-                    'study_alias': 'Cannabis Soils',
-                    'study_id': 1,
-                    'most_recent_contact': datetime(2014, 5, 19, 16, 11),
-                    'publications': [['10.100/123456', '123456'],
-                                     ['10.100/7891011', '7891011']],
-                    'num_samples': 27,
-                    'study_title': 'Identification of the Microbiomes for '
-                                   'Cannabis Soils',
-                    'number_samples_collected': 27}}
+                'status': 'private',
+                'spatial_series': False,
+                'study_description': 'Analysis of the Cannabis Plant '
+                                     'Microbiome',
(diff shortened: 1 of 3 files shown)


qiita_pet/handlers/api_proxy/tests/test_studies.py
```diff
@@ -52,28 +52,79 @@ def test_study_get_req(self):
...
-                    'study_alias': 'Cannabis Soils',
-                    'study_id': 1,
-                    'most_recent_contact': datetime(2014, 5, 19, 16, 11),
-                    'publications': [['10.100/123456', '123456'],
-                                     ['10.100/7891011', '7891011']],
-                    'num_samples': 27,
-                    'study_title': 'Identification of the Microbiomes for '
-                                   'Cannabis Soils',
-                    'number_samples_collected': 27}}
+                'status': 'private',
```
josenavas: The changes here are just fixing the indentation (PEP8 didn't flag it but it was incorrect).

qiita_pet/handlers/api_proxy/tests/test_studies.py
```diff
@@ -52,28 +52,79 @@ def test_study_get_req(self):
...
+                'most_recent_contact': datetime(2014, 5, 19, 16, 11),
+                'publications': [['10.100/123456', '123456'],
+                                 ['10.100/7891011', '7891011']],
+                'num_samples': 27,
+                'study_title': 'Identification of the Microbiomes for '
+                               'Cannabis Soils',
+                'number_samples_collected': 27}}
         self.assertEqual(obs, exp)
 
+        # Test with no lab person
```
josenavas: From here below is the new test that I've added.

ackermag: I think this is a good thing to have because PI denotes lab submitting, lab
personnel maybe may be changed to 'submitter' who uploaded the data.  This
allows us to identify source of data (PI) and the person who uploaded the
study.  Often uploader moves to new lab (new position, where they may be
PI) and continues to use qiita.  I believe it is good to track this
history.  It also has an implication for seq_center, center_name because it
will allow us to solve problems with processing.
Do not remove this requirement - it is very easy to add PI even if they
have no account

On Thu, Mar 24, 2016 at 8:54 PM, Jose Navas <notifications@github.com>
wrote:

> In qiita_pet/handlers/api_proxy/tests/test_studies.py
> <https://github.com/biocore/qiita/pull/1708#discussion_r57416936>:
>
> >          self.assertEqual(obs, exp)
> >
> > +        # Test with no lab person
>
> From here below is the new test that I've added.
>
> —
> You are receiving this because you are subscribed to this thread.
> <https://github.com/biocore/qiita/pull/1708/files/a6067ee7ec0f54b4c9572a5fe16bc862b45818b7#r57416936>
>



-- 
Gail Ackermann
Knight Lab
UCSD

josenavas: @ackermag Lab person was never required. When you create a study, you are forced to introduce a PI but not a Lab person. Since it is not required there, I can't make sure that it exists here, which breaks the new interface. Making lab person required involves more changes, which I think out of the scope of this PR and the changes performed here - can you open an issue to discuss this in there? You can also bring it up in the next Qiita meeting.

Note that lab person doesn't necessary lists who uploaded the data - that is the owner of the study (or the people who has access to the study). As an example, I can create an study, add Bob as a PI and Alice as lab person, w/o Bob or Alice being involved at all in the study.

ackermag:
ok

PR #1716, opened 2016-03-28
Import artifact

josenavas:
This PR enables importing an artifact from another study that the user owns.

### DIFF
--- a/qiita_db/user.py
+++ b/qiita_db/user.py
@@ -31,6 +31,8 @@ from __future__ import division
 from re import sub
 from datetime import datetime
 
+from future.utils import viewitems
+
 from qiita_core.exceptions import (IncorrectEmailError, IncorrectPasswordError,
                                    IncompetentQiitaDeveloperError)
 from qiita_core.qiita_settings import qiita_config
@@ -475,6 +477,45 @@ class User(qdb.base.QiitaObject):
             return qdb.sql_connection.TRN.execute_fetchindex()
 
     # ------- methods ---------
+    def user_artifacts(self, artifact_type=None):
+        """Returns the artifacts owned by the user, grouped by study
+
+        Parameters
+        ----------
+        artifact_type : str, optional
+            The artifact type to retrieve. Default: retrieve all artfact types
+
+        Returns
+        -------
+        dict of {qiita_db.study.Study: list of qiita_db.artifact.Artifact}
+            The artifacts owned by the user
+        """
+        with qdb.sql_connection.TRN:
+            sql_args = [self.id, qiita_config.portal]
+            sql_a_type = ""
+            if artifact_type:
+                sql_a_type = " AND artifact_type = %s"
+                sql_args.append(artifact_type)
+
+            sql = """SELECT study_id, array_agg(
+                        artifact_id ORDER BY artifact_id)
+                     FROM qiita.study
+                        JOIN qiita.study_portal USING (study_id)
+                        JOIN qiita.portal_type USING (portal_type_id)
+                        JOIN qiita.study_artifact USING (study_id)
+                        JOIN qiita.artifact USING (artifact_id)
+                        JOIN qiita.artifact_type USING (artifact_type_id)
+                        WHERE email = %s AND portal = %s{0}
+                        GROUP BY study_id""".format(sql_a_type)
+            qdb.sql_connection.TRN.add(sql, sql_args)
+            db_res = dict(qdb.sql_connection.TRN.execute_fetchindex())
+            res = {}
+            for s_id, artifact_ids in viewitems(db_res):
+                res[qdb.study.Study(s_id)] = [
+                    qdb.artifact.Artifact(a_id) for a_id in artifact_ids]
+
+            return res
+
     def change_password(self, oldpass, newpass):
         """Changes the password from oldpass to newpass
 
--- a/qiita_db/user.py
+++ b/qiita_db/user.py
@@ -506,7 +506,8 @@ class User(qdb.base.QiitaObject):
                         JOIN qiita.artifact USING (artifact_id)
                         JOIN qiita.artifact_type USING (artifact_type_id)
                         WHERE email = %s AND portal = %s{0}
-                        GROUP BY study_id""".format(sql_a_type)
+                        GROUP BY study_id
+                        ORDER BY study_id""".format(sql_a_type)
             qdb.sql_connection.TRN.add(sql, sql_args)
             db_res = dict(qdb.sql_connection.TRN.execute_fetchindex())
             res = {}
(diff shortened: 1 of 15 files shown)


qiita_pet/handlers/api_proxy/artifact.py
```diff
@@ -250,6 +250,8 @@ def artifact_post_req(user_id, filepaths, artifact_type, name,
         Name to give the artifact
     prep_template_id : int or str castable to int
         Prep template to attach the artifact to
+    artifact_id : str, optional
```
antgonza: Is this really a str vs an int?

qiita_pet/handlers/api_proxy/artifact.py
```diff
@@ -267,41 +269,55 @@ def artifact_post_req(user_id, filepaths, artifact_type, name,
...
-    try:
-        artifact = Artifact.create(cleaned_filepaths, artifact_type, name=name,
-                                   prep_template=prep)
-    except Exception as e:
-        # We should hit this exception rarely (that's why it is an exception)
-        # since at this point we have done multiple checks. However, it can
-        # occur in weird cases, so better let the GUI know that this failed
-        return {'status': 'error',
-                'message': "Error creating artifact: %s" % str(e)}
+    else:
```
antgonza: Just want to confirm that there are tests for the try/except blocks for this section.

qiita_pet/handlers/study_handlers/base.py
```diff
@@ -78,9 +78,6 @@ def get(self):
         atype = self.get_argument('artifact_type')
         pt_id = self.get_argument('prep_template_id')
 
-        res = study_files_get_req(study_id, pt_id, atype)
+        res = study_files_get_req(self.current_user.id, study_id, pt_id, atype)
```
antgonza: I was wondering about this, why send the user.id vs. the user object?

qiita_db/user.py
```diff
@@ -475,6 +477,46 @@ def unread_messages(self):
...
+
+        Returns
+        -------
+        dict of {qiita_db.study.Study: list of qiita_db.artifact.Artifact}
+            The artifacts owned by the user
+        """
+        with qdb.sql_connection.TRN:
+            sql_args = [self.id, qiita_config.portal]
+            sql_a_type = ""
+            if artifact_type:
```
ElDeveloper: Question: incorrect artifact types here will go unnoticed i.e. `>>> .....user_artifacts(artifact_type='B10M')`, this will return an empty dictionary, would we want to alert consumers that this is actually an artifact type not recognized by the system or is that not at all expected/supported.

qiita_pet/handlers/api_proxy/studies.py
```diff
@@ -242,8 +246,20 @@ def study_files_get_req(study_id, prep_template_id, artifact_type):
     # because selected is initialized to the empty list
     file_types.insert(0, (first[0], first[1], selected))
 
+    # Create a list of artifacts that the user has access to, in case that
+    # he wants to import the files from another artifact
+    user = User(user_id)
+    artifact_options = []
+    for study, artifacts in viewitems(user.user_artifacts(
```
antgonza: This makes sense for most users but for an admin you will like to see not only your artifacts but also the artifacts that belong to the owner of the study you are checking. For example, let's say that you are helping a user and you added an artifact in prep X, then you want to reuse it for prep Y, with this structure you can't.

qiita_pet/handlers/api_proxy/artifact.py
```diff
@@ -267,41 +269,55 @@ def artifact_post_req(user_id, filepaths, artifact_type, name,
...
-    # This should never happen, but it doesn't hurt to actually have
-    # a explicit check, in case there is something odd with the JS
-    if not cleaned_filepaths:
-        return {'status': 'error',
-                'message': "Can't create artifact, no files provided."}
+    if artifact_id:
+        # if the artifact id has been provided, import the artifact
+        try:
+            artifact = Artifact.copy(Artifact(artifact_id), prep)
+        except Exception as e:
```
ElDeveloper: This pattern seems to be replicated in two different locations, the reality of it is that if anything goes wrong, we want the interface to know about that, so perhaps the thing to do here is to follow the decorator pattern and decorate these methods to catch a general exception and return the appropriate dictionary with the encoded information we need? Not blocking, but I thought I would bring this up now.

ElDeveloper:
Looks good, just a few non-blocking comments, and fyi the tests are failing.

qiita_db/user.py
josenavas: I see your point. This has been added mainly for the needs in the interface. I don't think though that an empty dict is a bad answer. What do others think?

qiita_pet/handlers/api_proxy/artifact.py
josenavas: This comes from the handler so it actually comes as a string.

qiita_pet/handlers/api_proxy/artifact.py
josenavas: That is actually a good point - however I think it shouldn't be blocking now and we can fix later. Can you open an issue?

qiita_db/user.py
antgonza: I think is fine to return an empty dict. This is also happening in other sections of the code.

qiita_pet/handlers/api_proxy/artifact.py
antgonza: OK but note that prep_template_id also does and it can be int or str.

qiita_db/user.py
ElDeveloper: Sounds good.

qiita_pet/handlers/api_proxy/artifact.py
ElDeveloper: See: https://github.com/biocore/qiita/issues/1717

qiita_pet/handlers/api_proxy/artifact.py
josenavas: I had for the one below, not for the one above - just added.

qiita_pet/handlers/api_proxy/artifact.py
josenavas: Good point

qiita_pet/handlers/api_proxy/studies.py
josenavas: Just to confirm - you wanna add to the list the artifacts of the current study - correct?

qiita_pet/handlers/study_handlers/base.py
josenavas: For consistency with the other calls - when these calls where put in place, it was assumed that they'll be replaced by REST api calls, hence the idea on just passing the ids. I don't think that is gonna happen, b/c this calls are specialized for the interface but I want to be consistent at this point.

PR #1718, opened 2016-03-29
adding tgz

antgonza:


antgonza:
This is ready for review. It adds the possibility of downloading all files from an analysis and the failures from close-ref-picking via a new filepath_type = tgz.

qiita_db/analysis.py
```diff
@@ -508,6 +508,29 @@ def mapping_file(self):
             return join(base_fp, mapping_fp[0][0])
 
     @property
+    def tgz(self):
```
josenavas: Can you use the function `retrieve_filepaths` from [here](https://github.com/biocore/qiita/blob/master/qiita_db/util.py#L648) ? This way it will look up the mountpoint correctly. This will fail, for example, in the test environment.

qiita_plugins/target_gene/tgp/pick_otus.py
```diff
@@ -155,17 +176,25 @@ def pick_closed_reference_otus(qclient, job_id, parameters, out_dir):
...
 
-    qclient.update_job_step(job_id, "Step 3 of 3: Executing OTU picking")
+    qclient.update_job_step(job_id, "Step 3 of 4: Executing OTU picking")
     std_out, std_err, return_value = system_call(command)
     if return_value != 0:
         error_msg = ("Error running OTU picking:\nStd out: %s\nStd err: %s"
                      % (std_out, std_err))
         return format_payload(False, error_msg=error_msg)
 
+    qclient.update_job_step(job_id, "Step 4 of 4: tgz sortmerna folder")
```
josenavas: `"Step 4 of 4: tgz sortmerna folder"` -> `"Step 4 of 4: Generating tgz sortmerna folder"`

qiita_ware/analysis_pipeline.py
```diff
@@ -89,6 +89,45 @@ def _finish_analysis(analysis, **kwargs):
...
+    kwargs : ignored
+        Necessary to have in parameters to support execution via moi.
+    """
+    from qiita_db.sql_connection import TRN
+    from qiita_db.util import get_mountpoint, get_mountpoint_path_by_id
+    from os.path import join
+    from subprocess import Popen, PIPE
+
+    fps_ids = analysis.all_associated_filepath_ids
+    with TRN:
```
josenavas: This is `qiita_ware`, in `qiita_ware` shouldn't be any SQL so please remove this from here. My recommendation will be to add a new method to analysis `generate_tgz` that performs this task. This way, everything regarding the data stored in analysis is managed internally by the analysis object.

josenavas:
Few comments - plus still errors

qiita_db/support_files/patches/python_patches/41.py
```diff
@@ -0,0 +1,109 @@
...
+                # removing / from the path if it exists
+                to_tgz = fp[:-1] if fp[-1] == '/' else fp
+                break
+
+        if to_tgz is None:
+            continue
+
+        tgz = to_tgz + '.tgz'
+        if not exists(tgz):
+            cmd = 'tar zcf %s %s' % (tgz, to_tgz)
```
ElDeveloper: Is there any benefit to use a subprocess instead of the `tarfile` module?

qiita_plugins/target_gene/tgp/tests/test_pick_otus.py
```diff
@@ -66,6 +67,14 @@ def test_generate_pick_closed_reference_otus_cmd(self):
         self.assertEqual(obs, exp)
         self.assertEqual(obs_dir, join(output_dir, 'cr_otus'))
 
+    def test_generate_sortmerna_tgz(self):
+        outdir = mkdtemp()
+        self._clean_up_files.append(outdir)
+        obs = generate_sortmerna_tgz(outdir)
+        exp = ("tar zcf %s/sortmerna_picked_otus.tgz "
```
ElDeveloper: We should never hit the situation where there's spaces in the filepaths right? or if we could maybe worth enclosing the paths between quotes?

qiita_db/support_files/patches/python_patches/41.py
antgonza: I don't think so ... perhaps not loading into memory the files and let the system deal with that?

ElDeveloper:
Looks good, just two comments.

qiita_plugins/target_gene/tgp/tests/test_pick_otus.py
antgonza: All the files are created by the system and all our filenames are only alphanumeric. If you feel strong about it I can add quotes.

qiita_db/support_files/patches/python_patches/41.py
ElDeveloper: Is that a known issue? It just seems unnecessary to use a subprocess, since there's already a module to do this.

antgonza: No that I know of ... changing.

ElDeveloper: Thanks!

ElDeveloper:
:+1:

Issue #1719, opened 2016-03-30
Offload prep template creation/deletion from the webserver


josenavas:

antgonza:
Solved here: https://github.com/biocore/qiita/pull/1720

antgonza:
Not true, we still need to fix this one.

antgonza:
AFAIK, this has been solved. Please reopen if that's not the case.

Issue #1717, opened 2016-03-28
Catch errors in the API-proxy methods 


ElDeveloper:
As it stands right now, there are individual try/except blocks spread through some methods in the API proxy module, these methods should instead be decorated so that there's a general try/except block that catches any unexpected error and responds with the appropriate error message.

For more information see this discussion:
https://github.com/biocore/qiita/pull/1716/files/8365dceee043c29c50ea3a58e61887e96b3f99a2#r57650625

PR #1720, opened 2016-03-30
Offload template processing from webserver

josenavas:
:finnadie::finnadie::finnadie::finnadie::finnadie::finnadie::finnadie::finnadie:

This PR offloads the sample template create/update/delete and the prep template udpate from the webserver. The create/delete prep template are a bit more tricky given that those operations perform bigger changes in the interface, so I open an issue to keep track of it: https://github.com/biocore/qiita/issues/1719

Some of the fun stuff that starts happening here. The system becomes now way more distributed, as some of the processing is happening through moi and the interface is just checking it is done or not. It becomes really interesting at the time of testing the handlers, since they just submit and return right away, which means that we can't reset the DB at this point until the process running the job is done. Thus, I needed to put some waiting code on those tests.

In order to make the interface more responsive to this changes, we will need to introduce websockets, but I don't think it is blocking ATM.

### DIFF
--- a/qiita_pet/handlers/api_proxy/prep_template.py
+++ b/qiita_pet/handlers/api_proxy/prep_template.py
@@ -7,19 +7,25 @@
 # -----------------------------------------------------------------------------
 from __future__ import division
 import warnings
-
 from os import remove
 from os.path import basename
+from json import loads
+
 from natsort import natsorted
+from moi import r_client
 
 from qiita_core.util import execute_as_transaction
 from qiita_pet.handlers.api_proxy.util import check_access, check_fp
+from qiita_ware.context import safe_submit
+from qiita_ware.dispatchable import update_prep_template
 from qiita_db.metadata_template.util import load_template_to_dataframe
 from qiita_db.util import convert_to_id, get_files_from_uploads_folders
 from qiita_db.study import Study
 from qiita_db.user import User
 from qiita_db.ontology import Ontology
 from qiita_db.metadata_template.prep_template import PrepTemplate
+
+PREP_TEMPLATE_KEY_FORMAT = 'prep_template_%s'
 
 
 def _get_ENA_ontology():
@@ -124,6 +130,25 @@ def prep_template_ajax_get_req(user_id, prep_id):
 
     ontology = _get_ENA_ontology()
 
+    job_id = r_client.get(PREP_TEMPLATE_KEY_FORMAT % prep_id)
+    if job_id:
+        redis_info = loads(r_client.get(job_id))
+        processing = redis_info['status_msg'] == 'Running'
+        if processing:
+            alert_type = 'info'
+            alert_msg = 'This prep template is currently being updated'
+        else:
+            alert_type = redis_info['return']['status']
+            alert_msg = redis_info['return']['message']
+    else:
+        processing = False
+        alert_type = ''
+        alert_msg = ''
+
+    alert_msg = alert_msg.replace('\n', '</br>')
+
+    editable = Study(study_id).can_edit(User(user_id)) and not processing
+
     return {'status': 'success',
             'message': '',
             'name': name,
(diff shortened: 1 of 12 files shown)


qiita_pet/handlers/api_proxy/prep_template.py
```diff
@@ -124,6 +130,25 @@ def prep_template_ajax_get_req(user_id, prep_id):
...
+            alert_msg = 'This prep template is currently being updated'
+        else:
+            alert_type = redis_info['return']['status']
+            alert_msg = redis_info['return']['message']
+    else:
+        processing = False
+        alert_type = ''
+        alert_msg = ''
+
+    alert_msg = alert_msg.replace('\n', '</br>')
```
antgonza: Do we really need this? Perhaps it will make more sense to have it in line 143.

qiita_pet/handlers/api_proxy/sample_template.py
```diff
@@ -214,38 +214,64 @@ def sample_template_summary_get_req(samp_id, user_id):
...
+            alert_type = 'info'
+            alert_msg = 'This sample template is currently being processed'
+        else:
+            alert_type = redis_info['return']['status']
+            alert_msg = redis_info['return']['message']
+    else:
+        processing = False
+        alert_type = ''
+        alert_msg = ''
+    alert_msg = alert_msg.replace('\n', '</br>')
```
antgonza: Perhaps in line 231?

qiita_ware/context.py
```diff
@@ -262,3 +264,23 @@ def _submit(ctx, channel, f, *args, **kwargs):
 # likely want this in qiita_ware.__init__
 context = Dispatch()
 submit = partial(_submit, context)
+
+
+def safe_submit(*args, **kwargs):
+    """Safe wraper for the submit function
+
+    There are cases in which a race condition may occur: submit returns the
+    job id but moi still did not had time to actually submit the job. In some
```
antgonza: rewrite

antgonza:
A few comments.

qiita_pet/handlers/api_proxy/prep_template.py
josenavas: move to 142

qiita_pet/handlers/api_proxy/sample_template.py
josenavas: move to 230

qiita_ware/context.py
```diff
@@ -262,3 +264,23 @@ def _submit(ctx, channel, f, *args, **kwargs):
...
+
+def safe_submit(*args, **kwargs):
+    """Safe wraper for the submit function
+
+    There are cases in which a race condition may occur: submit returns the
+    job id but moi hasn't submitted the job. In some cases this is not
+    acceptable, so this wrapper makes sure that the job_id
+    is returned only once the job has already been submitted.
+
+    From previous tests, the while loop is executed ~2 times, so there is no
```
ElDeveloper: no -> not

qiita_ware/dispatchable.py
```diff
@@ -42,3 +42,170 @@ def create_raw_data(filetype, prep_template, filepaths):
 def copy_raw_data(prep_template, artifact_id):
     """Creates a new raw data by copying from artifact_id"""
     Artifact.copy(Artifact(artifact_id), prep_template)
+
+
```
ElDeveloper: It seems like all these functions could also be benefitted from an approach like #1717

qiita_ware/test/test_dispatchable.py
```diff
@@ -0,0 +1,76 @@
...
+                          'differences between the data stored in the DB and '
+                          'the new data provided'}
+        self.assertEqual(obs['status'], exp['status'])
+        self.assertItemsEqual(obs['message'].split('\n'),
+                              exp['message'].split('\n'))
+
+    def test_delete_sample_template(self):
+        obs = delete_sample_template(1)
+        exp = {'status': 'danger',
+               'message': 'Sample template can not be erased because there '
```
ElDeveloper: can not -> cannot

ElDeveloper:
Looks good!

qiita_ware/context.py
josenavas: Done

qiita_ware/dispatchable.py
josenavas: Good point, thanks for linking to the issue

PR #1727, opened 2016-03-31
Fix user verify

josenavas:


qiita_db/test/test_user.py
```diff
@@ -305,12 +305,6 @@ def test_verify_code(self):
                  WHERE email = 'new@test.bar' AND dflt = true"""
         self.assertEqual(self.conn_handler.execute_fetchone(sql)[0], 2)
 
-        # Make sure new system messages are linked to user
```
squirrelo: This needs to be put in test_create to make sure this test is still done.

squirrelo:
General question: Why move this one piece from verify to create? It was originally in verify because a user can't log in until they are verified, so we saved extra message rows for users that sat unverified. Also, any messages that expire before a user verifies shouldn't matter since they expired before the user can log in and see them.

antgonza:
@squirrelo the original code has a pathological case: if a user registers, then messages get added, and the user tries to verify their email, the system will fail cause the messages have already been added:
```
ValueError: Error running SQL query:
Query: INSERT INTO qiita.message_user (email, message_id)
                             SELECT %s, message_id FROM qiita.message
                             WHERE expiration > %s
Arguments: ['my_email@email.com', datetime.datetime(2016, 3, 31, 4, 49, 43, 242531)]
Error: duplicate key value violates unique constraint "idx_message_user"
DETAIL:  Key (email, message_id)=(michael.vermeulen0@outlook.com, 16) already exists.
```

qiita_db/test/test_user.py
antgonza: Agree and more importantly we need to make sure the pathological case doesn't happens. I have created: https://github.com/biocore/qiita/pull/1731

antgonza:
Closing in favor of: https://github.com/biocore/qiita/pull/1731

PR #1741, opened 2016-04-02
Delete artifacts from new gui

antgonza:


qiita_pet/handlers/api_proxy/artifact.py
```diff
@@ -84,7 +84,8 @@ def artifact_summary_get_request(user_id, artifact_id):
 
     buttons = []
     btn_base = (
-        '<button onclick="set_artifact_visibility(\'%s\', {0})" '
+        '<button onclick="if (confirm(\'Are you sure you want to %s '
+        'artifact id: {0}?\')) {{ set_artifact_visibility(\'%s\', {0}) }}" '
```
wasade: are the jinja templates dynamic?

wasade: (confused with the `{{ ... }}`)

antgonza: Yes, they are dynamic and the {{ is to scape  one {

josenavas:
:+1:

Issue #1395, opened 2015-07-23
The study metadata should be added to the QIIME mapping file


ElDeveloper:
The information about the PI, study title, study alias, etc should be added as extra columns to the QIIME mapping file.

PR #1742, opened 2016-04-02
fix #1693

### LINKED ISSUE #1693: Automatically including TITLE in metaanalysis would be handy
...as it makes partitioning the samples trivial

antgonza:


qiita_db/analysis.py
```diff
@@ -1011,6 +1011,10 @@ def _build_mapping_file(self, samples, rename_dup_samples=False):
                     samps = set(samps) - all_ids
                     all_ids.update(samps)
 
+                # appending the study title
+                qm['qiita_study_title'] = qdb.artifact.Artifact(
```
ElDeveloper: Historically (I think) the name of the column has varied from `title` to `study_title`, etc, do we want to add yet another column name? I think the answer maybe is yes, because some studies have a legacy `title` column that is inaccurate right?

ElDeveloper:
:+1: looks good, just one small question that shouldn't be blocking.

qiita_db/analysis.py
antgonza: Correct. Basically, this way we make sure that legacy and new ones have it. Also the reason I prepended with qiita_

antgonza:
After going over the issues I realized that a few extra minor changes will also fix #1395 so decided to add them. Note that the delete of the prep template also had a small bug that is being fixed here.

qiita_db/analysis.py
```diff
@@ -1011,6 +1011,18 @@ def _build_mapping_file(self, samples, rename_dup_samples=False):
                     samps = set(samps) - all_ids
                     all_ids.update(samps)
 
+                # appending study metadata to the analsis
```
wasade: Typo in comment

qiita_pet/templates/study_ajax/prep_summary.html
```diff
@@ -242,9 +242,8 @@
     if(confirm("Are you sure you want to delete prep template " + prep_id + "?")) {
       // Perform the AJAX call to delete
       $.ajax({
-        url: '/prep_template/',
+        url: '/prep_template/?prep-template-id=' + prep_id,
```
wasade: Isn't having "prep-template-id" redundant? Why not just "ID"?

wasade:
Two small comments

qiita_db/analysis.py
antgonza: Fixing.

qiita_pet/templates/study_ajax/prep_summary.html
antgonza: Perhaps but that's what the full system expects at this point. If you feel strongly about this, could you open an issue?

qiita_db/analysis.py
```diff
@@ -1011,6 +1011,18 @@ def _build_mapping_file(self, samples, rename_dup_samples=False):
...
+                study = qdb.artifact.Artifact(aid).study
+                study_owner = study.owner
+                study_info = study.info
+                pi = study_info['principal_investigator']
+                qm['qiita_study_title'] = study.title
+                qm['qiita_study_alias'] = study.info['study_alias']
+                qm['qiita_owner'] = '%s (%s)' % (study_owner.info['name'],
+                                                 study_owner.email)
+                qm['qiita_principal_investigator'] = '%s (%s)' % (
+                    pi.name, pi.email)
```
ElDeveloper: If we include the email, then we are absolutely including the `@` symbol, which is an unsupported character as per the QIIME mapping file description, I think this is fine with just the PI's name.

antgonza: Excellent point, removing.

ElDeveloper:
:+1: ready for merge once tests pass.

ackermag:
so delete is available?

antgonza:
Yup, note that I just realized that it's blocking; will need to fix that.

PR #1749, opened 2016-04-05
small fix

antgonza:
Turns out we still have none valid ascii chars in old studies and when you try to make an analysis it fails with:
```
UnicodeEncodeError: 'ascii' codec can't encode character u'\xca' in position 576: ordinal not in range(128)
```

This solves it.

### DIFF
--- a/qiita_db/analysis.py
+++ b/qiita_db/analysis.py
@@ -1039,7 +1039,7 @@ class Analysis(qdb.base.QiitaStatusObject):
             _, base_fp = qdb.util.get_mountpoint(self._table)[0]
             mapping_fp = join(base_fp, "%d_analysis_mapping.txt" % self._id)
             merged_map.to_csv(mapping_fp, index_label='#SampleID',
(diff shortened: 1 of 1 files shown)


ElDeveloper:
Can we add a test for this?

Yoshiki Vázquez-Baeza

antgonza:
Sure, how?

antgonza:
Expanding: didn't add it cause I didn't see a way to do it. Also, cause the fix is adding a parameter in an external tool ....

ElDeveloper:
Is it not possible to add a study with non-ASCII characters and create a "test" analysis that would have otherwise failed during this step of the analysis creation?

antgonza:
At this point, I don't think that's actually possible without hacking something out: bypassing the current tests in place so this doesn't get inserted. The issue came from old studies ....

ElDeveloper:
Oh, OK. 👍 then.

squirrelo:
:+1:

PR #1750, opened 2016-04-05
fixing empty table error

antgonza:
This error is that in the previous version we will stop analysis execution if all the samples within a study were rarefied out. However, this is not desirable. In this new version, it will raise an error only if the final merged biom there are no samples. Note that the merged table can be the result of only one biom.

### DIFF
--- a/qiita_db/analysis.py
+++ b/qiita_db/analysis.py
@@ -946,9 +946,10 @@ class Analysis(qdb.base.QiitaStatusObject):
                         new_table = new_table.merge(biom_table)
 
                 if not new_table or len(new_table.ids()) == 0:
-                        raise RuntimeError(
-                            "All samples filtered out from Artifact %s due "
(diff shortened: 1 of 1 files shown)


qiita_db/analysis.py
```diff
@@ -947,6 +945,11 @@ def _build_biom_tables(self, grouped_samples, rarefaction_depth=None,
                     else:
                         new_table = new_table.merge(biom_table)
 
+                if not new_table or len(new_table.ids()) == 0:
+                        raise RuntimeError(
+                            "All samples filtered out from Artifact %s due "
```
squirrelo: This text is not clear. "All sample filtered from artifact %s: No samples in artifact matched selected samples."

squirrelo:
One comment.

ElDeveloper:
👍 ready for merge, though the tests seem to still be queued.

PR #1755, opened 2016-04-08
Share studies in studies

squirrelo:
Fixes #1730 and #1092 
This adds the ability to share studies from the actual study page, as well as fixing the issue of showing all emails in the HTML of the page. Now the autocomplete is done using AJAX and the select2 javascript library.

screen shot 2016-04-08 at 00 24 09

The share modal looks and acts the same
screen shot 2016-04-08 at 00 24 31

PR #1758, opened 2016-04-08
Share analyses

### LINKED ISSUE #1615: allow sharing analysis output
It makes sense to share the (meta)analysis result to other users instead of re-creating the analysis every time various collaborators need. Gail believe qiita doesn't allow it yet.

squirrelo:
Fixes #1615 

Requires #1755 

This adds the ability to share analyses to the completed analysis page. It also adds messaging to sharing, so users will receive a message when a study or analysis is shared or unshared with them. 

screen shot 2016-04-08 at 11 16 56

### DIFF
--- a/qiita_db/analysis.py
+++ b/qiita_db/analysis.py
@@ -684,13 +684,14 @@ class Analysis(qdb.base.QiitaStatusObject):
             The user to share the analysis with
         """
         # Make sure the analysis is not already shared with the given user
-        if user.id in self.shared_with or user.id == self.owner:
+        if user.id == self.owner:
             return
 
-        sql = """INSERT INTO qiita.analysis_users (analysis_id, email)
-                 VALUES (%s, %s)"""
-        qdb.sql_connection.TRN.add(sql, [self._id, user.id])
-        qdb.sql_connection.TRN.execute()
+        with qdb.sql_connection.TRN:
+            sql = """INSERT INTO qiita.analysis_users (analysis_id, email)
+                     VALUES (%s, %s)"""
+            qdb.sql_connection.TRN.add(sql, [self._id, user.id])
+            qdb.sql_connection.TRN.execute()
 
     def unshare(self, user):
         """Unshare the analysis with another user
@@ -700,10 +701,11 @@ class Analysis(qdb.base.QiitaStatusObject):
         user: User object
             The user to unshare the analysis with
         """
-        sql = """DELETE FROM qiita.analysis_users
-                 WHERE analysis_id = %s AND email = %s"""
-        qdb.sql_connection.TRN.add(sql, [self._id, user.id])
-        qdb.sql_connection.TRN.execute()
+        with qdb.sql_connection.TRN:
+            sql = """DELETE FROM qiita.analysis_users
+                     WHERE analysis_id = %s AND email = %s"""
+            qdb.sql_connection.TRN.add(sql, [self._id, user.id])
+            qdb.sql_connection.TRN.execute()
 
     def add_samples(self, samples):
         """Adds samples to the analysis
--- a/qiita_db/analysis.py
+++ b/qiita_db/analysis.py
@@ -683,17 +683,14 @@ class Analysis(qdb.base.QiitaStatusObject):
         user: User object
             The user to share the analysis with
         """
-        with qdb.sql_connection.TRN:
-            self._lock_check()
+        # Make sure the analysis is not already shared with the given user
+        if user.id in self.shared_with or user.id == self.owner:
+            return
 
-            # Make sure the analysis is not already shared with the given user
-            if user.id in self.shared_with:
-                return
-
-            sql = """INSERT INTO qiita.analysis_users (analysis_id, email)
-                     VALUES (%s, %s)"""
-            qdb.sql_connection.TRN.add(sql, [self._id, user.id])
-            qdb.sql_connection.TRN.execute()
+        sql = """INSERT INTO qiita.analysis_users (analysis_id, email)
+                 VALUES (%s, %s)"""
+        qdb.sql_connection.TRN.add(sql, [self._id, user.id])
+        qdb.sql_connection.TRN.execute()
 
     def unshare(self, user):
         """Unshare the analysis with another user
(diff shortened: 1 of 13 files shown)


antgonza:
Conflicts, please solve.

squirrelo:
done, ready for review.

qiita_db/analysis.py
```diff
@@ -683,13 +683,11 @@ def share(self, user):
...
             The user to share the analysis with
         """
-        with qdb.sql_connection.TRN:
-            self._lock_check()
-
-            # Make sure the analysis is not already shared with the given user
-            if user.id in self.shared_with:
-                return
+        # Make sure the analysis is not already shared with the given user
+        if user.id == self.owner:
```
antgonza: I think you are missing `and user.id in self.shared_with`, right?

qiita_pet/handlers/util.py
```diff
@@ -89,3 +89,32 @@ def to_int(value):
     except ValueError:
         raise HTTPError(400, "%s cannot be converted to an integer" % value)
     return res
+
+
+@execute_as_transaction
+def get_shared_links(obj):
+    """Createse HTML links for users obj is shared with
```
antgonza: Rewrite please.

qiita_pet/handlers/util.py
```diff
@@ -89,3 +89,32 @@ def to_int(value):
     except ValueError:
         raise HTTPError(400, "%s cannot be converted to an integer" % value)
     return res
+
+
+@execute_as_transaction
+def get_shared_links(obj):
```
antgonza: why rewrite this one vs. using the one for studies?

antgonza:
A few comments.

qiita_pet/handlers/util.py
squirrelo: This is the one from studies. listing_handlers.py has it removed and moved here, since this function is now used in more than just that file.

qiita_db/analysis.py
squirrelo: fixed.

squirrelo:
Comments addressed

qiita_pet/handlers/analysis_handlers.py
```diff
@@ -349,3 +355,49 @@ class AnalysisSummaryAJAX(BaseHandler):
...
+        analysis = Analysis(analysis_id)
+        if self.current_user != analysis.owner:
+            raise HTTPError(403, 'User %s does not have permissions to share '
+                            'analysis %s' % (
+                                self.current_user.id, analysis.id))
+
+        selected = self.get_argument('selected', None)
+        deselected = self.get_argument('deselected', None)
+
+        if selected is not None:
```
josenavas: What happens if both are `None`?

squirrelo: No share or unshare actions, but the user email links and user list is still returned. This is leveraged for initial population of the shared list in the page.

qiita_pet/static/js/sharing.js
```diff
@@ -18,18 +18,18 @@ $(document).ready(function () {
...
   });
 
   $('#shares-select').on("select2:unselect", function (e) {
-    update_share({deselected: e.params.data.text});
+    update_share(e.target.classList[0], {deselected: e.params.data.text});
   });
 });
 
-function modify_sharing(study_id) {
+function modify_sharing(share_type, id) {
```
josenavas: Looks like the indentation in this function is messed up.

qiita_pet/static/js/sharing.js
```diff
@@ -1,4 +1,4 @@
-var current_study = null;
+var current_id = null;
```
josenavas: Can the global variable be removed? Recommend taking a look at the [data](  http://www.w3schools.com/tags/att_global_data.asp) variables. They're pretty useful and they allowed me to remove all the global variables that was in the previous code.

josenavas:
A couple of comments

qiita_pet/static/js/sharing.js
squirrelo: This was discussed with @ElDeveloper in the last PR and it is not feasible to do at this time.

josenavas: All global variables are possible to remove if you want to - please check the link I sent above and remove them.

qiita_pet/static/js/sharing.js
squirrelo: fixed

squirrelo:
Comments addressed

qiita_pet/templates/analysis_results.html
```diff
@@ -1,8 +1,15 @@
...
 {%block head%}
-
-
+<link rel="stylesheet" href="/static/vendor/css/select2.min.css" type="text/css">
+<script type="text/javascript" src="/static/vendor/js/select2.min.js"></script>
+<script type="text/javascript" src="/static/js/sharing.js"></script>
+<script type="text/javascript">
+$(document).ready(function () {
+  current_id = {{analysis_id}};
+  update_share('analysis');
```
josenavas: Pass the `current_id` as a parameter.

qiita_pet/templates/study_ajax/base_info.html
```diff
@@ -3,8 +3,8 @@
 <script type="text/javascript" src="/static/js/sharing.js"></script>
 <script type="text/javascript">
 $(document).ready(function () {
-  current_study = {{study_info['study_id']}};
-  update_share();
+	current_id = {{study_info['study_id']}};
+	update_share('study');
```
josenavas: Same here and fix indentation.

qiita_pet/templates/analysis_results.html
squirrelo: This does not work for the on select and deselect that are in the sharing.js

josenavas: Can you elaborate? Are you saying that you can't modify the functions that you created?

squirrelo: No, I'm saying that the calls on line 21 and 25 of sharing.js have no idea what the current_id is if it is not a global variable, so this suggestion will not work.

josenavas: I don't see current_id used in those lines.

squirrelo: Exactly. There is no way of passing in current_id there as a parameter if it is not a global variable since it is unknown, so modifying the function will not work. I have a way of using data variables I am testing right now.

squirrelo:
Now uses data attribute instead of global.

qiita_pet/static/js/sharing.js
```diff
@@ -1,5 +1,3 @@
-var current_study = null;
-
 $(document).ready(function () {
```
josenavas: Actually I just saw this - this is a library, why does it has a `$(document).ready` call? It seems weird that this is in the library and this may be the source of the difficulties that you where finding. If this is used to initialize the different parts of this library, it may be useful to actually have a `sharing_init` call that you call in the `$(document).ready` of the actual HTML pages that is using this library.

squirrelo: Because you can have as many document.ready calls in a single page as needed, and this also standardizes naming and initialization of the sharing dropdown for use in the other helper functions.

squirrelo:
Closed to merge onto portals branch.

PR #1759, opened 2016-04-08
Fix showing metadata header problems

### LINKED ISSUE #1745: Incorrectly formatted headers break the system
Having a space or a slash in the header names breaks the system to the point of not showing errors and breaking the javascript + database for submitting new sample metadata files to correct the situation.

squirrelo:
This fixes #1745 by escaping the text to be shown in the prep and sample metadata messages before sending into the javascript, then unescaping it before being shown. This removes all characters that could interfere with the javascript working.

See the broken /slash and broken \slash columns above that would otherwise break the system.
screen shot 2016-04-08 at 14 53 39

### DIFF
--- a/qiita_pet/handlers/study_handlers/prep_template.py
+++ b/qiita_pet/handlers/study_handlers/prep_template.py
@@ -9,6 +9,7 @@ from __future__ import division
 from os.path import join
 
 from tornado.web import authenticated
+from tornado.escape import url_escape
 import pandas as pd
 
 from qiita_pet.handlers.util import to_int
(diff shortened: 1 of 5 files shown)


antgonza:
Could you add a patch to current messages? Basically, a loop over the redis messages so they can be displayed. Thanks.

squirrelo:
Not sure what you are asking, could you clarify? The system messages display fine:
screen shot 2016-04-09 at 08 11 55

However, with or without a system message the error messages are showing up above the nav bar, which shouldn't happen. This is also happening on the main deploy currently. I'll try and figure out why this is happening and fix it.

antgonza:
Sorry for the poor wording. What I was asking is to add a fix for the current messages in the system, not only the new ones. However, I think [this line](https://github.com/biocore/qiita/pull/1759/files#diff-4d2364edf48c2787073cdcdc4cc54d77R97) does this, is that right?

squirrelo:
Yes, that is correct. It url encodes any message as it comes out of the system, so it will work with current as well as future messages.

antgonza:
👍

PR #1762, opened 2016-04-11
Offload artifact deletion from webserver

### LINKED ISSUE #1747: artifact delete is blocking in the current interface
Need to fix ASAP.

josenavas:
Fixes #1747


qiita_pet/templates/study_ajax/artifact_summary.html
```diff
@@ -1,5 +1,24 @@
 <script type="text/javascript">
   /*
+   * Function to delete an artifact
+   *
+   * @param aid The artifact to be removed
+   *
+   * This function executes an AJAX query to delete an artifact from the system
+   */
+  function delete_artifact(aid) {
+    $.post("/artifact/", {artifact_id: +aid })
```
antgonza: Needs to change based on portal changes. I think we should merge the other branch before adding new functionality, do you agree?

josenavas: Yup not a problem

qiita_pet/handlers/api_proxy/tests/test_artifact.py
```diff
@@ -310,20 +318,6 @@ def test_artifact_patch_request_errors(self):
                           'operations: replace'}
         self.assertEqual(obs, exp)
 
-    def test_artifact_delete_req(self):
```
squirrelo: Why delete these tests? They're still important for making sure the call returns properly.

PR #1789, opened 2016-04-15
Storing template errors/warnings correctly in redis

josenavas:
Before we where "trusting" moi to store these error messages. However, moi sets an expiration time in the keys, which makes the interface to break once the key expires. This changes the way the keys are stored to actually store the information in new keys, which do not expire.

### DIFF
--- a/qiita_pet/handlers/api_proxy/prep_template.py
+++ b/qiita_pet/handlers/api_proxy/prep_template.py
@@ -118,9 +118,11 @@ def prep_template_ajax_get_req(user_id, prep_id):
                 alert_type = 'info'
                 alert_msg = 'This prep template is currently being updated'
             elif redis_info['status_msg'] == 'Success':
+                alert_type = redis_info['return']['status']
+                alert_msg = redis_info['return']['message']
                 payload = {'job_id': None,
-                           'status': redis_info['return']['status'],
-                           'message': redis_info['return']['message']}
+                           'status': alert_type,
+                           'message': alert_msg}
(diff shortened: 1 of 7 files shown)


antgonza:
👍  once tests pass - there is a real minor error.

josenavas:
Fixed - let the tests run since I couldn't reproduce in my system, but I knew it was a race condition as soon as I saw the code.

josenavas:
@ElDeveloper @mortonjt any chance you can take a look?

qiita_pet/handlers/api_proxy/prep_template.py
```diff
@@ -107,16 +107,33 @@ def prep_template_ajax_get_req(user_id, prep_id):
...
-        if processing:
-            alert_type = 'info'
-            alert_msg = 'This prep template is currently being updated'
+    job_info = r_client.get(PREP_TEMPLATE_KEY_FORMAT % prep_id)
+    if job_info:
+        job_info = loads(job_info)
+        job_id = job_info['job_id']
+        if job_id:
+            redis_info = loads(r_client.get(job_id))
+            processing = redis_info['status_msg'] == 'Running'
```
ElDeveloper: Is `processing` being used anywhere else? Would likely make sense to just drop `redis_info.....` into the `if`.

ElDeveloper: If that's the case, then it would be worth checking if the same is possible in `sample_template.py`.

ElDeveloper: nix that, it is being used in the return.

PR #1793, opened 2016-04-16
Fixing artifact REST api

josenavas:


antgonza:
👍

qiita_db/handlers/artifact.py
```diff
@@ -124,32 +112,24 @@ def get(self, artifact_id):
...
         -------
         dict
-            Format:
-            {'success': bool,
-             'error': str,
-             'mapping': str}
-             - success: whether the request is successful or not
-             - error: in case that success is false, it contains the error msg
-             - mapping: the filepath to the mapping file
+            {'mapping': str}
```
ElDeveloper: Should this still be a dictionary if it is only a string? i.e. should this function just return the filepath instead?

ElDeveloper: Upon further reading, yeah this makes sense.

PR #1795, opened 2016-04-16
Fixing QiitaClient

josenavas:
Depends on #1794 so review/merge that one first.

The QiitaClient is improved to reduce workload on the plugin developer.

PR #1796, opened 2016-04-16
Fixing BIOM plugin

josenavas:
This depends on #1795 so review/merge that one first


antgonza:
👍 @ElDeveloper, could you take a look?

ElDeveloper:
👍

PR #1799, opened 2016-04-18
Fix target gene type plugin

josenavas:
This depends on https://github.com/biocore/qiita/pull/1796 so review/merge that one first.

PR #1800, opened 2016-04-18
Fix target gene plugin

josenavas:
This depends on https://github.com/biocore/qiita/pull/1799 so review/merge that one first


antgonza:
👍 @ElDeveloper, could you take a look?

ElDeveloper:
👍

PR #1804, opened 2016-04-19
Fix 1783: display EBI values

antgonza:
This adds to the study page the EBI status and links once it's successful and also adds the status of the preps. After discussing with @ackermag we agreed that for the preps displaying a message was enough => not displaying all the accessions.


josenavas:
👍

qiita_pet/templates/study_ajax/prep_summary.html
```diff
@@ -353,7 +353,10 @@
 <!-- #samples and #columns -->
 <div class="row">
   <div class="col-md-12">
-    There are <b>{{num_samples}}</b> samples and <b>{{num_columns}}</b> columns in this preparation.
+    There are <b>{{num_samples}}</b> samples and <b>{{num_columns}}</b> columns in this preparation.</br>
+    {% if ebi_experiment_accessions %}
+      <b>EBI status: </b> Submitted
```
ElDeveloper: We have other statuses available right? Any reason why we are not showing the current status instead of just Submitted?

ElDeveloper:
This looks fine 👍

qiita_pet/templates/study_ajax/prep_summary.html
antgonza: Not really, this is for preps. This is, the study is what's being submitted/updated and the prep is only submitted/not-submitted and we know is submitted if the samples have ebi_experiment_accessions.

ElDeveloper: Got it.

PR #1806, opened 2016-04-22
Do not delete jobs when the artifact is removed

josenavas:
This will prevent one of the errors appearing in the system, while at the same time we will be able to gather some information about system usage, % of jobs failed, etc.

This can be useful for multiple reasons: check which commands the users use more, which is the command that generates more errors so we need to improve docs or the command itself, or even report utilization of the system in general.


antgonza:
👍

PR #1821, opened 2016-05-02
Improve study list

antgonza:
This improves the speed of generating the list of studies and solves an issue with the sample count. I plan to continue solving the other issues we chatter last week but these are already a lot of changes. 

In the test environment the study display for my user came down from 9.6 to 3.2 seconds.

### DIFF
--- a/qiita_db/util.py
+++ b/qiita_db/util.py
@@ -42,6 +42,7 @@ Methods
 
 from __future__ import division
 from future.builtins import zip
+from future.utils import viewitems
 from random import choice
 from string import ascii_letters, digits, punctuation
 from binascii import crc32
@@ -1227,7 +1228,7 @@ def generate_study_list(study_ids, build_samples):
     Parameters
     ----------
     study_ids : list of ints
-        The study ids to look for. Not existing ids will be ignored
+        The study ids to look for. Non-existing ids will be ignored
     build_samples : bool
         If true the sample information for each process artifact within each
         study will be included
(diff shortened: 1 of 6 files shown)


qiita_db/util.py
```diff
@@ -1219,3 +1219,159 @@ def supported_filepath_types(artifact_type):
...
+            del info['artifacts_visibility']
+
+            # pi info
+            info["pi"] = (info['pi_email'], info['pi_name'])
+            del info["pi_email"]
+            del info["pi_name"]
+
+            # shared with
+            info['shared'] = []
+            if info['shared_with_name']:
```
josenavas: Do you need the `if` statement? You should probably do the for loop directly, if it is an empty list it will not execute any iteration.

On the other hand, the `if` is over `shared_with_name` but from the for loop below it looks like the name may not exist. Should be better to do the if over `shared_with_email` which looks like is ensured to exist?

It is also possible that I'm missing something 😄

antgonza: You are right, I can remove this if.

qiita_db/util.py
```diff
@@ -1219,3 +1219,159 @@ def supported_filepath_types(artifact_type):
...
+            info['proc_data_info'] = []
+            if build_samples and info['artifact_biom_ids']:
+                to_loop = zip(
+                    info['artifact_biom_ids'], info['artifact_biom_dts'],
+                    info['artifact_biom_ts'], info['artifact_biom_params'])
+                for artifact_id, dt, ts, params in to_loop:
+                    proc_info = {'processed_date': str(ts)}
+                    proc_info['pid'] = artifact_id
+                    proc_info['data_type'] = dt
+                    del params['input_data']
```
josenavas: This is hard coded and it is not ensured that the parameter name will be `input_data`. I think it should check for the parameters that are artifacts and remove them.

Also, all the lines below here, are hard-coded for target gene - and I don't think they work in case that the reference used is UNITE (AFAIK there is no tree associated with it).

What is the outcome once we start adding the other processing pipelines? This may not work for a table resulting from HUMANN2.

What is going to happen with the BIOM artifacts uploaded by the user? They may not have this information.

qiita_pet/handlers/study_handlers/listing_handlers.py
```diff
@@ -327,11 +203,29 @@ def get(self, ignore):
...
+        for i in range(len_info):
+            info[i]['shared'] = ", ".join([study_person_linkifier(element)
+                                           for element in info[i]['shared']])
+            info[i]['pmid'] = ", ".join([pubmed_linkifier([element])
+                                         for element in info[i]['pmid']])
+            info[i]['publication_doi'] = ", ".join([
+                doi_linkifier([element])
+                for element in info[i]['publication_doi']])
+            info[i]['pi'] = study_person_linkifier(info[i]['pi'])
+            info[i]['ebi_info'] = '%s (%s)' % (
```
josenavas: should we create an `ebi_linkifier` function for consistency with the above?

qiita_pet/handlers/study_handlers/listing_handlers.py
```diff
@@ -327,11 +203,29 @@ def get(self, ignore):
             study_proc = proc_samples = None
         info = _build_study_info(self.current_user, study_proc=study_proc,
                                  proc_samples=proc_samples)
+        # linkifying data
```
josenavas: Can all this be done at the end of _build_study_info instead that in the handler? I think it will be easier to debug and even to offload at some point if we want to.

qiita_pet/templates/list_studies.html
```diff
@@ -82,17 +82,16 @@
...
-              { "data": "num_raw_data" },
               { "data": "shared" },
               { "data": "pi" },
               { "data": "pmid" },
-              { "data": "status" }
+              { "data": "status" },
+              { "data": "ebi_info" }
             ],
-            order: [[11, "desc"], [ 2, "asc" ]],
+            // order: [[11, "desc"], [ 2, "asc" ]],
```
josenavas: If this is not needed, should be removed?

josenavas:
Few comments. The SQL looks pretty scary and I'm unsure if it is worth it to add some comments on it to try to explain a bit everything that is done on it. Probably a bullet list explaining a bit the inner `SELECT`s may be helpful for rapid debugging and developer guidance - what others think?

qiita_db/util.py
antgonza: 1. I will add a way to automatically remove everything that is artifact in the parameter_type from qiita.command_parameter
2. If there is no tree it will return a None, which is fine.
3-4. Other processing pipelines - direct BIOM uploads: I guess that depends on how those are implemented and we will need to solve once they are added.

qiita_pet/handlers/study_handlers/listing_handlers.py
antgonza: I don't have strong feeling about this. Note that we have tests for this and _build_study_info so we are testing both the formatted and not formatted lists.

qiita_pet/handlers/study_handlers/listing_handlers.py
antgonza: sure

qiita_db/util.py
josenavas: BIOM uploads are already implemented - the parameters are None

qiita_db/util.py
```diff
@@ -1219,3 +1219,159 @@ def supported_filepath_types(artifact_type):
...
         return qdb.sql_connection.TRN.execute_fetchindex()
+
+
+def generate_study_list(study_ids, build_samples):
+    """Get general study information
+
+    Parameters
+    ----------
+    study_ids : list of ints
+        The study ids to look for. Not existing ids will be ignored
```
ElDeveloper: `Not` -> `Non-`

qiita_pet/templates/list_studies.html
```diff
@@ -82,17 +82,16 @@
...
             columnDefs: [
-              {type:'natural', targets:[4,5,6,10]},
+            //   {type:'natural', targets:[4,5,6,10]},
```
ElDeveloper: can you remove this line?

ElDeveloper:
👍

qiita_db/util.py
antgonza: BIOM: Cool, so they fall into 2, so we are fine.

josenavas:
Thanks @antgonza

PR #1823, opened 2016-05-04
Allowing multiple SFF files per run prefix

josenavas:
Right now, the value `raw_sff` is hardcoded, so the interface is fixed and those files can be processed.
We may want to discuss which is a better way of handling this in a general manner - i.e. is this case going to occur more often? If so, should we store that info in the DB? Others?

### DIFF
--- a/qiita_pet/handlers/api_proxy/studies.py
+++ b/qiita_pet/handlers/api_proxy/studies.py
@@ -226,9 +226,9 @@ def study_files_get_req(user_id, study_id, prep_template_id, artifact_type):
     uploaded = get_files_from_uploads_folders(study_id)
     pt = PrepTemplate(prep_template_id).to_dataframe()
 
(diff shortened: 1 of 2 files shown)


qiita_pet/handlers/api_proxy/studies.py
```diff
@@ -226,8 +226,9 @@ def study_files_get_req(user_id, study_id, prep_template_id, artifact_type):
     uploaded = get_files_from_uploads_folders(study_id)
     pt = PrepTemplate(prep_template_id).to_dataframe()
 
-    if (any(ft.startswith('raw_') for ft, _ in supp_file_types) and
-            'run_prefix' in pt.columns):
+    ftypes_it = (ft.startswith('raw_') for ft, _ in supp_file_types
```
antgonza: ftypes_if ??

antgonza:
One minor comment and then 👍  once the tests pass.

PR #1824, opened 2016-05-04
BLD: New pandas 🐼

ElDeveloper:


josenavas:
All test are passing - what do we do with this? We merge it or should be upgrade the test environment to use the last pandas and play with it in there?

ElDeveloper:
Hmmm, after seeing the error that @antgonza sent around, I would say
it's only worth adding this restriction if this solves the problem we
saw in the study that was not being updated.

PR #1829, opened 2016-05-05
update figure1.png and add source

colinbrislawn:


antgonza:
Thanks for updating. A couple of comments, you changed the font and color of each study title. I don't have a preference in color but white over light green makes it hard to read on the screen and I can imagine even more in a presentation or poster. Also, as you changed the font and you are using something not standard AI has to outline it -> it becomes not text, which makes it hard to edit. Could you change?

josenavas:
Agree with @antgonza comments. @antgonza once those comments are fixed - feel free to merge :) 👍

colinbrislawn:
Thanks! I'll update colors now... 

Unfortunately I don't have an AI license, and fonts across systems are hard. That font _should be_ Arial, and I assumed it would be ubiquitous. I'll try baking the font into the PDF file... Other suggestions welcome.

PR #1822, opened 2016-05-04
adding a prep/biom combination to the test db

antgonza:
After working on the study list page we realized that there is no direct biom/prep combination in the test environment. This PR add that and fixes all tests.

PR #1831, opened 2016-05-06
public/user listing studies

antgonza:
This solves an issue introduced in the initial listing PR, adds 2 lists with only one control set, and fixes some indentation issues in the js.

### DIFF
--- a/qiita_db/util.py
+++ b/qiita_db/util.py
@@ -1391,11 +1391,12 @@ def generate_study_list(study_ids, build_samples):
 
             # shared with
             info['shared'] = []
-            for name, email in zip(info['shared_with_name'],
-                                   info['shared_with_email']):
-                if not name:
-                    name = email
-                info['shared'].append((email, name))
+            if info['shared_with_name'] and info['shared_with_email']:
+                for name, email in zip(info['shared_with_name'],
+                                       info['shared_with_email']):
+                    if not name:
+                        name = email
+                    info['shared'].append((email, name))
             del info["shared_with_name"]
             del info["shared_with_email"]
 
@@ -1456,12 +1457,12 @@ def generate_study_list(study_ids, build_samples):
 
                     info["proc_data_info"].append(proc_info)
 
-        del info["artifact_biom_ids"]
-        del info["artifact_biom_dts"]
-        del info["artifact_biom_ts"]
-        del info["artifact_biom_params"]
-        del info['artifact_biom_cmd']
+            del info["artifact_biom_ids"]
+            del info["artifact_biom_dts"]
+            del info["artifact_biom_ts"]
+            del info["artifact_biom_params"]
+            del info['artifact_biom_cmd']
 
-        infolist.append(info)
+            infolist.append(info)
 
     return infolist
--- a/qiita_pet/handlers/study_handlers/listing_handlers.py
+++ b/qiita_pet/handlers/study_handlers/listing_handlers.py
@@ -180,10 +180,7 @@ class SearchStudiesAJAX(BaseHandler):
         if user != self.current_user.id:
             raise HTTPError(403, 'Unauthorized search!')
         if search_type not in ['user', 'public']:
-            self.clear()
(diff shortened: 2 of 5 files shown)


antgonza:
Depends on https://github.com/biocore/qiita/pull/1822, merge that one first

antgonza:
Ready for review, thanks!

ElDeveloper:
👍

qiita_db/util.py
```diff
@@ -1391,11 +1391,12 @@ def generate_study_list(study_ids, build_samples):
 
             # shared with
             info['shared'] = []
-            for name, email in zip(info['shared_with_name'],
-                                   info['shared_with_email']):
-                if not name:
-                    name = email
-                info['shared'].append((email, name))
+            if info['shared_with_name'] and info['shared_with_email']:
```
josenavas: It looks like these lines of code where bugs introduced in the previous PR, which means that the tests do not capture this case. I recommend to add a specific test that will capture this case so it doesn't repeat.

qiita_pet/handlers/study_handlers/listing_handlers.py
```diff
@@ -168,10 +174,16 @@ class SearchStudiesAJAX(BaseHandler):
     def get(self, ignore):
         user = self.get_argument('user')
         query = self.get_argument('query')
+        search_type = self.get_argument('search_type')
         echo = int(self.get_argument('sEcho'))
 
         if user != self.current_user.id:
             raise HTTPError(403, 'Unauthorized search!')
+        if search_type not in ['user', 'public']:
+            self.clear()
```
josenavas: `raise HTTPError(400, 'Not a valid search type')`

qiita_pet/templates/list_studies.html
```diff
@@ -21,7 +21,8 @@
 <script src="{% raw qiita_config.portal_dir %}/static/vendor/js/select2.min.js"></script>
 <script src="{% raw qiita_config.portal_dir %}/static/js/sharing.js"></script>
 <script type="text/javascript">
-var ajaxURL = "{% raw qiita_config.portal_dir %}/study/search/?&user={{current_user.id}}&sEcho=" + Math.floor(Math.random()*1001);
+var user_studies_ajaxURL = "{% raw qiita_config.portal_dir %}/study/search/?&user={{current_user.id}}&search_type=user&sEcho=" + Math.floor(Math.random()*1001);
```
josenavas: Can we avoid using global variables? Specially this values are used only in a single place.

josenavas:
Few comments. However, the sample selection doesn't work with this branch. Unsure where the bug is, but neither the `Add to analysis` nor the individual `Add` buttons for each biom table work.

qiita_db/util.py
antgonza: k

qiita_pet/handlers/study_handlers/listing_handlers.py
antgonza: The issue is that that error is logged in the DB and not shown in the GUI. Thus decided to go this route so it's shown.

qiita_pet/templates/list_studies.html
antgonza: yup, fixing.

qiita_pet/handlers/study_handlers/listing_handlers.py
josenavas: A user should never hit this error - this is more a safeguard for developers - or am I missing something?

antgonza: I was thinking in the restapi and that this should be the best method to access the list of studies.

josenavas: In that case you should be able to use the code above - is what I've been using in the qiita_db handlers. If the error message is not sent, then there is a problem in the function `write_error` on the base handler. I think the one in the qiita_db subdir is different and does it correctly.

PR #1835, opened 2016-05-11
Add cookie secret as part of config

squirrelo:
does what it says on the tin.

### DIFF
--- a/qiita_core/configuration_manager.py
+++ b/qiita_core/configuration_manager.py
@@ -10,6 +10,9 @@ from functools import partial
 from os.path import join, dirname, abspath, isdir
 from os import environ
 from future import standard_library
+from base64 import b64encode
+from uuid import uuid4
+import warnings
 
 from .exceptions import MissingConfigSection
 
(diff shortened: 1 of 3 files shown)


antgonza:
👍  once tests pass.

antgonza:
BTW this was deployed in the test environment and it works fine.

josenavas:
@antgonza I'm confused on the tests passing here - the encoded config file is not changed. I'm assuming that what is happening is that is just using 'None' as the COOKIE_SECRET. What should we do when the COOKIE_SECRET is not provided in the config file? Should we fail or generate one at random?

antgonza:
Nop, here the COOKIE_SECRET = SECRET. The failure is gonna be when we merge to master.

josenavas:
I guess I then misunderstood how it works - I though that the PR against master are the ones already using the encoded config file - otherwise the changes in the previous `portals` branch would not have failed.

qiita_core/configuration_manager.py
```diff
@@ -178,6 +178,7 @@ def _get_main(self, config):
         if not self.certificate_file:
             self.certificate_file = join(install_dir, 'qiita_core',
                                          'support_files', 'server.crt')
+        self.cookie_secret = config.get('main', 'COOKIE_SECRET')
```
josenavas: IMO, we should check if `self.cookie_secret` is `None` (as it is done for `self.key_file` below). If it is None, I would go back to the current behavior, generate a random one, but also raising a Warning to the user to let him now that it was not set, in case that he actually needs to set it (e.g. load balancing).

I don't think it is critical to the point of raising an error, since the usage of that config value is specific to some functionality (we didn't realize that we needed until we wanted to do load balancing), but not signing at all I think it's bad.

squirrelo:
Comment addressed

qiita_core/configuration_manager.py
squirrelo: Functionality added

josenavas:
👍

PR #1840, opened 2016-05-18
Removing qiita-client

josenavas:
It already exists in its own repository, here: https://github.com/qiita-spots/qiita_client

I need to do some modifications to the qiita-client and I realized that it is still living here, so in order to avoid duplication of code/effort I'm deleting it from the qiita repository.

The modifications that I need to do in the Qiita client are just adding some extra convenient functions for the plugin developers.


HannesHolste:
Please remember to remove this from `INSTALL.md` too:
```bash
 cd qiita_client
 pip install -e .
 ```

Thanks!

josenavas:
@antgonza possible a quick review in here?

PR #1843, opened 2016-05-23
rm unused handlers

antgonza:
@josenavas and I are validating the access rights/permissions of all pages and he identified a list of handlers that are not used anymore due to recent changes. 

Removing:
PreprocessingSummaryHandler
ComputeCompleteHandler
CreateRawData
PreprocessHandler
ProcessHandler


josenavas:
👍

PR #1845, opened 2016-05-25
Just updating to master

antgonza:
... so we avoid conflicts in future changes


josenavas:
Is it worth to wait for the tests or can we merge?

antgonza:
I think is fine either way, I'm not going to have the next PR until the end of the day ...

Issue #1847, opened 2016-05-26
Fix setup.py and update INSTALL.md installation documentation accordingly


HannesHolste:
The [ current one ](https://github.com/biocore/qiita/commit/378be825d8b4d147180774606a589f9f2051909f) is outdated, causing users to encounter various errors when trying to follow the tutorial.

I will update these instructions and add notes on the use of conda/miniconda and Mac OSX installation.

Can an admin assign me to this in GitHub?

antgonza:
Great, thanks!

HannesHolste:
Almost finished.

Currently wrangling with this issue: the package 'biom' uses
`from future.utils import string_types`
which apparently has been removed after future v.0.13.1 (see: https://github.com/biocore/scikit-bio/issues/691)

However, simply downgrading the local package version of future as suggested [here](https://github.com/biocore/scikit-bio/issues/691#issuecomment-57902757) did not work; it raises the same error.

`> qiita pet webserver
Traceback (most recent call last):
  File "/Users/hannes/miniconda2/envs/qiita/bin/qiita", line 24, in <module>
    import qiita_db as qdb
  File "/Users/hannes/miniconda2/envs/qiita/lib/python2.7/site-packages/qiita_db/__init__.py", line 13, in <module>
    import analysis
  File "/Users/hannes/miniconda2/envs/qiita/lib/python2.7/site-packages/qiita_db/analysis.py", line 25, in <module>
    from biom import load_table
  File "/Users/hannes/miniconda2/envs/qiita/lib/python2.7/site-packages/biom/__init__.py", line 51, in <module>
    from .table import Table
  File "/Users/hannes/miniconda2/envs/qiita/lib/python2.7/site-packages/biom/table.py", line 188, in <module>
    from future.utils import string_types
ImportError: cannot import name string_types
(qiita)`


@ElDeveloper any insight into this?

antgonza:
Strange, my guess is that you have something leftover because in my installation I have:
```bash
(qiita)14:24:15 qiita@validate-prep-fixes$ python -c 'import future; print future.__version__'
0.15.2
(qiita)14:24:41 qiita@validate-prep-fixes$ python -c 'import biom; print biom.__version__'
2.1.5
```

ElDeveloper:
[...]

PR #1850, opened 2016-05-27
Add missing definitions of package directories to setup.py

HannesHolste:
See [here](https://github.com/biocore/qiita/issues/1847#issuecomment-222020146).


setup.py
```diff
@@ -42,11 +42,13 @@
       test_suite='nose.collector',
       packages=['qiita_core',
                 'qiita_db',
+                'qiita_db/handlers',
                 'qiita_db/metadata_template/',
-                'qiita_pet',
+                'qiita_pet/',
```
antgonza: Why do you need the final slash here but not on the other folders?

HannesHolste: typo because I was testing exactly that. Turns out it doesn't matter if it has slash or no slash.

antgonza: OK, should be consistent, just because? No strong feelings about it so up to you ...

antgonza:
👍 once tests pass.

PR #1851, opened 2016-05-27
validating prep templates

antgonza:
As you can see this became way larger than expected. 

Anyway, the main tasks here are:
- _clean_validate_template doesn't check restriction anymore, there is a new validate method that generates the warnings based on sample/value. Next PR will add to DB and we will need to discuss how will it be the best way to deal with this.
- _clean_validate_template now also checks for PgSQL reserve words and invalid chars in the column names
- the previous change made impossible to test that on sample/prep info creation failure the rollback works fine. After discussing with @josenavas we decided that this was fine cause this was an extra test for the actual rollback function
- the new validate function uses the data type (this had to change to fit python vs. PgSQL) of the restrictions to try to cast
- due to the new validations I had to change a lot of tests that were relying on data cast for them to work. Mainly date and NULL value conversions.
- I think that's it.


antgonza:
This should be ready for initial review. I have fixed all but one errors as fixing the last one will require changes to the DB, which I will do in the next PR. The failure is because lat/long are still defined as float and it fails when searching with the new ebi-ena null values:
```
Error: invalid input syntax for type double precision: "not applicable"
```

In the last commit I also removed one test from update_category where we used to check that if a column was int it will not allow to insert a string => basically why we are doing all these changes.

qiita_db/metadata_template/base_metadata_template.py
```diff
@@ -507,15 +508,13 @@ def _clean_validate_template(cls, md_template, study_id, restriction_dict,
...
-        restriction_dict : dict of {str: Restriction}
-            A dictionary with the restrictions that apply to the metadata
         current_columns : iterable of str, optional
             The current list of metadata columns
 
         Returns
         -------
         md_template : DataFrame
-            Cleaned copy of the input md_template
+            ceed copy of the input md_template
```
josenavas: `ceed` -> `Cleaned`?

qiita_db/metadata_template/base_metadata_template.py
```diff
@@ -541,38 +540,35 @@ def _clean_validate_template(cls, md_template, study_id, restriction_dict,
         # we don't modify the user one
         md_template = md_template.copy(deep=True)
 
+        # In the database, all the column headers are lowercase
+        md_template.columns = [c.lower() for c in md_template.columns]
+        # validating pgsql reserved words not to be column headers
+        current_headers = set(md_template.columns.values)
+        reserved_words = qdb.metadata_template.util.get_pgsql_reserved_words()
+        overlap = reserved_words & current_headers
+        if overlap:
```
josenavas: Can this check be included in `get_invalid_column_names`?

qiita_db/metadata_template/constants.py
```diff
@@ -14,32 +14,32 @@
 # A dict containing the restrictions that apply to the sample templates
 SAMPLE_TEMPLATE_COLUMNS = {
     # The following columns are required by EBI for submission
-    'EBI': Restriction(columns={'collection_timestamp': 'timestamp',
-                                'physical_specimen_location': 'varchar',
-                                'taxon_id': 'integer',
-                                'scientific_name': 'varchar'},
+    'EBI': Restriction(columns={'collection_timestamp': 'datetime',
+                                'physical_specimen_location': 'str',
```
josenavas: Instead of storing the types as strings, use the actual types. This way, when checking the types in [this line](https://github.com/biocore/qiita/pull/1851/files#diff-fe8c1935d9373ee4e64757a20e79c142R1464) yo don't need to call `eval`, you will be able to do `datatype(val)`

qiita_db/metadata_template/base_metadata_template.py
```diff
@@ -1426,3 +1422,53 @@ def _update_accession_numbers(self, column, values):
...
+                for column, datatype in viewitems(restriction.columns):
+                    for sample, val in viewitems(self.get_category(column)):
+                        # ignore if valid null value
+                        if val in valid_null:
+                            continue
+                        # test values
+                        if datatype == 'datetime':
+                            val = str(val)
+                            try:
+                                datetime.strptime(val, '%m/%d/%y %H:%M:%S')
```
josenavas: I would like the input from @ackermag, if I recall correctly the column `collection_timestamp` was going to change and it can store either a full timestamp or just the date.

qiita_db/metadata_template/test/test_sample_template.py
```diff
@@ -238,50 +239,47 @@ class BaseTestSampleTemplate(TestCase):
...
-                        datetime(2014, 5, 29, 12, 24, 51),
+                        'collection_timestamp': '05/29/14 12:24:15',
                         'host_subject_id': 'NotIdentified',
                         'Description': 'Test Sample 1',
                         'str_column': 'Value for sample 1',
-                        'int_column': 1,
-                        'latitude': 42.42,
-                        'longitude': 41.41,
-                        'taxon_id': 9606,
+                        'int_column': '1',
```
josenavas: Am I missing something? I though this changes where done in a previous PR...

antgonza: The previous PR change the populate and the test that failed but not the creation/updates within tests as the constrains were not there.

qiita_db/metadata_template/test/test_util.py
```diff
@@ -246,6 +246,13 @@ def test_get_get_invalid_sample_names_mixed(self):
         obs = qdb.metadata_template.util.get_invalid_sample_names(one_invalid)
         self.assertItemsEqual(obs, [' ', ' ', ' '])
 
+    def test_get_get_invalid_column_names(self):
+        invalid = ['tax on', 'bla.', '.', '{', 'this|is']
```
josenavas: another invalid column type is if it starts with a number

qiita_db/metadata_template/util.py
```diff
@@ -337,6 +331,33 @@ def get_invalid_sample_names(sample_names):
...
+        List of str objects where each object is an invalid column name.
+
+    References
+    ----------
+    .. [1] postgresql SQL-SYNTAX-IDENTIFIERS: https://goo.gl/EF0cUV.
+    """
+    valid = set(letters+digits+'_')
+    inv = []
+
+    for s in column_names:
```
josenavas: I would check if the first character is a digit, since it is not allowed.

qiita_db/metadata_template/util.py
```diff
@@ -453,3 +474,17 @@ def strip_f(x):
             "No data found in mapping file.")
 
     return mapping_data, header, comments
+
+
+def get_pgsql_reserved_words():
```
josenavas: 💥

qiita_db/support_files/test_data/analysis/1_analysis_mapping.txt
```diff
@@ -0,0 +1,6 @@
+#SampleID	BarcodeSequence	LinkerPrimerSequence	center_name	emp_status	experiment_center	experiment_design_description	experiment_title	illumina_technology	instrument_model	library_construction_protocol	pcr_primers	platform	run_center	run_date	run_prefix	samp_size	sample_center	sequencing_meth	study_center	target_gene	target_subfragment	altitude	anonymized_name	assigned_from_geo	collection_timestamp	common_name	country	depth	description_duplicate	elevation	env_biome	env_feature	has_extracted_data	has_physical_specimen	host_subject_id	host_taxid	latitude	longitude	ph	physical_location	samp_salinity	sample_type	season_environment	taxon_id	temp	texture	tot_nitro	tot_org_carb	water_content_soil	qiita_study_title	qiita_study_alias	qiita_owner	qiita_principal_investigator	Description
```
josenavas: Why is this file being added?

antgonza: cause I always forget to remove it 😞 ... will do in the next commit

qiita_db/metadata_template/test/test_sample_template.py
josenavas: oh - ok, thanks!

qiita_db/support_files/test_data/analysis/1_analysis_mapping.txt
josenavas: Hahaha - it's ok. My guess is that there is one of the tests that is not cleaning the environment correctly, probably in `qiita_db/test/test_analysis.py` given the file that is modifying.

josenavas:
Few comments - thanks for working on this! My biggest concern is being too strict parsing dates/timestamps.

antgonza:
I could add other formats. Currently, just added what EBI expects.

qiita_db/metadata_template/base_metadata_template.py
antgonza: The problem is that then we will not know which comes from invalid chars or from pgsl without making a lot of changes and I'm not sure is worth it. Do you feel strong about it?

josenavas: If `get_invalid_column_names` is only used here I think we can make it return the error message (if any) instead of the list of invalid columns. @ElDeveloper what do you think?

josenavas:
That's interesting - if EBI is requiring that format, should we standardize on that one? cc @ackermag

qiita_db/metadata_template/base_metadata_template.py
```diff
@@ -1426,3 +1422,53 @@ def _update_accession_numbers(self, column, values):
...
+                                datatype(val)
+                            except ValueError:
+                                warning_msg.append('%s, wrong value "%s"' % (
+                                    sample, val))
+
+        if warning_msg:
+            warnings.warn(
+                "Some functionality will be disabled due to missing "
+                "columns:\n\t%s.\nSee the Templates tutorial for a description"
+                " of these fields." % ";\n\t".join(warning_msg),
```
wasade: why the trailing `\t`?

qiita_db/metadata_template/test/test_prep_template.py
```diff
@@ -653,73 +653,26 @@ def test_clean_validate_template_error_bad_chars(self):
         self.metadata.index = ['o()xxxx[{::::::::>', 'sample.1', 'sample.3']
```
wasade: while that does look like a sword, it is not the only thing it resembles at first

qiita_db/metadata_template/test/test_sample_template.py
```diff
@@ -1848,63 +1759,63 @@ def test_extend_new_samples_and_columns(self):
         self.assertItemsEqual(st.categories(), exp_categories)
         exp_dict = {
             "%s.Sample1" % st.id: {
-                'collection_timestamp': datetime(2014, 5, 29, 12, 24, 51),
+                'collection_timestamp': '05/29/14 12:24:15',
```
wasade: this change didn't cause an issue? the seconds went from 51 to 15

qiita_db/metadata_template/test/test_sample_template.py
```diff
@@ -1848,63 +1759,63 @@ def test_extend_new_samples_and_columns(self):
...
+                'physical_specimen_remaining': 'true',
                 'sample_type': "type1",
                 'str_column': "Value for sample 1",
-                'taxon_id': 9606,
+                'taxon_id': '9606',
                 'scientific_name': 'homo sapiens',
                 'newcol': 'val1'},
             "%s.Sample2" % st.id: {
-                'collection_timestamp': datetime(2014, 5, 29, 12, 24, 51),
+                'collection_timestamp': '05/29/14 12:24:15',
```
wasade: same

qiita_db/metadata_template/test/test_sample_template.py
```diff
@@ -1848,63 +1759,63 @@ def test_extend_new_samples_and_columns(self):
...
+                'int_column': '4',
                 'str_column': 'Value for sample 4',
                 'physical_specimen_location': 'location1',
-                'physical_specimen_remaining': True,
-                'dna_extracted': True,
+                'physical_specimen_remaining': 'true',
+                'dna_extracted': 'true',
                 'sample_type': 'type1',
-                'collection_timestamp': datetime(2014, 5, 29, 12, 24, 51),
+                'collection_timestamp': '05/29/14 12:24:15',
```
wasade: there seems to be a theme...

qiita_db/metadata_template/test/test_util.py
```diff
@@ -289,58 +302,58 @@ def test_parse_mapping_file(self):
     "sample_name\tcollection_timestamp\tdescription\thas_extracted_data\t"
     "has_physical_specimen\thost_subject_id\tint_column\tlatitude\tlongitude\t"
     "physical_location\trequired_sample_info_status\tsample_type\tstr_column\n"
-    "2.Sample1\t2014-05-29 12:24:51\tTest Sample 1\tTrue\tTrue\tNotIdentified"
+    "2.Sample1\t05/29/2014 12:24:51\tTest Sample 1\tTrue\tTrue\tNotIdentified"
```
wasade: ya, all these times are 51 seconds

qiita_db/metadata_template/util.py
```diff
@@ -337,6 +331,36 @@ def get_invalid_sample_names(sample_names):
...
+    -------
+    list
+        List of str objects where each object is an invalid column name.
+
+    References
+    ----------
+    .. [1] postgresql SQL-SYNTAX-IDENTIFIERS: https://goo.gl/EF0cUV.
+    """
+    valid_initial_char = letters
+    valid_rest = set(letters+digits+'_')
```
wasade: spaces please

qiita_db/metadata_template/util.py
```diff
@@ -337,6 +331,36 @@ def get_invalid_sample_names(sample_names):
     return inv
 
 
+def get_invalid_column_names(column_names):
```
wasade: should this assert ascii or is unicode okay?

wasade:
just a few minor comments

qiita_db/metadata_template/base_metadata_template.py
antgonza: @wasade, what do you think?

qiita_db/metadata_template/base_metadata_template.py
antgonza: No idea but that's in the [original code](https://github.com/biocore/qiita/pull/1851/files/65eea0cfb5b28450de90bb2b80b53de39b5a1d25#diff-fe8c1935d9373ee4e64757a20e79c142L572) and I rather leave it as is, is that fine?

qiita_db/metadata_template/test/test_prep_template.py
antgonza: 😞

qiita_db/metadata_template/test/test_sample_template.py
antgonza: We had some that were 51 and others that were 15, now all are the same.

qiita_db/metadata_template/util.py
antgonza: spaces are not valid names for columns.

qiita_db/metadata_template/util.py
antgonza: I think `letters` only return ascii so we are fine, right?

qiita_db/metadata_template/test/test_sample_template.py
antgonza: After a second review, I'm wrong but now all tests pass, except the one that I said that requires DB changes.

qiita_db/metadata_template/base_metadata_template.py
wasade: i agree that it seems cleaner to do this within `get_invalid_column_names`, but if that triggers an annoying refactor, the perhaps best to just make an issue and move on

qiita_db/metadata_template/base_metadata_template.py
wasade: k

qiita_db/metadata_template/util.py
wasade: between the `+` marks. I thought pep8 would flag this?

qiita_db/metadata_template/util.py
wasade: given the import, my guess is that's only safe if py3. but i see where letters is coming from now so i htink this is fine

qiita_db/metadata_template/util.py
antgonza: Excellent point but it doesn't: https://travis-ci.org/biocore/qiita/builds/133477520

josenavas: Although pep8 is not flagging it, we should change it for readability...

qiita_db/metadata_template/base_metadata_template.py
antgonza: Added an issue: https://github.com/biocore/qiita/issues/1854

PR #1852, opened 2016-05-28
Fixed setup.py and completely rewrote and tested installation docs (Fixes #1847)

### LINKED ISSUE #1847: Fix setup.py and update INSTALL.md installation documentation accordingly

HannesHolste:
20 hours later... dependency hell figured out!

See #1847.

Only minimal changes to setup.py made – mostly adding missing package definitions. The only changes to the python dependencies made were removing the moi version constraint and adding the coveralls package (because travis config specified it too).

These instructions were tested successfully on Mac OS X El Capitan 10.11.4 and a clean installation of Ubuntu 12.04 LTS (precise) using conda v4.0.7 and cloning from qiita master branch commit #a9e4e03 on 5/27/2016.


INSTALL.md
```diff
@@ -1,115 +1,226 @@
+
 Qiita installation
 ==================
 
-Qiita is pip installable, but depends on some non-python packages that must be installed first.
+Qiita is pip installable, but depends on some non-python packages that must be installed first. We strongly recommend using virtual environments; the solution we recommend ot manage them is [miniconda](http://conda.pydata.org/miniconda.html), a lightweight version of the virtual environment, python distribution, and package manager anaconda.
```
antgonza: I'm not sure if miniconda is the version we recommend ... I would say that is the version we will use in this manual. Note that we use virtualenv in all our internal installs.

antgonza: we recommend ot -> we recommend to

INSTALL.md
```diff
@@ -1,115 +1,226 @@
...
+
+These instructions were tested successfully by [@HannesHolste](github.com/HannesHolste) on Mac OS X El Capitan 10.11.4 and a clean installation of Ubuntu 12.04 LTS (precise) using conda v4.0.7 and cloning from [qiita master branch commit #a9e4e03](https://github.com/biocore/qiita/commit/a9e4e03ecd781d3985abc03d15f2248143e565d7) on 5/27/2016.
+
+## Install miniconda
+
+Download the appropriate installer [here](http://conda.pydata.org/miniconda.html) corresponding to your operating system and execute it.
+
+Ensure that your `~/.bash_profile` (for default bash users) and/or `~/.zshrc` (for zshell users) contains the following line, which prepends a path leading to miniconda's binaries to the `$PATH` variable.
```
antgonza: for default -> default for

INSTALL.md
```diff
@@ -1,115 +1,226 @@
...
+
+```bash
+# added by Miniconda2 installer
+export PATH="/Users/hannes/miniconda2/bin:$PATH"
+```
```
antgonza: Perhaps add a command like grep to help novice users to find this line.

INSTALL.md
```diff
@@ -1,115 +1,226 @@
...
+
+Finally, restart your shell or reload your profile/rc file to ensure your `$PATH` variable was updated:
+
+````bash
+source ~/.bash_profile
+````
+
+## Setup miniconda environment for QIITA
+
+To ensure that QIITA can install its dependencies – some of which require older or specific distributions of pip packages – without interfering with any pre-existing, globally installed pip packages you might have, you will setup a virtual environment in conda that you must *activate* whenever you are working with or running QIITA.
```
antgonza: This makes it sound like this is something specific about Qiita but it's really something that you need to do for all packages. Perhaps move to previous section.

antgonza: Change all QIITA -> Qiita.

INSTALL.md
```diff
@@ -1,115 +1,226 @@
...
+
+```bash
+conda create --yes --name qiita python=2.7 pip nose flake8 pyzmq networkx pyparsing natsort mock future libgfortran seaborn 'pandas>=0.18' 'matplotlib>=1.1.0' 'scipy>0.13.0' 'numpy>=1.7' 'h5py>=2.3.1' --channel https://conda.anaconda.org/OpenMDAO
+```
+
+If you receive an error message about conda being unable to find one of the specified packages in its repository, you will have to manually search for them (see troubleshooting miniconda below).
```
antgonza: Add a link?

INSTALL.md
```diff
@@ -1,115 +1,226 @@
...
+
+Next, activate your newly created virtual environment for qiita:
+
+```bash
+source activate qiita
+```
+
+(When you want to deactivate this environment, e.g. to return to a different project or back to your global python and pip packages, run `source deactivate`)
```
antgonza: Also, for all conda envs, not only Qiita.

INSTALL.md
```diff
@@ -1,115 +1,226 @@
...
+
+If your new conda environment is functioning correctly, you should see this kind of output when you run `which python`, indicating that the `python` command now refers to the python binary in your new virtual environment, rather than a previous global default such as `/usr/bin/python`:
```
antgonza: Same here.

INSTALL.md
```diff
@@ -1,115 +1,226 @@
...
+
+If you don't see this output, your `$PATH` variable was setup incorrectly or you haven't restarted your shell.
+
+As long as you are in the active qiita environment, commands such as `pip install` or `python` will refer to and be contained within this virtual environment.
+
 
 Install the non-python dependencies
 -----------------------------------
 
 * [PostgreSQL](http://www.postgresql.org/download/) (minimum required version 9.3.0, we have tested most extensively with 9.3)
```
antgonza: This is now 9.3.5 and 9.3.6

INSTALL.md
```diff
@@ -1,115 +1,226 @@
...
 * [redis-server](http://redis.io) (we have tested most extensively with 2.8.17)
+* [hdf5](https://www.hdfgroup.org/HDF5/)
```
antgonza: hdf5 is required by biom, no? Also this should be added to your env before the qiita/biom, no?

INSTALL.md
```diff
@@ -1,115 +1,226 @@
...
+/Users/your_username/miniconda2/envs/qiita/bin/python
+(qiita)
+```
+
+If you don't see this output, your `$PATH` variable was setup incorrectly or you haven't restarted your shell.
+
+As long as you are in the active qiita environment, commands such as `pip install` or `python` will refer to and be contained within this virtual environment.
+
 
 Install the non-python dependencies
```
antgonza: I think this section should go before installing Qiita.

INSTALL.md
```diff
@@ -1,115 +1,226 @@
...
 -----------------------------------
 
 * [PostgreSQL](http://www.postgresql.org/download/) (minimum required version 9.3.0, we have tested most extensively with 9.3)
 * [redis-server](http://redis.io) (we have tested most extensively with 2.8.17)
+* [hdf5](https://www.hdfgroup.org/HDF5/)
+
+There are several options to install these dependencies depending on your needs:
+
+- You could install them via conda, however, the anaconda repository may not have the exact versions of these that you want.
```
antgonza: Can you rewrite this line?

INSTALL.md
```diff
@@ -1,115 +1,226 @@
...
 -----------------------------------
 
 * [PostgreSQL](http://www.postgresql.org/download/) (minimum required version 9.3.0, we have tested most extensively with 9.3)
 * [redis-server](http://redis.io) (we have tested most extensively with 2.8.17)
+* [hdf5](https://www.hdfgroup.org/HDF5/)
+
+There are several options to install these dependencies depending on your needs:
+
+- You could install them via conda, however, the anaconda repository may not have the exact versions of these that you want. 
+- You could install the exact versions we tested as per instructions on their websites, and thus make their available globally on your operating system. This may or may not cause problems if other software on your system requires a differing, specific version of the packages. These instructions were tested with this method.
```
antgonza: We suggest installing the exact versions in these instructions by following the instructions of the provided links and making them globally available in your machine. However, this might interfere with other apps that might require different versions. ----- Make it the first option

INSTALL.md
```diff
@@ -1,115 +1,226 @@
...
+- You could setup a full development environment with [Vagrant](https://www.vagrantup.com/), and continue using conda under it to primarily manage python dependencies.
```
antgonza: Add at the end: Note that we don't cover Vagrant in this installation.

INSTALL.md
```diff
@@ -1,115 +1,226 @@
...
+You'll then need to ensure that the postgres binaries (for example, ``psql``) are in your executable search path (``$PATH`` environment variable). If you are using Postgres.app on OS X, you can do this by running the following, though you may have to replace`~/.bash_profile`with `~/.zshrc` if you're using zshell rather than the built-in bash, and you may have to change the version number `Versions/9.3/` to the exact one that you are installing:
 
 ```bash
-export PATH=$PATH:/Applications/Postgres.app/Contents/Versions/9.3/bin/
+echo 'export PATH="$PATH:/Applications/Postgres.app/Contents/Versions/9.3/bin/"' >> ~/.bash_profile
 ```
 
-Install Qiita and its python dependencies
------------------------------------------
+### Redis-server installation on Mac OS X
```
antgonza: I really don't like homebrew, how do others feel? I think redis is super simple to build ... just a make. Note that you should add that you need the dev tools for Ubuntu and OS installed.

antgonza: Also, if we keep homebrew you should mention that you need to install it before proceeding with the instructions

INSTALL.md
```diff
@@ -1,115 +1,226 @@
...
-pip install numpy
-pip install https://github.com/biocore/mustached-octo-ironman/archive/master.zip --no-deps
-pip install qiita-spots
+brew update
+brew install homebrew/versions/redis28
 ```
 
-Install Qiita development version and its python dependencies
--------------------------------------------------------------
+### hdf5 installation on Mac OS X
```
antgonza: Same comment about homebrew/redis.

INSTALL.md
```diff
@@ -1,115 +1,226 @@
...
 
 ```bash
 git clone https://github.com/biocore/qiita.git
 ```
 
-Install Qiita:
+Install Qiita (this occurs through setuptools' `setup.py` file in the qiita directory):
 ```bash
 cd qiita
 pip install .
```
antgonza: Add a `-e` so when you update the changes are reflected in your install

INSTALL.md
```diff
@@ -1,115 +1,226 @@
...
 ```
 
-You will also need to install the target gene plugin for Qiita to be fully functional.
-However, the plugin is already included in the repository so to install it simply execute:
+You will also need to install several plugins for Qiita to be fully functional. As they are already included in the repository, to install, simply execute:
```
antgonza: Note that this is gonna change in like 1-2 weeks.

INSTALL.md
```diff
@@ -1,115 +1,226 @@
...
 ```bash
-cd qiita_plugins/target_gene
+pushd qiita_plugins/target_gene
 pip install .
+popd
+pushd qiita_plugins/qiita_client
+pip install .
```
antgonza: Don't forget the `-e`.

INSTALL.md
```diff
@@ -1,115 +1,226 @@
...
+
 Qiita configuration
 ===================
+
 After these commands are executed, you will need to:
-1. Download a [sample Qiita configuration file](https://github.com/biocore/qiita/blob/master/qiita_core/support_files/config_test.cfg).
 
-  ```bash
-  cd
+Download a [sample Qiita configuration file](https://github.com/biocore/qiita/blob/master/qiita_core/support_files/config_test.cfg) by running these commands (assuming you have curl installed).
```
antgonza: This is also in the repo that you cloned and you could simply edit it.

INSTALL.md
```diff
@@ -1,115 +1,226 @@
...
+**Q: Conda – What if I already have python installed? Or homebrew on OS  X? And a bunch of pip packages??**
+
+This isn't a problem because conda *prepends* its binaries to the `$PATH` variable.
+
+**Q: I get the error** `ConfigParser.NoOptionError: No option 'cookie_secret' in section: 'main'`
+
+This means `cookie_secret` is missing on the qiita configuration file `~/config_test.cfg`. Try updating it as per above instructions, or manually adding the missing key-value to the config file.
+
+**Q: I get the error:** `Error: database "qiita_test" already exists`
+
```
antgonza: Perhaps add the instructions to drop it?

INSTALL.md
```diff
@@ -1,115 +1,226 @@
...
+# this builds documentation before starting the server
+# alternatively: qiita pet webserver --no-build-docs start
+qiita pet webserver start
+```
 If all the above commands executed correctly, you should be able to go to http://localhost:21174 in your browser, to login use `test@foo.bar` and `password` as the credentials. (In the future, we will have a *single user mode* that will allow you to use a local Qiita server without logging in. You can track progress on this on issue [#920](https://github.com/biocore/qiita/issues/920).)
 
-## Installation issues on Ubuntu 14.04
+
+
+# Frequently Asked Questions and Troubleshooting
```
antgonza: Should these be split or perhaps an OS flag?

antgonza:
A few comments, thanks!

INSTALL.md
```diff
@@ -1,115 +1,226 @@
+
 Qiita installation
 ==================
 
-Qiita is pip installable, but depends on some non-python packages that must be installed first.
+Qiita is pip installable, but depends on some non-python packages that must be installed first. We strongly recommend using virtual environments; the solution we recommend ot manage them is [miniconda](http://conda.pydata.org/miniconda.html), a lightweight version of the virtual environment, python distribution, and package manager anaconda.
+
+These instructions were tested successfully by [@HannesHolste](github.com/HannesHolste) on Mac OS X El Capitan 10.11.4 and a clean installation of Ubuntu 12.04 LTS (precise) using conda v4.0.7 and cloning from [qiita master branch commit #a9e4e03](https://github.com/biocore/qiita/commit/a9e4e03ecd781d3985abc03d15f2248143e565d7) on 5/27/2016.
+
+## Install miniconda
```
josenavas: Should we point to the miniconda installation instructions? We should not be maintaining the way they install it, but make it easier for user to find the instructions

INSTALL.md
josenavas: Agree with @antgonza 

This make sound Qiita bad, but that is a common thing on software packages, I would remove this sentence completely and argument that miniconda is recommended because Qiita and its plugins may have conflicting dependencies.

INSTALL.md
josenavas: Agree - the Qiita installation will fail if this packages are not previously installed

INSTALL.md
josenavas: also, hdf5 can be installed with conda

INSTALL.md
josenavas: This is the first time I heard about Vagrant. I don't see the advantage of adding it. Conda is already an environment manager (not only for python).

INSTALL.md
josenavas: Agree with @antgonza on not using homebrew - I don't use it and when I use it I had more problems than solutions. This dependencies are extremely easy to install so I would not add homebrew just for this.

INSTALL.md
josenavas: It can be installed with conda, so you can add it to the installations once conda is installed.

INSTALL.md
josenavas: 👍 * 1000!

INSTALL.md
```diff
@@ -1,115 +1,226 @@
...
+
+# Frequently Asked Questions and Troubleshooting
+
+As a first line of defense, ensure that you activated the qiita conda environment (assuming you're using conda as per the above instruction). If it wasn't previously active, try re-running your commands.
+
+```bash
+source activate qiita
+```
+
+**Q: Conda – What if I already have python installed? Or homebrew on OS  X? And a bunch of pip packages??**
```
josenavas: That's conda specific - we shouldn't have this.

INSTALL.md
```diff
@@ -1,115 +1,226 @@
...
+
+This isn't a problem because conda *prepends* its binaries to the `$PATH` variable.
+
+**Q: I get the error** `ConfigParser.NoOptionError: No option 'cookie_secret' in section: 'main'`
```
josenavas: If this is still happening is a bug in our code, and we should fix it rather than letting users deal with it.

INSTALL.md
josenavas: `qiita-env clean_test`?

INSTALL.md
```diff
@@ -136,6 +249,40 @@ It might be necessary to restart postgresql: `sudo service postgresql restart`.
 
 Furthermore, the `pg_hba.conf` file can be modified to change authentication type for local users to trust (rather than, e.g., md5) but we haven't tested this solution.
 
+### `Error: You need to install postgresql-server-dev-X.Y for building a server-side extension or libpq-dev for building a client-side application.`
```
josenavas: All these "troubleshooting" on Ubuntu. It looks like is a result of incomplete instructions rather than actual problems.

INSTALL.md
```diff
@@ -146,6 +293,38 @@ If running on OS X you should make sure that the Xcode and the Xcode command lin
 
 If you are using Postgres.app on OSX, a database user will be created with your system username. If you want to use this user account, change the `USER` and `ADMIN_USER` settings to your username under the `[postgres]` section of your Qiita config file.
 
+### conda
```
josenavas: Again, I am hesitant on supporting installation instructions for packages that we don't support at all - those can change without us noticing and these instructions get outdated. Point to the troubleshooting conda official instructions.

josenavas:
Thanks for working on this!

INSTALL.md
HannesHolste: @josenavas: conda appears primarily built to support python use cases. It does not have well maintained or older versions of non-python dependencies that Qiita recommends, such as Postgres 9.3. 

Vagrant is like virtualenv on an operating system level with two advantages:
1. You can provide a vagrant 'box' (basically an image) with a pre-built ubuntu server with the correct versions of hdf5, postgres, etc. pre-built and pre-configured as well as conda with all the Qiita packages already installed. So no rebuilding or system config needed.
2. So it's very easy to download the image and you can launch it by doing `vagrant up` in the Qiita directory, and `vagrant ssh` into the virtual machine. The Qiita directory on your local machine is kept in sync with the one in the virtual machine.
3. Every Qiita developer can theoretically very easily replicate the same OS and conda environment to develop Qiita.

INSTALL.md
HannesHolste: On OS X, to build redis, you'd need to have gcc and a bunch of xcode developer tools installed (as far as I can remember). Most of us devs already have this, but it may be an added hassle for new devs or non-dev users.

josenavas: @biocore/qiita-dev I though we were moving away from officially supporting laptop installations besides for developers?

INSTALL.md
HannesHolste: I removed most of this section and am referring users to the conda docs as @josenavas suggested.

INSTALL.md
HannesHolste: Thanks @josenavas. Removed hdf5 instructions and added to `conda create` cmd.

INSTALL.md
HannesHolste: What to do? Keep it in there and add a note that this'll change?

INSTALL.md
HannesHolste: Not a problem anymore in the latest config file.

INSTALL.md
HannesHolste: Yeah, these are primarily a result of really poor Postgres instructions for installing older (9.3.x) version of its database (not our responsibility per se). I think it doesn't hurt to keep these here for now.

INSTALL.md
HannesHolste: I think the postgres section existed before I edited it.

IMO the conda section is generic enough to stay here, because Qiita requires `libgfortran` so it's directly relevant. Plus, most OSX users without a preconfigured system will run into this issue because libgfortran is most likely not in the mainstream conda channels.

INSTALL.md
HannesHolste: If that's the case, then we should really consider Vagrant in addition to miniconda. Practically, many devs will be using laptops to run Qiita, so vagrant would allow us to stop 'officially' supporting laptop installations while still providing consistent and easy-to-setup dev environment (based on say, ubuntu 14 server) that devs can use on their laptops. 

I'll make a note of this to discuss next Qiita meeting.

INSTALL.md
HannesHolste: Not sure what you mean – the highlighted section is only discussing how to activate and deactivate qiita conda env, no?

INSTALL.md
HannesHolste: Good point. I reworded this to make it clearer.

antgonza: Do we really need the parenthesis?

INSTALL.md
```diff
@@ -1,117 +1,196 @@
...
+
+Activate your newly created virtual environment for qiita whenever you want to run or develop for it:
+
+```bash
+source activate qiita
+```
+
+(When you want to deactivate this environment, e.g. to return to a different project or back to your global python and pip packages, run `source deactivate`)
+
+After activating your new environment, you should see this kind of output when you run `which python`, indicating that the `python` command now refers to the python binary in your new virtual environment, rather than a previous global default such as `/usr/bin/python`. For example, assuming you installed miniconda in `Users/your_username/`:
```
antgonza: Users/ -> /Users/

INSTALL.md
antgonza: I think this is fixed. Basically, we were saying that this section should be towards its beginning rather than the end.

INSTALL.md
antgonza: Yup.

antgonza:
A few more comments.

HannesHolste:
Cool, thank you for your feedback @antgonza and @josenavas, it's looking a lot better now!

INSTALL.md
```diff
@@ -1,117 +1,200 @@
...
+$ which python
+/Users/your_username/miniconda2/envs/qiita/bin/python
+(qiita)
+```
+
+If you don't see this output, your `$PATH` variable was setup incorrectly or you haven't restarted your shell. Consult the [conda documentation](http://conda.pydata.org/docs/install/quick.html).
+
+As long as you are in the active qiita environment, commands such as `pip install` or `python` will refer to and be contained within this virtual environment.
+
+When you want to deactivate your current condo environment, e.g. to return to a different project or back to your global python and pip packages, run:
```
antgonza: condo -> conda

HannesHolste:
Fixed typo. 

@antgonza

josenavas:
👍

PR #1853, opened 2016-05-31
Simplify rest api

josenavas:
First step for simplifying the REST api.

The current REST api for the artifacts was getting complex with a lot of endpoints. Right now, it has been reduced to a single endpoint to get all the information.

I've started modifying some of the plugins but not all of them. Doing the PR against a different branch given our last discussion in the Qiita meeting so we can get some discussion started about the API.


qiita_db/handlers/artifact.py
```diff
@@ -43,39 +44,62 @@ def _get_artifact(a_id):
...
-            The id of the artifact whose filepath information is being
-            retrieved
+            The id of the artifact whose information is being retrieved
 
         Returns
         -------
         dict
-            {'filepaths': list of (str, str)}
-            The filepaths attached to the artifact and their filepath types
+            {''}
```
antgonza: Should we just remove this?

antgonza:
One minor comment. @ElDeveloper or @squirrelo could you check this PR?

qiita_plugins/target_gene_type/target_gene_type/summary.py
```diff
@@ -103,35 +103,37 @@ def _summary_not_demultiplexed(artifact_type, filepaths):
...
+                hash_md5 = md5()
+                for chunk in iter(lambda: f.read(4096), b""):
+                    hash_md5.update(chunk)
+
+            # getting head of the files
+            header = []
+            if artifact_type not in FILEPATH_TYPE_TO_NOT_SHOW_HEAD:
+                # we need to encapsulate the full for loop because gzip will
+                # not raise an error until you try to read
+                try:
```
ElDeveloper: Instead of doing a `try/except` you could read in the magic number right? It would probably amount to the same lines of code in total, but at least there wouldn't be an except `IOError` (which can also be raised if the file doesn't exist IIRC).

ElDeveloper:
Looks good.

antgonza:
The commits will not change the passing status, merging.

PR #1855, opened 2016-06-01
adding dtype=str to all from_dict 

antgonza:


antgonza:
@josenavas and @HannesHolste will review, thanks.

josenavas:
Thanks @antgonza the changes here look good.

I'm assuming that the checks for making sure that everything is a string will come in a different PR? If so, I would also recommend removing the functions `type_loookup`, `get_datatypes`, `cast_to_python` and `as_python_types` from `qiita_db.metadata_template.util`; since those are the ones that are related with the type inference.

qiita_db/metadata_template/test/test_sample_template.py
```diff
@@ -1342,20 +1343,22 @@ def test_update_numpy(self):
...
+        metadata = pd.DataFrame.from_dict(metadata_dict, orient='index',
+                                          dtype=str)
         npt.assert_warns(qdb.exceptions.QiitaDBWarning, st.update, metadata)
 
         sql = "SELECT * FROM qiita.sample_{0}".format(st.id)
         obs = self.conn_handler.execute_fetchall(sql)
-        exp = [['2.Sample1', False, datetime(2015, 9, 1)],
-               ['2.Sample2', False, datetime(2015, 9, 1)]]
+        exp = [['2.Sample1', 'false', '2015-09-01 00:00:00'],
+               ['2.Sample2', 'false', '2015-09-01 00:00:00']]
```
HannesHolste: `2015-09-01 00:00:00` for sample2 -- should this correspond to the date for sample2 on line 1344 above?
```
'Sample2': {'bool_col': np.bool_(False),
                          'date_col': np.datetime64(datetime(2015, 8, 1))}
```
Current one is 9/1/15 and the other is 8/1/15

HannesHolste:
Overall, looks good! 

(I assume you checked the locations in code beyond the scope of this that could be affected by changes in datatype – I'm still familiarizing myself with Qiita so I'm not sure where to look beyond the diffs here)

qiita_db/metadata_template/test/test_sample_template.py
antgonza: It changes bak to 09 in likne 1351.

antgonza:
That's right, @josenavas! Thanks.

PR #1856, opened 2016-06-06
rm inter sample/prep tables and fixing tests

antgonza:
Removing:
- study_sample_columns table
- prep_columns table
- type_loookup 
- get_datatypes 
- cast_to_python
- as_python_types 
- convert_type

This also includes the code to transform all tables to varchar but is commented cause it increases the testing time dramatically.


qiita_db/metadata_template/base_metadata_template.py
```diff
@@ -1044,7 +985,7 @@ def to_dataframe(self):
             meta = qdb.sql_connection.TRN.execute_fetchindex()
 
             # Create the dataframe and clean it up a bit
-            df = pd.DataFrame((list(x) for x in meta), columns=cols)
+            df = pd.DataFrame((list(x) for x in meta), columns=cols, dtype=str)
```
josenavas: The only case that I'm concerned here is if there is a `None` in `meta`:
```python
In [20]: l = [[1, 3], [None, 2]]

In [21]: df = pd.DataFrame(l, columns=['c1', 'c2'], dtype=str)

In [22]: df
Out[22]: 
    c1 c2
0    1  3
1  NaN  2
```
It changes the `None` to `NaN`. Can this case occur in here?

qiita_db/metadata_template/sample_template.py
```diff
@@ -57,13 +56,12 @@ class SampleTemplate(MetadataTemplate):
...
     _table_prefix = "sample_"
-    _column_table = "study_sample_columns"
     _id_column = "study_id"
     _sample_cls = Sample
     _filepath_table = 'sample_template_filepath'
 
-    @staticmethod
-    def metadata_headers():
+    @classmethod
+    def metadata_headers(cls):
```
josenavas: Should this method be moved to the base class so it also becomes available for the PrepTemplate class?

antgonza: Currently is not used in the PrepTemplate class so I think we should move once is required there, no?

qiita_db/support_files/patches/python_patches/38.py
```diff
@@ -0,0 +1,21 @@
+#
+# This is a really heavy step and it will be better not to run after each test
```
josenavas: Uncomment this. When this is executed in the tests, the database is empty, since it is populated after all patches are applied.

qiita_db/metadata_template/base_metadata_template.py
antgonza: I guess that's possible if something in the database is NULL but if it is, my guess is that it's fine ...

qiita_db/metadata_template/sample_template.py
josenavas: Ok - just added an issue: https://github.com/biocore/qiita/issues/1857

qiita_db/support_files/patches/python_patches/38.py
antgonza: Excellent point!

qiita_db/metadata_template/base_metadata_template.py
josenavas: I'm concerned about the warning at the end of [this section](http://pandas.pydata.org/pandas-docs/stable/missing_data.html#values-considered-missing) on the pandas docs.

antgonza: Not sure what will be the solution to your concern ...

josenavas: Ideally, all `NaN`s should be `None`s, but pandas for whatever reason it's applying some kind of transformation although you're passing `dtype=str`. I think the only option would be to force everything to be `None` afterwards using this:
```python
In [32]: df
Out[32]: 
    c1 c2
0    1  3
1  NaN  2

In [33]: df.where(pd.notnull(df), None)
Out[33]: 
     c1 c2
0     1  3
1  None  2
```

antgonza: K

josenavas:
👍

antgonza:
@ElDeveloper, @wasade, @HannesHolste, do you guys have a (few) second(s) to review?

josenavas:
👍 once test pass

ElDeveloper:
👍 once tests pass.

HannesHolste:
Looks good.

Issue #1863, opened 2016-06-16
Missing moi installation instructions on the install page


josenavas:
In the current install instructions, the release version of moi is installed. However, we should be installing the development version of moi, so it is missing this command before installing Qiita:
``` bash
pip install https://github.com/biocore/mustached-octo-ironman/archive/master.zip --no-deps
```

@HannesHolste can you fix?

HannesHolste:
Good catch – I tested it with the release version moi originally, but is it better to use the moi dev version because it's more up to date than the release version, or because of something else?

Right now `moi` is listed under `install_requires` in `setup.py`.  I think I struggled with trying to get it to prefer the `moi-dev` version in `setup.py`. So can I completely remove moi from this?:
```python
install_requires=['psycopg2', 'click >= 3.3', 'future',
                        'bcrypt', 'pandas >= 0.17', 'numpy >= 1.7',
                        'tornado==3.1.1', 'toredis', 'redis', 'six',
                        'ipython[all] >= 2.4.1, < 2.5', 'pyparsing',
                        'h5py >= 2.3.1', 'biom-format', 'natsort', 'networkx',
                        'scikit-bio >= 0.2.3, < 0.3.0', 'wtforms == 2.0.1',
                        'qiime >= 1.9.0, < 1.10.0', 'moi',
                        'sphinx-bootstrap-theme', 'Sphinx >= 1.2.2',
                        'gitpython'],
```

josenavas:
There are a couple of small fixes that were introduced in the dev version.

No, don't remove moi from that list. install_requires should list all the packages that Qiita depends on. Once we do an actual release of Qiita we will also do a release of moi.

antgonza:
Fixed in https://github.com/biocore/qiita/pull/1864

PR #1864, opened 2016-06-16
Replace current pip moi with moi-dev version #1863

HannesHolste:
see #1863


setup.py
```diff
@@ -108,7 +108,7 @@
                         'ipython[all] >= 2.4.1, < 2.5', 'pyparsing',
                         'h5py >= 2.3.1', 'biom-format', 'natsort', 'networkx',
                         'scikit-bio >= 0.2.3, < 0.3.0', 'wtforms == 2.0.1',
-                        'qiime >= 1.9.0, < 1.10.0', 'moi',
+                        'qiime >= 1.9.0, < 1.10.0',
```
josenavas: Please add moi back, it is still a dependency of Qiita.

josenavas:
👍 Thank for fixing #1863 so fast!

PR #1871, opened 2016-06-28
Adding prep template data handler to REST api

josenavas:
This endpoint was required by the validate functions of the type plugins but was never added.

### DIFF
--- a/qiita_db/handlers/prep_template.py
+++ b/qiita_db/handlers/prep_template.py
@@ -1,0 +1,66 @@
+# -----------------------------------------------------------------------------
+# Copyright (c) 2014--, The Qiita Development Team.
+#
+# Distributed under the terms of the BSD 3-clause License.
+#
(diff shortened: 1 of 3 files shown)


qiita_db/handlers/prep_template.py
```diff
@@ -0,0 +1,65 @@
...
+# -----------------------------------------------------------------------------
+
+from tornado.web import HTTPError
+
+import qiita_db as qdb
+from .oauth2 import OauthBaseHandler, authenticate_oauth
+
+
+def _get_prep_template(p_id):
+    """Returns the prep template with the given id if it exists
```
antgonza: given p_id?

antgonza: In previous code we normally use pid but up to you. Perhaps, we should create a formal dictionary ...

josenavas: Done

PR #1872, opened 2016-06-28
Add prep test API creation

josenavas:
This depends on #1871 so review/merge that one first.

This adds an endpoint so plugins can create a prep template.

### DIFF
--- a/qiita_db/handlers/prep_template.py
+++ b/qiita_db/handlers/prep_template.py
@@ -1,0 +1,66 @@
+# -----------------------------------------------------------------------------
+# Copyright (c) 2014--, The Qiita Development Team.
+#
+# Distributed under the terms of the BSD 3-clause License.
+#
+# The full license is in the file LICENSE, distributed with this software.
+# -----------------------------------------------------------------------------
+
+from tornado.web import HTTPError
+from collections import defaultdict
+
+import qiita_db as qdb
+from .oauth2 import OauthBaseHandler, authenticate_oauth
+
+
+def _get_prep_template(p_id):
+    """Returns the prep template with the given id if it exists
+
+    Parameters
+    ----------
+    p_id : str
+        The prep template id
+
+    Returns
+    -------
+    qiita_db.metadata_template.prep_template.PrepTemplate
+        The requested prep template
+
+    Raises
+    ------
+    HTTPError
+        If the prep template does not exist, with error code 404
+        If there is a problem instantiating the template, with error code 500
+    """
+    try:
+        p_id = int(p_id)
+        pt = qdb.metadata_template.prep_template.PrepTemplate(p_id)
+    except qdb.exceptions.QiitaDBUnknownIDError:
+        raise HTTPError(404)
+    except Exception as e:
+        raise HTTPError(500, 'Error instantiating prep template %s: %s'
+                             % (p_id, str(e)))
+
+    return pt
+
+
+class PrepTemplateDataHandler(OauthBaseHandler):
+    @authenticate_oauth
+    def get(self, prep_id):
+        """Retrieves the prep contents
+
+        Parameters
+        ----------
+        prep_id : str
+            The id of the prep template whose information is being retrieved
+
+        Returns
+        -------
+        dict
+            The contents of the prep information keyed by sample id
+        """
+        with qdb.sql_connection.TRN:
+            pt = _get_prep_template(prep_id)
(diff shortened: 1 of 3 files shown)


qiita_db/handlers/prep_template.py
```diff
@@ -63,3 +66,16 @@ def get(self, prep_id):
             response = {'data': pt.to_dataframe().to_dict(orient='index')}
 
         self.write(response)
+
+
+class PrepTemplateAPItestHandler(OauthBaseHandler):
+    @authenticate_oauth
+    def post(self):
+        prep_info_dict = loads(self.get_argument('prep_info'))
```
antgonza: Is the size of the metadata something we should be concern about? It seems like if we have a large study we can hit the maximum parameter size really quickly.

antgonza:
1 comment.

ElDeveloper:
👍

qiita_db/handlers/prep_template.py
josenavas: That's a good question - I will open an issue to keep track of this. In this case, this endpoint is tailored only for being able to execute the plugin tests and it will not be exposed in a live environment. Do you agree with this solution?

josenavas: Issue: https://github.com/biocore/qiita/issues/1873

antgonza: I think it's worth testing with the largest study in our current version of Qiita before deploying because if it breaks we need to rethink this section. Now, this can happen before merging this or after as this is going to its own branch. If you decide we should merge, could you add a point to the list so we make sure it's done and tested before this branch is merged to master?

josenavas: Fair point - I've investigated it with @ElDeveloper 
The name of the function (`get_argument`) seems a bit misleading here. The information is sent in the body of the request, so we are not limited at the argument of the URL. Here is the code snippet that we use to test this, sending a fake template of ~74Mb (which I don't know how often we see templates this big):
```python
import requests
from json import dumps
from qiita_client import QiitaClient

metadata_dict = {
    'SKB8.640193': {
        'primer': 'GTGCCAGCMGCCGCGGTAA',
        'barcode': 'GTCCGCAAGTTA',
        'platform': 'ILLUMINA',
        'instrument_model': 'Illumina MiSeq'},
    'SKD8.640184': {
        'primer': 'GTGCCAGCMGCCGCGGTAA',
        'barcode': 'GTCCGCAAGTTA',
        'platform': 'ILLUMINA',
        'instrument_model': 'Illumina MiSeq'}}
# Create a mapping file of ~76MB
metadata_dict['SKB8.640193']['primer'] = metadata_dict['SKB8.640193']['primer'] * 200000
metadata_dict['SKB8.640193']['primer'] = metadata_dict['SKB8.640193']['primer'] * 20
data = {
    'prep_info': dumps(metadata_dict),
    'study': 1,
    'data_type': '16S'}
CLIENT_ID = '19ndkO3oMKsoChjVVWluF7QkxHRfYhTKSFbAVt8IhK7gZgDaO4'
CLIENT_SECRET = ('J7FfQ7CQdOxuKhQAf1eoGgBAE81Ns8Gu3EKaWFm3IO2JKh'
                 'AmmCWZuabe0O5Mp28s1')
qclient = QiitaClient("https://localhost:21174", CLIENT_ID, CLIENT_SECRET,
                      server_cert='/Users/jose/qiime_software/QiiTa/qiita_core/support_files/server.crt')
r = qclient._request_oauth2(requests.post, 'https://localhost:21174/apitest/prep_template/', data=data, verify=False)
r.request.headers

Out[24]: {'Content-Length': '76000409', 'Accept-Encoding': 'gzip, deflate', 'Accept': '*/*', 'User-Agent': 'python-requests/2.8.1', 'Connection': 'keep-alive', 'Content-Type': 'application/x-www-form-urlencoded', 'Authorization': u'Bearer ZVvdQbf0Xyp5WWzqH0XRuYikfktsVF5jJ2hJBfJNjoEwup4GVQJKfbG'}

r.request.url
Out[28]: 'https://localhost:21174/apitest/prep_template/'
```

antgonza: 👍  thanks for making sure. In this kind of things is better to be sure about them working fine.

josenavas: Sure! Can we merge if there are no more comments?

PR #1877, opened 2016-07-05
Adding prep template get handler

josenavas:
Adding an endpoint to the restAPI that returns the information of the prep info object.

### DIFF
--- a/qiita_db/handlers/prep_template.py
+++ b/qiita_db/handlers/prep_template.py
@@ -80,6 +80,8 @@ class PrepTemplateDBHandler(OauthBaseHandler):
                 'study': pt.study_id,
                 'status': pt.status,
                 'qiime-map': pt.qiime_map_fp,
+                # The first element in the prep_files is the newest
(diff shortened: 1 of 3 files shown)


qiita_db/handlers/prep_template.py
```diff
@@ -46,6 +47,45 @@ def _get_prep_template(pid):
...
+            prep_files = [fp for _, fp in pt.get_filepaths()
+                          if 'qiime' not in basename(fp)]
+            response = {
+                'data_type': pt.data_type(),
+                'artifact': pt.artifact.id,
+                'investigation_type': pt.investigation_type,
+                'study': pt.study_id,
+                'status': pt.status,
+                'qiime-map': pt.qiime_map_fp,
+                'prep-file': prep_files[0]
```
antgonza: Could you add why [0]?

antgonza:
One minor comment.

antgonza:
The tests are not going to change as the only modification was in documentation. Thus, merging.

PR #1878, opened 2016-07-06
Fix reference handler

josenavas:
~~Depends on #1877, so review/merge that one first~~

Fixes the reference handler to return a dict rather than a list


ElDeveloper:
👍

PR #1882, opened 2016-07-06
fix #1857

### LINKED ISSUE #1857: Move `metadata_headers` to the base metadata class
Currently it is in the sample template class, but it can be generalized and made available on the PrepTemplate class if it is moved to the base class

antgonza:
Also fixing one of the tests for timestamp conversions.

### DIFF
--- a/qiita_db/metadata_template/base_metadata_template.py
+++ b/qiita_db/metadata_template/base_metadata_template.py
@@ -601,7 +601,10 @@ class MetadataTemplate(qdb.base.QiitaObject):
                         FROM information_schema.columns
                         WHERE table_name LIKE '{0}%' AND
                               table_name != 'sample_template_filepath' AND
-                              table_name != 'prep_template_filepath'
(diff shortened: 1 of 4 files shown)


qiita_db/metadata_template/base_metadata_template.py
```diff
@@ -586,6 +587,24 @@ def _common_creation_steps(cls, md_template, obj_id):
...
+        Returns
+        -------
+        list
+            Alphabetical list of all metadata headers available
+        """
+        with qdb.sql_connection.TRN:
+            sql = """SELECT DISTINCT column_name
+                        FROM information_schema.columns
+                        WHERE table_name LIKE '{0}%' AND
+                              table_name != 'sample_template_filepath'
```
josenavas: should `prep_template_filepath` also be excluded here, now that this function is in the base class?

Having a test for the prep template case may help to detect other potential tables that should be ignored here.

josenavas:
👍

PR #1884, opened 2016-07-06
Add artifact creation test handler

josenavas:
Depends on #1878 so review/merge that one first

This adds a new endpoint for the apitest restAPI in which plugin developers can create new artifacts

### DIFF
--- a/qiita_db/handlers/artifact.py
+++ b/qiita_db/handlers/artifact.py
@@ -146,7 +146,7 @@ class ArtifactAPItestHandler(OauthBaseHandler):
 
         Parameters
         ----------
-        filepahts : str
+        filepaths : str
             Json string with a list of filepaths and its types
         type : str
             The artifact type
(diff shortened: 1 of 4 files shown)


antgonza:
could you pull from master?

josenavas:
Done!

antgonza:
Looks good. Do you think it's worth adding a test for when the artifact can't be created?

josenavas:
Good call! There was a bug in oauth which prevented to send a useful error to the consumer. That may be why we were sometimes not seeing the error messages on the plugin.

qiita_db/handlers/artifact.py
```diff
@@ -136,3 +137,43 @@ def patch(self, artifact_id):
...
+
+
+class ArtifactAPItestHandler(OauthBaseHandler):
+    @authenticate_oauth
+    def post(self):
+        """Creates a new artifact
+
+        Parameters
+        ----------
+        filepahts : str
```
mortonjt: filepaths?

PR #1886, opened 2016-07-08
Remove plugins directory

josenavas:
In this PR:
 - The qiita_plugins folder is removed
 - Qiime is removed from setup.py, since it is no longer a dependency of Qiita
 - Travis is fixed to not install the local plugins
 - The install.md instructions have been updated

Once this is merged into master - I highly recommend that you remove the qiita_plugins directory from your local machines, so the *pyc files do not interfere...


INSTALL.md
```diff
@@ -115,11 +115,14 @@ Install Qiita (this occurs through setuptools' `setup.py` file in the qiita dire
...
 
-You will also need to install several plugins for Qiita to be fully functional. As they are already included in the repository, to install, simply execute:
+You will also need to install several plugins for Qiita to be fully functional.
+To install those plugins, simply execute
+
 ```bash
-pushd qiita_plugins/target_gene
-pip install -e .
-popd
+pip install https://github.com/qiita-spots/qiita_client
```
antgonza: These are the base plugins so perhaps it will be better to add them to the setup.py file, right?

antgonza:
One comment.

INSTALL.md
josenavas: I would prefer not to add them in the setup.py. The reason why I'm installing them here is because ATM 2 of these plugins depend on Qiita, so it makes sense to have them in the same environment. However, once those plugins don't depend on Qiita, I would like to update the installation instructions so they are not installed in the same environment.

antgonza: That makes sense but without them Qiita has no plugins AKA not functional, right? So perhaps, making them default will be a better way to go ...

INSTALL.md
```diff
@@ -115,8 +115,14 @@ Install Qiita (this occurs through setuptools' `setup.py` file in the qiita dire
 pip install -e .
 ```
 
-You will also need to install several plugins for Qiita to be fully functional.
-To install those plugins, simply execute
+At this point, Qiita will be installed and the system will start. However,
+you will need to install plugins in order to process any kind of data. For a list
+of available plugins, visit the [Qiita Spots](https://github.com/qiita-spots)
+github organization. At this moment, the `Type Plugins`
```
antgonza: At this moment, -> Currently,

INSTALL.md
```diff
@@ -115,11 +115,20 @@ Install Qiita (this occurs through setuptools' `setup.py` file in the qiita dire
 pip install -e .
 ```
 
-You will also need to install several plugins for Qiita to be fully functional. As they are already included in the repository, to install, simply execute:
+At this point, Qiita will be installed and the system will start. However,
+you will need to install plugins in order to process any kind of data. For a list
+of available plugins, visit the [Qiita Spots](https://github.com/qiita-spots)
+github organization. At this moment, the `Type Plugins`
```
antgonza: At this moment, -> Currently,

antgonza:
qiita and flake8 errors.

INSTALL.md
josenavas: Done

PR #1893, opened 2016-07-11
fix #1885

### LINKED ISSUE #1885: Can we change the expected time stamp format?
Currently these are required for validation:
mm/dd/yyyy hh:mm:ss or mm/dd/yyyy hh:mm or mm/dd/yyyy hh or mm/dd/yyyy or mm/yyyy or yyyy.

None of these are compatible with Excel, requiring individual curation.  Please...or should I just ignore the warnings ( in the template the year is  yyyy but it is shortened).
[804_20160706-235613.txt](https://github.com/biocore/qiita/files/351544/804_20160706-235613.txt)

antgonza:
Also adding some actual testing for the warnings raised on the tests.


ElDeveloper:
👍 looks great!

PR #1896, opened 2016-07-12
fixing errors

antgonza:
Tests are broken, this fixes them.


ElDeveloper:
👍

ElDeveloper:
@antgonza, just in case you missed this, there are 2 real failures in the unit tests.

antgonza:
Yes, they are random (even in my computer) and I think we need input from @josenavas to find the best way to fix.

PR #1899, opened 2016-07-18
Delete column/sample 

antgonza:
This still a WIP but wanted to see how long the PR is going to be.

### DIFF
--- a/qiita_db/metadata_template/base_metadata_template.py
+++ b/qiita_db/metadata_template/base_metadata_template.py
@@ -620,16 +620,15 @@ class MetadataTemplate(qdb.base.QiitaObject):
 
         Raises
         ------
-        QiitaDBColumnError
+        QiitaDBUnknownIDError
             If the `sample_name` doesn't exist
         """
         if sample_name not in self.keys():
-            raise qdb.exceptions.QiitaDBColumnError(
-                "'%s' not in info file %d" % (sample_name, self._id))
+            raise qdb.exceptions.QiitaDBUnknownIDError(sample_name, self._id)
 
         with qdb.sql_connection.TRN:
-            sql = 'DELETE FROM qiita.{0}{1} WHERE sample_id=%s'.format(
-                self._table_prefix, self._id)
+            sql = 'DELETE FROM qiita.{0} WHERE sample_id=%s'.format(
+                self._table_name(self._id))
             qdb.sql_connection.TRN.add(sql, [sample_name])
 
             sql = "DELETE FROM qiita.{0} WHERE sample_id=%s".format(
@@ -651,15 +650,16 @@ class MetadataTemplate(qdb.base.QiitaObject):
         Raises
         ------
         QiitaDBColumnError
-            If a the info file can't be updated
             If the `column_name` doesn't exist
+        QiitaDBOperationNotPermittedError
+            If a the info file can't be updated
         """
-        if not self.can_be_updated(columns={column_name}):
-            raise qdb.exceptions.QiitaDBColumnError(
-                '%s cannot be deleted' % column_name)
         if column_name not in self.categories():
             raise qdb.exceptions.QiitaDBColumnError(
                 "'%s' not in info file %d" % (column_name, self._id))
+        if not self.can_be_updated(columns={column_name}):
+            raise qdb.exceptions.QiitaDBOperationNotPermittedError(
+                '%s cannot be deleted' % column_name)
         with qdb.sql_connection.TRN:
             sql = 'ALTER TABLE qiita.%s%d DROP COLUMN %s' % (
                 self._table_prefix, self._id, column_name)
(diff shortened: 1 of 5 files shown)


qiita_db/metadata_template/base_metadata_template.py
```diff
@@ -609,6 +610,64 @@ def metadata_headers(cls):
...
+        ------
+        QiitaDBColumnError
+            If the `sample_name` doesn't exist
+        """
+        if sample_name not in self.keys():
+            raise qdb.exceptions.QiitaDBColumnError(
+                "'%s' not in info file %d" % (sample_name, self._id))
+
+        with qdb.sql_connection.TRN:
+            sql = 'DELETE FROM qiita.{0}{1} WHERE sample_id=%s'.format(
```
josenavas: Change by:
```python
sql = 'DELETE FROM qiita.{0} WHERE sample_id=%s'.format(self._table_name(self._id))
```

qiita_db/metadata_template/sample_template.py
```diff
@@ -162,6 +162,26 @@ def columns_restrictions(self):
...
+            The sample name to be deleted
+
+        Raises
+        ------
+        QiitaDBColumnError
+            If the `sample_name` has been used in a prep info file
+        """
+        if any([pt.get(sample_name) != None
+                for pt in qdb.study.Study(self.study_id).prep_templates()]):
+            raise qdb.exceptions.QiitaDBColumnError(
```
josenavas: this should be a `QiitaDBOperationNotPermittedError`

qiita_db/metadata_template/test/test_prep_template.py
```diff
@@ -1464,6 +1464,32 @@ def test_validate_template_warning_missing(self):
...
+        with self.assertRaises(qdb.exceptions.QiitaDBColumnError):
+            pt.delete_column('ph')
+
+    def test_delete_sample(self):
+        pt = qdb.metadata_template.prep_template.PrepTemplate(2)
+        pt.delete_sample('1.SKM5.640177')
+        self.assertNotIn('1.SKM5.640177', pt.keys())
+
+        pt = qdb.metadata_template.prep_template.PrepTemplate(1)
+        with self.assertRaises(qdb.exceptions.QiitaDBColumnError):
```
josenavas: This is the same test as before, since it first checks if the prep info file has been processed. By changing the error type as I suggested in my previous comment you will see that this test will fail (or at least it should).

qiita_db/metadata_template/base_metadata_template.py
```diff
@@ -609,6 +610,64 @@ def metadata_headers(cls):
...
+        r"""Executes the common delete sample steps
+
+        Parameters
+        ----------
+        sample_name : str
+            The sample name to be erased
+
+        Raises
+        ------
+        QiitaDBColumnError
```
ElDeveloper: Wouldn't a `QiitaDBUnknownIDError` be more appropriate here?

qiita_db/metadata_template/base_metadata_template.py
```diff
@@ -609,6 +610,64 @@ def metadata_headers(cls):
...
+            The column name to be deleted
+
+        Raises
+        ------
+        QiitaDBColumnError
+            If a the info file can't be updated
+            If the `column_name` doesn't exist
+        """
+        if not self.can_be_updated(columns={column_name}):
+            raise qdb.exceptions.QiitaDBColumnError(
```
josenavas: this should be a `QiitaDBOperationNotPermittedError`

josenavas:
Few comments.

qiita_db/metadata_template/base_metadata_template.py
```diff
@@ -609,6 +610,64 @@ def metadata_headers(cls):
...
+                '%s cannot be deleted' % column_name)
+        if column_name not in self.categories():
```
ElDeveloper: Hmmm, not very sure about this one, but I think the check to see if the column exists or not should come before the check to see if the column can be updated.

qiita_db/metadata_template/sample_template.py
```diff
@@ -162,6 +162,26 @@ def columns_restrictions(self):
...
+        ----------
+        sample_name : str
+            The sample name to be deleted
+
+        Raises
+        ------
+        QiitaDBColumnError
+            If the `sample_name` has been used in a prep info file
+        """
+        if any([pt.get(sample_name) != None
```
ElDeveloper: I believe the pythonic way says that this should be `is not None`.

qiita_db/metadata_template/sample_template.py
```diff
@@ -162,6 +162,26 @@ def columns_restrictions(self):
...
+                for pt in qdb.study.Study(self.study_id).prep_templates()]):
+            raise qdb.exceptions.QiitaDBColumnError(
+                "'%s' has been linked in a prep template" % (sample_name))
```
ElDeveloper: It would perhaps be handy to know what prep templates it has been linked to, for troubleshooting?

ElDeveloper:
Same here, overall looks good. :+1:

josenavas:
Good to merge when tests pass

PR #1901, opened 2016-07-18
BUG: Return error message with invalid code

### LINKED ISSUE #1839: if a user's email is verified and it reloads the page you get a 500 error
We should decide on what to do, I like the idea of saying: email already verified.

ElDeveloper:
When a user registers with an already verified code, the server will no
longer return a 500 error, instead it will return a 200 code with an
appropriate error message.

Fixes #1839


josenavas:
👍

PR #1904, opened 2016-07-19
fix #1897

### LINKED ISSUE #1897: main site link in the main site is broken
It takes you to a page that doesn't exist.

antgonza:
Once merged we will need to update the config files (for each site) in the qiita main system deploy.

### DIFF
--- a/qiita_pet/support_files/config_portal.cfg
+++ b/qiita_pet/support_files/config_portal.cfg
@@ -18,7 +18,7 @@ TEXT = <p align="justify">
   point for all their analyses.
   </p>
   <p align="justify">
-  Qiita’s <a href=”qiita.microbio.me”>main site</a> provides database and
+  Qiita’s <a href='http://qiita.microbio.me'>main site</a> provides database and
   compute resources to the global community, alleviating the technical
   burdens, such as familiarity with the command line or access to compute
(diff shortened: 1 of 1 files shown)


qiita_pet/support_files/config_portal.cfg
```diff
@@ -32,7 +32,11 @@ TEXT = <p align="justify">
   </p>
   <p align="justify">
   For more information about how to use Qiita visit the
-  <a href=”static/doc/html/index.html”>documentation</a>
+  <a href=”static/doc/html/index.html”>documentation</a>.
+  </p>
+  <p align="justify">
+  Note that you should be log into the system to access to any studies and
```
ElDeveloper: log -> logged

ElDeveloper: to any -> the

josenavas:
AFAIK we will not need to change anything. This file is the one used by default and the portals have a different one that does not contain this information, so I think once we deploy the change will be automatically done.

josenavas:
👍

josenavas:
I don't think we need for the tests to pass on this since it is only changing documentation that does not affect the tests. @ElDeveloper feel free to merge once your comments are addressed.

PR #1906, opened 2016-07-20
Make sure that the error message encoded is in utf8

josenavas:
While helping a user in Qiita I found that if a sample information file contained non-utf8 characters, it will break the communication between the ipython workers and the main server. This PR makes sure that the error message is a utf8 valid string, by replacing the non-utf8 characters with the official U+FFFD REPLACEMENT CHARACTER, as seen below: 

screen shot 2016-07-20 at 2 45 37 pm


antgonza:
Tests passed locally, merging due to restart.

PR #1908, opened 2016-07-26
Plugin installation db

josenavas:
Note that this PR is going to its own branch (plugin installation). In that branch I will add everything that is needed so plugins can register themselves.

Here I'm modifying the database to add a new column to the software and software_command tables (active) so a given plugin or command can be deactivated and hence not run (this is specially needed for QIIME2).

In the HTML and qdb file I've also removed the tables that @antgonza dropped in his stringify work but the schema was not updated (shame on us for forgetting).

I've added a new value in the config file for specifying where the plugins files can live (this will be used later). I've realized that the ConfigurationManager did not have tests so I've added them. That's why I'm issuing this PR before it becomes too big.

### DIFF
--- a/qiita_core/configuration_manager.py
+++ b/qiita_core/configuration_manager.py
@@ -7,7 +7,7 @@
 # -----------------------------------------------------------------------------
 
 from functools import partial
-from os.path import join, dirname, abspath, isdir
+from os.path import join, dirname, abspath, isdir, expanduser
 from os import environ
 from future import standard_library
 from base64 import b64encode
@@ -116,6 +116,8 @@ class ConfigurationManager(object):
         The filepath to the portal styling config file
     plugin_launcher : str
         The script used to start the plugins
+    portal_dir : str
+        The path to the directory containing the plugin configuration files
     """
     def __init__(self):
         # If conf_fp is None, we default to the test configuration file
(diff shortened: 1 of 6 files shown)


qiita_core/configuration_manager.py
```diff
@@ -168,14 +170,18 @@ def _get_main(self, config):
...
+        if not self.plugin_dir:
+            self.plugin_dir = join(expanduser('~'), '.qiita_plugins')
 
         self.valid_upload_extension = [ve.strip() for ve in config.get(
             'main', 'VALID_UPLOAD_EXTENSION').split(',')]
         if (not self.valid_upload_extension or
            self.valid_upload_extension == ['']):
             self.valid_upload_extension = []
-            print 'No files will be allowed to be uploaded.'
+            warnings.warn('No files will be allowed to be uploaded.')
```
antgonza: Should this be a warning or an error? Perhaps an error and if the admin passes a parameter it changes to a warning? My concern is that the user will install by mistake without upload extensions ...

antgonza:
A comment. Do we need 2 reviewers?

qiita_core/configuration_manager.py
josenavas: Before it was just a print - I am ok to make it an error since you will not able to upload any kind of data.

antgonza: I think making it a requirement is the way to go even if the extension is duplicated from another package, which is fine as they are independent, right?

josenavas: Note that this is the general Qiita config, independent of the plugins. However, you actually raised a good point and probably plugins should be the ones that describe which extensions are valid for upload, and not the Qiita config. I will keep this in mind when finishing up this branch, but I will not include that here to not explode the size of this PR.

josenavas:
Regarding 2 reviewers - the changes in this PR are small (just a bunch of tests and a couple of column additions to the DB) so I don't think it is needed.

antgonza:
OK, makes sense. Merging as it's going to its own branch

PR #1909, opened 2016-07-27
Plugin installation

### LINKED ISSUE #1625: Adding necessary mechanisms to install a plugin
Include, but not limited to:
 - [ ] Create software
 - [ ] Create commands (with their parameters and default parameters set)
 - [ ] Add new filepath types
 - [ ] Add new artifact types
 - [ ] Add filepath type - artifact type relationship

josenavas:
Branch that allows plugins to register themselves.

TODOs:
- [x] Modify the DB so plugins can be active/inactive
- [x] Create a parser for the files in the plugins dir to install/(de)activate plugins (https://github.com/biocore/qiita/pull/1923)
- [x] Add CLI to update plugins and modify webserver start to activate only available plugins.
- [x] Add REST api for plugins so they can register artifact types, file types/extensions and commands
- [x] Update relevant GUI so only active plugins/commands/types are visible

If you think that there is something missing in this TODO list, please add it. If you have a concern about this, please come and talk to me.

Fixes #1625


josenavas:
Note: tests here are failing because I changed the contents of the config file (added an entry). For fixing that I need to update the encoded config file for travis, but I'm not going to do that until this PR is ready to merge, so other PR can be done against master.

qiita_core/configuration_manager.py
```diff
@@ -116,6 +116,8 @@ class ConfigurationManager(object):
         The filepath to the portal styling config file
     plugin_launcher : str
         The script used to start the plugins
+    portal_dir : str
```
ElDeveloper: Shouldn't this be `plugin_dir`?

qiita_core/configuration_manager.py
```diff
@@ -168,14 +170,18 @@ def _get_main(self, config):
                              self.working_dir)
         self.max_upload_size = config.getint('main', 'MAX_UPLOAD_SIZE')
         self.require_approval = config.getboolean('main', 'REQUIRE_APPROVAL')
+
         self.plugin_launcher = config.get('main', 'PLUGIN_LAUNCHER')
+        self.plugin_dir = config.get('main', 'PLUGIN_DIR')
+        if not self.plugin_dir:
+            self.plugin_dir = join(expanduser('~'), '.qiita_plugins')
```
ElDeveloper: Should this check that the folder exists? and if it doesn't create it?

ElDeveloper:
👍 just a few minor comments/questions.

qiita_core/configuration_manager.py
josenavas: You're right!

qiita_core/tests/test_configuration_manager.py
```diff
@@ -0,0 +1,348 @@
...
+        conf_setter = partial(self.conf.set, 'main')
+        conf_setter('COOKIE_SECRET', '')
+        conf_setter('BASE_DATA_DIR', '')
+        conf_setter('PLUGIN_DIR', '')
+        conf_setter('CERTIFICATE_FILE', '')
+        conf_setter('KEY_FILE', '')
+
+        # Warning raised if No files will be allowed to be uploaded
+        # Warning raised if no cookie_secret
+        with warnings.catch_warnings(record=True) as warns:
```
mortonjt: Not really a comment - but some motivation on switching over to PY3

We can substitute this with [assertWarns](https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertWarns)

josenavas: Does assertWarns allow you to check the contents of the warning? We have been using `npt.assert_warns` in other places of the code, but here I decided to use the catch warnings approach so I can test that the contents of the warnings is what I expect.

qiita_core/configuration_manager.py
antgonza: Note that currently, if there is no PLUGIN_DIR in the configuration this fails.

antgonza: Also, if it doesn't exist, it fails.

PR #1910, opened 2016-07-27
Addressing comments

josenavas:
@ElDeveloper the other PR is a WIP PR done from upstream so I've addressed your comments here - if you agree with them, just merge them so the branch gets updated. Thanks!


ElDeveloper:
Thanks!

PR #1917, opened 2016-08-30
Add delete buttons to GUI

josenavas:
Several things done in this PR as during the adding of the GUI several smaller bugs were found:
 - the `to_dataframe` function was enabling a race condition. Note that the functionality itself was correct because it followed the rule of "multiple reads - single write". However, if somebody else wrote the table in the middle the function crashed. The new implementation is not only thread safe but also at least 2X faster.
 - The prep template sample deletion was not safe. Having an artifact attached is enough to not allow sample deletion. Fixed and test added.
 - Connected the GUI with the backend to delete samples/columns from the templates. @antgonza noticed that it can be slow, so we offload it to the cluster. To simplify code, a single dispatchable function has been added that can be used for both template types and both functionalities, simplifying a lot of code duplication.
 - The HTML code that @antgonza added had a fair amount of code duplication due to the previous structure of the code - I've reorganize it to reduce the amount of code duplication.

### DIFF
--- a/qiita_db/metadata_template/base_metadata_template.py
+++ b/qiita_db/metadata_template/base_metadata_template.py
@@ -1059,15 +1059,19 @@ class MetadataTemplate(qdb.base.QiitaObject):
             The metadata in the template,indexed on sample id
         """
         with qdb.sql_connection.TRN:
-            cols = sorted(qdb.util.get_table_cols(self._table_name(self._id)))
-            # Get all metadata for the template
-            sql = "SELECT {0} FROM qiita.{1}".format(", ".join(cols),
-                                                     self._table_name(self.id))
-            qdb.sql_connection.TRN.add(sql, [self._id])
+            # Retrieve all the information from the database
+            sql = "SELECT * FROM qiita.{0}".format(self._table_name(self._id))
+            qdb.sql_connection.TRN.add(sql)
             meta = qdb.sql_connection.TRN.execute_fetchindex()
 
-            # Create the dataframe and clean it up a bit
-            df = pd.DataFrame((list(x) for x in meta), columns=cols, dtype=str)
+            # When we create the dataframe, we are providing a matrix (list of
+            # lists) with all the data, so we need to make sure that all the
+            # rows contain the columns in the same order
+            cols = sorted(meta[0].keys())
+            meta_matrix = [[r[c] for c in cols] for r in meta]
+            df = pd.DataFrame(meta_matrix, columns=cols, dtype=str)
+
+            # Make sure that we are changing np.NaN by Nones
             df.where((pd.notnull(df)), None)
             df.set_index('sample_id', inplace=True, drop=True)
 
--- a/qiita_db/metadata_template/test/test_prep_template.py
+++ b/qiita_db/metadata_template/test/test_prep_template.py
@@ -1478,15 +1478,18 @@ class TestPrepTemplateReadWrite(BaseTestPrepTemplate):
 
     def test_delete_sample(self):
         QE = qdb.exceptions
-        pt = qdb.metadata_template.prep_template.PrepTemplate(2)
-        pt.delete_sample('1.SKM5.640177')
-        self.assertNotIn('1.SKM5.640177', pt.keys())
+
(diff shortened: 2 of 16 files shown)


qiita_pet/handlers/api_proxy/sample_template.py
```diff
@@ -189,36 +190,8 @@ def sample_template_category_get_req(category, samp_id, user_id):
...
-        Format {num_samples: value,
-                category: [(val1, count1), (val2, count2), ...], ...}
-    """
-    access_error = check_access(samp_id, user_id)
-    if access_error:
-        return access_error
-
-    job_info = r_client.get(SAMPLE_TEMPLATE_KEY_FORMAT % samp_id)
+def get_sample_template_processing_status(st_id):
+    job_info = r_client.get(SAMPLE_TEMPLATE_KEY_FORMAT % st_id)
```
mortonjt: Do we want a test for this?

qiita_pet/handlers/api_proxy/sample_template.py
```diff
@@ -250,6 +223,41 @@ def sample_template_summary_get_req(samp_id, user_id):
         alert_type = ''
         alert_msg = ''
 
+    return processing, alert_type, alert_msg
+
+
+def sample_template_summary_get_req(samp_id, user_id):
```
mortonjt: Do we want a test for this?

qiita_db/metadata_template/base_metadata_template.py
```diff
@@ -1059,15 +1059,19 @@ def to_dataframe(self):
...
+            # When we create the dataframe, we are providing a matrix (list of
+            # lists) with all the data, so we need to make sure that all the
+            # rows contain the columns in the same order
+            cols = sorted(meta[0].keys())
+            meta_matrix = [[r[c] for c in cols] for r in meta]
+            df = pd.DataFrame(meta_matrix, columns=cols, dtype=str)
+
+            # Make sure that we are changing np.NaN by Nones
             df.where((pd.notnull(df)), None)
             df.set_index('sample_id', inplace=True, drop=True)
```
mortonjt: How confident are we that this exists?

qiita_db/metadata_template/test/test_prep_template.py
```diff
@@ -1478,15 +1478,18 @@ def test_delete_column(self):
...
         QE = qdb.exceptions
-        pt = qdb.metadata_template.prep_template.PrepTemplate(2)
-        pt.delete_sample('1.SKM5.640177')
-        self.assertNotIn('1.SKM5.640177', pt.keys())
+
+        pt = qdb.metadata_template.prep_template.PrepTemplate.create(
+            self.metadata, self.test_study, self.data_type)
+        sample_id = '%s.SKB8.640193' % self.test_study.id
+        pt.delete_sample(sample_id)
+        self.assertNotIn(sample_id, pt.keys())
```
mortonjt: Do we need a test to delete something that doesn't exist?

mortonjt:
Some comments

qiita_db/metadata_template/base_metadata_template.py
josenavas: 100% :-) is the primary key on the table

qiita_db/metadata_template/test/test_prep_template.py
josenavas: Tested on line 1489

qiita_pet/handlers/api_proxy/sample_template.py
josenavas: Good catch - added

qiita_pet/handlers/api_proxy/sample_template.py
josenavas: Github mess - this function already existed, I just pulled out the above function to avoid code duplication. It is already tested.

antgonza:
👍  will merge once tests pass, thanks!!!

antgonza:
@josenavas actual errors!

josenavas:
Tests are passing!

PR #1921, opened 2016-09-07
fixing stats page - ignore not numeric

antgonza:


qiita_db/support_files/patches/39.sql
```diff
@@ -0,0 +1,15 @@
+-- Sep 7, 2016
+-- Adding a is numeric function to avoid problems with texts
+-- http://stackoverflow.com/a/16206123
+
+CREATE OR REPLACE FUNCTION isnumeric(text) RETURNS BOOLEAN AS $$
```
josenavas: Can you add a test for this function in here? https://github.com/biocore/qiita/blob/master/qiita_db/test/test_sql.py

ElDeveloper:
👍

PR #1923, opened 2016-09-08
Plugin installation parser

josenavas:
This PR allows to install/activate plugins from a config file generated by the plugins.


antgonza:
@josenavas, there are actual errors here, what's the plan?

josenavas:
I just restarted the build - I was not able to see the actual errors.

antgonza:
flake8 fixed, tests failing:
```
ERROR: test_get_main (qiita_core.tests.test_configuration_manager.ConfigurationManagerTests)
  File "/home/travis/build/biocore/qiita/qiita_core/tests/test_configuration_manager.py", line 37, in tearDown
    environ['QIITA_CONFIG_FP'] = self.old_conf_fp
  File "/home/travis/miniconda3/envs/env_name/lib/python2.7/os.py", line 473, in __setitem__
    putenv(key, item)
TypeError: putenv() argument 2 must be string, not None
```

antgonza:
👍 , @wasade, could you review?

qiita_db/util.py
```diff
@@ -43,7 +43,7 @@
 from __future__ import division
 from future.builtins import zip
 from future.utils import viewitems
-from random import choice
+from random import SystemRandom
```
wasade: why the change?

josenavas: There is a warning in the python documentation that said that random should not be used for security purposes, and it recommends the usage of SystemRandom. Since now I'm using the generate random string function to generate the client id / client secret pair for oauth2, I think is better to use SystemRandom.

wasade:
thanks!

PR #1926, opened 2016-09-14
Fixing flake8 error in master

josenavas:


josenavas:
I think it is ok to merge once the faster build passes, since it is already checking flake8

PR #1929, opened 2016-09-15
Speedup metadata template tests

josenavas:
This was the files that I expected to be harder to fix (I may be mistaken though...)

There was a small bug on how I created the tearDownClass decorator, the order of the decorators matter and there is no need to call the super, and before was calling the wrong one.

I found a bug when deleting a sample from a prep template: the bug was deleting the sample from all sample templates in the study, not only from that sample template. I could discover this thanks to not reseting the database and seeing how one test was affecting others although I was not expecting it to affect. I've fixed it and add a small test to make sure that we can catch this in an easy way.

### DIFF
--- a/qiita_db/metadata_template/test/test_prep_template.py
+++ b/qiita_db/metadata_template/test/test_prep_template.py
@@ -23,7 +23,8 @@ from qiita_core.exceptions import IncompetentQiitaDeveloperError
 import qiita_db as qdb
 
 
-class BaseTestPrepSample(TestCase):
+@qiita_test_checker()
+class TestPrepSample(TestCase):
     def setUp(self):
         self.prep_template = \
             qdb.metadata_template.prep_template.PrepTemplate(1)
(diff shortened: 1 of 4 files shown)


qiita_db/metadata_template/test/test_prep_template.py
```diff
@@ -378,19 +377,14 @@ def tearDown(self):
...
-
     def test_study_id(self):
         """Ensure that the correct study ID is returned"""
         self.assertEqual(self.tester.study_id, 1)
 
     def test_init_unknown_error(self):
         """Init raises an error if the id is not known"""
         with self.assertRaises(qdb.exceptions.QiitaDBUnknownIDError):
-            qdb.metadata_template.prep_template.PrepTemplate(3)
+            qdb.metadata_template.prep_template.PrepTemplate(30000)
```
antgonza: What's the benefit of using 30K vs 3? Avoid collision of ids?

josenavas: Correct - since we are not reseting the database and we are creating a lot of templates for testing, the pt id with id 3 ends up existing, and hence making this test to fail. By changing it to a really big number we avoid that collision.

PR #1931, opened 2016-09-15
WIP test_sample_template.py

antgonza:


qiita_pet/handlers/api_proxy/tests/test_sample_template.py
```diff
@@ -65,6 +65,8 @@ def tearDown(self):
         if exists(self.new_study_fp):
             remove(self.new_study_fp)
 
+        qdb.study.Study.delete(self.new_study.id)
```
josenavas: @antgonza is the problem here that the tests fail because the study "Some New Study" already exists?

antgonza: Yup

PR #1933, opened 2016-09-15
Fixing artifact and processing job tests

josenavas:
I have added a new property "outputs" to the processing jobs that returns the outputs of a given processing job. This was needed to be able to fix one of the artifact tests.

### DIFF
--- a/qiita_db/processing_job.py
+++ b/qiita_db/processing_job.py
@@ -620,6 +620,29 @@ class ProcessingJob(qdb.base.QiitaObject):
         for c in ready:
             c.submit()
 
+    @property
+    def outputs(self):
+        """The outputs of the job
+
+        Returns
+        -------
+        dict of {str: qiita_db.artifact.Artifact}
+            The outputs of the job keyed by output name
+        """
+        with qdb.sql_connection.TRN:
+            if self.status != 'success':
+                raise qdb.exceptions.QiitaDBOperationNotPermittedError(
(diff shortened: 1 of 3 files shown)


qiita_db/processing_job.py
```diff
@@ -620,6 +620,29 @@ def _update_and_launch_children(self, mapping):
...
+    def outputs(self):
+        """The outputs of the job
+
+        Returns
+        -------
+        dict of {str: qiita_db.artifact.Artifact}
+            The outputs of the job keyed by output name
+        """
+        with qdb.sql_connection.TRN:
+            if self.status != 'success':
```
antgonza: Should this really raise an error? I think sending an empty array should suffice, no?

antgonza: Also, this is not being tested, right?

ElDeveloper:
👍 so as long as @antgonza can confirm he is OK with the outputs method raising an exception, this is ready to be merged.

qiita_db/processing_job.py
josenavas: It is tested here: https://github.com/biocore/qiita/pull/1933/files#diff-0e7761c945e8eddac7dcc4afcb8f1a35R437

The only reason why I preferred to raise an error is to differentiate the case in which the job did not finish from the cases in which (1) the job did not generate any output (not sure if this will ever happen) or (2) the output artifact has been removed

PR #1934, opened 2016-09-15
Fix commands, ontology and portal tests

josenavas:
I did one exception an in one of the tests classes of commands the db is still reseted after each test. The reason is because it is the patch command and I'm unsure on how to safely change the tests, so I think it is ok to reset in that case since there are few tests.


ElDeveloper:
👍

PR #1938, opened 2016-09-19
fixing util tests

josenavas:


qiita_db/test/test_util.py
```diff
@@ -156,24 +156,26 @@ def test_insert_filepaths(self):
...
-            "SELECT count(1) FROM qiita.filepath")[0]
-        obs = qdb.util.insert_filepaths([(fp, 1)], 1, "raw_data", "filepath")
+            "SELECT last_value FROM qiita.filepath_filepath_id_seq")[0]
+        obs = qdb.util.insert_filepaths([(fp, 1)], 2, "raw_data", "filepath")
         self.assertEqual(obs, [exp_new_id])
 
         # Check that the files have been copied correctly
         exp_fp = join(qdb.util.get_db_files_base_dir(), "raw_data",
-                      "1_%s" % basename(fp))
+                      "2_%s" % basename(fp))
```
antgonza: Is #2 going to cause problems later on?

josenavas: It shouldn't. The only reason why it is a problem right now is because the files modified here was colliding with another tests, so modifying artifact 2 rather than artifact 1 made the tests independent.

PR #1939, opened 2016-09-19
test_job.py

antgonza:
- Added creation of new job on setUp so tests are performed over it - alpha div.
- Always adding 2 results: 1 file and 1 folder.
- For simplicity and avoid duplication of code: joined testing files and folders in to single tests.

### DIFF
--- a/qiita_db/test/test_job.py
+++ b/qiita_db/test/test_job.py
@@ -21,12 +21,36 @@ class JobTest(TestCase):
     """Tests that the job object works as expected"""
 
     def setUp(self):
-        self.job = qdb.job.Job(1)
-        self.options = {"option1": False, "option2": 25, "option3": "NEW"}
         self._delete_path = []
         self._delete_dir = []
+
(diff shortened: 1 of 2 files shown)


qiita_db/test/test_job.py
```diff
@@ -21,12 +21,36 @@ class JobTest(TestCase):
...
+                                  qdb.reference.Reference(1),
+                                  qdb.software.Command(3))
+        self.job = qdb.job.Job.create(*self.job_create_params)
+        self.job_id = self.job.id
+
+        # adding filepaths
+        # file
+        self.fp = join(self._job_folder, "%d_job_result.txt" % self.job_id)
+        if not exists(self.fp):
+            with open(self.fp, 'w') as f:
```
josenavas: Should these files be removed in the tearDown?

antgonza: Just tested and they get removed once the job is deleted so they are automatically removed.

Issue #1914, opened 2016-08-22
resize the pipeline window


antgonza:
When the screen is large, the processing window uses all the space which means that you can't scroll down pass it. We should make it smaller so you can scroll down while on the corners.

antgonza:
Fixed in #1947

PR #1947, opened 2016-09-22
Fix 1914

antgonza:
Fix for #1914. Basically, making the graph a bit smaller and giving space so people can scroll from the sides. Note that I tried using the .graph style but it's not actually modifying this so decided to go with direct style. Also decided that the chars were too big so reduce the size a little bit.


wasade:
:+1: ?

ElDeveloper:
This looks approved by two people (@wasade and @tkosciol), can one of you merge?

PR #1948, opened 2016-09-22
fix #1912

### LINKED ISSUE #1912: Sample information summary
If the user has a really long list of samples, each value could be different for a particular column. It might be a good idea to allow each column to be collapsible.

antgonza:
By default the values are now hidden and you have to click on a "values" button so they are shown, see image below. This changed hit https://github.com/twbs/bootstrap/issues/12593 so had to upgrade bootstrap. 

screen shot 2016-09-22 at 2 42 08 pm


qiita_pet/templates/study_ajax/sample_summary.html
```diff
@@ -145,30 +145,34 @@
...
+            </td>
+            {% elif len(set([row[1] for row in summary])) == 1 %}
+                <b>{{category}}</b>: All the values in this category are different.
+              </td>
+            {% else %}
+                <b>{{category}}<b>
+                &nbsp;&nbsp;&nbsp;&nbsp;
+                <button type="button" class="btn" data-toggle="collapse" data-target=".{{i}}collapsed">values</button>
+              </td>
+              {% for row in summary %}
```
wasade: would it make sense to expand the tuple here instead of indexing on lines 173 and 174?

antgonza: Not sure but if you feel strong about it, I can add.

wasade: meh

mortonjt:
I'm curious - what changed in the css files (the diff is too big to view).

ElDeveloper:
@mortonjt, sorry, didn't see your comment while I was reviewing and merged. My bad! As @antgonza points out in his opening message, he updated bootstrap to deal with a bug that was introduced as part of these changes.

mortonjt:
Oh ok.  I think it is good then.

PR #1949, opened 2016-09-22
fix #1849

### LINKED ISSUE #1849: Make "16S" link a button-style link
It would be helpful if the 16S link appeared like a "raised" button (like the ones above) when there was data there. It could have the current "flat" appearance when there's no data.

antgonza:
I added a "click on the tabs" message and added a list glyphicon to make the links more obvios, see image below.

screen shot 2016-09-22 at 3 00 52 pm


wasade:
:+1:

PR #1954, opened 2016-09-26
fix #1268

### LINKED ISSUE #1268: If a title of a study contains a single quote, it will break the JavaScript
More concrete, it will break the function `delete_study` from the `study_description.html` template, in [this](https://github.com/biocore/qiita/blob/master/qiita_pet/templates/study_description.html#L313) confirm call.

We should scape those characters to avoid breaking the JavaScript. I'm not sure how to do it, so although I think it is probably an easy-fix I have not labeled it until we're sure that it is an easy fix.

antgonza:
Additionally to fixing that issue in the study description page (delete), I also fixed the edit page and found that qiita_pet/templates/study_description.html was a leftover from previous GUI versions (thus, remove it).


mortonjt:
So if I'm understanding, the study_description page is now obselete, which is why you are deleting it.

If so, think this ok 👍

antgonza:
@mortonjt Yes, that's one thing the other is that I'm fixing #1268 (JS errors due to single ' ")

ElDeveloper:
👍

Issue #1553, opened 2015-11-27
Investigate why the information column from the qiita.logger table can't be NULL


josenavas:
See discussion here: https://github.com/biocore/qiita/pull/1552/files#r46072008

antgonza:
Fixed in #1956

PR #1956, opened 2016-09-26
fix 1553

antgonza:
Now it accepts NULL and I have added a case for qiita_db/support_files/populate_test_db.sql.

### DIFF
--- a/qiita_db/logger.py
+++ b/qiita_db/logger.py
@@ -157,7 +157,14 @@ class LogEntry(qdb.base.QiitaObject):
                      logging_id = %s""".format(self._table)
             qdb.sql_connection.TRN.add(sql, [self.id])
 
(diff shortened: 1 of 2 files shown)


mortonjt:
@antgonza really don't have enough context on this to merge this.  But I don't see any problems with this.  @josenavas want to merge if there are no problems?

antgonza:
@mortonjt Perhaps this will help. #1553 was opened cause I realized that the error information couldn't be NULL (allowed by the db), while working on another issue. Thus, we needed to figure out and fix why was this failing. The issue was we weren't testing if the return value was NULL so we could return an empty dict. Now, we do. 😄

mortonjt:
Ok, I think that makes sense.   Merging.

PR #1958, opened 2016-09-26
Plugin installation endpoints

josenavas:
This adds the needed endpoints so plugins can register themselves using the REST api.

@antgonza there are 2 missing things in here:
- [x] In the class CommandListHandler, perform the call to create, passing `parameters` using the correct format
- [x] ~~Add a new function in the class qiita_db.software.Command called `add_default_parameter_sets` that adds new parameter sets to the command. Then you can call this function in the previous handler once the command is created.~~ Actually not needed, there is the DefaultParameters.create call


qiita_db/handlers/plugin.py
```diff
@@ -0,0 +1,186 @@
...
+    ------
+    HTTPError
+        If the plugin does not exist, with error code 404
+        If there is a problem instantiating the plugin, with error code 500
+    """
+    try:
+        plugin = qdb.software.Software.from_name_and_version(name, version)
+    except qdb.exceptions.QiitaDBUnknownIDError:
+        raise HTTPError(404)
+    except Exception as e:
```
antgonza: It seems like all 500 errors are missing tests. Could you double check?

josenavas: I'm unsure on how to force those errors to happen. I put those as a security guard in case that something unexpected happens (that's why they catch any exception). Other endpoints are similar to this.
Any idea on how to force the error?

antgonza: If it's not possible to generate them, why test them? 

Now, this is working on `from_name_and_version`, which only raises `QiitaDBUnknownIDError`, so the second except is working on:
```
            sql = """SELECT software_id
                     FROM qiita.software
                     WHERE name = %s AND version = %s"""
            qdb.sql_connection.TRN.add(sql, [name, version])
            res = qdb.sql_connection.TRN.execute_fetchindex()
```
which will only fail if the DB fails or the structure changes, right? If that's correct I think we can remove them, right?

josenavas: If the DB fails, what will be then returned to the client?

antgonza: If the DB fails, the Qiita server will fail, right? So no return at all ...

josenavas: Not necessarily - our fault tolerant system allows the webserver to still be alive even when the DB is dead.

antgonza: K

PR #1960, opened 2016-09-27
Adds missing bits

josenavas:
Adds the missing code to be able to successfully install commands and types and adds some endpoints so the plugins can correctly run the tests.

### DIFF
--- a/qiita_db/artifact.py
+++ b/qiita_db/artifact.py
@@ -100,6 +100,57 @@ class Artifact(qdb.base.QiitaObject):
             qdb.sql_connection.TRN.add(sql)
             return qdb.sql_connection.TRN.execute_fetchindex()
 
+    @staticmethod
+    def create_type(name, description, can_be_submitted_to_ebi,
+                    can_be_submitted_to_vamps, filepath_types):
+        """Creates a new artifact type in the system
+
+        Parameters
+        ----------
+        name : str
+            The artifact type name
+        description : str
+            The artifact type description
+        can_be_submitted_to_ebi : bool
+            Whether the artifact type can be submitted to EBI or not
+        can_be_submitted_to_vamps : bool
+            Whether the artifact type can be submitted to VAMPS or not
+        filepath_types : list of (str, bool)
+            The list filepath types that the new artifact type supports, and
+            if they're required or not in an artifact instance of this type
+
+        Raises
+        ------
+        qiita_db.exceptions.QiitaDBDuplicateError
+            If an artifact type with the same name already exists
+        """
+        with qdb.sql_connection.TRN:
+            sql = """SELECT EXISTS(
+                        SELECT *
+                        FROM qiita.artifact_type
+                        WHERE artifact_type=%s)"""
+            qdb.sql_connection.TRN.add(sql, [name])
+            if qdb.sql_connection.TRN.execute_fetchlast():
+                raise qdb.exceptions.QiitaDBDuplicateError(
+                    'artifact type', 'name: %s' % name)
+            sql = """INSERT INTO qiita.artifact_type
+                        (artifact_type, description, can_be_submitted_to_ebi,
+                         can_be_submitted_to_vamps)
+                     VALUES (%s, %s, %s, %s)
+                     RETURNING artifact_type_id"""
+            qdb.sql_connection.TRN.add(
+                sql, [name, description, can_be_submitted_to_ebi,
+                      can_be_submitted_to_vamps])
+            at_id = qdb.sql_connection.TRN.execute_fetchlast()
+            sql = """INSERT INTO qiita.artifact_type_filepath_type
+                        (artifact_type_id, filepath_type_id, required)
+                     VALUES (%s, %s, %s)"""
+            sql_args = [
+                [at_id, qdb.util.convert_to_id(fpt, 'filepath_type'), req]
+                for fpt, req in filepath_types]
+            qdb.sql_connection.TRN.add(sql, sql_args, many=True)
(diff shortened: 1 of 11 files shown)


qiita_db/handlers/plugin.py
```diff
@@ -184,3 +190,35 @@ def get(self, plugin_name, plugin_version, cmd_name):
...
+            The command name
+        """
+        with qdb.sql_connection.TRN:
+            cmd = _get_command(plugin_name, plugin_version, cmd_name)
+            cmd.activate()
+
+        self.finish()
+
+
+class ReloadPluginAPItestHandler(OauthBaseHandler):
```
antgonza: This is missing a test, right?

qiita_db/software.py
```diff
@@ -281,6 +284,17 @@ def create(cls, software, name, description, parameters):
                     for at in atypes]
                 qdb.sql_connection.TRN.add(sql_type, sql_params, many=True)
 
+            # Add the outputs to the command
+            if outputs:
```
antgonza: Just want to confirm that now there are no tests without output. I think is fine due to how it work but want to double check it.

qiita_db/software.py
```diff
@@ -425,6 +439,31 @@ def outputs(self):
             qdb.sql_connection.TRN.add(sql, [self.id])
             return qdb.sql_connection.TRN.execute_fetchindex()
 
+    @property
+    def active(self):
```
antgonza: This doesn't have an actual tests, right?

qiita_db/handlers/plugin.py
josenavas: Added

qiita_db/software.py
josenavas: Added

qiita_db/software.py
josenavas: I tested it through the function `test_activate` so it is implicitly tested.

antgonza: K

antgonza:
👍 , @wasade, could you review? Thanks.

qiita_db/artifact.py
```diff
@@ -100,6 +100,57 @@ def types():
...
+            The artifact type description
+        can_be_submitted_to_ebi : bool
+            Whether the artifact type can be submitted to EBI or not
+        can_be_submitted_to_vamps : bool
+            Whether the artifact type can be submitted to VAMPS or not
+        filepath_types : list of (str, bool)
+            The list filepath types that the new artifact type supports, and
+            if they're required or not in an artifact instance of this type
+
+        Raises
```
wasade: is the only type of unacceptable artifact a duplicate one?

qiita_db/handlers/plugin.py
wasade: is this method idempotent? if not, should it be?

wasade: same question regarding `CommandActivateHandler`

qiita_db/test/test_artifact.py
```diff
@@ -42,14 +42,31 @@ def test_iter_public(self):
...
                ['SFF', None], ['per_sample_FASTQ', None]]
         self.assertItemsEqual(obs, exp)
 
+        qdb.artifact.Artifact.create_type(
+            "NewType", "NewTypeDesc", False, False,
+            [("log", False), ("raw_forward_seqs", True)])
+
+        obs = qdb.artifact.Artifact.types()
+        exp = [['BIOM', 'BIOM table'],
+               ['Demultiplexed', 'Demultiplexed and QC sequeneces'],
```
wasade: typo in sequences

qiita_pet/webserver.py
```diff
@@ -152,10 +154,13 @@ def __init__(self):
             (r"/qiita_db/jobs/(.*)/step/", ActiveStepHandler),
             (r"/qiita_db/jobs/(.*)/complete/", CompleteHandler),
             (r"/qiita_db/jobs/(.*)", JobHandler),
+            (r"/qiita_db/artifacts/types/", ArtifactTypeHandler),
             (r"/qiita_db/artifacts/(.*)/", ArtifactHandler),
             (r"/qiita_db/prep_template/(.*)/data/", PrepTemplateDataHandler),
             (r"/qiita_db/prep_template/(.*)/", PrepTemplateDBHandler),
             (r"/qiita_db/references/(.*)/", ReferenceHandler),
+            (r"/qiita_db/plugins/(.*)/(.*)/commands/(.*)/activate/",
```
wasade: just curious, why not do named regex? eg `?P<foo>[^/]+/?P<bar>[^/]+/`?

qiita_db/artifact.py
josenavas: This is creating a new artifact type (not a new artifact) and I couldn't think in any other case where a new type is unacceptable.

qiita_db/handlers/plugin.py
josenavas: This method is idempotent as long as the files in `qiita_config.plugin_dir` do not change.
`CommandActivateHandler` is idempotent.

qiita_pet/webserver.py
josenavas: I'm not sure what is the benefit?

qiita_db/test/test_artifact.py
antgonza: Good catch, opened https://github.com/biocore/qiita/issues/1961 cause this is a larger issue.

antgonza:
@wasade thanks for review. Merging so we can continue reviews.

PR #1962, opened 2016-09-28
DOC: Add note about reserved words

ElDeveloper:


antgonza:
👍

antgonza:
@tanaes I _think_ I just gave you merge permissions, can you try merging?

tanaes:
Still seeing 'Only those with write access can...' 

But Travis is still in progress, so maybe after it finishes it will let me.

antgonza:
@tanaes, it was something else with the permissions, you should be able now.

tanaes:
thanks @antgonza !

PR #1964, opened 2016-09-29
fixing some typos

mortonjt:


antgonza:
Thanks @mortonjt, I have always use the version without `git+`, what's the difference?

mortonjt:
@antgonza that doesn't sound right: http://stackoverflow.com/a/20101940/1167475

It should not work if you pip install directly from a github page without the `git+` tag

antgonza:
@mortonjt the example you sent is for an specific branch not master.

mortonjt:
That doesn't matter - it works the same way.

I just know that the instructions have never worked for my machine.  And pip install has never worked for any other machine that I have seen as shown in the qiita.  I'm not sure exactly what you did on your machine, but I'd be curious to see how you have configured your machine.

antgonza:
@mortonjt you are right, I was reading that wrongly. I normally point to the zip file: `pip install https://github.com/qiita-spots/qiita_client/archive/master.zip` and not the repo.

👍

PR #1965, opened 2016-09-30
Allowing updating the oauth2 info

josenavas:
I don't know why I was not allowing the update of the oauth2 info before, but I allow it now. This will allow to run the plugin tests easier.

### DIFF
--- a/qiita_db/software.py
+++ b/qiita_db/software.py
@@ -579,10 +579,27 @@ class Software(qdb.base.QiitaObject):
 
                 if (client_id != instance.client_id or
                         client_secret != instance.client_secret):
-                    raise qdb.exceptions.QiitaDBOperationNotPermittedError(
-                        'The (client_id, client_secret) pair of the plugin '
-                        '"%s" version "%s" does not match the one in the '
-                        'system' % (name, version))
+                    if update:
+                        sql = """INSERT INTO qiita.oauth_identifiers
+                                    (client_id, client_secret)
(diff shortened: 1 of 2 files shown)


antgonza:
👍 , when tests pass. @ElDeveloper, could you review?

qiita_db/software.py
```diff
@@ -579,10 +579,27 @@ def from_file(cls, fp, update=False):
 
                 if (client_id != instance.client_id or
                         client_secret != instance.client_secret):
-                    raise qdb.exceptions.QiitaDBOperationNotPermittedError(
-                        'The (client_id, client_secret) pair of the plugin '
-                        '"%s" version "%s" does not match the one in the '
-                        'system' % (name, version))
+                    if update:
+                        sql = """INSERT INTO qiita.oauth_identifiers
```
wasade: does this stand to make multiple inserts and is that expected?

josenavas: I think I don't understand your question? This is inserting if it doesn't already exists.

josenavas: I think I understand your question now (thanks @antgonza ). The select that can return a lot of row is inside an `EXISTS` and it is only used to check if that row already exists or not. The first select, can only return 1 row, since the values are passes as a parameter, i.e. it is effectively doing something similar to `SELECT 42`.

wasade: Okay thanks

PR #1966, opened 2016-09-30
Plugin install gui

josenavas:
Depends on #1965 so review/merge that one first.

Modifies the command retrieval to retrieve only active commands (this way only active commands are shown in the GUI).

Makes sure that when Qiita starts, the plugins are called so they register their commands.

The call to create types need to be idempotent, instead of failing. So I changed it here.

### DIFF
--- a/qiita_db/handlers/artifact.py
+++ b/qiita_db/handlers/artifact.py
@@ -197,11 +197,6 @@ class ArtifactTypeHandler(OauthBaseHandler):
         filepath_types : list of (str, bool)
             The list filepath types that the new artifact type supports, and
             if they're required or not in an artifact instance of this type
-
-        Raises
-        ------
-        HTTPError
-            If the artifact type already exists, with error code 400
         """
         a_type = self.get_argument('type_name')
         a_desc = self.get_argument('description')
@@ -212,7 +207,7 @@ class ArtifactTypeHandler(OauthBaseHandler):
         try:
             qdb.artifact.Artifact.create_type(a_type, a_desc, ebi, vamps,
                                               fp_types)
-        except qdb.exceptions.QiitaDBDuplicateError as e:
-            raise HTTPError(400, str(e))
+        except qdb.exceptions.QiitaDBDuplicateError:
+            pass
 
         self.finish()
--- a/qiita_db/handlers/artifact.py
+++ b/qiita_db/handlers/artifact.py
@@ -208,6 +208,8 @@ class ArtifactTypeHandler(OauthBaseHandler):
             qdb.artifact.Artifact.create_type(a_type, a_desc, ebi, vamps,
                                               fp_types)
         except qdb.exceptions.QiitaDBDuplicateError:
+            # Ignoring this error as we want this endpoint in the rest api
+            # to be idempotent.
             pass
 
         self.finish()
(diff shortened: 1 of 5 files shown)


qiita_db/handlers/artifact.py
```diff
@@ -212,7 +207,7 @@ def post(self):
         try:
             qdb.artifact.Artifact.create_type(a_type, a_desc, ebi, vamps,
                                               fp_types)
-        except qdb.exceptions.QiitaDBDuplicateError as e:
-            raise HTTPError(400, str(e))
+        except qdb.exceptions.QiitaDBDuplicateError:
+            pass
```
antgonza: Could you comment why this change?

qiita_db/handlers/artifact.py
```diff
@@ -212,7 +207,9 @@ def post(self):
         try:
             qdb.artifact.Artifact.create_type(a_type, a_desc, ebi, vamps,
                                               fp_types)
-        except qdb.exceptions.QiitaDBDuplicateError as e:
-            raise HTTPError(400, str(e))
+        except qdb.exceptions.QiitaDBDuplicateError:
+            # Ignoring this error as we want this endpoint in the rest api
+            # to be idempotent.
+            pass
```
ElDeveloper: I think you can also set the `_reason` to the response so that there's a difference between it already exists and everything went fine. Not blocking though. http://www.tornadoweb.org/en/stable/web.html#tornado.web.RequestHandler.set_status

qiita_db/software.py
```diff
@@ -42,13 +44,15 @@ class Command(qdb.base.QiitaObject):
...
     @classmethod
-    def get_commands_by_input_type(cls, artifact_types):
+    def get_commands_by_input_type(cls, artifact_types, active_only=True):
         """Returns the commands that can process the given artifact types
 
         Parameters
         ----------
         artifact_type : list of str
             The artifact types
+        active_only : bool
```
ElDeveloper: `bool, optional`.

qiita_db/software.py
```diff
@@ -61,7 +65,10 @@ def get_commands_by_input_type(cls, artifact_types):
                         JOIN qiita.parameter_artifact_type
                             USING (command_parameter_id)
                         JOIN qiita.artifact_type USING (artifact_type_id)
+                        JOIN qiita.software_command USING (command_id)
                      WHERE artifact_type IN %s"""
+            if active_only:
+                sql = "%s AND active = True" % sql
```
ElDeveloper: I almost thought there was a bug in there, was expecting something like `sql += ' AND active = True'`.

scripts/qiita
```diff
@@ -437,7 +438,14 @@ def start(port, master):
...
 
     click.echo("Qiita started on port %d" % port)
-    tornado.ioloop.IOLoop.instance().start()
+    ioloop = tornado.ioloop.IOLoop.instance()
+
+    def callback_function():
+        for s in qdb.software.Software.iter_active():
+            s.register_commands()
+
+    ioloop.add_timeout(time() + 0.5, callback_function)
```
ElDeveloper: Maybe more convenient to use tornado's time? http://www.tornadoweb.org/en/stable/ioloop.html#tornado.ioloop.IOLoop.time

Not suer if it is available in the version of tornado we are using.

qiita_db/handlers/artifact.py
josenavas: Done

qiita_db/software.py

qiita_db/software.py

scripts/qiita

ElDeveloper:
Thanks @josenavas

PR #1968, opened 2016-10-05
Fix 1836

antgonza:
- make both lists start with 5 element
Fixed via datatables config 

-  filter user study list to not contain public studies
Fixed in qiita_pet/handlers/study_handlers/listing_handlers.py by removing public studies from the list

- check why everything in the user study list is awaiting approval
This one was trickier than expected: (1) there was an error in the SQL, we were retrieving all available statuses vs. all statuses of all the artifacts in that study; and (2) the infer_status method was expecting a list of list with one string element but it really returned a list of strings, we were comparing against the first letter. Had to change some of the tests. 

Note that this was tested in the test environment and it works as expected, still deploy if you want to take a look.

### DIFF
--- a/qiita_db/test/test_util.py
+++ b/qiita_db/test/test_util.py
@@ -766,16 +766,16 @@ class UtilTests(TestCase):
         obs = qdb.util.infer_status([])
         self.assertEqual(obs, 'sandbox')
 
-        obs = qdb.util.infer_status([['private']])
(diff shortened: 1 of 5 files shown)


ElDeveloper:
It all looks good to me! Just kinda surprised that none of the tests changed with the changes to util.py. 👍

PR #1972, opened 2016-10-17
get_validator

antgonza:


qiita_db/software.py
```diff
@@ -109,6 +109,41 @@ def get_html_generator(cls, artifact_type):
...
+            The artifact type to search the Validate for
+
+        Returns
+        -------
+        qiita_db.software.Command
+            The newly created command
+
+        Raises
+        ------
+        qdb.exceptions.QiitaDBError
```
josenavas: Can you add an explanation on when is this error raised?

qiita_db/software.py
```diff
@@ -109,6 +109,41 @@ def get_html_generator(cls, artifact_type):
...
+                        JOIN qiita.software_artifact_type USING (software_id)
+                        JOIN qiita.artifact_type USING (artifact_type_id)
+                     WHERE artifact_type = %s
+                        AND name = 'Validate'"""
+            qdb.sql_connection.TRN.add(sql, [artifact_type])
+            try:
+                res = qdb.sql_connection.TRN.execute_fetchlast()
+            except IndexError:
+                raise qdb.exceptions.QiitaDBError(
+                    "There is no command to generate the Validate for "
```
josenavas: `...Validate command...`

antgonza: I rather leave as is as: "There is no command to generate the Validate _command_" sounds weird ...

PR #1974, opened 2016-10-20
Validate uploaded files

### LINKED ISSUE #1085: Allow uploading a processed data (BIOM)
Users should be able to upload a processed data (BIOM tables) to use in their meta-analysis. This requires:
  - Check if the sample ids in the BIOM table correspond to the IDs in the DB
  - Modify the UI interface to support this (the back end already supports this once the previous point is solved)
  - Modify study listing page to check if it has processing information or not.

josenavas:
Fixes #1085

### DIFF
--- a/qiita_core/testing.py
+++ b/qiita_core/testing.py
@@ -1,0 +1,35 @@
+# -----------------------------------------------------------------------------
+# Copyright (c) 2014--, The Qiita Development Team.
+#
+# Distributed under the terms of the BSD 3-clause License.
+#
+# The full license is in the file LICENSE, distributed with this software.
+# -----------------------------------------------------------------------------
+
+from json import loads
+from time import sleep
+
+from moi import r_client
+
+from qiita_db.processing_job import ProcessingJob
+
+
+def wait_for_prep_information_job(prep_id, raise_if_none=True):
+    res = r_client.get('prep_template_%d' % prep_id)
+
+    if raise_if_none:
+        assert res is not None, "unexpectedly None"
+
+    if res is not None:
+        payload = loads(res)
+        job_id = payload['job_id']
+        if payload['is_qiita_job']:
+            job = ProcessingJob(job_id)
+            while job.status not in ('success', 'error'):
+                sleep(0.05)
+        else:
+            redis_info = loads(r_client.get(job_id))
+            while redis_info['status_msg'] == 'Running':
+                sleep(0.05)
+                redis_info = loads(r_client.get(job_id))
+        sleep(0.05)
--- a/qiita_core/testing.py
+++ b/qiita_core/testing.py
@@ -32,7 +32,7 @@ def wait_for_prep_information_job(prep_id, raise_if_none=True):
     """
     res = r_client.get('prep_template_%d' % prep_id)
 
-    if raise_if_none:
+    if raise_if_none and res is None:
         raise AssertionError("unexpectedly None")
 
     if res is not None:
--- a/qiita_core/testing.py
+++ b/qiita_core/testing.py
@@ -15,10 +15,25 @@ from qiita_db.processing_job import ProcessingJob
 
 
 def wait_for_prep_information_job(prep_id, raise_if_none=True):
+    """Waits until a prep information job is completed
+
+    Parameters
+    ----------
+    prep_id : int
+        Prep template id
+    raise_if_none : bool, optional
+        If True, raise an AssertionError if the correspondent redis key
+        is empty. Default: True
+
+    Raises
+    ------
+    AssertionError
+        If `raise_if_none` is True and the correspondent redis key is not set
+    """
     res = r_client.get('prep_template_%d' % prep_id)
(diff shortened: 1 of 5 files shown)


josenavas:
Yay!! Can I have a review in here?

qiita_pet/handlers/api_proxy/tests/test_artifact.py
```diff
@@ -159,6 +159,22 @@ def tearDown(self):
...
+        if payload['is_qiita_job']:
+            job = ProcessingJob(job_id)
+            while job.status not in ('success', 'error'):
+                sleep(0.05)
+        else:
+            redis_info = loads(r_client.get(job_id))
+            while redis_info['status_msg'] == 'Running':
+                sleep(0.05)
+                redis_info = loads(r_client.get(job_id))
+        sleep(0.05)
```
mortonjt: Is it just me,  or is this pattern being copied and pasted?

qiita_pet/handlers/study_handlers/tests/test_artifact.py
```diff
@@ -126,16 +127,19 @@ def test_post_artifact(self):
...
+        job_id = payload['job_id']
+        if payload['is_qiita_job']:
+            job = ProcessingJob(job_id)
+            while job.status not in ('success', 'error'):
+                sleep(0.05)
+        else:
+            redis_info = loads(r_client.get(job_id))
+            while redis_info['status_msg'] == 'Running':
+                sleep(0.05)
+                redis_info = loads(r_client.get(job_id))
```
mortonjt: Here

qiita_pet/handlers/study_handlers/tests/test_artifact.py
```diff
@@ -149,10 +153,17 @@ def test_delete_artifact(self):
...
```
mortonjt: Here

mortonjt: Here

qiita_pet/handlers/study_handlers/tests/test_artifact.py
antgonza: It is but not sure what the solutions could be as the code is duplicated in two different tests files, do you have an idea of how to solve it?

mortonjt: I'd build up a library of testing functions in a separate python script to store these sorts of functions.

But its definitely not blocking - if you guys think this is a good idea, I guess this can be raised as a issue to refactor later.

ElDeveloper:
@josenavas, I'm reviewing this right now.

qiita_pet/handlers/api_proxy/artifact.py
```diff
@@ -318,21 +319,32 @@ def artifact_post_req(user_id, filepaths, artifact_type, name,
...
             return {'status': 'error',
                     'message': "Can't create artifact, no files provided."}
 
-        job_id = safe_submit(user_id, create_raw_data, artifact_type, prep,
-                             cleaned_filepaths, name=name)
+        command = Command.get_validator(artifact_type)
+        job = ProcessingJob.create(
+            User(user_id),
+            Parameters.load(command, values_dict={
+                'template': prep_template_id,
```
ElDeveloper: Super minor but I think this could be prep.id instead of needing this new variable, right?

qiita_core/testing.py
```diff
@@ -0,0 +1,35 @@
...
+
+from json import loads
+from time import sleep
+
+from moi import r_client
+
+from qiita_db.processing_job import ProcessingJob
+
+
+def wait_for_prep_information_job(prep_id, raise_if_none=True):
```
ElDeveloper: Any reason why this function is not documented?

qiita_core/testing.py
```diff
@@ -0,0 +1,35 @@
...
+    res = r_client.get('prep_template_%d' % prep_id)
+
+    if raise_if_none:
+        assert res is not None, "unexpectedly None"
```
ElDeveloper: This may be overcautious, but what about raising an actual AssertionError, when we use `assert` we run the risk that if python is being executed in an "optimized" mode, these statements will be ignored. I suggest this be changed to:

```python
if raise_if_none and res is None
    raise AssertionError('Unexpectedly None')
```

qiita_core/testing.py
josenavas: Oops I forgot! Thanks!

qiita_core/testing.py
josenavas: This is actually a really good point - changed!

qiita_pet/handlers/api_proxy/artifact.py
josenavas: `prep_template_id` is a parameter to this function, so there is no new variable created.

ElDeveloper:
Thanks @josenavas!

PR #1976, opened 2016-10-24
Running validate after a command finishes

josenavas:
This PR removes the assumption that plugins will generate correct artifacts, so after a command finishes, it will call the validate command of the given artifact type. This allows us to make sure that each artifact is fully correct, as well as enabling generating some stuff automatically (e.g. the HTML summary).

Please, review this PR but hold a minute before merging - I'm testing it on the test environment, but if this gets reviewed we get a head start!


qiita_db/processing_job.py
```diff
@@ -332,6 +332,116 @@ def submit(self):
         p = Process(target=_job_submitter, args=(self, cmd))
         p.start()
 
+    def _complete_artifact_definition(self, artifact_data):
+        """"Performs the needed steps to complete an artifact definition job
+
+        In order to complete an artifact definition job we ned to create
```
ElDeveloper: ned -> need

qiita_db/processing_job.py
```diff
@@ -332,6 +332,116 @@ def submit(self):
...
+            # changes that need to be done on the plugins, so we are going to
+            # restrict the number of templates to one. Note that at this moment
+            # there is no way of generating an artifact from 2 or more
+            # artifacts, so we can impose this limitation now and relax it
+            # later.
+            templates = set()
+            for artifact in self.input_artifacts:
+                templates.update(pt.id for pt in artifact.prep_templates)
+            if len(templates) > 1:
+                raise qdb.exceptions.QiitaDBError(
```
ElDeveloper: Can you add this exception in the `Raises` section of the docstring?

ElDeveloper:
👍 looks good to me.

antgonza:
The error is real (already restarted the job once), @josenavas can you fix it?

josenavas:
Yeah - I'm working on it - while testing on the test environment I found an issue and I'm still trying to figure out where the issue resides, but I think it is a bug in the qiita plugins.

qiita_db/support_files/patches/42.sql
```diff
@@ -0,0 +1,15 @@
...
+-- Adding provenance parameter to validate commands
+
+DO $do$
+DECLARE
+    cmd  RECORD;
+BEGIN
+    FOR cmd IN
+        SELECT command_id FROM qiita.software_command WHERE name = 'Validate'
+    LOOP
+        INSERT INTO qiita.command_parameter (command_id, parameter_name, parameter_type, required, default_value)
```
antgonza: Should this add the actual provenance value to the already created artifacts?

antgonza:
Looks good, one question.

qiita_db/support_files/patches/42.sql
josenavas: Note that this is adding a parameter called "provenance" to the validate command. The command itself doesn't use it but it is used internally in Qiita to differentiate a validate command that has been generated as a result of the user uploading data to Qiita from a validate command that has been generated as a result of an "artifact transformation" job finishing. This doesn't really contain the provenance of the artifact, so no need to add to the already created artifacts.

Note that the provenance of the artifacts is stored on the DAG structure in the DB.

antgonza: Got it, thanks. Just to confirm, this value is not required or defined by the artifacts, they are defined as part of the Validate command that generated those artifacts, right?

josenavas: Correct

PR #1977, opened 2016-10-26
WIP: Allowing analysis on single standing bioms

antgonza:
Still figuring out the amount of changes.


antgonza:
Decided to close this and add to: https://github.com/biocore/qiita/pull/1979 so we can test better in the qiita-test system.

PR #1992, opened 2016-11-07
Fix 633

antgonza:
After discussing with @josenavas we agreed that the easiest solution was to simply add those non public studies to the admins listing.

### DIFF
--- a/qiita_pet/handlers/study_handlers/listing_handlers.py
+++ b/qiita_pet/handlers/study_handlers/listing_handlers.py
@@ -22,6 +22,7 @@ from qiita_db.search import QiitaStudySearch
 from qiita_db.logger import LogEntry
 from qiita_db.exceptions import QiitaDBIncompatibleDatatypeError
 from qiita_db.util import (add_message, generate_study_list)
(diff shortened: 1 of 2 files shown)


qiita_pet/handlers/study_handlers/listing_handlers.py
```diff
@@ -72,8 +73,12 @@ def _build_study_info(user, search_type, study_proc=None, proc_samples=None):
 
     # get list of studies for table
     if search_type == 'user':
-        user_study_set = user.user_studies.union(user.shared_studies)
-        study_set = user_study_set - Study.get_by_status('public')
+        if user.level != 'admin':
+            user_study_set = user.user_studies.union(user.shared_studies)
+            study_set = user_study_set - Study.get_by_status('public')
+        else:
+            portal_set = Portal(qiita_config.portal).get_studies()
```
josenavas: Why is this line needed? `Study.get_by_status` should already return the studies that are only visible in the current portal. If not, there is a bug in that function so I would prefer to change that function rather than patch it here.

PR #1995, opened 2016-11-13
Fix documentation link on homepage.

srh:
Replaces smart quotes with ASCII quotes so that the link works.


antgonza:
@srh, thanks for fixing.

PR #2002, opened 2016-11-20
Fixing 1998

josenavas:
`generate_study_list` was returning all the artifacts, w/o looking at the status of them. Now, if the list being created is the public list, it skips the BIOM artifacts that are not public.


antgonza:
👍 , @ElDeveloper, could you take a look?

PR #2004, opened 2016-11-24
Offloading file creation form the webserver

josenavas:
I think there was an issue open but I can't find it - if somebody knows which one it is, it will be nice if it can be linked here.

Already noted in the code but explaining it here to. It is moving this processing to the ipyhon cluster, which means that qiita_db.handlers is importing from qiita_ware. We had a rule to not import from qiita_ware in qiita_db, but I think in this case is acceptable given that it's the webserver.

This may point out that we probably want to move this handlers out of qiita_db, and create a new module called qiita_rest.

### DIFF
--- a/qiita_core/testing.py
+++ b/qiita_core/testing.py
@@ -49,7 +49,7 @@ def wait_for_prep_information_job(prep_id, raise_if_none=True):
 
 
 def wait_for_processing_job(job_id):
-    """Waits unitl a processing job is completed
+    """Waits until a processing job is completed
 
     Parameters
     ----------
--- a/qiita_core/testing.py
+++ b/qiita_core/testing.py
@@ -39,12 +39,24 @@ def wait_for_prep_information_job(prep_id, raise_if_none=True):
         payload = loads(res)
         job_id = payload['job_id']
         if payload['is_qiita_job']:
-            job = ProcessingJob(job_id)
-            while job.status not in ('success', 'error'):
-                sleep(0.05)
+            wait_for_processing_job(job_id)
         else:
             redis_info = loads(r_client.get(job_id))
             while redis_info['status_msg'] == 'Running':
                 sleep(0.05)
                 redis_info = loads(r_client.get(job_id))
         sleep(0.05)
+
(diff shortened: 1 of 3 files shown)


qiita_core/testing.py
```diff
@@ -39,12 +39,24 @@ def wait_for_prep_information_job(prep_id, raise_if_none=True):
...
         else:
             redis_info = loads(r_client.get(job_id))
             while redis_info['status_msg'] == 'Running':
                 sleep(0.05)
                 redis_info = loads(r_client.get(job_id))
         sleep(0.05)
+
+
+def wait_for_processing_job(job_id):
+    """Waits unitl a processing job is completed
```
antgonza: until

qiita_db/handlers/processing_job.py
```diff
@@ -10,6 +10,11 @@
 
 from tornado.web import HTTPError
 
+# We agreed before that qiita db should never import from
+# qiita_ware. However, this is part of the rest API and I think it
+# is acceptable to import from qiita_ware, specially to offload
+# processed to the ipython cluster
```
ElDeveloper: processed -> processing (?)

qiita_db/handlers/processing_job.py
```diff
@@ -45,6 +50,30 @@ def _get_job(job_id):
...
+    if payload_success:
+        artifacts = payload['artifacts']
+        error = None
+    else:
+        artifacts = None
+        error = payload['error']
+    job = qdb.processing_job.ProcessingJob(job_id)
+    try:
+        job.complete(payload_success, artifacts, error)
+    except Exception as e:
```
ElDeveloper: Would it be worth to save the traceback itself? Or is this information presented to the user?

josenavas: It may be shown to the user, but if it fails at this point is better for us if it is actually store the traceback, so thanks for pointing that out!

ElDeveloper:
Looks great, thanks for making the changes @josenavas!

Issue #1024, opened 2015-03-31
Impossible to test data patches


squirrelo:
Test data is loaded into the database after applying all patches, so any patches that rely on data to create new rows/alter existing rows will never be tested for correctness.

antgonza:
This is the broader comment of: https://github.com/biocore/qiita/issues/717

josenavas:
This is also a broader discussion - if we can switch the tests for not relying on SQL and do them independent of the actual data held in SQL, we can create a populate test db that loads at the very beginning and then all patches are applied.

squirrelo:
This has actually worked out really well in practice for AG tests, so I'd say go that way.

antgonza:
Group resolution: this will be a valuable addition but it's low priority. The idea will be to populate the DB, then make sure that the latests tests pass (meaning that the test were successful after the patch).

antgonza:
This was fixed here: https://github.com/biocore/qiita/pull/2006

PR #2006, opened 2016-11-30
Fix 1996

antgonza:
This PR will:
- [X] Change the way we are testing. From now on (patch 43) we will load the test DB and then patch for testing. This is something we discussed in the past and which will allow us to test the patches with the current setting. Basically, fixing: https://github.com/biocore/qiita/issues/1024 
- [X] Fix the GUI so the user can move and zoom the pipeline summary
- [X] Add a new SQL function that returns the descendant artifact ids and the job ids that generated them of a given id
- [X] Change the API call to use this new function and modify the GUI so it takes advantage of this
- [X] Figure out if we can have automatic toggling: the automatic clustering doesn't really work as expected. I think that the new visual makes it easier to read and follow so no need for the toggling but open to discussion. 

~~As you can see this is incomplete but want to let the test run while I work on the final touches of the GUI~~


antgonza:
This is ready for review: @tanaes, @ElDeveloper or @josenavas. Note that this has been deployed in the test environment:
- Screenshot of a study with not a lot of processing
screen shot 2016-12-01 at 9 17 42 am
- Zoom/iteraction, etc:

qiita_db/environment_manager.py
```diff
@@ -387,6 +383,10 @@ def patch(patches_dir=PATCHES_DIR, verbose=False):
 
         for sql_patch_fp in sql_patch_files[next_patch_index:]:
             sql_patch_filename = basename(sql_patch_fp)
+
+            if sql_patch_filename == '43.sql' and test:
```
ElDeveloper: Can you add a comment explaining that patch 43 is when we begun testing patches, etc?

qiita_db/support_files/patches/python_patches/43.py
```diff
@@ -0,0 +1,68 @@
...
+            c = nodes[child]
+            cmd_out = c.artifact_type
+            if cmd_out == 'Demultiplexed':
+                cmd_out = 'demultiplexed'
+            elif cmd_out == 'BIOM':
+                cmd_out = 'OTU table'
+            else:
+                # the actual DB has other possible values in
+                # artifact_type
+                print cmd_out
```
ElDeveloper: Do you want to leave this line in case there are unexpected artifact types? Or should this be removed?

qiita_db/test/test_artifact.py
```diff
@@ -320,6 +320,45 @@ def test_descendants(self):
...
+        # as jobs are created at random we will only check that the artifacts
+        # are there and that the number of jobs matches
+        exp_nodes = [('artifact', A(1)), ('artifact', A(2)),
+                     ('artifact', A(3)), ('artifact', A(4)),
+                     ('artifact', A(5)), ('artifact', A(6))]
+        for e in exp_nodes:
+            self.assertIn(e, obs_nodes)
+        self.assertEqual(5, len([e for dt, e in obs_nodes if dt == 'job']))
+        obs_edges = obs.edges()
+        # as jobs are created at random we will only the number of pairs
```
ElDeveloper: only -> only check

qiita_db/support_files/patches/python_patches/43.py
antgonza: I left it there to make sure there was only 1 occurrence in the test env but it's not really needed, removing.

PR #2007, opened 2016-12-01
fix #1928

### LINKED ISSUE #1928: Clarify license of data set uploaded to qiita
When someone uploads their data set to Qiita, what is the license on that data set?

[Github encourages licenses on repos](https://help.github.com/articles/open-source-licensing/) to decrease ambiguity about terms of (re)use.
[Zenodo requires users to select one of the 'Creative Commons Attribution-*` licenses](https://zenodo.org/deposit/new), but only if the data set is going to be made public.

So what's up with the data sets on Qiita?

Edit: So the SRA is run by INSDC, and looks like INSDC data may be public-domain:
From the SRA paper [10.1093/nar/gkq1019](https://www.ncbi.nlm.nih.gov/pmc/articles/PMC3013647/): 

> The mission of INSDC is to preserve _public-domain_ sequencing data ...

From the [INSDC Policy page](http://www.insdc.org/policy.html):

> The INSD will not attach statements to records that restrict access to the data, limit the use of the information in these records, or prohibit certain types of publications based on these records. Specifically, _no use restrictions or licensing requirements_ will be included in any sequence data records, and no restrictions or licensing fees will be placed on the redistribution or use of the database by any party. 

Emphasis mine.

antgonza:
This adds our terms/conditions to qiita. The link is added to the email sent to new users and present on the footer of all pages in the system.

### DIFF
--- a/qiita_pet/static/qiita_data_terms_of_use.html
+++ b/qiita_pet/static/qiita_data_terms_of_use.html
@@ -117,7 +117,7 @@
             impairment of, or the assessment of the health of, human beings.
           </li>
           <li>
-            You may consult Qitta, including the Data generated as part of a
+            You may consult Qiita, including the Data generated as part of a
             study that you conduct using Qiita, solely for your own personal
             or, as applicable, organizational purposes in accordance with
             these Terms of Use.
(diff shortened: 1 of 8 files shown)


ElDeveloper:
👍 (note that I've only verified the validity of the source and not of the content of the source).

qiita_pet/static/qiita_data_terms_of_use.html
```diff
@@ -0,0 +1,249 @@
...
+      <li><u>Restrictions on Use; No Clinical Use.</u></li>
+        <ol type="a">
+          <li>
+            Qiita is a research tool only, and you acknowledge and agree that
+            you shall not use Qiita for any clinical purpose, including for
+            the diagnosis, prevention, or treatment of any disease of
+            impairment of, or the assessment of the health of, human beings.
+          </li>
+          <li>
+            You may consult Qitta, including the Data generated as part of a
```
mortonjt: `Qitta` -> `Qiita`

mortonjt:
Other than the single typo, it looks good to m.

PR #2016, opened 2016-12-09
Changed the pip command in order to have qiita-files installed

mdehollander:
Make sure pip installs qiita-files


antgonza:
Thanks @mdehollander

PR #2024, opened 2016-12-15
fix #388

### LINKED ISSUE #388: Make sure that PrepTemplate and SampleTemplate does not have columns in common
as part of the metadata validation.

antgonza:
Decided to add _check_duplicated_columns cause it avoids duplication of code, which is an internal method in the prep_template.py.

### DIFF
--- a/qiita_db/metadata_template/prep_template.py
+++ b/qiita_db/metadata_template/prep_template.py
@@ -13,6 +13,7 @@ from os.path import join
 from time import strftime
 from copy import deepcopy
 import warnings
+from skbio.util import find_duplicates
 
 import pandas as pd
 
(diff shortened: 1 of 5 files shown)


antgonza:
Forgot to mention that I also have code to deal with the current duplicated fields, following Gail's suggestions, so the patch shouldn't do anything in the real system. That code is being tested in the test environment and applied before deploying this patch.

josenavas:
Tests are failing. Code looks good. It looks like all the python patch is dealing with SQL only - is it possible to write that code in pure SQL? If it is going to be a pain - fine as it is.

antgonza:
Restarted tests, it was the random postgres error. I would prefer to leave the patch as is.

ElDeveloper:
👍

josenavas:
@antgonza it seems that there is a real error

antgonza:
Yes, the error was that one of our tests was adding a duplicated column 😄

PR #2025, opened 2016-12-16
Fixing bug with summary jobs

josenavas:
Found this bug while developing the qtp-biom plugin. Basically, the "Generate HTML summary" jobs do not generate new artifacts. This bug was not affecting any functionality per se, but it was leaving jobs in the DB in the "runnning" state although they successfully completed.

Added a specific test for this issue.

### DIFF
--- a/qiita_db/test/test_processing_job.py
+++ b/qiita_db/test/test_processing_job.py
@@ -394,6 +394,16 @@ class ProcessingJobTest(TestCase):
         # Implicitly tested by "test_complete"
         pass
 
+    def test_complete_no_artifact_data(self):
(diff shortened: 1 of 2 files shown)


qiita_db/test/test_processing_job.py
```diff
@@ -448,7 +458,9 @@ def test_complete_success(self):
         alljobs = set(self._get_all_job_ids())
 
         job.complete(True, artifacts_data=artifacts_data)
-        self.assertTrue(job.status, 'success')
+        # The validator job spawn blocks this job in running state until
```
antgonza: Can you rewrite this? Is not really clear.

PR #2034, opened 2016-12-28
fix #1609

### LINKED ISSUE #1609: Banner message masks text and appears out of scale
This is following the update from Jan. 14, 2016 on the production system. 
screen shot 2016-01-16 at 3 07 13 pm

antgonza:
Turns out that the fix was easier than expected.

Now it looks like this with just the alert message:
Example 1:
screen shot 2016-12-27 at 8 17 18 pm
Example 2:
screen shot 2016-12-27 at 8 17 03 pm
and with the extra message:
screen shot 2016-12-27 at 8 16 53 pm
and adding samples to analysis:
screen shot 2016-12-27 at 8 17 51 pm

### DIFF
--- a/qiita_pet/templates/sitebase.html
+++ b/qiita_pet/templates/sitebase.html
@@ -8,6 +8,22 @@
 {% set user = current_user %}
 {% set qiita_version, qiita_sha = get_qiita_version() %}
 
+{% set level = globals().get('level', '') %}
+{% if level not in {'danger', 'success', 'info', 'warning'} %}
+    {% set level = 'info' %}
+{% end %}
+{% set message = globals().get('message', '') %}
+{% if maintenance is not None %}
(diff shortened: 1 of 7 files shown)


qiita_pet/templates/sitebase.html
```diff
@@ -8,6 +8,22 @@
 {% set user = current_user %}
 {% set qiita_version, qiita_sha = get_qiita_version() %}
 
+{% set level = globals().get('level', '') %}
+{% if level not in {'danger', 'success', 'info', 'warning'} %}
+    {% set level = 'info' %}
+{% end %}
+{% set message = globals().get('message', '') %}
+{% if maintenance is not None %}
+  {# maintenance mode takes precidence #}
```
wasade: precedence

PR #2035, opened 2016-12-31
adding User.jobs()

antgonza:
This retrieves the jobs for a user. Note that this only retrieves the processing jobs, AKA no moi jobs. However, AFAIK only the analysis is currently using moi jobs and @josenavas is working on removing that as part of the refactor. I'll do another PR to add the GUI part of this functionality. This should be reviewed and merged before the other PR.

### DIFF
--- a/qiita_db/processing_job.py
+++ b/qiita_db/processing_job.py
@@ -866,8 +866,8 @@ class ProcessingJob(qdb.base.QiitaObject):
             qdb.sql_connection.TRN.add(sql, [self.id])
             r = qdb.sql_connection.TRN.execute_fetchindex()
 
-            return (None if not r
-                    else qdb.processing_job.ProcessingWorkflow(r[0][0]))
+            return (qdb.processing_job.ProcessingWorkflow(r[0][0]) if r
+                    else None)
 
 
 class ProcessingWorkflow(qdb.base.QiitaObject):
--- a/qiita_db/processing_job.py
+++ b/qiita_db/processing_job.py
@@ -850,6 +850,25 @@ class ProcessingJob(qdb.base.QiitaObject):
                 name: qdb.artifact.Artifact(aid)
                 for aid, name in qdb.sql_connection.TRN.execute_fetchindex()}
 
+    @property
+    def processing_job_worflow(self):
+        """The processing job worflow
+
+        Returns
+        -------
+        ProcessingWorkflow
+            The processing job workflow the job
+        """
+        with qdb.sql_connection.TRN:
+            sql = """SELECT processing_job_workflow_id
+                     FROM qiita.processing_job_workflow_root
+                     WHERE processing_job_id = %s"""
+            qdb.sql_connection.TRN.add(sql, [self.id])
+            r = qdb.sql_connection.TRN.execute_fetchindex()
+
+            return (None if not r
+                    else qdb.processing_job.ProcessingWorkflow(r[0][0]))
+
 
 class ProcessingWorkflow(qdb.base.QiitaObject):
     """Models a workflow defined by the user
--- a/qiita_pet/handlers/api_proxy/user.py
+++ b/qiita_pet/handlers/api_proxy/user.py
@@ -29,7 +29,7 @@ def user_jobs_get_req(user):
      'template': {sample: {column: value, ...}, ...}
     """
 
-    response = {}
+    response = []
     cmds = {}
     for j in user.jobs():
         cmd = j.command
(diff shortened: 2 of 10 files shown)


antgonza:
Turns out that you _can_ remove processingjobs via the processingworkflow method, which is much cleaner than adding a processingjob remove. However, the processing job didn't have a method to return its processingworkflow. Just added it.

qiita_db/processing_job.py
```diff
@@ -850,6 +850,25 @@ def outputs(self):
...
+            The processing job workflow the job
+        """
+        with qdb.sql_connection.TRN:
+            sql = """SELECT processing_job_workflow_id
+                     FROM qiita.processing_job_workflow_root
+                     WHERE processing_job_id = %s"""
+            qdb.sql_connection.TRN.add(sql, [self.id])
+            r = qdb.sql_connection.TRN.execute_fetchindex()
+
+            return (None if not r
```
wasade: what about `return qdb.processing_job.ProcessingWorkflow(r[0][0]) if r else None`? I don't think the use of a negative conditional is necessary

qiita_pet/handlers/api_proxy/user.py
```diff
@@ -0,0 +1,53 @@
...
+    {'status': status,
+     'message': message,
+     'template': {sample: {column: value, ...}, ...}
+    """
+
+    response = []
+    cmds = {}
+    for j in user.jobs():
+        cmd = j.command
+        if cmd not in cmds:
```
wasade: is there a valid situation in which the same command will appear multiple times?

qiita_pet/handlers/api_proxy/user.py
```diff
@@ -0,0 +1,53 @@
...
+            cmds[cmd] = cmd
+        ccmd = cmds[cmd]
```
wasade: `ccmd is cmd -> True`, is there a need for this additional variable?

qiita_pet/handlers/user_handlers.py
```diff
@@ -185,3 +186,10 @@ def post(self):
 
         self.render("user_messages.html",
                     messages=self.current_user.messages())
+
+
+class UserJobs(BaseHandler):
+    @authenticated
+    def get(self):
+        response = user_jobs_get_req(self.current_user)
```
wasade: this executes as a transaction and has the potential to block as it is performing IO. Should this method be a tornado `coroutine` instead?

qiita_pet/handlers/api_proxy/user.py
antgonza: yes, multiple and the idea of cache; basically, the command is the definition of the command (split libs, demux, etc) and the job is the command with the specific params.

qiita_pet/handlers/api_proxy/user.py
antgonza: Not necessary but I like it cause that makes the subsequent calls smaller. However, we are only retrieving the name of the command so change it a little bit.

qiita_pet/handlers/user_handlers.py
antgonza: sure, changing

qiita_pet/handlers/api_proxy/user.py
wasade: I don't see how this caches anything though? `cmds` is reset prior to the forloop so the only compute that is avoided is the assignment of a value to a key in a dict?

wasade: ...on closer inspection, it doesn't look like `cmds` is needed at all

qiita_pet/handlers/user_handlers.py
```diff
@@ -185,3 +187,11 @@ def post(self):
 
         self.render("user_messages.html",
                     messages=self.current_user.messages())
+
+
+class UserJobs(BaseHandler):
+    @authenticated
+    @coroutine
+    def get(self):
+        response = user_jobs_get_req(self.current_user)
```
wasade: I believe this needs take the result of a `yield`? 

http://www.tornadoweb.org/en/stable/gen.html

qiita_pet/handlers/user_handlers.py
```diff
@@ -1,8 +1,10 @@
 from tornado.web import authenticated, HTTPError
+from tornado.gen import coroutine
```
josenavas: There are flake8 errors because this import is not being used.

PR #2036, opened 2017-01-03
Fix 1505

antgonza:
This is the graphical part of https://github.com/biocore/qiita/pull/2035 so review that one first. This has been deployed in the test environment. After review there we should discuss:
- [x] displaying all jobs is basically useless/impossible(?) for admins, should we always display only the latests 10 jobs? ===> 30, and avoid displaying successful jobs
- [x] some jobs don't have workflows so deleting via workflows is not always possible. We could add deletion of jobs via workflows, even if the workflow doesn't exist or differentiate between the jobs that have a workflow or not - I like the first option ==> as we can only remove jobs in_construction, all those jobs have a workflow; not a problem anymore.  
- [x] how often should we "refresh" the page, currently 1 second ===> 30 seconds
- [ ] some of the jobs don't have heartbeat, what should we do? ===> they don't have a heartbeat cause they are not running.
- [x] error messages are only displayed on reload, thus not showed with this new code. ==> review
- [x] the GUI ==> remove params and just display when on top

Note that here I'm adding [vue.min.js](https://vuejs.org/) and it's fully used by the new GUI, which we should also discuss where we want to use.

The example:
qiita-processingjob


josenavas:
@antgonza merge conflicts here, but I think we are talking about this PR today during the meeting?

antgonza:
This is ready for another review, all comments addressed. Main changes:
- Only updates the jobs when the div is shown, see the small delay to show the list in the example below
- Update every 30 secs
- Only refreshes the view if the data actually changes
- Fixed issues and followed GUI recommendations discussed in today's meeting
- Delete button only on those jobs that can be deleted.

Example from qiita-test deploy:
qiita-processingjob

antgonza:
@josenavas, could you review?

PR #2040, opened 2017-01-09
Analysis refactor db

josenavas:
Doing the PR against the correct branch.

Modifies the DB to transfer the analysis to the plugin system.

Sorry for the length of the SQL patch.


qiita_db/support_files/patches/47.sql
```diff
@@ -0,0 +1,404 @@
...
+    -- makes the process of trasnferring the data from the old structure
+    -- to the new one a bit more challenging, as we will need to handle
+    -- different special cases.
+
+    -- Special case 1: there are jobs in the database that do not contain
+    -- any information about the options used to process those parameters.
+    -- However, these jobs do not have any results and all are marked either
+    -- as queued or error, although no error log has been saved. Since these
+    -- jobs are mainly useleess, we are going to remove them from the system
+    DELETE FROM qiita.analysis_job
```
antgonza: I _think_ these are the base for new analysis creation and done at User confirmation: https://github.com/biocore/qiita/blob/master/qiita_db/user.py#L320

qiita_db/support_files/patches/python_patches/47.py
```diff
@@ -0,0 +1,50 @@
...
+    for a_id in TRN.execute_fetchflatten():
+        a = Artifact(a_id)
+        params = a.processing_parameters.values
+        # load the biom table to check the rarefaction depth
+        # Magic numbers: since we added the artifacts on the patch 47.sql, we
+        # know that these artifacts have only 1 file, hence the first 0.
+        # Each element of the filepath list is a 3-tuple with filepath_id,
+        # filepath and filepath type, and we are interested in the filepath
+        # (index 1)
+        t = load_table(a.filepaths[0][1])
```
antgonza: This assumes that all analyses were rarefied, which is not the case. You need to check some of the samples and decide if it's rarefied or not.

qiita_db/support_files/patches/47.sql
```diff
@@ -0,0 +1,404 @@
...
+            -- Step 1.1.b: Associate the artifact with the analysis
+            INSERT INTO qiita.analysis_artifact (analysis_id, artifact_id)
+                VALUES (analysis.analysis_id, initial_biom_id);
+            -- Step 1.1.c: Link the artifact with its file
+            INSERT INTO qiita.artifact_filepath (artifact_id, filepath_id)
+                VALUES (initial_biom_id, biom_data.filepath_id);
+
+            -- Step 1.2: Create the single rarefaction job
+            -- Step 1.2.a: Add the row in the procesisng job table
+            -- depth 1000 -> the depth information is not stored in the database
```
wasade: so the depth information that gets stored here may be incorrect?

qiita_db/support_files/patches/47.sql
```diff
@@ -0,0 +1,76 @@
+-- Jan 5, 2017
+-- Move the analysis to the plugin system. This is a major rewrite of the
+-- database backend that supports the analysis pipeline.
+-- After exploring the data on the database, we realized that
+-- there are a lot of inconsistencies in the data. Unfortunately, this
+-- makes the process of trasnferring the data from the old structure
```
antgonza: trasnferring -> transferring

qiita_db/support_files/patches/python_patches/47.py
```diff
@@ -0,0 +1,677 @@
...
+                    JOIN qiita.artifact USING (artifact_id)
+                 WHERE analysis_id = %s AND data_type_id = %s
+                 GROUP BY artifact_id"""
+        TRN.add(sql, [analysis['analysis_id'], biom_data['data_type_id']])
+        samples_by_artifact = TRN.execute_fetchindex()
+
+        # Create an empty BIOM table to be the new master table
+        new_table = Table([], [], [])
+        ids_map = {}
+        for a_id, samples in samples_by_artifact:
```
antgonza: This has an extra level of complexity, right? You need to merge based on the processing params of the artifact, right?

antgonza: Scratch that, this is for the already existing bioms so there we should had done that and if not we are simply recovering so we _should_ be fine. Leaving here so we keep in mind.

qiita_db/support_files/patches/python_patches/47.py
```diff
@@ -0,0 +1,677 @@
...
+                              '"min_rare_depth":"10",'
+                              '"max_rare_depth":"Default"}'
+                              % (initial_biom_id, tree_fp))
+                    output_artifact_type_id = rc_atype_id
+                    cmd_out_id = arare_cmd_out_id
+
+                transfer_job()
+
+with TRN:
+    # Unlink the analysis from the biom table filepaths
```
antgonza: Due to permissions in our DB, could you put this in a try/except and if it fails print the SQL commands?

antgonza:
@wasade, could you take another look just to make sure that the changes make sense? Thanks!

qiita_db/support_files/patches/python_patches/47.py
josenavas: Correct, we should've stored the original table and applied rarefaction afterwards. The way is done now is doing this.

qiita_db/support_files/patches/python_patches/47.py
josenavas: Sure

josenavas:
@wasade @mortonjt @ElDeveloper available for a review in here?

josenavas:
ping @biocore/qiita-dev

qiita_db/support_files/patches/python_patches/47.py
```diff
@@ -0,0 +1,680 @@
...
+    sql = """INSERT INTO qiita.command_parameter
+                (command_id, parameter_name, parameter_type, required)
+             VALUES (%s, %s, %s, %s)"""
+    sql_args = [(validate_cmd_id, 'files', 'string', True),
+                (validate_cmd_id, 'artifact_type', 'string', True),
+                (html_summary_cmd_id, 'input_data', 'artifact', True)]
+    TRN.add(sql, sql_args, many=True)
+
+    # Step 4: Add the new artifact types
+    sql = """INSERT INTO qiita.artifact_type (
```
wasade: This is a general comment. One thing which came up in Krumlov, which almost resulted in @ElDeveloper's death, was that "artifact" has a very negative meaning to biologists. We pinged Greg about it regarding its use in QIIME2 and it is under discussion for possibly changing that the name used.

qiita_db/support_files/patches/python_patches/47.py
```diff
@@ -0,0 +1,680 @@
...
+                # The BIOM table was rarefied
+                # Create the initial unrarefied artifact
+                initial_biom_artifact_id = create_non_rarefied_biom_artifact(
+                    analysis, biom_data, table)
+                # Create the rarefaction job
+                rarefaction_job_id, params = create_rarefaction_job(
+                    depths.pop(), initial_biom_artifact_id, analysis,
+                    srare_cmd_id)
+                # Create the rarefied artifact
+                rarefied_biom_artifact_id = create_rarefied_biom_artifact(
```
wasade: is this creating a new rarefied version of the unrarefied table? since this is a random process, won't this table be different than what `table` currently is?

qiita_db/support_files/patches/python_patches/47.py
```diff
@@ -0,0 +1,680 @@
...
+                    cmd_id = sum_taxa_cmd_id
+                    params = ('{"biom_table":%d,"metadata_category":"",'
+                              '"sort":false}' % initial_biom_id)
+                    output_artifact_type_id = ts_atype_id
+                    cmd_out_id = sum_taxa_cmd_out_id
+                elif job_data['command_id'] == 2:
+                    # Beta diversity
+                    cmd_id = bdiv_cmd_id
+                    tree_fp = loads(job_data['options'])['--tree_fp']
+                    params = ('{"biom_table":%d,"tree":"%s","metrics":'
```
wasade: the default bdiv metrics differ -- is it the case that all analyses here have a tree?

wasade:
@antgonza @josenavas , I'm not very comfortable commenting on the PR

qiita_db/support_files/patches/python_patches/47.py
josenavas: Thanks - that make sense. I think we can normalize once a decision is made.

qiita_db/support_files/patches/python_patches/47.py
josenavas: Good question - the name of the function may be misleading. The function is gathering the biom file that is attached to the analysis and creating an artifact with that file (so no new BIOM is created).

qiita_db/support_files/patches/python_patches/47.py
josenavas: Now that you mention it I think I remember @antgonza adding the functionality to run other metrics if the tree didn't exist for the given processing parameters. Will change to take that into account.

josenavas:
I've fixed @wasade's comments regarding the non-phylogenetic metrics.

PR #2044, opened 2017-01-16
Analysis refactor fix analysis

josenavas:
Depends on https://github.com/biocore/qiita/pull/2040

Fixes the artifact object to be using the plugins


antgonza:
@josenavas can you update/resolve-conflicts?

josenavas:
@antgonza done

Unsure why this is still showing all the commits - although the files changed show only the changes made in here.

qiita_db/analysis.py
```diff
@@ -897,16 +775,17 @@ def build_files(self,
...
-            self._build_biom_tables(grouped_samples, rarefaction_depth,
-                                    rename_dup_samples)
+            biom_files = self._build_biom_tables(
+                grouped_samples, rename_dup_samples)
 
-    def _build_biom_tables(self, grouped_samples, rarefaction_depth=None,
-                           rename_dup_samples=False):
+            return biom_files
+
+    def _build_biom_tables(self, grouped_samples, rename_dup_samples=False):
```
antgonza: Note that this doesn't take into account grouping by processing parameter, which will merge deblur and close ref tables, fine to leave as is if it will be addressed in a future pr.

qiita_db/artifact.py
```diff
@@ -269,107 +284,163 @@ def create(cls, filepaths, artifact_type, name=None, prep_template=None,
...
 
-                # Check that all parents have the same data type
+        with qdb.sql_connection.TRN:
+            if parents:
                 dtypes = {p.data_type for p in parents}
-                if len(dtypes) > 1:
+                # If an artifact has parents, it can be either from the
+                # processing pipeline or the analysis pipeline. Decide which
+                # one here
+                studies = set()
```
antgonza: This is something I'm not sure but wondering, is it faster to create this as an array and then do a set on it or this?

qiita_db/artifact.py
```diff
@@ -269,107 +284,163 @@ def create(cls, filepaths, artifact_type, name=None, prep_template=None,
...
+                analyses = set()
+                for p in parents:
+                    s = p.study
+                    a = p.analysis
+                    if s is not None:
```
antgonza: Out of curiosity, which cases will a selected study/analysis be None?

qiita_db/handlers/tests/test_analysis.py
```diff
@@ -0,0 +1,106 @@
...
+import qiita_db as qdb
+
+
+class UtilTests(TestCase):
+    def test_get_analysis(self):
+        obs = _get_analysis(1)
+        exp = qdb.analysis.Analysis(1)
+        self.assertEqual(obs, exp)
+
+        # It doesn't eist
```
antgonza: eist?

qiita_db/analysis.py
josenavas: I was not modifying the functionality done in here, so I will take a note to fix it in a later PR.

qiita_db/artifact.py
josenavas: This is faster:
```python
In [3]: def f():
   ...:     s = set()
   ...:     for i in range(10000):
   ...:         s.add(i)
   ...:     return s
   ...: 

In [4]: def g():
   ...:     s = []
   ...:     for i in range(10000):
   ...:         s.append(i)
   ...:     return set(s)
   ...: 

In [5]: %timeit f()
1000 loops, best of 3: 958 µs per loop

In [6]: %timeit g()
1000 loops, best of 3: 1.09 ms per loop
```

qiita_db/artifact.py
josenavas: An artifact can be attached to a study OR to an analysis, but not both. Hence one or the other will be None.

qiita_db/handlers/tests/test_analysis.py
josenavas: exist - fixed

josenavas:
@antgonza the functionality/checks removed here are not needed anymore. The only one that it still may be useful but it needs to completely be rewritten is the generate TGZ file, but I want to do that at the end since now an analysis will look the same way as a study, so it is not that easy to generate the TGZ. Also, once we have the TGZ for the analysis, we can add it for the studies too.

qiita_db/analysis.py
```diff
@@ -204,15 +205,20 @@ def delete(cls, _id):
...
                 raise qdb.exceptions.QiitaDBUnknownIDError(_id, "analysis")
 
+            # Check if the analysis has any artifact
+            sql = """SELECT EXISTS(SELECT *
+                                   FROM qiita.analysis_artifact
+                                   WHERE analysis_id = %s)"""
+            qdb.sql_connection.TRN.add(sql, [_id])
+            if qdb.sql_connection.TRN.execute_fetchlast():
+                raise qdb.exceptions.QiitaDBOperationNotPermittedError(
+                    "Can't delete analysis %d, has artifacts attached")
```
ElDeveloper: This format string is missing the analysis identifier, can you add the appropriate formatting?

qiita_db/analysis.py
```diff
@@ -619,22 +509,47 @@ def pmid(self, pmid):
...
             sql = """UPDATE qiita.{0} SET pmid = %s
                      WHERE analysis_id = %s""".format(self._table)
             qdb.sql_connection.TRN.add(sql, [pmid, self._id])
             qdb.sql_connection.TRN.execute()
 
-    # @property
-    # def parent(self):
-    #     """Returns the id of the parent analysis this was forked from"""
-    #     return QiitaDBNotImplementedError()
+    # ---- Functions ----
```
ElDeveloper: Should this be Methods instead of Functions? Also, any reason why this is being explicitly highlighted here?

qiita_db/artifact.py
```diff
@@ -269,107 +284,163 @@ def create(cls, filepaths, artifact_type, name=None, prep_template=None,
             raise qdb.exceptions.QiitaDBArtifactCreationError(
                 "at least one filepath is required.")
 
-        # Parents or prep template must be provided, but not both
-        if parents and prep_template:
+        # Check that the combination of parameters is correct
+        counts = Counter([bool(parents or processing_parameters),
```
ElDeveloper: Would casting each of these boolean values to integers make sense instead? That way you would be able to sum all three values and if the result is not equal to one then you would know more than one parameter was specified. Not blocking, just an alternative to using `Counter`.

qiita_db/artifact.py
ElDeveloper: Why not both :neckbeard: ?

```python
In [11]: %paste
def f():
    s = set()
    for i in range(10000):
        s.add(i)
    return s


def g():
    s = []
    for i in range(10000):
        s.append(i)
    return set(s)

def h():
    return {i for i in range(10000)}
## -- End pasted text --

In [12]: %timeit f()
1000 loops, best of 3: 1.11 ms per loop

In [13]: %timeit g()
1000 loops, best of 3: 1.22 ms per loop

In [14]: %timeit h()
1000 loops, best of 3: 573 µs per loop
```

qiita_db/analysis.py
josenavas: Good catch!

qiita_db/analysis.py
josenavas: Nah, it was just old documentation that was there. Removed as I don't think it divides the file between properties and methods anymore...

qiita_db/artifact.py
josenavas: Good call, I think when I started writing this I had cases for the counts but it end up evolving and at the end I don't need it - changed to a simple addition of ints (although with a few casts...)

qiita_db/artifact.py
josenavas: I actually decided to make a more realistic test and it turns out that it doesn't matter because the bulk of the time is spent on accessing the DB:
```python

In [33]: artifacts = [Artifact(randint(1, 9)) for i in range(1000)]

In [34]: %cpaste
Pasting code; enter '--' alone on the line to stop or use Ctrl-D.
:def f(parents):
:    studies = set()
:    analyses = set()
:    for p in parents:
:        s = p.study
:        a = p.analysis
:        if s is not None:
:            studies.add(s.id)
:        if a is not None:
:            analyses.add(a.id)
:    return studies, analyses
:
:def g(parents):
:    studies = {p.study for p in parents}
:    analyses = {p.analysis for p in parents}
:    studies.discard(None)
:    analyses.discard(None)
:    studies = {s.id for s in studies}
:    analyses = {a.id for a in analyses}
:    return studies, analyses
:--

In [35]: %timeit f(artifacts)
1 loops, best of 3: 1.25 s per loop

In [36]: %timeit g(artifacts)
1 loops, best of 3: 1.26 s per loop

```

Anyways I changed to the "g" version since it is more compact

qiita_db/handlers/analysis.py
```diff
@@ -0,0 +1,69 @@
...
+            The contents of the analysis keyed by sample id
+        """
+        with qdb.sql_connection.TRN:
+            a = _get_analysis(analysis_id)
+            mf_fp = a.mapping_file
+            response = None
+            if mf_fp is not None:
+                df = qdb.metadata_template.util.load_template_to_dataframe(
+                    mf_fp, index='#SampleID')
+                response = df.to_dict(orient='index')
```
josenavas: tornado magic! When you do `self.write(response)` as long as `response` is a dict it gets JSON-ified 😄

ElDeveloper:
Thanks @josenavas!

PR #2045, opened 2017-01-16
Remove old job code

josenavas:
Depends on #2044 so review/merge that one first.

Removes the old job.py code and its tests.

PR #2046, opened 2017-01-16
Fixes Qiita db

josenavas:
Depends on #2045 so review/merge that one first

Fixes most of the tests in qiita db


josenavas:
Ready for review - This is fixing the rest of qiita_db to be compliant with the new analysis.

Note: The QiitaStatusObject is no longer used in the system so I removed it here - more code cleaning 😄

qiita_db/test/test_meta_util.py
```diff
@@ -46,13 +46,13 @@ def test_get_accessible_filepath_ids(self):
         obs = qdb.meta_util.get_accessible_filepath_ids(
             qdb.user.User('shared@foo.bar'))
         self.assertItemsEqual(obs, {
-            1, 2, 3, 4, 5, 9, 12, 13, 14, 15, 16, 17, 18, 19, 20, 21})
+            1, 2, 3, 4, 5, 9, 12, 16, 17, 18, 19, 20, 21})
```
antgonza: Why are these being deleted? Are the analysis permissions be different from now on?

josenavas: Those filepaths ids no longer exists. The jobs that where inserted in the DB didn't had any kind of command information, so the patch removes them.

antgonza: Just to confirm, what you are saying is that those filepaths_ids where pushed but are not actually valid, right? If that's the case, are you creating new ones to test that the validate works or what's the plan? 

My concern is that: we don't add some kind of permission checking for the analysis.

josenavas: The permission checking on the analysis is done through the artifacts so that has been already tested. These files no longer exist, and 1 new file has been added (and it is seen in a test below)

antgonza: Thanks for confirming! As the changes are mainly deletions I think 1 review is enough. Merging.

PR #2047, opened 2017-01-16
fix #1987

### LINKED ISSUE #1987: should new artifacts created from public ones be public?
Currently, if you generate a new artifact from a public artifact, the new one is private. IMOO it should be public -- the artifact is being generated from a public one using a plugin available systemwide.

antgonza:


qiita_db/support_files/patches/python_patches/47.py
```diff
@@ -0,0 +1,30 @@
+from qiita_db.study import Study
+
+
+class ForRecursion(object):
```
josenavas: I think you are facing the same issue that I was facing when doing the analysis refactor. Try changing this:

https://github.com/biocore/qiita/pull/2040/files#diff-694f6ed1529987bc3435746a74426405R408

then the functions should work

antgonza:
Perhaps but I think the current method works fine, right?

josenavas:
I was looking at the code and I'm not sure if we really want to go and modify the visibility of the current artifacts. I think it is ok for moving forward inherit the lowest visibility of the parents, but I don't think I agree with the python patch.

Basically, it is possible that the raw data is private, but the final BIOM is public, and it has been used in an analysis. According the current code, the BIOM would become private, and people that had access before to that BIOM and did analysis with it, no longer have access. I would not modify the current status of the artifacts in the DB.

antgonza:
Good point but I don't think that case is actually possible, right? See: https://github.com/biocore/qiita/blob/master/qiita_db/artifact.py#L621. I could change it to check if any children in the "family" is public/private and if it is make them all follow that visibility.

qiita_db/support_files/patches/python_patches/47.py
```diff
@@ -0,0 +1,30 @@
...
+    work"""
+
+    @classmethod
+    def change_status(cls, artifact, status):
+        for a in artifact.children:
+            try:
+                a.visibility = status
+            except:
+                # print so we know which changes failed and we can deal by hand
+                print "failed aid: %d, status %s" % (artifact.id, status)
```
tanaes: Failed "aid?" Not sure I would understand what that was saying if I got this message.

PR #2052, opened 2017-01-20
fix get_studies

antgonza:


antgonza:
Assuming the test pass, which they should (?). This fixes why some studies are not being shown. 

- Old version
screen shot 2017-01-19 at 6 23 44 pm

- New version in test environment. Note that this is faster as the UNION is only returning unique values
screen shot 2017-01-19 at 6 23 51 pm

ackermag:
That looks a lot better... Is there any way you can check that there are no samples from North Korea (we have had problems before where N America samples are assigned to N Korea due to incorrect co-ordinates (+/-))?

antgonza:
None. 
screen shot 2017-01-19 at 7 03 44 pm

antgonza:
BTW we can wait for the key to die so the system updates or we can delete any of the stats keys:
- To find: `KEYS *:stats:*`
- To delete: `DEL *:stats:*`

PR #2056, opened 2017-01-21
fix #1913

### LINKED ISSUE #1913: field with new line breaks preprocessing
if we have a column with a new line in this format:

```
"This is my line.
"
```

preprocessing (split libraries) fails.

antgonza:
- Fixed #1913 + added test
- Reduced the loops over the headers and full files

### DIFF
--- a/qiita_db/metadata_template/util.py
+++ b/qiita_db/metadata_template/util.py
@@ -60,16 +60,13 @@ def prefix_sample_names_with_id(md_template, study_id):
         md_template.index.name = None
 
 
-def load_template_to_dataframe(fn, strip_whitespace=True, index='sample_name'):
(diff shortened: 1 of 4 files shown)


qiita_db/metadata_template/util.py
```diff
@@ -132,12 +116,30 @@ def load_template_to_dataframe(fn, strip_whitespace=True, index='sample_name'):
         holdfile.insert(0, "%s\n" % '\t'.join(headers))
         # The QIIME parser fixes the index and removes the #
         index = 'SampleID'
+    else:
```
josenavas: Any reason why these checks are done only if the file is not a mapping file? Note that users can upload a QIIME mapping file instead of sample + prep information, so I think it is better if these checks are always executed (as they're before).

PR #2057, opened 2017-01-24
Automatic jobs & new stats

antgonza:
This adds automatic jobs (which should be added to a cron job by the sys admin) and new stats.

### DIFF
--- a/qiita_db/meta_util.py
+++ b/qiita_db/meta_util.py
@@ -25,6 +25,16 @@ Methods
 # -----------------------------------------------------------------------------
 from __future__ import division
 
+from moi import r_client
+from os import stat
+from time import strftime, localtime
+import matplotlib.pyplot as plt
+import matplotlib as mpl
+from base64 import b64encode
+from urllib import quote
+from StringIO import StringIO
+from future.utils import viewitems
+
 from qiita_core.qiita_settings import qiita_config
 import qiita_db as qdb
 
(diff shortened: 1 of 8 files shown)


antgonza:
Tested in the test-env and seems to be working fine, runs < 2min. Once, we agree on what other stats we want to add (via the internal email) I will add the code so stats run from the new redis variables so nothing is calculated on the fly. 

Also, this is how the storage plot looks like:
screen shot 2017-01-24 at 6 43 59 am

antgonza:
This is ready for review, full code deployed in the test env.

qiita_db/meta_util.py
```diff
@@ -122,6 +133,150 @@ def get_accessible_filepath_ids(user):
...
+
+    plot = StringIO()
+    plt.savefig(plot, format='png')
+    plot.seek(0)
+    img = 'data:image/png;base64,' + quote(b64encode(plot.buf))
+
+    time = datetime.now().strftime('%m-%d-%y %H:%M:%S')
+
+    portal = qiita_config.portal
+    keys = [
```
josenavas: Note that in this for loop, each key enters in a single value of the if, which means that a for loop is not needed. I think a better way of doing this is:

```python
keys = ['number_studies', 'number_of_samples', 'num_users', 'lat_longs', 'num_studies_ebi', 'num_samples_ebi', 'img', 'time']
values = [number_studies, number_of_samples, num_users, lat_longs, num_studies_ebi, num_samples_ebi, img, time]
funcs = [r_client.hmset, r_client.hmset, r_client.set, r_client.set, r_client.set, r_client.set, r_client.set, r_client.set]

for k, v, f in zip(keys, values, funcs):
    redis_key = '%s:stats:%s' % (portal, k)
    f(redis_key, v)
```

qiita_db/test/test_meta_util.py
```diff
@@ -180,6 +181,59 @@ def test_get_lat_longs_EMP_portal(self):
...
+    def test_update_redis_stats(self):
+        qdb.meta_util.update_redis_stats()
+
+        # checking values from redis
+        portal = qiita_config.portal
+        keys = [
+            'number_studies', 'number_of_samples', 'num_users', 'lat_longs',
+            'num_studies_ebi', 'num_samples_ebi', 'img']
+
+        for k in keys:
```
josenavas: Same comment as above re doing the for loop.

qiita_db/util.py
```diff
@@ -714,7 +714,17 @@ def path_builder(db_dir, filepath, mountpoint, subdirectory, obj_id):
...
+    if exists(fp):
+        if isdir(fp):
+            func = rmtree
+        else:
+            func = remove
+        TRN.add_post_commit_func(func, fp)
+
+
+def purge_filepaths(delete_files=True):
     r"""Goes over the filepath table and remove all the filepaths that are not
```
josenavas: Can you update the docstring with the new parameter?

qiita_db/util.py
```diff
@@ -714,7 +714,17 @@ def path_builder(db_dir, filepath, mountpoint, subdirectory, obj_id):
                 for fpid, fp, fp_type_, m, s in results]
 
 
-def purge_filepaths():
+def _rm_files(TRN, fp):
```
josenavas: I don't think you need to pass TRN as a parameter, given that it is a global variable in the system, you can just use it.

qiita_pet/templates/stats.html
```diff
@@ -49,26 +49,40 @@
...
+      <tr>
+        <th>Studies</th>
+        <th>Samples</th>
+        <th>Users</th>
+      </tr>
+    </thead>
+      <tr>
+        <td>
+          {% for k in number_studies %}
+            <i>{{k}}</i>: {{number_studies[k]}}
```
josenavas: Can you put each of this values in one row? I think it looks better than one next to each other.

scripts/qiita-cron-job
```diff
@@ -0,0 +1,49 @@
...
+#    qiita.filepath and are present in the filesystem
+# 2. empty_trash_upload_folder: remove files that are present in the trash
+#    of the upload folders
+# 3. update_redis_stats: updates the redis stats information
+#
+# Note that is responsability of the Qiita install system admin to add to a
+# cron job this script and responsible to define how often it should run
+
+
+def call_purge_filepaths():
```
josenavas: Why do you have this intermediate functions? You can call them directly under main.

qiita_db/meta_util.py
```diff
@@ -122,6 +133,150 @@ def get_accessible_filepath_ids(user):
...
+    ebi_samples = {}
+    for k, sts in viewitems(studies):
+        number_of_samples[k] = 0
+        for s in sts:
+            st = s.sample_template
+            if st is not None:
+                number_of_samples[k] += len(list(st.keys()))
+
+            ebi_samples_count = 0
+            for pt in s.prep_templates():
```
antgonza: Excellent point, will add both.

qiita_db/util.py
antgonza: Turns out that you do, if not:
```
ERROR: test_purge_filepaths_null_cols (qiita_core.util.DecoratedClass)
  File "qiita_db/test/test_util.py", line 393, in test_purge_filepaths_null_cols
    self._common_purge_filpeaths_test()
  File "qiita_db/test/test_util.py", line 342, in _common_purge_filpeaths_test
    qdb.util.purge_filepaths()
  File "/Users/antoniog/svn_programs/qiita/qiita_db/util.py", line 768, in purge_filepaths
    _rm_files(fp)
  File "/Users/antoniog/svn_programs/qiita/qiita_db/util.py", line 724, in _rm_files
    TRN.add_post_commit_func(func, fp)
NameError: global name 'TRN' is not defined
```

josenavas: Oh, well, yeah, in that case you need to use it as:
`qdb.sql_connection.TRN` not just `TRN`.

josenavas: Not blocking though...

qiita_db/meta_util.py
```diff
@@ -122,6 +133,147 @@ def get_accessible_filepath_ids(user):
...
+    missing_files = []
+    for k, sts in viewitems(studies):
+        for s in sts:
+            for a in s.artifacts():
+                for _, fp, dt in a.filepaths:
+                    try:
+                        s = stat(fp)
+                        stats.append((dt, s.st_size, strftime('%Y-%m',
+                                      localtime(s.st_ctime))))
+                    except:
```
ElDeveloper: Can you specify the exception name here?

qiita_db/util.py
ElDeveloper: From looking at the file, I think this is because you need to import the TRN object, right @josenavas?

qiita_db/util.py
```diff
@@ -739,30 +754,58 @@ def purge_filepaths():
...
         sql = "DELETE FROM qiita.filepath WHERE filepath_id=%s"
         db_results = qdb.sql_connection.TRN.execute_fetchindex()
         for fp_id, fp, fp_type, dd_id in db_results:
-            qdb.sql_connection.TRN.add(sql, [fp_id])
+            if delete_files:
+                qdb.sql_connection.TRN.add(sql, [fp_id])
+                fp = join(get_mountpoint_path_by_id(dd_id), fp)
+                _rm_files(qdb.sql_connection.TRN, fp)
+            else:
+                print fp, fp_type
```
ElDeveloper: Is this print statement needed? Doesn't seem to be described in the docs.

qiita_db/util.py
```diff
@@ -739,30 +754,58 @@ def purge_filepaths():
...
+
+        for mp in qdb.sql_connection.TRN.execute_fetchflatten():
+            for path, dirs, files in walk(gfp(mp)):
+                if path.endswith('/trash'):
+                    if delete_files:
+                        for f in files:
+                            fp = join(path, f)
+                            _rm_files(qdb.sql_connection.TRN, fp)
+                    else:
+                        print files
```
ElDeveloper: Same as the other print.

scripts/qiita-cron-job
```diff
@@ -0,0 +1,33 @@
...
+# Distributed under the terms of the BSD 3-clause License.
+#
+# The full license is in the file LICENSE, distributed with this software.
+# -----------------------------------------------------------------------------
+
+from qiita_db.util import purge_filepaths, empty_trash_upload_folder
+from qiita_db.meta_util import update_redis_stats
+
+
+# This script will perform these jobs:
```
ElDeveloper: ok

qiita_db/meta_util.py
```diff
@@ -122,6 +133,147 @@ def get_accessible_filepath_ids(user):
...
+    plt.xticks(all_dates_axis, all_dates)
+    plt.legend()
+    plt.grid()
+    ax = plt.gca()
+    ax.yaxis.set_major_formatter(mpl.ticker.FuncFormatter(sizeof_fmt))
+    plt.xlabel('Date')
+    plt.ylabel('Storage space per data type')
+
+    plot = StringIO()
+    plt.savefig(plot, format='png')
```
ElDeveloper: Sounds good then.

qiita_db/util.py
ElDeveloper: Sounds good.

PR #2059, opened 2017-01-26
fix #1010

### LINKED ISSUE #1010: Refactor get_accessible_filepath_ids

antgonza:
To solve this issue we can either do a complex SQL command or simply verify per filepath_id and return true/false. For simplicity and to keep a similar-to-previous/detailed-structure decided to go with the latter option.


qiita_db/meta_util.py
```diff
@@ -48,7 +47,7 @@ def _get_data_fpids(constructor, object_id):
         return {fpid for fpid, _, _ in obj.get_filepaths()}
 
 
-def get_accessible_filepath_ids(user):
+def validate_filepath_access_by_user(user, filepath_id):
     """Gets all filepaths that this user should have access to
```
josenavas: Can you update the one sentence description of the documentation?

qiita_db/meta_util.py
```diff
@@ -48,78 +47,117 @@ def _get_data_fpids(constructor, object_id):
...
+            (SELECT count(*) FROM qiita.analysis_filepath
+             WHERE filepath_id = {0}) AS analysis""".format(filepath_id)
+        TRN.add(sql)
+
+        arid, sid, pid, jid, anid = TRN.execute_fetchflatten()
+
+        # artifacts
+        if arid:
+            # check the public artifacts
+            public_artifacts = qdb.artifact.Artifact.iter_public()
```
josenavas: I think I have not explained myself correctly on the previous comments. As it is currently written here, you're still iterating over all the artifacts. However, at this point you can now which artifacts this have been seen:
```python
sql = """SELECT artifact_id FROM qiita.artifact_filepath WHERE filepath_id = %s"""
```

Similarly you can do for the other entries. You may be able to modify the above call to return the values in an array so you already have them.

qiita_db/support_files/populate_test_db.sql
```diff
@@ -11,7 +11,7 @@ INSERT INTO qiita.qiita_user (email, user_level_id, password, name,
     '$2a$12$gnUi8Qg.0tvW243v889BhOBhWLIHyIJjjgaG6dxuRJkUM8nXG9Efe', 'Dude',
     'Nowhere University', '123 fake st, Apt 0, Faketown, CO 80302',
     '111-222-3344'),
-    ('shared@foo.bar', 3,
+    ('shared@foo.bar', 4,
```
antgonza: level 3 is superuser and 4 is use, shared user should be regular user so we can actually test that sharing/etc works ... let's see which other tests will fail ...

antgonza:
Ready for another review.

qiita_db/meta_util.py
```diff
@@ -59,78 +58,97 @@ def _get_data_fpids(constructor, object_id):
...
+        elif pid:
+            # the prep access is given by it's artifacts, if the user has
+            # access to any artifact, it should have access to the prep
+            # [0] cause we should only have 1
+            a = qdb.metadata_template.prep_template.PrepTemplate(
+                pid[0]).artifact
+            if (a.visibility == 'public' or a.study.has_access(user)):
+                return True
+            else:
+                for c in a.children:
```
josenavas:
```python
for c in a.descendants.nodes():
```
`children` is only checking at the immediate children but not to the entire descendants.

qiita_db/support_files/patches/50.sql
```diff
@@ -0,0 +1,15 @@
+-- Feb 1, 2017
```
josenavas: Please remove this file as I don't think it belongs to this PR.

wasade:
nothing stood out to me, :+1:

PR #2061, opened 2017-01-27
fix #1837

### LINKED ISSUE #1837: better alternative to 'click and drag your uploaded files to the correct file type'
I feel this may be covered in another issue, so feel free to point me in the right direction!

When setting up the Prep Info with paired reads, is there a good way to get the reads in the right column? My forward reads are `*.f.fastq.gz` and my reverse reads are `*.r.fastq.gz` files. 

screen shot 2016-05-11 at 2 27 45 pm

Thank you for your time!

antgonza:
Turns out that qiita_pet/templates/study_ajax/add_prep_artifact.html was left behind by mistake so removing it. Also, during testing I realized that you could send an unrelated study_id/prep_info_id and the system will not complain, added a raise and a test. However, this meant that a few of the tests not didn't make sense and to be honest it seems like they weren't testing anything important/new; let me know if I read this incorrectly. Finally, simplified the code a bit.


antgonza:
@tanaes, this is the functionality you were requesting, could you take a look?

antgonza:
While working on other stuff I realized that this functionality will be nice on all cases we have a run prefix so the files are auto-selected so decided to make it work that way:
my-first-video

josenavas:
I am concerned of the approach taken here. We tried to do this at the beginning, but the problem is that you don't know which one is the barcode files and which one is the forward reads. For example, in the GIF that you have in the comments, it shows the R1 in barcodes and the R2 in forward reads, which I don't think is the correct assignment...

antgonza:
After offline discussion we agreed that this is the best we can come up with and that for all the _real_ cases we know this will work. However, we should expand the Note to say that the user should verify the file distribution.

antgonza:
Message looks like this now:
screen shot 2017-02-01 at 2 25 09 pm

josenavas:
@tanaes can you review this PR when you have a second?

qiita_pet/handlers/api_proxy/studies.py
```diff
@@ -234,31 +235,42 @@ def study_files_get_req(user_id, study_id, prep_template_id, artifact_type):
...
+        prep_prefixes = sorted(prep_prefixes, key=len, reverse=True)
+        # group files by prefix
+        sfiles = {p: [f for _, f in uploaded if f.startswith(p)]
+                  for p in prep_prefixes}
+        inuse = [y for x in sfiles.values() for y in x]
+        remaining.extend([f for _, f in uploaded if f not in inuse])
+
+        for k, v in viewitems(sfiles):
+            len_files = len(v)
+            if len_files != 1 and len_files != 2:
```
wasade: would it be possible to have a comment here about this?

qiita_pet/templates/study_ajax/artifact_file_selector.html
```diff
@@ -195,7 +195,13 @@
 </div>
 <div id="drag-files-div" class="row">
   <div class="col-md-12">
-    <b>Or click and drag your uploaded files to the correct file type:</b><br/>
+    <b>or click and drag your uploaded files to the correct file type</b><br/>
+      <div class="blinking-message">
+        Please make sure that the correct files are in the correct column.<br/>
+        Note: the system will try to auto select the files based on run_prefix, if that doesn't work, either the type you selected doesn't support
+        it or the run_prefix is wrong
```
wasade: what is "it" here?

qiita_pet/handlers/api_proxy/studies.py
antgonza: Good catch, added an explanation, improved the code and added a test.

qiita_pet/templates/study_ajax/artifact_file_selector.html
antgonza: well, it's it ... fixing.

PR #2062, opened 2017-01-29
WIP: Fix 1805

antgonza:
Based on https://github.com/biocore/qiita/pull/2059, so merge that one first.
The idea is, as defined in the issue, to download all file paths from the BIOM artifacts that the user has access to via nginx zip.

ToDo:
- [x] figure out how to test failures, place holders in place
- [ ] test/tweak in the test-env until it actually works as expected 

The new button is show here:
screen shot 2017-01-28 at 6 32 49 pm


josenavas:
Closing in favor of https://github.com/biocore/qiita/pull/2087

PR #2063, opened 2017-01-29
fix #1816

### LINKED ISSUE #1816: raw data names are not complete in new GUI
if the raw data names are too long the names are cut out which makes selection difficult, we should display the full name on selection or hover.

antgonza:
Now it looks like this:
my-first-video


josenavas:
How does it looks with very long file names?

antgonza:
Same but longer? I think those files have the standard file size ...

PR #2065, opened 2017-01-30
Analysis refactor gui (Part 1)

josenavas:
This depends on #2070, so review/merge that one first.

Initial pass of the GUI updates.

There is still some parts that need to be updated but I want to keep the PRs small.

Once this is ready for review, I will comment pointing out the most important things to focus reviews on.

PR #2068, opened 2017-02-01
Analysis refactor gui Part 2

josenavas:
This depends on https://github.com/biocore/qiita/pull/2065 so review/merge that one first.

This connects the creation of the analysis with the new analysis description page and makes that page reactive in the sense that it is not showing the processing graph workflow until the initial jobs have been generated, without needing to reload the page. I'm using vue.js to accomplish this.

An example:
creating_analysis


josenavas:
Closing in favor of #2077

PR #2070, opened 2017-02-03
Analysis refactor master

josenavas:
Previous PR was done against the wrong branch.

This PR syncs the analysis-refactor branch with the current status of master. I will comment in the code about the specific things that I would like people to take a look, since most of the things here are actually changes that have been already merged in master.


qiita_db/artifact.py
```diff
@@ -420,6 +421,15 @@ def _associate_with_analysis(instance, analysis_id):
                 sql_args = [(instance.id, p.id) for p in parents]
                 qdb.sql_connection.TRN.add(sql, sql_args, many=True)
 
+                # inheriting visibility
```
josenavas: @antgonza can you review this section? This was added in master and I had a bit of trouble solving the conflicts in here but I think I end up placing this code in the correct location.

antgonza: 👍

josenavas:
This is not in sync with the latest master!

PR #2071, opened 2017-02-03
database changes to fix 969

antgonza:
These are the DB changes so we can add the tagging system to Qiita. Note that I decided to keep track of which user adds which tag so we can allow users to add tags and still catch any abuse.

### DIFF
--- a/qiita_db/study.py
+++ b/qiita_db/study.py
@@ -986,7 +986,6 @@ class Study(qdb.base.QiitaObject):
         with qdb.sql_connection.TRN:
             sql = """DELETE FROM qiita.per_study_tags WHERE study_id = %s"""
             qdb.sql_connection.TRN.add(sql, [self._id])
-            qdb.sql_connection.TRN.execute()
 
             if tag_ids:
                 sql = """INSERT INTO qiita.per_study_tags
@@ -1000,7 +999,8 @@ class Study(qdb.base.QiitaObject):
                                 )"""
                 sql_args = [[tid, self._id, tid, self._id] for tid in tag_ids]
                 qdb.sql_connection.TRN.add(sql, sql_args, many=True)
-            return qdb.sql_connection.TRN.execute()
+
+            qdb.sql_connection.TRN.execute()
 
 # --- methods ---
     def artifacts(self, dtype=None, artifact_type=None):
--- a/qiita_db/study.py
+++ b/qiita_db/study.py
@@ -464,6 +464,42 @@ class Study(qdb.base.QiitaObject):
 
             qdb.sql_connection.TRN.execute()
 
+    @classmethod
+    def get_tags(cls):
+        """Returns the available study tags
+
+        Returns
+        -------
+        list of DictCursor
+            Table-like structure of metadata, one tag per row. Can be
+            accessed as a list of dictionaries, keyed on column name.
+        """
+        with qdb.sql_connection.TRN:
+            sql = """SELECT study_tag_id, study_tag
+                     FROM qiita.study_tags"""
+
+            qdb.sql_connection.TRN.add(sql)
+            return qdb.sql_connection.TRN.execute_fetchindex()
+
+    @classmethod
+    def insert_tags(cls, user, tags):
+        """Insert available study tags
+
+        Parameters
+        ----------
+        user : qiita_db.user.User
+            The user adding the tags
+        tags : list of str
+            The list of tags to add
+        """
+        with qdb.sql_connection.TRN:
+            sql = """INSERT INTO qiita.study_tags (email, study_tag)
+                     VALUES (%s, %s)"""
+            sql_args = [[user.email, tag] for tag in tags]
+
+            qdb.sql_connection.TRN.add(sql, sql_args, many=True)
+            qdb.sql_connection.TRN.execute()
+
 
 # --- Attributes ---
     @property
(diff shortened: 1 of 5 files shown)


qiita_db/study.py
```diff
@@ -921,7 +957,46 @@ def ebi_submission_status(self, value):
...
+            qdb.sql_connection.TRN.add(sql)
+            return qdb.sql_connection.TRN.execute_fetchindex()
+
+    @tags.setter
+    def tags(self, tag_ids):
+        """Sets the tags of the study
+
+        Parameters
+        ----------
+        tag_ids : list of int
```
josenavas: Can we provide the list of tags rather than the list of tag ids? We have been removing the exposure of DB ids for these types of thins as much as we could, so I would like not to re-add some of them back

qiita_db/study.py
```diff
@@ -921,7 +957,46 @@ def ebi_submission_status(self, value):
...
+        with qdb.sql_connection.TRN:
+            sql = """SELECT study_tag_id, study_tag
+                        FROM qiita.study_tags
+                        LEFT JOIN qiita.per_study_tags USING (study_tag_id)
+                        WHERE study_id = {0}""".format(self._id)
+            qdb.sql_connection.TRN.add(sql)
+            return qdb.sql_connection.TRN.execute_fetchindex()
+
+    @tags.setter
+    def tags(self, tag_ids):
```
josenavas: This is not setting the tags correctly. If a tag has been removed from the list it doesn't get removed from the DB.

Would it be more useful to add a couple of functions `add_tags` and `remove_tags`?

qiita_db/study.py
```diff
@@ -921,7 +957,46 @@ def ebi_submission_status(self, value):
...
+    def tags(self):
+        """Returns the tags of the study
+
+        Returns
+        -------
+        list of str
+            The study tags
+        """
+        with qdb.sql_connection.TRN:
+            sql = """SELECT study_tag_id, study_tag
```
josenavas: Just return the "study_tag" (not the id)

qiita_db/test/test_study.py
```diff
@@ -832,6 +832,23 @@ def test_environmental_packages_sandboxed(self):
...
+        qdb.study.Study.insert_tags(user, tags)
+
+        # testing that insertion went fine
+        obs = qdb.study.Study.get_tags()
+        exp = [[i + 1, tag] for i, tag in enumerate(tags)]
+        self.assertEqual(obs, exp)
+
+        # assigning the tags to study
+        study = qdb.study.Study(1)
+        study.tags = [tig for tig, tag in obs]
```
josenavas: can you add a:
```python
self.asertEqual(study.tags, [])
```

before this line

qiita_db/test/test_study.py
```diff
@@ -832,6 +832,23 @@ def test_environmental_packages_sandboxed(self):
...
+        # and checking that everything went fine
+        self.assertEqual(exp, study.tags)
```
josenavas: Add another case in which a tag is removed from the list.

qiita_db/study.py
ElDeveloper: 👍

qiita_db/study.py
```diff
@@ -921,7 +957,52 @@ def ebi_submission_status(self, value):
...
+        tag_ids : list of int
+            The tag ids of the study
+        """
+        with qdb.sql_connection.TRN:
+            sql = """DELETE FROM qiita.per_study_tags WHERE study_id = %s"""
+            qdb.sql_connection.TRN.add(sql, [self._id])
+            qdb.sql_connection.TRN.execute()
+
+            if tag_ids:
+                sql = """INSERT INTO qiita.per_study_tags
```
ElDeveloper: If something goes wrong when this query is executed, we will have lost the tags 😱. IIRC the API should allow you to only execute at the very end of this function, such that if something fails the whole transaction is rolled back.

qiita_db/study.py
```diff
@@ -921,7 +957,52 @@ def ebi_submission_status(self, value):
...
+                         SELECT %s, %s
+                            WHERE
+                                NOT EXISTS (
+                                    SELECT study_tag_id, study_id
+                                    FROM qiita.per_study_tags
+                                    WHERE study_tag_id = %s AND study_id = %s
+                                )"""
+                sql_args = [[tid, self._id, tid, self._id] for tid in tag_ids]
+                qdb.sql_connection.TRN.add(sql, sql_args, many=True)
+            return qdb.sql_connection.TRN.execute()
```
ElDeveloper: What's being returned here? ... should anything be returned at all?

qiita_db/support_files/patches/50.sql
```diff
@@ -0,0 +1,19 @@
+-- Feb 3, 2017
+-- adding study tagging system
+
+CREATE TABLE qiita.study_tags (
+  study_tag_id bigserial NOT NULL,
+  email varchar NOT NULL,
+  study_tag varchar NOT NULL,
```
ElDeveloper: Can/should we guarantee unique tags?

qiita_db/study.py
antgonza: Just FYI: We (@josenavas and me) decided via offline chat that it was fine to leave as is because that way we can link the ids in the templates and the text can be escaped in the templates, which makes our live easier.

qiita_db/study.py
antgonza: good point, deleting this execute

qiita_db/study.py
antgonza: good point, removing the return

qiita_db/support_files/patches/50.sql
antgonza: yes, 1 line below: `CONSTRAINT pk_study_tag UNIQUE ( study_tag_id ),`

PR #2072, opened 2017-02-06
fix generate_biom_and_metadata_release

antgonza:
found a small bug during today's deploy


ElDeveloper:
Looks good, but worth adding a test?

antgonza:
What kind of test? The problem was that the code was using bdir_len vs os.path internal functions so if the system configuration was `path` vs. `path/` the files will not be found.

ElDeveloper:
A regression test would catch that right? Changing the config.

ElDeveloper:
👍 thanks @antgonza!!

PR #2075, opened 2017-02-09
Fix timestamp / -> -

antgonza:
@wasade and @josenavas this is what we discussed during the last meeting about the new timestamp formats.


josenavas:
@antgonza unfortunately changing `/` to `-` are not enough since the actual proposed change also involved changing the order of the data: `%m/%d/%Y %H:%M:%S` -> `%Y-%m-%d %H:%M:%S`

antgonza:
This is ready for review: @josenavas, @wasade, @ElDeveloper 

A lot of the changes were to update the tests to match the new date formats. Remember that now we are actually patching the test DB/files so this change also implies that I needed to change how we are testing some functionality. However, I think there are not big changes.

qiita_db/support_files/patches/python_patches/51.py
```diff
@@ -0,0 +1,107 @@
...
+
+    # loop over the old formats to see which one is it
+    date = None
+    for i, fmt in enumerate(old_formats):
+        try:
+            date = datetime.strptime(value, fmt)
+            break
+        except ValueError:
+            pass
+    if date is not None:
```
wasade: this block is rather confusing...

qiita_db/test/test_util.py
```diff
@@ -750,33 +750,55 @@ def test_generate_biom_and_metadata_release(self):
...
+        fn = ('processed_data/1_study_1001_closed_reference_otu_table_Silva.'
+              'biom')
+        self.assertTrue(fn in tgz_obs)
+        tgz_obs.remove(fn)
+        # now let's check prep info files based on their suffix, just take
+        # the first one and check/rm the occurances of that file
+        fn_prep = [f for f in tgz_obs
+                   if f.startswith('templates/1_prep_1_')][0]
+        # 3 times
+        self.assertTrue(fn_prep in tgz_obs)
```
wasade: Why would the item continue to exist after being removed?

qiita_pet/support_files/doc/source/tutorials/prepare-information-files.rst
```diff
@@ -72,14 +72,14 @@ These are the columns required for successfully submit your data to EBI:
...
-|                                  | or ``mm/dd/yyyy hh:mm`` |                                                                                                                                                     |
-|                                  | or ``mm/dd/yyyy hh``    |                                                                                                                                                     |
-|                                  | or ``mm/dd/yyyy``       |                                                                                                                                                     |
-|                                  | or ``mm/yyyy``          |                                                                                                                                                     |
+| ``collection_timestamp``         | ``yyyy-mm-dd hh:mm:ss`` | The time stamp (preferred) of when the sample was collected. Several format are accepted, all ISO 8601.                                             |
```
wasade: timezone or are these UTC?

qiita_db/support_files/patches/python_patches/51.py
antgonza: Any suggestions? Basically, is testing which is the fmt of the date within the old formats, if it's found it using the new format. Now, new format doesn't accept 2 digits so I'm simply reassigning to one with 4 digits if one of the 2 year digits is found ...

qiita_db/test/test_util.py
antgonza: because we created the tests we added more artifacts and didn't assign new file paths so we reused the same ones

qiita_pet/support_files/doc/source/tutorials/prepare-information-files.rst
antgonza: User defined? We never got to discuss this in the meetings, do you, @josenavas have a preference?

qiita_db/support_files/patches/python_patches/51.py
wasade: Ah. What about a format mapping: `{old_fmt: new_format}`, and in the case of unacceptable, just specify `None` as the new format?

qiita_pet/support_files/doc/source/tutorials/prepare-information-files.rst
josenavas: No preference, whatever the standard says.

PR #2076, opened 2017-02-09
Analysis refactor GUI part1

josenavas:
The file `qiita_pet/handlers/analysis_handlers/listing_handlers.py` contains handlers that already existed before, I just moved them to create a better structure of the code base so it is easier to maintain. They may change a bit in a later PR.

The main changes are done in the file `qiita_pet/handlers/analysis_handlers/base_handlers.py`.

This gif shows how the new analysis description page looks like.

new_analysis_desc_page

The artifact page is not loading ATM because there are some changes that I need to do on that handler - however this PR is already big enough that I refer those changes for a future PR.

@antgonza this is ready for review.

PR #2077, opened 2017-02-09
Analysis refactor GUI part2

josenavas:
Depends on https://github.com/biocore/qiita/pull/2076 so review/merge that one first.

This PR modifies the study description page so it is responsive to the status of the analysis. When the analysis is created, the initial set of (BIOM) files may not exist since they're still being generated. The changes here take into account this fact and makes the interface responsive to the status of the jobs, so once the initial set of files for the analysis is generated, the page is updated to show the graph. Example:

analysis_creation


josenavas:
This is ready for review

PR #2078, opened 2017-02-15
Analysis refactor gui part3

josenavas:
Depends on #2077 so review merge that one first.

This PR partially fixes the artifact pages so they work both in the study pipeline and in the analysis pipeline. I still need to fix a couple of functions and handlers but I want to keep the PRs small.

I've moved the artifact handlers out from the study_handlers folder and created a new artifact_handlers folder, since now it doesn't make sense to have them under the study specific list.
I haven't deleted the old code (will do in another PR).

The functionalities changed here are: (1) being able to load the page, (2) modify the artifact name, and (3) being able to submit the artifact summary.

I've done some minor modifications to the artifact interface, which in my opinion improves it: (1) I moved the visibility information next to the name and ID and I put all the buttons next to each other in the same row.

Here is a GIF which shows the functionality and the GUI changes:

artifact_update

PR #2079, opened 2017-02-15
Analysis refactor gui part4

josenavas:
~~This depends on #2078 so review/merge that one first.~~

This PR fixes the "delete artifact" and "change visibility" functionality from the artifact page. Now this page can be used from the analysis and from the study pipeline.

There was also a bug on the artifact deletion that I fixed here.

Here is a screenshot on how the UI from the study part changed a bit so the buttons are better aligned.

screen shot 2017-02-15 at 1 48 58 am


josenavas:
@antgonza this is ready for review, tests are still expected to fail. Just to provide a bit more background on the tests failures, I stubbed a couple of functions in my machine so I can check how the interface would look like but they're not complete, hence not adding them to the repo until I complete them. I have a few other PR's on the interface side before I fix them.

PR #2087, opened 2017-03-22
Fix 1805

josenavas:
This grab's @antgonza's code and fixes it so the download through nginx works correctly.

@antgonza @ElDeveloper are you able to do a review?


qiita_pet/handlers/download.py
```diff
@@ -37,3 +41,77 @@ def get(self, filepath_id):
...
+        # loop over artifacts and retrieve those that we have access to
+        to_download = []
+        vfabu = validate_filepath_access_by_user
+        for a in study.artifacts():
+            if a.artifact_type == 'BIOM':
+                to_add = True
+                for i, (fid, path, data_type) in enumerate(a.filepaths):
+                    # validate access only of the first artifact filepath,
+                    # the rest have the same permissions
+                    if (i == 0 and not vfabu(user, fid)):
```
antgonza: Can you add a tests for this case?

qiita_pet/handlers/download.py
```diff
@@ -37,3 +41,77 @@ def get(self, filepath_id):
...
+                        to_add = False
+                        break
+                    if data_type == 'directory':
```
antgonza: ... and one for this?

qiita_pet/handlers/download.py
```diff
@@ -37,3 +41,77 @@ def get(self, filepath_id):
...
+                                fullpath = join(dp, fname)
+                                spath = fullpath
+                                if fullpath.startswith(basedir):
+                                    spath = fullpath[basedir_len:]
+                                to_download.append((fullpath, spath, spath))
+                    elif path.startswith(basedir):
+                        spath = path[basedir_len:]
+                        to_download.append((path, spath, spath))
+                    else:
+                        to_download.append((path, path, path))
```
antgonza: ... and this?

qiita_pet/handlers/download.py
```diff
@@ -37,3 +41,77 @@ def get(self, filepath_id):
...
+                        if qmf is not None:
+                            sqmf = qmf
+                            if qmf.startswith(basedir):
+                                sqmf = qmf[basedir_len:]
+                            to_download.append(
+                                (qmf, sqmf, 'mapping_files/%s_mapping_file.txt'
+                                            % a.id))
+
+        # If we don't have nginx, write a file that indicates this
+        all_files = '\n'.join(["- %s /protected/%s %s" % (getsize(fp), sfp, n)
```
antgonza: I liked having the message "This installation of Qiita was not equipped with nginx, ... ", is it possible to add it?

ElDeveloper:
👍

qiita_pet/handlers/download.py
josenavas: Done

qiita_pet/handlers/download.py
josenavas: Done

qiita_pet/handlers/download.py
josenavas: Not sure how to add a test, added a comment explaining why

qiita_pet/handlers/download.py
josenavas: According the the mod_zip documentation you cannot have any extra lines, so I don't think it is possible to add that information.

qiita_pet/handlers/download.py
```diff
@@ -37,3 +42,82 @@ def get(self, filepath_id):
...
+                            sqmf = qmf
+                            if qmf.startswith(basedir):
+                                sqmf = qmf[basedir_len:]
+                            to_download.append(
+                                (qmf, sqmf, 'mapping_files/%s_mapping_file.txt'
+                                            % a.id))
+
+        # If we don't have nginx, write a file that indicates this
+        all_files = '\n'.join(["%s %s /protected/%s %s"
+                               % (compute_checksum(fp), getsize(fp), sfp, n)
```
wasade: `compute_checksum` is going to be expensive as it requires reading each file in whole from the filesystem. This is multiplied by the potential number of files collected from the deeply nested if statements. Are you _sure_ this should be executed by tornado?

antgonza: Agree with @wasade. Note that this was a test and we are removing.

josenavas:
Is there anything else needed in this PR?

antgonza:
Just to keep everyone in the loop: We realized that in some occasions the zip downloaded will not open in Macs because due to the size of the files it requires unzip >= 6.00. Thus, we decided to host a Mac version of unzip 6.00 in the FTP (cause the only way to install it is via brew), and add a GUI line under the button pointing to new documentation explaining the issue and fix.

antgonza:
Closing as I'm gonna create another PR with the requested fixes as @josenavas requested it.

ElDeveloper:
Oh, I think I have seen this before, what about using gzip?

My past experience was that large zip files could not be opened unless I used `jar xvf files.zip` or something like that.

antgonza:
mod_gzip for nginx doesn't really work like mod_zip. In other words, we can't use mod_gzip for what we are doing here ...

PR #2088, opened 2017-03-22
Adding tags

antgonza:
Based on reviews during the Qiita meeting, I'm changed the way this feature works. You can see how it works in the attachment below. Another big change from original definition is that now we are using REST calls to get the tags and the updates. 

Note that this adds the possibility of adding tags to studies and keeping track of them by user. However, we are still not showing or filtering the studies in the study list by this. I'll make another PR adding that functionality as this PR is pretty long (but not terrible) already.

tags


antgonza:
@biocore/qiita-admin, wanna take a look/review?

qiita_db/study.py
```diff
@@ -1152,6 +1137,75 @@ def unshare(self, user):
             qdb.sql_connection.TRN.add(sql, [self._id, user.id])
             qdb.sql_connection.TRN.execute()
 
+    def add_tags(self, user, tags):
```
josenavas: Should this be named `update_tags`, given that it can add and remove tags?

qiita_db/study.py
```diff
@@ -1152,6 +1137,75 @@ def unshare(self, user):
...
+        """Sets the tags of the study
+
+        Parameters
+        ----------
+        user: User object
+            The user to unshare the study with
```
josenavas: Update doc

qiita_db/study.py
```diff
@@ -1152,6 +1137,75 @@ def unshare(self, user):
...
+        tags : list of str
+            The tags to add to the study
```
josenavas: The tags to attach to the study - or something along those lines (since it is not only adding)

qiita_db/study.py
```diff
@@ -1152,6 +1137,75 @@ def unshare(self, user):
...
+                        sql = """DELETE FROM qiita.per_study_tags
+                                     WHERE study_id = %s AND study_tag IN %s"""
+                        qdb.sql_connection.TRN.add(
+                            sql, [self._id, tuple(to_delete)])
+
+                if to_add:
+                    if user_level != 'admin':
+                        admin_tags = to_add & system_tags_admin
+                        if admin_tags:
+                            message += ('Only admins can assing: '
```
josenavas: `assing` -> `assign`

wasade:
Seems reasonable

qiita_db/study.py
antgonza: k

qiita_db/study.py
antgonza: k

qiita_db/study.py
antgonza: k

qiita_db/study.py
antgonza: k

qiita_db/study.py
```diff
@@ -1152,6 +1137,75 @@ def unshare(self, user):
             qdb.sql_connection.TRN.add(sql, [self._id, user.id])
             qdb.sql_connection.TRN.execute()
 
+    def updata_tags(self, user, tags):
```
josenavas: updata -> update

qiita_pet/handlers/study_handlers/tests/test_base.py
```diff
@@ -59,5 +59,37 @@ def test_get(self):
...
+
+
+class TestStudyTags(OauthTestingBase):
+    def test_get(self):
+        response = self.get('/study/tags/1')
+        exp = ('{"status": "success", "message": "", "tags": []}')
+        self.assertEqual(response.code, 200)
+        self.assertEqual(response.body, exp)
+
+    def test_patch(self):
```
wasade: same question as in #2092, should there be negative tests?

antgonza: I don't think is required as all possible options are tested here: https://github.com/biocore/qiita/pull/2088/files#diff-a0b367a8e2e2e1d72fee16533d797415R541, what do you think?

wasade: I think this is logically distinct. The goal of a negative test here, and with `test_get`, is to verify the status codes and response messages are rational for malformed or inappropriate REST requests

PR #2092, opened 2017-03-26
Listing tags

antgonza:
ToDo:
- [x] make tests pass
- [x] test in qiita-test env


antgonza:
This should be ready for review after https://github.com/biocore/qiita/pull/2088 is merged. This code has been deployed in the test env and seems to be working fine. Note that I had to update some js and css libraries

Example of functionality, see how the "regular" search and the tags interact nicely:
tags

qiita_pet/handlers/study_handlers/tests/test_base.py
```diff
@@ -59,5 +59,37 @@ def test_get(self):
...
+class TestStudyGetTags(TestHandlerBase):
+    def test_get(self):
+        response = self.get('/study/get_tags/')
+        exp = ('{"status": "success", "message": "", "tags": '
+               '{"admin": [], "user": []}}')
+        self.assertEqual(response.code, 200)
+        self.assertEqual(response.body, exp)
+
+
+class TestStudyTags(OauthTestingBase):
```
wasade: should there be negative test cases?

antgonza:
Thanks @wasade note that this is based on: https://github.com/biocore/qiita/pull/2088 ....

wasade:
Should I not review this then?

josenavas:
@antgonza can you pull from upstream/master?

antgonza:
pull from upstream/master ready. Note that currently we are using tag-it also to display it and it looks like this:
screen shot 2017-03-27 at 4 55 55 pm
but I could simplify things so it looks like this:
screen shot 2017-03-27 at 4 56 00 pm
I really don't have a preference but wanted to ask ...

antgonza:
@wasade, @josenavas, @ElDeveloper, could you review?

josenavas:
Thanks @antgonza code looks good!

Answering your last question - I think either looks fine, I don't have a strong preference. What I don't really like is the position of the filter. i.e. I don't like it showing inside the table. Is it easy to have on top of the table (i.e. above the headers?). The reason why I don't like it there is because it looks like only applies to the top table, but it should be applied to both tables (private and public studies).

Do you think that change can be done easily?

antgonza:
Not difficult at all, we have 3 options:
1. what we have now
2. this: screen shot 2017-03-29 at 11 42 40 am
3. or this: screen shot 2017-03-29 at 11 42 13 am

ElDeveloper:
Option 2 looks great 👍

josenavas:
My vote goes to 2 - I like when things are modular and can be moved around!!! :D

antgonza:
This is the pushed version: screen shot 2017-03-29 at 11 50 07 am

wasade:
merge?

ElDeveloper:
:shipit:

PR #2093, opened 2017-03-27
Fix 1805

antgonza:
Adding the missing documentation and fixing a link that we missed on previous reviews.

Previous PRs: https://github.com/biocore/qiita/pull/2087 & https://github.com/biocore/qiita/pull/2062.

@josenavas & @wasade, wanna do another review? 

BTW, the new link is shown as:
screen shot 2017-03-27 at 10 24 33 am
and the documentations looks like:
screen shot 2017-03-27 at 10 24 11 am


qiita_pet/handlers/download.py
```diff
@@ -16,8 +19,9 @@ def get(self, filepath_id):
         fid = int(filepath_id)
 
         if not validate_filepath_access_by_user(self.current_user, fid):
-            raise QiitaPetAuthorizationError(
-                self.current_user, 'filepath id %s' % str(fid))
+            raise HTTPError(
+                404, "%s doesn't have access to "
```
wasade: should this be a 403?

josenavas:
@antgonza and I pair-programmed this - do we think that we need another reviewer or with @wasade 's review is enough?

antgonza:
I think if @wasade or someone else reviews we should be OK.

PR #2094, opened 2017-03-28
A minimal REST API for Qiita

ElDeveloper:
A minimal API to support the functionality for the plate mapper. Whenever possible we've used JSON as the way to represent the transferred data. The endpoints are as follows:

| action | URI | description |
|--------|-----|-------------|
|GET |  `/api/v1/study/<int>` | get study details (title, contacts, abstract, description, alias and efo) |
|GET | `/api/v1/study/<int>/samples` | get samples associated with a study and the available metadata headers for these samples |
|GET | `/api/v1/study/<int>/samples?categories=foo,bar` | get metadata categories foo and bar for all samples in the study | 
|GET | `/api/v1/study/<int>/status` | the status of a study (whether or not the study: is public, has sample information, sample information has warnings and a list of preparations |
|GET | `/api/v1/person?name=foo&affiliation=bar` | see if a person exists |
|POST| `/api/v1/study` | create a study (mirrors study creation on qiita UI with minimal requirements) |
|POST| `/api/v1/person?name=foo&affiliation=bar&email=address` | create a study person (ie lab person or PI) |
|POST| `/api/v1/study/<int>/preparation` | associate a prep with a study |
|POST| `/api/v1/study/<int>/preparation/<int>/artifact` | associate filepaths to a preparation, assuming this filepaths are present in the uploads folder |

------------

Open questions ( `¯\_(ツ)_/¯` )
===========

* ~~How to properly document this for consumers and developers.~~
* ~~How should these handlers be name-spaced so they can be imported in `webserver.py`.~~
* ~~Per-user authentication.~~
* ~~What other information do we need to include in the study status.~~


ElDeveloper:
One very important thing I forgot to mention: @wasade and I pair-programmed everything in this PR. 🎉

qiita_pet/handlers/rest/study.py
```diff
@@ -0,0 +1,155 @@
...
+        if not User.exists(owner):
+            self.fail('Unknown user', 403)
+            return
+        else:
+            owner = User(owner)
+
+        efo = payload['efo']
+        contacts = payload['contacts']
+
+        if Study.exists(title):
```
antgonza: Should all the errors/warnings be passed at once?

qiita_pet/handlers/rest/study.py
```diff
@@ -0,0 +1,155 @@
...
+                    sample_information_warnings = False
+        else:
+            sample_information_warnings = False
+
+        preparations = []
+        for prep in study.prep_templates():
+            pid = prep.id
+            art = prep.artifact is not None
+            # TODO: unclear how to test for warnings on the preparations as
+            # it requires knowledge of the preparation type. It is possible
```
antgonza: I'm not sure what this refers to? data_type, information_type, other?

qiita_pet/test/rest/test_study.py
```diff
@@ -0,0 +1,167 @@
...
+        response = self.get('/api/v1/study/-1', headers=self.headers)
+        self.assertEqual(response.code, 404)
+        # not asserting the body content as this is not a valid URI according
+        # to the regex associating the handler to the webserver
+
+    def test_get_invalid_namespace(self):
+        response = self.get('/api/v1/study/1.11111', headers=self.headers)
+        self.assertEqual(response.code, 404)
+        # not asserting the body content as this is not a valid URI according
+
```
antgonza: I guess: "# to the regex associating the handler to the webserver" is missing ...

qiita_pet/test/rest/test_study_preparation.py
```diff
@@ -0,0 +1,180 @@
...
+        prep = StringIO(EXP_PREP_TEMPLATE.format(1))
+        prep_table = load_template_to_dataframe(prep)
+
+        response = self.post('/api/v1/study/1/preparation?data_type=16S',
+                             data=prep_table.T.to_dict(),
+                             headers=self.headers, asjson=True)
+        prepid = json_decode(response.body)['id']
+
+        uri = '/api/v1/study/1/preparation/%d/artifact' % prepid
+        # 1 -> fwd or rev sequences in fastq
```
antgonza: I guess this info needs to be added to the documentation too, right? Or will it need to be part of the API, something that returns the artifact_type, artifact_id?

qiita_pet/webserver.py
```diff
@@ -164,7 +174,21 @@ def __init__(self):
              CommandActivateHandler),
             (r"/qiita_db/plugins/(.*)/(.*)/commands/(.*)/", CommandHandler),
             (r"/qiita_db/plugins/(.*)/(.*)/commands/", CommandListHandler),
-            (r"/qiita_db/plugins/(.*)/(.*)/", PluginHandler)
+            (r"/qiita_db/plugins/(.*)/(.*)/", PluginHandler),
```
antgonza: Could you add a comment to separate the plugin handlers from the API?

qiita_pet/handlers/rest/study.py
wasade: This is the only means we could find to determine if a study already existed. Not sure what else can be passed back.

qiita_pet/handlers/rest/study.py
wasade: The call to `validate` requires knowledge of what the preparation actually is, and we didn't want to replicate a bunch of functionality for this.

qiita_pet/test/rest/test_study.py
wasade: :)

qiita_pet/test/rest/test_study_preparation.py
wasade: Not clear what the best direction here is. Accepting internal IDs like this sucks, so having an informative endpoint to get these names probably makes sense and then just pass in strings. But, we opted to consider this low priority right now since this API is scoped for internal use only at the moment.

qiita_pet/webserver.py
wasade: Done

wasade:
On the additional comments:

- TODOs, no strong feelings
- v1, it's fairly common to see use this convention but also no strong feelings
- Docs, yup, no really great solution within the current qiita ecosystem as numpydoc doesn't lend itself properly and the endpoints are detached from the handler logic anyway. Would you be okay with docstrings off the handlers? @ElDeveloper and I came up with a possible solution which places the endpoint definition in proximity to the handler logic, but we opted to not include it in this PR as its scope really ought to be much larger than just these endpoints.
- name-spaced: the names of the handlers on import in `webserver.py`, although we can keep them as is... 
- authentication: an API key mechanism would be fantastic. At this point, I do not believe qiita has privilege granularity over the keys though, so an API key I believe grants full access.

qiita_pet/handlers/rest/study.py
antgonza: I was thinking in getting the warnings like this, unknown principal investigator, lab person etc all at once vs. each test by itself.

qiita_pet/handlers/rest/study.py
antgonza: Got it! Depending on the data_type is how the validation happen. One simply option is to return all the validations avaialble? Not pretty but this will allow us to test and figure out a better solution. Other option will be to force the user to pass that, I guess a similar issue than artifact_type/id. BTW I think this is important cause the first use case will be either target gene or shotgun sequencing.

qiita_pet/test/rest/test_study_preparation.py
antgonza: Agree, perhaps next step if this is done? @josenavas, what do you think?

wasade:
Our impression was that keys are full access and not user specific at this
time?

antgonza:
Yes, that's how it currently is. Sorry for not being clear. When I said: "(part of your user info page)" I meant extend that functionality so each user has its own.

qiita_pet/test/rest/test_study_samples.py
```diff
@@ -0,0 +1,216 @@
...
+    def test_get_no_category(self):
+        exp = {'message': 'No categories specified'}
+        response = self.get('/api/v1/study/1/samples/categories=',
+                            headers=self.headers)
+        self.assertEqual(response.code, 405)
+        obs = json_decode(response.body)
+        self.assertEqual(obs, exp)
+
+    def test_get_no_samples(self):
+        # /api/v1/study/%d/samples/info -> {'number-of-samples':<int>,
```
SantiagoTorres: image

qiita_pet/handlers/rest/study.py
wasade: That would complicate the logic quite a bit as some checks can't be done if others fail

qiita_pet/handlers/rest/study.py
wasade: We'd need to hardcode the variable names of all the prep template constants which can be passed to `validate`, unless there is a mapping somewhere already? It would also make the response from this API variable as different requirements are added in which could easily cause issue with consumers.

qiita_pet/test/rest/test_study_preparation.py
josenavas: I don't think passing the ids is required. IIRC you can pass the actual string and the code internally will change to the ID. So instead of 3 you can pass 'raw_barcodes'

qiita_pet/webserver.py
```diff
@@ -57,6 +57,16 @@
     ReloadPluginAPItestHandler)
 from qiita_pet import uimodules
 from qiita_db.util import get_mountpoint
+from qiita_pet.handlers.rest import (
+    StudyHandler as WeDontKnowWhatToCallThis,
```
josenavas: What about prefixing all the handlers with API or REST? It will avoid collisions with anything else and it will be clear that is part of the rest api

qiita_pet/handlers/rest/study.py
```diff
@@ -0,0 +1,155 @@
...
+                    'study_description': info['study_description'],
+                    'study_alias': info['study_alias'],
+                    'efo': study.efo})
+        self.finish()
+
+
+class StudyCreatorHandler(RESTHandler):
+
+    @authenticate_oauth
+    def post(self, *args, **kwargs):
```
josenavas: Are we expecting any args/kwargs in here? If not, why having them here?

qiita_pet/handlers/rest/study.py
ElDeveloper: The awkward thing about this is that we would have to mix the error codes, and this is also the way that `Study.create` does things i.e. by raising one exception at a time.

qiita_db/study.py
```diff
@@ -1214,6 +1214,33 @@ def exists(cls, name, affiliation):
...
+        with qdb.sql_connection.TRN:
+            if not cls.exists(name, affiliation):
+                raise qdb.exceptions.QiitaDBLookupError(
+                        'Study person does not exist')
+
+            sql = """SELECT study_person_id FROM qiita.{0}
+                        WHERE name = %s
+                     AND affiliation = %s""".format(cls._table)
+            qdb.sql_connection.TRN.add(sql, [name, affiliation])
+            return cls(int(qdb.sql_connection.TRN.execute_fetchlast()))
```
ElDeveloper: Great point, not sure where this came from, fixed!

qiita_pet/handlers/rest/rest_handler.py
```diff
@@ -0,0 +1,28 @@
...
+# The full license is in the file LICENSE, distributed with this software.
+# -----------------------------------------------------------------------------
+from qiita_db.study import Study
+from qiita_db.exceptions import QiitaDBUnknownIDError
+from qiita_pet.handlers.util import to_int
+from qiita_pet.handlers.base_handlers import BaseHandler
+
+
+class RESTHandler(BaseHandler):
+    def fail(self, msg, status):
```
ElDeveloper: Sounds like a great idea! The reason why we had been putting the error message in the body is because we were modeling the API's *etiquette* after GitHub's API, where they always return either the data or a dictionary with **just** a message, and they even return a link to their API's documentation. So what about modifying `QiitaHTTPError` to write the message to the body?

Just for a reference, this is how it looks when you use curl to pull some data and the request fails:

```
(qiita)yoshikivazquezbaeza:qiita@death-march$ curl -ik 'https://127.0.0.1:21174/api/v1/person?name=LabDude&affiliation=knight%20lab2'
HTTP/1.1 404 Not Found
Date: Thu, 30 Mar 2017 17:59:45 GMT
Content-Length: 31
Content-Type: application/json; charset=UTF-8
Server: TornadoServer/3.1.1

{"message": "Person not found"}
```


Thinking more, should we be logging these errors into the database, or is the Tornado logging sufficient?

qiita_pet/handlers/rest/rest_handler.py
```diff
@@ -0,0 +1,28 @@
...
+        self.write({'message': msg})
+        self.set_status(status)
+        self.finish()
+
+    def study_boilerplate(self, study_id):
```
ElDeveloper: Sounds good, I've renamed the method to safe_get_study.

qiita_pet/handlers/rest/study.py
```diff
@@ -0,0 +1,155 @@
...
+                                     pi.affiliation,
+                                     pi.email],
+                                 'lab_person': [
+                                     lp.name,
+                                     lp.affiliation,
+                                     lp.email]},
+                    'study_abstract': info['study_abstract'],
+                    'study_description': info['study_description'],
+                    'study_alias': info['study_alias'],
+                    'efo': study.efo})
```
ElDeveloper: Sounds good, I've just removed this.

qiita_pet/handlers/rest/study.py
```diff
@@ -0,0 +1,155 @@
...
+
+    @authenticate_oauth
+    def post(self, *args, **kwargs):
+        try:
+            payload = json_decode(self.request.body)
+        except ValueError:
+            self.fail('Could not parse body', 400)
+            return
+
+        required = {'title', 'study_abstract', 'study_description',
```
ElDeveloper: Sounds good, I've just removed this.

qiita_pet/handlers/rest/study.py
```diff
@@ -0,0 +1,155 @@
...
+    def get(self, study_id):
+        study = self.study_boilerplate(study_id)
+        if study is None:
+            return
+
+        public = study.status == 'public'
+        st = study.sample_template
+        sample_information = st is not None
+        if sample_information:
+            with warnings.catch_warnings():
```
ElDeveloper: Awesome, thanks for creating that issue, I just added a link to it.

qiita_pet/handlers/rest/study_preparation.py
```diff
@@ -0,0 +1,87 @@
...
+from qiita_db.artifact import Artifact
+from qiita_pet.handlers.util import to_int
+from qiita_db.exceptions import QiitaDBUnknownIDError, QiitaError
+from qiita_db.metadata_template.prep_template import PrepTemplate
+from qiita_db.handlers.oauth2 import authenticate_oauth
+from .rest_handler import RESTHandler
+
+
+class StudyPrepCreatorHandler(RESTHandler):
+    # TODO: do something smart about warnings, perhaps this should go in its
```
ElDeveloper: Yes, something like that would work! I've linked to #2096 just in case.

qiita_pet/handlers/rest/study_preparation.py
```diff
@@ -0,0 +1,87 @@
...
+
+        self.write({'id': p.id})
+        self.set_status(201)
+        self.finish()
+
+
+class StudyPrepArtifactCreatorHandler(RESTHandler):
+
+    @authenticate_oauth
+    def post(self, study_id, prep_id):
```
ElDeveloper: It makes more sense for how the rest endpoint is constructed, we too noticed this, but didn't quite know what to do  `¯\_(ツ)_/¯`, so we left it as is, is this OK?

qiita_pet/handlers/rest/study_samples.py
```diff
@@ -0,0 +1,81 @@
...
+        study = self.study_boilerplate(study_id)
+        if study is None:
+            return
+
+        if study.sample_template is None:
+            samples = []
+        else:
+            samples = list(study.sample_template.keys())
+
+        self.write(json_encode(samples))
```
ElDeveloper: You are correct, unless you pass a dict, `write` won't convert to JSON.

qiita_pet/handlers/rest/study_samples.py
```diff
@@ -0,0 +1,81 @@
...
+            return
+
+        study = self.study_boilerplate(study_id)
+        if study is None:
+            return
+
+        categories = categories.split(',')
+
+        if study.sample_template is None:
+            self.fail('Category not found', 404)
```
ElDeveloper: The ID is already implied in the URI itself, so it seems a bit redundant. Also, after modeling from GitHub's API, none of their messages seemed to imply any specific information about the requested resource, they were succinct and generic.

qiita_pet/handlers/rest/study_samples.py
```diff
@@ -0,0 +1,81 @@
...
+            return
+
+        available_categories = set(study.sample_template.categories())
+        if not set(categories).issubset(available_categories):
+            self.fail('Category not found', 404)
```
ElDeveloper: Maybe as an additional attribute? The response would be something like:

```
{'message': 'category not found', 'categories_not_found': ['foo', 'bar', 'baz']}
```

qiita_pet/handlers/rest/study.py
josenavas: Ok - sounds good - I don't think is that critical here as it is during metadata addition.

qiita_pet/handlers/rest/rest_handler.py
josenavas: I'm unsure if you are going to be able to modify the QiitaHTTPError to write the message in the body (is not a handler class so I'm unsure on how tornado is handling that internally). It may be worth testing how the current status of the QiitaHTTPError shows when using curl.

I don't think registering these errors in the DB is needed as these are misuses of the API and generally user generated. That would be an easy way of blowing up the size of our database.

qiita_pet/handlers/rest/study_preparation.py
josenavas: Yeah - I think it is more important to have a consistent endpoint :-)

qiita_pet/handlers/rest/study_samples.py
josenavas: ha - tornado hacks... thanks!

qiita_pet/handlers/rest/study_samples.py
josenavas: I was just thinking on differentiating this error from the one below

qiita_pet/handlers/rest/study_samples.py
josenavas: that sounds good!

qiita_pet/webserver.py
ElDeveloper: We've put all these under a constant, this avoids the namespace collisions :) we think this is better.

qiita_pet/handlers/rest/study.py
ElDeveloper: Removed 👍

qiita_pet/handlers/rest/study_preparation.py
ElDeveloper: 🍻

qiita_pet/handlers/rest/study_samples.py
ElDeveloper: 🍻

qiita_pet/handlers/rest/study_samples.py
ElDeveloper: Great point, fixed.

qiita_pet/webserver.py
josenavas: 🍻 I think that is an awesome solution! Never though about that and it is superclean!

qiita_pet/handlers/rest/study_samples.py
ElDeveloper: Great, fixed!

qiita_pet/test/rest/test_study_preparation.py
ElDeveloper: You are right @josenavas, this works with either the identifier or the string value, we've modified a test to show that this works either with identifiers or with ids, should we do anything else?

ElDeveloper:
I believe all comments should be addressed, we've added a minimal documentation block in the Qiita dev docs.

qiita_pet/test/rest/test_study_preparation.py
josenavas: no that I can think of here

ElDeveloper:
Thanks all!!

PR #2095, opened 2017-03-29
fix #2091

### LINKED ISSUE #2091: If you put up a FASTA file, and try to get prep info, you get a 500
```python
  File "/Users/mcdonadt/miniconda3/envs/qiita/lib/python2.7/site-packages/tornado/web.py", line 1141, in _when_complete
    callback()
  File "/Users/mcdonadt/miniconda3/envs/qiita/lib/python2.7/site-packages/tornado/web.py", line 1162, in _execute_method
  File "/Users/mcdonadt/miniconda3/envs/qiita/lib/python2.7/site-packages/tornado/web.py", line 2297, in wrapper
  File "/Users/mcdonadt/ResearchWork/software/qiita/qiita_core/util.py", line 121, in wrapper
    return func(*args, **kwargs)
  File "/Users/mcdonadt/ResearchWork/software/qiita/qiita_pet/handlers/download.py", line 18, in get
    if not validate_filepath_access_by_user(self.current_user, fid):
  File "/Users/mcdonadt/ResearchWork/software/qiita/qiita_db/meta_util.py", line 128, in validate_filepath_access_by_user
    if (a.visibility == 'public' or a.study.has_access(user)):
AttributeError: 'NoneType' object has no attribute 'visibility'
ERROR:tornado.access:500 GET /download/28 (::1) 121.62ms
```

antgonza:
@wasade, @josenavas, could you review?

### DIFF
--- a/qiita_db/meta_util.py
+++ b/qiita_db/meta_util.py
@@ -123,14 +123,21 @@ def validate_filepath_access_by_user(user, filepath_id):
             # the prep access is given by it's artifacts, if the user has
             # access to any artifact, it should have access to the prep
             # [0] cause we should only have 1
-            a = qdb.metadata_template.prep_template.PrepTemplate(
-                pid[0]).artifact
-            if (a.visibility == 'public' or a.study.has_access(user)):
-                return True
+            pt = qdb.metadata_template.prep_template.PrepTemplate(
+                pid[0])
+            a = pt.artifact
+            # however, the prep info file could not have any artifacts attached
+            # , in that case we will use the study access level
+            if a is None:
+                return qdb.study.Study(pt.study_id).has_access(user)
             else:
-                for c in a.descendants.nodes():
-                    if (c.visibility == 'public' or c.study.has_access(user)):
-                        return True
+                if (a.visibility == 'public' or a.study.has_access(user)):
+                    return True
+                else:
+                    for c in a.descendants.nodes():
+                        if ((c.visibility == 'public' or
+                             c.study.has_access(user))):
+                            return True
             return False
         # analyses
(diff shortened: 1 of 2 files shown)


qiita_db/meta_util.py
```diff
@@ -123,14 +123,21 @@ def validate_filepath_access_by_user(user, filepath_id):
...
+            if a is None:
+                return qdb.study.Study(pt.study_id).has_access(user)
             else:
-                for c in a.descendants.nodes():
-                    if (c.visibility == 'public' or c.study.has_access(user)):
-                        return True
+                if (a.visibility == 'public' or a.study.has_access(user)):
+                    return True
+                else:
+                    for c in a.descendants.nodes():
```
wasade: I don't understand how this fits in...?

wasade:
:+1: short of the question, and the present travis failures

qiita_db/meta_util.py
antgonza: This was there before and decided to leave it. Basically, it protects for the case that an artifact is not public but one of its children is. As you know, this could happen right now and we have an issue to solve it and it doesn't hurt. BTW and AFAIK this is still possible in the test db so it's being tested. 

The travis failure is the stochastic error we see ...

wasade:
It was? I didn't see it on review hence the question

qiita_db/meta_util.py
antgonza: You are right, it's not shown here ... but it does exist in the original code: https://github.com/biocore/qiita/blob/master/qiita_db/meta_util.py#L131, perhaps github doing some magic?

PR #2097, opened 2017-03-31
Update and add samples via REST

wasade:
Added a `PATCH` to `/api/v1/study/<int>/samples` to facilitate adding new samples to sample information (e.g., blanks, replicates, etc), or to update existing metadata for a sample (e.g., `physical_sample_id`)

This was pair programmed with @ElDeveloper


qiita_pet/handlers/rest/study_samples.py
```diff
@@ -27,6 +28,58 @@ def get(self, study_id):
...
+        data = pd.DataFrame.from_dict(json_decode(self.request.body),
+                                      orient='index')
+
+        if len(data.index) == 0:
+            self.fail('No samples provided', 400)
+            return
+
+        categories = set(study.sample_template.categories())
+
+        if set(data.columns) != categories:
```
antgonza: This expects to have the same columns, right? What about the case that metadata changed (perhaps out of the scope of this PR)? Anyway, why not use: the [extend](https://github.com/biocore/qiita/blob/master/qiita_db/metadata_template/base_metadata_template.py#L1114) or [update](https://github.com/biocore/qiita/blob/master/qiita_db/metadata_template/base_metadata_template.py#L1129) methods?

wasade: We do, see the lines below

From discussing with @josenavas, we decided that we will not allow for expansion of metadata, and that the client must specify a value for every column.

antgonza: OK, just want to confirm that we explicitly want to limit the REST functionality of something that is already possible in the GUI ...

ElDeveloper: Yes, I think this is correct @antgonza, as per @josenavas's request, we only need to add blank samples right now, and AFAIK they will inherit the existing metadata categories in the sample information.

antgonza: Thanks!

PR #2098, opened 2017-04-03
Changing how artifact visibility works

antgonza:
I'm almost sure some of the tests will fail and I can fix later but want to see if you guys agree with this approach before expending time fixing them. 

@wasade and @josenavas , what do you think?


qiita_db/artifact.py
```diff
@@ -617,19 +619,20 @@ def visibility(self, value):
         only applies when the new visibility is more open than before.
         """
         with qdb.sql_connection.TRN:
+            # In order to correctly propagate the visibility we need to find
+            # the root of this artifact and then propagate to all the artifacts
+            root = self.artifact
```
josenavas: What is self.artifact? 

Anyways, you can simplify this by:

```python
sql = "SELECT * FROM qiita.find_artifact_roots(%s)"
TRN.add(sql, [self.id])
root = TRN.execute_fetchlast()
```

qiita_db/support_files/patches/python_patches/53.py
```diff
@@ -0,0 +1,19 @@
...
+for rd in raw_data:
+    # getting the most open visibility of all the children in the pipeline
+    children = rd.descendants.nodes()
+    vis = [a.visibility for a in children]
+    vis.append(rd.visibility)
+
+    new_vis = 'sandbox'
+    if 'public' in vis:
+        new_vis = 'public'
+    elif 'private' in vis:
```
wasade: are there situations where an artifact was bumped to private but where the user expects all other artifacts to remain sandboxed?

qiita_db/artifact.py
antgonza: brain fart, thanks! Will fix ...

qiita_db/support_files/patches/python_patches/53.py
antgonza: in praxis no ... at least based on what we have seen so far ...

qiita_db/artifact.py
```diff
@@ -260,9 +260,11 @@ def create(cls, filepaths, artifact_type, name=None, prep_template=None,
 
         Notes
         -----
-        The visibility of the artifact is set by default to `sandbox`
-        The timestamp of the artifact is set by default to `datetime.now()`
-        The value of `submitted_to_vamps` is set by default to `False`
+        The visibility of the artifact is set by default to `sandbox` if
+        prep_template is passed but if parents is passed we will inherit the
+        most closed visibility.
```
ElDeveloper: most closed -> closest?

qiita_db/artifact.py
```diff
@@ -260,9 +260,11 @@ def create(cls, filepaths, artifact_type, name=None, prep_template=None,
 
         Notes
         -----
-        The visibility of the artifact is set by default to `sandbox`
-        The timestamp of the artifact is set by default to `datetime.now()`
-        The value of `submitted_to_vamps` is set by default to `False`
+        The visibility of the artifact is set by default to `sandbox` if
+        prep_template is passed but if parents is passed we will inherit the
```
ElDeveloper: `is passed but` -> `is passed and if`?

antgonza:
This is ready for actual review ...

antgonza:
@wasade, @ElDeveloper, @josenavas, wanna take a look?

qiita_db/artifact.py
ElDeveloper: I understand now, what you mean here.

qiita_db/artifact.py
ElDeveloper: I think I also understand this as well.

PR #2099, opened 2017-04-04
OAuth2 support allowing public and user access

wasade:
This was partially pair programmed with @ElDeveloper.

The goal of this PR is to lay the ground work for allowing handlers to optionally allow defaulting to a public user, and to also make the `User` available to the handler so that access control can be performed if necessary.

What has been done so far is:

- added a means to associate a oauth2 `client_id` with a `User`
- added a means to instantiate a `User` from a `client_id`  
- added tests for these `User` modifications
- decomposed the oauth2 handling logic
- added explicit tests for the oauth2 handling logic using a mocked out handler
- creation of a parameterized oauth2 decorator
- added tests for the parameterized oauth2 decorator using a mocked out handler
- replacement of the original `authenticate_oauth` decorator across all of Qiita

The way the new decorator works is to allow, per method, a means to describe whether it can default to access if a) a valid token is provided or b) if the `Authorization` header is not provided. If a oauth2 token is provided, and it is invalid, rate limited, etc, the decorator will error as expected. This approach requires the developer indicate explicitly that a method can default to public, and that statement is proximal to the function signature. Second, and per method as well, the `User` object associated with the oauth2 `client_id` can be made available by monkey patching `get_current_user`. As of right now, the public user is `demo@microbio.me`. This monkey patching is parameterized as the original `authenticate_oauth` decorator did not do this, so it was added to minimize possible side effects from replacing the authentication mechanism.


ElDeveloper:
:+1: looks good to me!!

qiita_db/handlers/oauth2.py
```diff
@@ -47,66 +47,138 @@ def _oauth_error(handler, error_msg, error):
...
+        """
+        @functools.wraps(f)
+        def wrapper(handler, *args, **kwargs):
+            errtype, errdesc, cid = _check_oauth2_header(handler)
+
+            if self.default_public:
+                # no error, or no authorization header. We should error if
+                # oauth is actually attempted but there was an auth issue
+                # (e.g., rate limit hit)
+                if errtype not in (None, 'invalid_request'):
```
antgonza: test?

qiita_db/handlers/oauth2.py
```diff
@@ -47,66 +47,138 @@ def _oauth_error(handler, error_msg, error):
...
-                r_client.decr(limit_key)
-                if int(r_client.get(limit_key)) <= 0:
-                    _oauth_error(
-                        handler, 'Oauth2 error: daily request limit reached',
-                        'invalid_grant')
+                if errtype is not None:
+                    _oauth_error(handler, errdesc, errtype)
                     return
+                if self.inject_user:
+                    if cid is None:
```
antgonza: test?

wasade:
Thanks @antgonza! Regarding your comments,

> for a couple of tests missing

[Here](https://github.com/biocore/qiita/pull/2099/files/b839bfeba09b7a0290603b60179a8eb57dafff92#diff-a966de12bcac4b803a12d6de44564e51R245), and one is not possible to reach so I'm not sure how to test. 

>  I think it's worth deploying in the test env and testing there

sounds good

> this will not really change anything in the front end, right? What will be missing to finish it?

It was not intended to change the front end, this was to allow for per-user support against the rest API.

> I think is worth removing the demo@microbio.me default

@josenavas suggested this. We're fine with anything but it does need to be a valid qiita account.

> it's not clear for me how this generated the client_id for existing users

We have not touched the existing code which already does this. What is missing though is a UI to generate per-user API tokens, which effectively needs to call the existing logic, and then issue a call to `User(foo).client_id = bar`.

wasade:
Oh, I see the test issue now. "Test?" Is somewhat ambiguous fyi... Ya, we
can add one

On Apr 5, 2017 06:33, "Antonio Gonzalez" <notifications@github.com> wrote:

*@antgonza* commented on this pull request.

Code looks fine except:

   - for a couple of tests missing
   - I think it's worth deploying in the test env and testing there
   - this will not really change anything in the front end, right? What
   will be missing to finish it?
   - I think is worth removing the demo@microbio.me default
   - it's not clear for me how this generated the client_id for existing
   users

------------------------------

In qiita_db/handlers/oauth2.py
<https://github.com/biocore/qiita/pull/2099#discussion_r109914532>:

> +        Notes
+        -----
+        If an error with oauth2 occurs, a status code of 400 is set, a message
+        about the error is sent out over `write` and the response is ended
+        with `finish`. This happens without control being passed to the
+        handler, and in this situation, the handler is not executed.
+        """
+        @functools.wraps(f)
+        def wrapper(handler, *args, **kwargs):
+            errtype, errdesc, cid = _check_oauth2_header(handler)
+
+            if self.default_public:
+                # no error, or no authorization header. We should error if
+                # oauth is actually attempted but there was an auth issue
+                # (e.g., rate limit hit)
+                if errtype not in (None, 'invalid_request'):

test?
------------------------------

In qiita_db/handlers/oauth2.py
<https://github.com/biocore/qiita/pull/2099#discussion_r109914567>:

>                      return
+                if self.inject_user:
+                    if cid is None:

test?

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/biocore/qiita/pull/2099#pullrequestreview-31046944>,
or mute
the thread

PR #2102, opened 2017-04-06
fix #2086

### LINKED ISSUE #2086: Download public studies release
As a group we decided that:
- [x] it should be a direct link from Qiita using sym links
- [x] we want to have md5s
- [x] README with the timestamp

antgonza:
A few changes from the original issue:
- The timestamp and md5 are stored in redis so we can quickly display
- I moved the test_generate_biom_and_metadata_release method to meta_util.py as it requires redis and we already are importing it there and the other "qiita-cron-jobs" are there ...

Note that I'm gonna deploy in qiita-test to check that everything works as expected.


antgonza:
Forgot to add that the new link looks like: 
screen shot 2017-04-06 at 9 53 39 am

qiita_core/util.py
```diff
@@ -141,3 +142,32 @@ def get_qiita_version():
         sha = ''
 
     return (qiita_pet_lib_version, sha)
+
+
+def get_release_info(study_status='public'):
+    """Returns the study status release MD5
+
+    level
```
ElDeveloper: This should be Parameters.

qiita_core/util.py
```diff
@@ -141,3 +142,32 @@ def get_qiita_version():
...
+    ----------
+    study_status : str, optional
+        The study status to search for. Note that this should always be set
+        to 'public' but having this exposed as helps with testing. The other
```
ElDeveloper: exposed as helps -> exposed helps

qiita_pet/templates/sitebase.html
```diff
@@ -510,6 +511,23 @@
...
               </li>
             </ul>
+            <ul class="nav navbar-nav">
+              <!-- downloads -->
+              <li class="dropdown">
+                <a href="#" data-toggle="dropdown" class="dropdown-toggle">Downloads<b class="caret"></b></a>
+                <ul class="dropdown-menu">
+                  <li>
+                    <a href="{% raw qiita_config.portal_dir %}/release/download/public">
+                      Public processed data
```
ElDeveloper: Maybe instead "Public BIOM and metadata files", or something a bit more specific?

antgonza:
@ElDeveloper, @wasade, @josenavas, this is ready for final review. I would like to have this reviewed/merged soon so I can add the bulk raw data download in a following PR (will issue soon!). Thanks.

PR #2103, opened 2017-04-09
Adding endpoint to retrieve list of person

josenavas:
Useful to pre-populate dropdown menus, etc,


josenavas:
@ElDeveloper @antgonza available for a review?

qiita_pet/handlers/rest/study_person.py
```diff
@@ -15,18 +17,26 @@
...
+            try:
+                p = StudyPerson.from_name_and_affiliation(name, affiliation)
+            except QiitaDBLookupError:
+                self.fail('Person not found', 404)
+                return
+            self.write({'address': p.address, 'phone': p.phone,
+                        'email': p.email, 'id': p.id})
+            self.finish()
+        else:
+            self.fail('Missing parameter', 400)
```
ElDeveloper: I think this should be a raise `MissingArgumentError`?

josenavas: Thanks! Forgot about that one :)

qiita_pet/test/rest/test_study_person.py
```diff
@@ -45,6 +54,11 @@ def test_get_invalid_query_string(self):
                             headers=self.headers)
         self.assertEqual(response.code, 400)
 
+    def test_get_invalid_query_string_2(self):
+        response = self.get('/api/v1/person?affiliation=knight%20lab',
+                            headers=self.headers)
+        self.assertEqual(response.code, 400)
```
ElDeveloper: I think this will need to be updated.

josenavas: Nope - the [response code of a missing argument is 400](http://www.tornadoweb.org/en/stable/web.html#tornado.web.MissingArgumentError) :-) I was basically replicating the behavior of that exception with less readable code xD

ElDeveloper: 👍

PR #2104, opened 2017-04-10
Download bulk raw data

antgonza:
Depends on https://github.com/biocore/qiita/pull/2102, please review/merge that one first.


qiita_pet/test/test_download.py
```diff
@@ -77,8 +77,6 @@ def test_download_study(self):
         with open(tgz, 'w') as f:
             f.write('\n')
 
-        self._clean_up_files.append(tmp_dir)
```
antgonza: rm cause it's twice, see line 65

qiita_pet/handlers/download.py
```diff
@@ -143,5 +143,87 @@ def get(self, extras):
...
+        # loop over artifacts and retrieve raw data (no parents)
+        to_download = []
+        vfabu = validate_filepath_access_by_user
+        for a in study.artifacts():
+            if not a.parents:
+                to_add = True
+                for i, (fid, path, data_type) in enumerate(a.filepaths):
+                    # validate access only of the first artifact filepath,
+                    # the rest have the same permissions
+                    if (i == 0 and not vfabu(user, fid)):
```
wasade: A user can only have access to only some of the filepaths associated with a study?

qiita_pet/handlers/download.py
```diff
@@ -143,5 +143,87 @@ def get(self, extras):
...
+
+        zip_fn = 'study_raw_data_%d_%s.zip' % (
+            study_id, datetime.now().strftime('%m%d%y-%H%M%S'))
+
+        self.set_header('Content-Description', 'File Transfer')
+        self.set_header('Expires', '0')
+        self.set_header('Cache-Control', 'no-cache')
+        self.set_header('X-Archive-Files', 'zip')
+        self.set_header('Content-Disposition',
+                        'attachment; filename=%s' % zip_fn)
```
wasade: where is the file created?

qiita_pet/handlers/download.py
antgonza: To clarify, validate_filepath_access_by_user checks that the specific user has access to the specific filepath id based on it's data type so we are checking that the user has access to that filepath and if it does it allows download to all the other filepaths within that artifact. Remember an artifact can be formed by several files (filepath ids).

qiita_pet/handlers/download.py
antgonza: Nowhere, is created on the fly by nginx. Basically, we send a list of filepaths to nginx using the "protected" filepath (only nginx has access) and the zip is created during download.

qiita_pet/handlers/download.py
wasade: There can be an artifact associated with a study that is inaccessible to a user who has access to the study? That seems weird to me

qiita_pet/handlers/download.py
wasade: Ah, thought that might be the case. Can you add a comment since it's implicit?

qiita_pet/handlers/download.py
antgonza: Good point. We "normally" use validate_filepath_access_by_user to check that a user has access to a giving filepath for downloading specific files; with that in mind we are reusing it for bulk downloads but we don't really need to as we are checking that the user has "full" access to the artifact. Now, we can leave to be over cautious or remove ... I don't have a preference, @wasade?

wasade: I don't know enough about the permissions model to understand the implications. Can someone else who is more familiar weigh in?

josenavas: Given the check on line 175 I don't think there is any case where `validate_filepath_access_by_user` will return false. I don't have a preference on leaving or removing.

qiita_pet/handlers/download.py
wasade: eh?

wasade:
Just looking for that one comment

wasade:
ah, missed it, thanks @antgonza for linking directly

PR #2108, opened 2017-04-18
Revised rst values used for section headers

wasade:


wasade:
looks like a build timeout

antgonza:
both @wasade and I checked in our local machines and looks fine, merging.

PR #2117, opened 2017-04-25
Analysis refactor gui part7

josenavas:
This is NOT ready for review

I will comment with relevant information about the PR and remove the WIP tag once this is ready for review


josenavas:
@antgonza ready for review. In this PR I've modified the command object so it can be defined in the analysis pipeline only. This PR is also adding some of the missing files that I had locally but not in the branch to be able to execute the analysis as plugins. I've also fixed a couple of small bugs that I found on the job object.

qiita_db/private.py
```diff
@@ -0,0 +1,74 @@
...
+                                  'provenance': dumps({'job': job.id,
+                                                       'data_type': dtype}),
+                                  'analysis': analysis_id})
+            val_jobs.append(qdb.processing_job.ProcessingJob.create(
+                analysis.owner, validate_params))
+
+        job._set_validator_jobs(val_jobs)
+
+        for j in val_jobs:
+            j.submit()
```
antgonza: Just thinking out loud (not sure if that applies to written stuff, anyway ...) but should we add the sleep directly to the j.submit?

qiita_db/processing_job.py
```diff
@@ -1196,7 +1208,7 @@ def _raise_if_not_in_construction(self):
                      WHERE processing_job_workflow_id = %s"""
             qdb.sql_connection.TRN.add(sql, [self.id])
             res = qdb.sql_connection.TRN.execute_fetchflatten()
-            if len(res) != 1 or res[0] != 'in_construction':
+            if (len(res) == 1 and res[0] != 'in_construction') or len(res) > 1:
```
antgonza: Could you add some documentation of each test? I'm not following why/when len(res) > 1 ...

qiita_pet/templates/artifact_ajax/processing_artifact.html
```diff
@@ -0,0 +1,425 @@
...
+      var sel = $('<select>').appendTo("#cmd-opts-div").attr('id', p_name).attr('name', p_name).addClass('required-parameter');
+      for(var key in sel_artifacts_info) {
+         if(allowed_types.indexOf(sel_artifacts_info[key].type) !== -1) {
+           sel.append($('<option>').attr('value', key).text(sel_artifacts_info[key].name));
+         }
+      }
+    }
+    else if (p_type == 'integer') {
+      var inp = $('<input>').appendTo("#cmd-opts-div").attr('id', p_name).attr('name', p_name).attr('type', 'number').addClass('required-parameter');
+    }
```
antgonza: should there be an else case?

qiita_pet/templates/artifact_ajax/processing_artifact.html
```diff
@@ -0,0 +1,425 @@
...
+   *
+   * @param p_nodes list The ids of the selected artifacts
+   *
+   * This function executes an AJAX call to retrieve all the commands that can
+   * process the selected artifacts. It generates the interface so the user
+   * can select which command should be added to the workflow
+   *
+   */
+  function load_artifact_type(p_nodes) {
+    var types = []
```
antgonza: missing ;

scripts/qiita-private-2
```diff
@@ -0,0 +1,30 @@
+#!/usr/bin/env python
```
antgonza: not sure qiita-private-2 is a good name, sorry no suggestions ...

qiita_db/private.py
josenavas: I would disagree because the sleep is only needed if submitting multiple jobs, if you're submitting a single one it is not needed, so no reason to stop the processing.

qiita_db/processing_job.py
josenavas: Added a comment

qiita_pet/templates/artifact_ajax/processing_artifact.html
josenavas: Oops - I moved this file and started working on this function before our meeting last week. This function is the one that I have to modify in the next functionality, so I would suggest leave as it is since I will be completing this function in my next PR. Does this sound good to you?

qiita_pet/templates/artifact_ajax/processing_artifact.html
josenavas: Added

josenavas: Added

scripts/qiita-private-2
josenavas: Agreed - renamed

qiita_db/private.py
wasade: why sleep?

qiita_db/private.py
```diff
@@ -0,0 +1,74 @@
...
+    """
+    if job_id == 'register':
+        # We don't need to do anything here if Qiita is registering plugins
+        return
+
+    job = qdb.processing_job.ProcessingJob(job_id)
+    job.update_heartbeat_state()
+
+    try:
+        task_name = job.command.name
```
wasade: Does the instantiation of this local variable need to be within the try?

qiita_db/private.py
```diff
@@ -0,0 +1,74 @@
...
+        TASK_DICT[task_name](job)
+    except Exception:
+        job.complete(False, error="Error executing private task: %s"
```
wasade: what happens if the job does not except, shouldn't there be a `job.complete(True, ...)`?

qiita_db/private.py
josenavas: In some cases we found that if we programmatically submit multiple jobs to the torque queue, the resource queue are not updated accordingly and the jobs end up not being spawned fairly across the cluster.

qiita_db/software.py
```diff
@@ -222,6 +226,9 @@ def create(cls, software, name, description, parameters, outputs=None):
         outputs : dict, optional
             The description of the outputs that this command generated. The
             format is: {output_name: artifact_type}
+        analysis_only : bool, optional
```
wasade: Does false mean the command is in the analysis pipeline and whatever other pipelines exist? Is it important to be able to restrict commands to non-analysis pipelines? Or, rather, is a two-state variable sufficient to describe the different scenarios faced?

qiita_db/software.py
```diff
@@ -508,6 +515,22 @@ def activate(self):
...
             return qdb.sql_connection.TRN.execute()
 
+    @property
+    def analysis_only(self):
+        """Returns if the command is an analysis-only command
+
+        Returns
+        -------
+        bool
+            Whether the commad is analysis only or not
```
wasade: command

qiita_db/software.py
```diff
@@ -508,6 +515,22 @@ def activate(self):
             qdb.sql_connection.TRN.add(sql, [True, self.id])
             return qdb.sql_connection.TRN.execute()
 
+    @property
+    def analysis_only(self):
+        """Returns if the command is an analysis-only command
```
wasade: Can a command be both an analysis command and "other"? I guess I don't understand what is under the umbrella of "analysis" and what is not

qiita_db/private.py
josenavas: Not really, moved outside of the try.

qiita_db/private.py
josenavas: Not necessarily, as it depends on the actual private task being executed (right now there is only one type of private tasks, but we will be adding more in the future).

Let's say that we have job A that generates one output O. The output O needs to be validated by a different job B. The way it is currently done is that job A is marked as 'running' until the validator job B has completed. This way, if the validator fails because the output is not correct, the job A is marked as failed (as it should be). However, if the validator B completes successfully, it automatically marks job A as completed and success.

The current private task (build_analysis_files) takes advantage of this behavior by creating the BIOM tables and submitting the validator jobs to the queue. Once those jobs are completed, they will set the correct output status to the current job. Does it make sense?

qiita_db/software.py
josenavas: We currently have two pipelines using the artifacts structure: the study pipeline and the (meta-)analysis pipeline.  In my opinion, commands in the study pipeline should be available in the meta-analysis pipeline (my example command was open-ref, although we are no longer adding it to the system), but we do want to restrict the analysis pipeline commands on the study pipeline. Given the current status of the project and what we have outlined so far for the future of Qiita I don't foresee adding any other pipeline to the system that will use the artifacts structure as it currently stands.

qiita_db/software.py
josenavas: There are currently two options:
(1) The command is available in the analysis pipeline and the study processing pipeline
(2) The command is available only in the analysis pipeline.

I don't think the option "Only on study processing pipeline" should be added as this can potentially limit the opportunities to run meta-analysis, see my previous comment.

qiita_db/software.py

qiita_db/private.py
wasade: The reason for the delays historically was to spread load on the resource manager and scheduler as they are singleton resources within a shared environment. Generally speaking, the delays were only important for large volumes of submissions (e.g., > 100). I don't think we're doing that here. If there are issues where the scheduler or resource manager are not handling a small number of jobs appropriately, then it suggests either the jobs themselves are flawed or there is a critical issue with the servers handling the requests. Since we routinely hammer these servers outside of Qiita without delays, it suggests the former is the more likely scenario.

If you're not comfortable removing the delay, then at least set it to <= 100ms as 1s is a lifetime.

I wasn't aware Qiita used a library that links to the Torque C API? Aren't these programatic requests actually system calls?

qiita_db/software.py
wasade: What is the logical difference between a study command and an analysis command?

PR #2118, opened 2017-04-26
Let's get redbiom-ing

antgonza:
Just installing redbiom, once the test pass I'll add a handler


antgonza:
I _think_ this is ready for review. Sorry for the amount of changes but as you can imagine, had to do a bunch of tests/changes based on these. Note that this goes to it's own branch so it should be fine to merge once is reviewed and changes are approved. 

At this point, we are using redbiom to _only_ get the sample_ids that match a giving query (metadata or sequence), to then use the database to find the artifacts (based on the preparations) and their information where those samples exist. 

The ToDo list is as follows, this list will be moved to the PR against master once this one is merged:
- [ ] Move out database code from qiita_pet.handler
- [ ] Decide what to do with the context limitation to search based on sequences. Let's discuss this on our next Qiita meeting.
- [ ] Decide what to do with displaying on which categories/artifacts the search query is found. Also, let's discuss on next Qiita meeting.  
- [ ] Fix GUI based on decisions on the last points.
- [ ] Others??

INSTALL.md
```diff
@@ -87,6 +88,26 @@ brew update
 brew install homebrew/versions/redis28
 ```
 
+### webdis
+
+Note that this is the only package that assumes that Qiita is already installed (due to library dependencies). Also, that the general suggestion is to have 2 redis servers running, one for webdis/redbiom and the other for Qiita. The default configuration has webdis/redbiom running on the default redis port and Qiita's on 7777.
```
wasade: Can you add:

"The reason for multiple redis servers is so that the redbiom cache can be flushed without impacting the operation of the qiita server itself."

INSTALL.md
```diff
@@ -163,6 +184,12 @@ Next, make a test environment:
 qiita-env make --no-load-ontologies
 ```
 
+Finally, redbiom rely on the REDBIOM_HOST environment variable to set the URL to query. By default is set to http://127.0.0.1:7379, which is the webdis default. For example you could:
```
wasade: rely -> relies

once we shift to a release, it's be super awesome to have a default `REDBIOM_HOST` for clients, and that its `qiita.ucsd.edu` with forwarding provided by nginx.

qiita_pet/handlers/qiita_redbiom.py
```diff
@@ -0,0 +1,118 @@
...
+    @execute_as_transaction
+    def _get_context(self, callback):
+        df = redbiom.summarize.contexts()
+        result = df.ContextName.values
+        callback(result)
+
+    @coroutine
+    @execute_as_transaction
+    def get(self, search):
+        contexts = yield Task(self._get_context)
```
wasade: just curious, how annoying was it to deal with these `Task`s?

qiita_pet/handlers/qiita_redbiom.py
```diff
@@ -0,0 +1,118 @@
...
+
+    @execute_as_transaction
+    def _redbiom_search(self, context, query, search_on, callback):
+        df = redbiom.summarize.contexts()
+        contexts = df.ContextName.values
+        if context not in contexts:
+            callback(([], 'The given context is not valid: %s - %s' % (
+                context, contexts)))
+        else:
+            if search_on == 'metadata':
```
wasade: we may want an additional `elif` here for searching over `categories`

qiita_pet/handlers/qiita_redbiom.py
```diff
@@ -0,0 +1,118 @@
...
+                samples = redbiom.search.metadata_full(query, categories=None)
```
wasade: `categories` is defined as a `bool`, can this be set to `categories=False`?

qiita_pet/handlers/qiita_redbiom.py
```diff
@@ -0,0 +1,118 @@
...
+            elif search_on == 'observations':
+                # from_or_nargs first parameter is the file handler so it uses
```
wasade: I don't understand this comment. `from_or_nargs` handles the case of passing in a list of things via a file or from the command line

qiita_pet/handlers/qiita_redbiom.py
```diff
@@ -0,0 +1,118 @@
...
+                # that as the query input. None basically will force to take
+                # the values from query
+                samples = [s.split('_', 1)[1]
+                           for s in redbiom.util.samples_from_observations(
+                               query.split(' '), True, context)]
```
wasade: the `True` here may not be always desirable but that can be deferred

qiita_pet/handlers/qiita_redbiom.py
```diff
@@ -0,0 +1,118 @@
...
+            callback(([], 'The given context is not valid: %s - %s' % (
+                context, contexts)))
+        else:
+            if search_on == 'metadata':
+                samples = redbiom.search.metadata_full(query, categories=None)
+            elif search_on == 'observations':
+                # from_or_nargs first parameter is the file handler so it uses
+                # that as the query input. None basically will force to take
+                # the values from query
+                samples = [s.split('_', 1)[1]
```
wasade: What this is doing is converting from a redbiom ID to an ambiguous ID, and it does not seem to be enforcing that the ID is unique through a `set` comprehension. You may want to use `redbiom.util.resolve_ambiguities` instead

qiita_pet/handlers/qiita_redbiom.py
```diff
@@ -0,0 +1,118 @@
...
+                           for s in redbiom.util.samples_from_observations(
+                               query.split(' '), True, context)]
+            else:
+                callback(([], 'Incorrect search by: you can use observations '
+                          'or metadata and you passed: %s' % search_on))
+
+            if bool(samples):
```
wasade: why cast to `bool`?

qiita_pet/static/js/qiita.js
```diff
@@ -117,3 +117,90 @@ function show_hide_process_list() {
     $("#qiita-processing").hide();
   }
 }
+
```
wasade: i may not be the best person for review of changes to this file

qiita_pet/templates/redbiom.html
```diff
@@ -0,0 +1,167 @@
...
+                       row.study_id + '" role="button">' + row.study_title + ' (ID: ' + row.study_id + ')</a>';
+          {% else %}
+            var text = row.study_title + ' (ID: ' + row.study_id + ')';
+          {% end %}
+          return text;
+        }, targets: [1]},
+        // render Artifact Processing
+        {"render": function ( data, type, row, meta ) {
+            var text = '';
+            if (!(!row.command)) {
```
wasade: Isn't a double negative a positive?

qiita_pet/handlers/qiita_redbiom.py
```diff
@@ -0,0 +1,118 @@
...
+    @coroutine
+    @execute_as_transaction
+    def get(self, search):
+        contexts = yield Task(self._get_context)
+        self.render('redbiom.html', contexts=contexts)
+
+    @execute_as_transaction
+    def _redbiom_search(self, context, query, search_on, callback):
+        df = redbiom.summarize.contexts()
+        contexts = df.ContextName.values
```
wasade: per gchat, we may want to filter these. And one other possibility in addition to what we discussed was is to filter on load of the cache as well.

INSTALL.md
antgonza: k

INSTALL.md
antgonza: k, makes sense for default installations but it might be better to leave as is as the idea of these instructions is to have local installations for development.

qiita_pet/handlers/qiita_redbiom.py
antgonza: Once you get the drift, pretty easy.

qiita_pet/handlers/qiita_redbiom.py
antgonza: yup, let's chat this afternoon

qiita_pet/handlers/qiita_redbiom.py
antgonza: yup, let's chat this afternoon

qiita_pet/handlers/qiita_redbiom.py
antgonza: k

qiita_pet/handlers/qiita_redbiom.py
antgonza: in one of the previous versions I was using from_or_nargs but then replaced with current code and forgot to remove that comment, removing.

qiita_pet/handlers/qiita_redbiom.py
antgonza: ok, this is one of the other things we should discuss this afternoon.

qiita_pet/handlers/qiita_redbiom.py
antgonza: yup, also within our discussion today.

qiita_pet/handlers/qiita_redbiom.py
antgonza: I read somewhere that was the safest way to test for an empty set ...

qiita_pet/static/js/qiita.js
antgonza: it's mainly moving code around so the libraries are available everywhere vs. just in the study listing

qiita_pet/templates/redbiom.html
antgonza: yup, again read somewhere that to test is a variable is not undefined or null the easiest is to do !row.command but didn't want to add an else so just negated that boolean ...

qiita_pet/templates/redbiom.html
```diff
@@ -0,0 +1,175 @@
...
+          }
+        });
+    });
+  });
+</script>
+
+{%end%}
+
+{%block content%}
+  <small>Some general instructions!</small>
```
josenavas: I'm assuming that this is a placeholder - right?

qiita_pet/templates/redbiom.html
```diff
@@ -0,0 +1,175 @@
...
+
+      <div class="col-xs-2">
+        <select class="selectpicker form-control col-xs-2" name="search_on" id="search_on">
+          <option value="metadata">Metadata</option>
+          <option checked value="observations">Observation</option>
+          <option checked value="categories">Metadata Categories</option>
+        </select>
+      </div>
+      <div class="col-xs-1">
+          <button class="btn btn-default" type="submit">GO!</button>
```
josenavas: `GO!` -> `Search`

qiita_pet/test/test_qiita_redbiom.py
```diff
@@ -0,0 +1,161 @@
+from unittest import main
```
josenavas: This file is missing the copyright notice

qiita_pet/test/test_qiita_redbiom.py
```diff
@@ -0,0 +1,161 @@
...
+        self.assertEqual(loads(response.body), exp)
+
+        post_args = {
+            'search': '4353076',
+            'search_on': 'metadata'
+        }
+        response = self.post('/redbiom/', post_args)
+        self.assertEqual(response.code, 200)
+        exp = {'status': 'success',
+               'message': ('Not a valid search: "4353076", are you sure this '
```
josenavas: Why is this not a valid metadata value?

qiita_pet/test/test_qiita_redbiom.py
```diff
@@ -0,0 +1,161 @@
...
+                           'is a valid metadata value?'), 'data': []}
+        self.assertEqual(loads(response.body), exp)
+
+    def test_post_observations(self):
+        post_args = {
+            'search': ('4479944'),
```
josenavas: Not sure why these values are in parenthesis? They are not needed. (Applies to all)

qiita_pet/test/test_qiita_redbiom.py
```diff
@@ -0,0 +1,161 @@
...
+        self.assertEqual(loads(response.body), exp)
+
+        post_args = {
+            'search': '4353076',
+            'search_on': 'categories'
+        }
+        response = self.post('/redbiom/', post_args)
+        self.assertEqual(response.code, 200)
+        exp = {'status': 'success',
+               'message': ('Not a valid search: "4353076", are you sure this '
```
josenavas: Same question as above - why it is not valid?

scripts/qiita-test-install
```diff
@@ -395,6 +430,8 @@ def main():
         ipython_running = False
         extra_info = ('IPython test failed, perhaps due to an old MOI '
                       'version, please review')
+
+    print qiita_config_header
```
josenavas: Remove print?

qiita_pet/handlers/qiita_redbiom.py
```diff
@@ -0,0 +1,163 @@
+from tornado.gen import coroutine, Task
+
+from qiita_core.util import execute_as_transaction
+
+from .base_handlers import BaseHandler
+from requests import ConnectionError
+import redbiom.summarize
```
wasade: `redbiom.__init__` does not import the other modules. This is not unusual:

```python
In [1]: import scipy

In [2]: scipy.sparse.csr_matrix
AttributeError                            Traceback (most recent call last)
<ipython-input-2-c4d1267aee28> in <module>()
----> 1 scipy.sparse.csr_matrix

AttributeError: module 'scipy' has no attribute 'sparse'

In [3]: import scipy.sparse

In [4]: scipy.sparse.csr_matrix
Out[4]: scipy.sparse.csr.csr_matrix
```

qiita_pet/templates/redbiom.html
```diff
@@ -0,0 +1,175 @@
...
+{% from future.utils import viewitems %}
+{% extends sitebase.html %}
+
+{%block head%}
+<link rel="stylesheet" href="{% raw qiita_config.portal_dir %}/static/vendor/css/select2.min.css" type="text/css">
+<script type="text/javascript" src="{% raw qiita_config.portal_dir %}/static/vendor/js/select2.min.js"></script>
+<script type="text/javascript" src="{% raw qiita_config.portal_dir %}/static/js/sharing.js"></script>
+<script type="text/javascript">
+  $(document).ready(function() {
+    moi.init(null, window.location.host + '{% raw qiita_config.portal_dir %}/study/list/socket/', function(){}, error, error);
```
antgonza: that's defined in qiita.js ...

qiita_pet/templates/redbiom.html
```diff
@@ -0,0 +1,175 @@
...
+      'footerCallback': function ( row, data, start, end, display ) {
+        var api = this.api(), data;
+        var total_samples = api
+          .column(4)
+          .data()
+          .reduce( function (a, b) { return a + b.samples.length; }, 0);
+
+        $('.footer').addClass("col-md-12 text-right");
+        if (total_samples == 0) {
+          text = ''
```
antgonza: k

qiita_pet/templates/redbiom.html
```diff
@@ -0,0 +1,175 @@
...
+        } else {
+          if ($("#search_on").val() == 'categories'){
+            text = 'Found ' + total_samples + ' categories.';
```
antgonza: k

qiita_pet/templates/redbiom.html
```diff
@@ -0,0 +1,175 @@
...
+          "search": "Filter results by column data (Study Name, Artifact Name, Software, etc):",
+          "loadingRecords": "Please wait - loading information ...",
+          "zeroRecords": "  "
+      },
+  });
+
+    $("#submitForm").submit(function(event){
+      event.preventDefault();
+
+      show_loading("redbiom-info")
```
antgonza: k

qiita_pet/templates/redbiom.html
antgonza: right!

qiita_pet/templates/redbiom.html
antgonza: I like GO! but changing ...

qiita_pet/templates/sitebase.html
```diff
@@ -139,6 +142,20 @@
...
+       *
+       * Function to show the loading gif in a given div
+       *
+       * @param div_name string with the name of the div to populate with the loading gif
+       *
+       * This function replaces the content of the given div with the
+       * gif to show that the section of page is loading
+       *
+       */
+      function show_loading(div_name) {
```
antgonza: k

qiita_pet/test/test_qiita_redbiom.py
antgonza: k

qiita_pet/test/test_qiita_redbiom.py
antgonza: cause it's not a stem ...

qiita_pet/test/test_qiita_redbiom.py
antgonza: you are right but they don't hurt or aren't flake8 complaint ... so?

qiita_pet/test/test_qiita_redbiom.py
antgonza: also not a stem ... I guess that it will become valid if a metadata category or value is 4353076, @wasade?

scripts/qiita-test-install
antgonza: it's actually needed so it's displayed, see line https://github.com/biocore/qiita/pull/2118/files#diff-bb7154348182dace785aedbad03c3a2bL341

qiita_pet/handlers/qiita_redbiom.py
```diff
@@ -0,0 +1,163 @@
...
+
+    @execute_as_transaction
+    def _redbiom_search(self, query, search_on, callback):
+        error = False
+        message = ''
+        results = []
+
+        try:
+            df = redbiom.summarize.contexts()
+        except ConnectionError:
```
josenavas: Ok, just wanted to make sure that this was the desired behavior.

qiita_pet/static/js/qiita.js
josenavas: Thanks!

qiita_pet/templates/redbiom.html
josenavas: oh - thanks!

qiita_pet/templates/sitebase.html
```diff
@@ -46,8 +46,11 @@
     <script src="{% raw qiita_config.portal_dir %}/static/vendor/js/chosen.jquery.min.js"></script>
     <script src="{% raw qiita_config.portal_dir %}/static/vendor/js/jquery.dataTables.min.js" type="text/javascript"></script>
     <script src="{% raw qiita_config.portal_dir %}/static/vendor/js/vue.min.js"></script>
+    <script src="{% raw qiita_config.portal_dir %}/static/vendor/js/moi.js"></script>
+    <script src="{% raw qiita_config.portal_dir %}/static/vendor/js/select2.min.js"></script>
     <script src="{% raw qiita_config.portal_dir %}/static/js/qiita.js"></script>
 
+    <script src="{% raw qiita_config.portal_dir %}/static/vendor/js/moi.js"></script>
```
josenavas: This one is added twice (see line 49)

qiita_pet/templates/redbiom.html
```diff
@@ -0,0 +1,175 @@
+{% from qiita_core.qiita_settings import qiita_config %}
+{% from future.utils import viewitems %}
+{% extends sitebase.html %}
+
+{%block head%}
+<link rel="stylesheet" href="{% raw qiita_config.portal_dir %}/static/vendor/css/select2.min.css" type="text/css">
```
josenavas: Just checked - `select2.min.js` it is added in `sitebase.html` but the css is not. I would suggest removing them from here and adding the css in the `sitebase.html` file. Since the JS is added there, somewhere in the code it is expected to be available but currently is not, which I can generate some problems down the line.

On the other hand, `sharing.js` is ok to leave here since it is only used here.

qiita_pet/test/test_qiita_redbiom.py
josenavas: In order to allow this to move forward, can this be added as a TODO item (even if it is only a discussion)? I don't understand why a number will not be a valid stem, or why searching over numerical values is not allowed. However, I don't want to block this PR for this discussion.

qiita_pet/templates/redbiom.html
antgonza: k, removing all duplicated imports.

qiita_pet/templates/sitebase.html
antgonza: k

qiita_pet/test/test_qiita_redbiom.py
antgonza: k, will do, now numerical values are find but have to be inclosed in a where, like where age_cat > 40

qiita_pet/test/test_qiita_redbiom.py
wasade: not exactly. we intentionally do not index things which look like numbers as i could not determine a reasonable strategy for determining when a number was or was not a good thing to index. in general, it is not and it would have very large impact on the key space. however, as a downside, important numbers like host taxon ID are not index. but, if a user knows they want to search for host taxon ID, they can still do that with `where host_taxid=foo`

qiita_pet/test/test_qiita_redbiom.py
wasade: @josenavas, that would mean we'd index every single pH value, elevation, lat/long, etc in the absence of context.

qiita_pet/test/test_qiita_redbiom.py
antgonza: OK, thanks for the explanation. I'll coordinate with you to write some of these explanations and add them to the instructions.

qiita_pet/test/test_qiita_redbiom.py
josenavas: Thanks for the explanation @wasade , that makes sense.

Issue #2069, opened 2017-02-02
If the study is public, the "share" button is not shown


josenavas:
But it should appear since you may want to add some collaborator to "expand" the study.

PR #2122, opened 2017-04-29
public studies are being shown in the user own studies

antgonza:
This fixes: 
- https://github.com/biocore/qiita/issues/2069, now the public study is shown in the owner/shared section so the button is automatically added
- and the idea that the owner's studies should always be in their own list even if public but for other users it should be in the public list

Note that these changes will make a lot of the tests fail but want to see which are they ...


antgonza:
This is ready for review.

PR #2123, opened 2017-04-29
Improving study list speed

antgonza:


antgonza:
After looking at https://github.com/biocore/qiita/pull/2122, I realized that perhaps the reason that the listing page is so slow is because we are retrieving all sample info for that listing and then displaying it in the GUI -- looping twice in a really large dataset. The original reason we had it like this was because we wanted to filter out certain samples based on the searching engine. Currently, we don't have that section open and redbiom should be used for that feature. Thus, decided that perhaps changing the current behavior to only return the artifact info and id should suffice. 

Now, to support this idea I decided to run some test in the test-env:
1. Original code without generating sample information:
```
In [17]: %timeit generate_study_list([s.id for s in studies], False, False)
1 loops, best of 3: 233 ms per loop

In [18]: %timeit generate_study_list([s.id for s in studies], False, True)
1 loops, best of 3: 232 ms per loop
```

2. With adding sample information:
```
In [19]: %timeit generate_study_list([s.id for s in studies], True, True)
1 loops, best of 3: 3.43 s per loop

In [20]: %timeit generate_study_list([s.id for s in studies], True, False)
1 loops, best of 3: 3.89 s per loop
```

3. New code always generating the artifact info - not the sample info:
```
In [4]: %timeit generate_study_list([s.id for s in studies], True)
1 loops, best of 3: 264 ms per loop

In [5]: %timeit generate_study_list([s.id for s in studies], False)
1 loops, best of 3: 260 ms per loop
```

Note that this PR to be finished we will need to:
- [ ] change GUI + handlers to be able to allow artifact addition vs. sample addition to the analysis
- [ ] fix the tests

Please contact me if you want to discuss further ....

antgonza:
This is ready for review and it's deployed in the test environment. 

The new time is:
```
In [3]: %timeit qdb.util.generate_study_list([s.id for s in qdb.study.Study.get_by_status('public')], False)
1 loops, best of 3: 513 ms per loop
```
Remember that the original time was >3 secs and now we are also displaying the software version and the artifact name so it's easier to select your artifact.

qiita_pet/handlers/study_handlers/artifact.py
```diff
@@ -107,6 +107,16 @@ def patch(self):
         self.write(response)
 
 
+class ArtifactGetSamples(BaseHandler):
+    @authenticated
+    def get(self):
+        aids = map(int, self.request.arguments.get('ids[]', []))
```
josenavas: is this equivalent to `aids = map(int, self.get_arguments('ids'))`? I'm just trying to understand the differences of using one or another if they're not the same.

antgonza: `self.get_arguments`: retrieves one value
`self.request.arguments.get`: retrieves a list

josenavas: Note that I used [self.get_arguments](http://www.tornadoweb.org/en/stable/web.html#tornado.web.RequestHandler.get_arguments) instead of [self.get_argument](http://www.tornadoweb.org/en/stable/web.html#tornado.web.RequestHandler.get_argument). According to the documention it should return a list of parameters.

antgonza: ohh, got it, worth changing? IMOO no, note that we use self.request.arguments.get in one other place ...

josenavas: No need to change. I used `self.get_arguments` heavily on labadmin and I was wondering if you knew about an issue that I didn't.

qiita_pet/handlers/study_handlers/tests/test_artifact.py
```diff
@@ -140,6 +140,33 @@ def test_delete_artifact(self):
         wait_for_prep_information_job(1)
 
 
+class ArtifactGetSamplesTest(TestHandlerBase):
+    def test_get(self):
+        response = self.get('/artifact/samples/', {'ids[]': [4, 5]})
+        self.assertEqual(response.code, 200)
+        exp = (
+            '{"status": "success", "msg": "", "data": {"4": ['
```
ElDeveloper: This test might fail if the order of the samples or of the keys changes, a better way to test this is to load the JSON string as a Python dictionary and assert on that object with `asserEqual`.

qiita_db/util.py
```diff
@@ -1408,6 +1405,12 @@ def generate_study_list(study_ids, build_samples, public_only=False):
                     LEFT JOIN qiita.artifact_type USING (artifact_type_id)
                     WHERE artifact_type='BIOM' AND
                         study_id = qiita.study.study_id) AS artifact_biom_ts,
+                (SELECT array_agg(name ORDER BY artifact_id)
```
ElDeveloper: This is rather odd but can you update the docstring's query with this new section? ` ¯\_(ツ)_/¯`

qiita_pet/handlers/study_handlers/tests/test_artifact.py
antgonza: thanks, note that we are certain it will not change cause that's exactly what's being created. Now, do you want me to change?

ElDeveloper: > we are certain it will not change

As far as I know, dictionary keys in python are not guaranteed to remain in the same order, so `artifact_get_prep_req` could return the exact same object but formatted differently, think of:

`'{"status": "success", "msg": "", "data": { ...'`

vs

`'{"msg": "", "status": "success", "data": { ...'`

While the object is the same, the test would fail, this wouldn't happen if the objects were loaded as dictionaries.

I would load the body of the response using `from json import loads` and store this exp either as a python dictionary or load it as well with `loads`.

antgonza: thanks for the explanation, however, this works fine in the tests in travis and my laptop ... thus, not sure if needs to be changed, should I?

ElDeveloper: I would recommend changing it, the fact that we haven't seen this fail yet doesn't mean that it can't fail as the assumption it relies on is not guaranteed to be consistent.

jdereus:
I don't respond often but would agree with Yoshiki.  I am not sure what tests have been performed.  Same always in travis?  Same on your laptop?  This should be examined further or changed to definitively ensure no failures.

JD

PR #2127, opened 2017-05-03
fixing download buttons show

antgonza:
Currently, the download bioms and raw data is always present, this will only show the buttons when there is data to download.

### DIFF
--- a/qiita_pet/handlers/api_proxy/studies.py
+++ b/qiita_pet/handlers/api_proxy/studies.py
@@ -105,6 +105,11 @@ def study_get_req(study_id, user_id):
     study_info['has_access_to_raw_data'] = study.has_access(
         User(user_id), True)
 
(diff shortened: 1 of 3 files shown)


qiita_pet/handlers/api_proxy/studies.py
```diff
@@ -105,6 +105,11 @@ def study_get_req(study_id, user_id):
     study_info['has_access_to_raw_data'] = study.has_access(
         User(user_id), True)
 
+    study_info['show_biom_download_button'] = 'BIOM' in [
+        a.artifact_type for a in study.artifacts()]
+    study_info['show_raw_download_button'] = bool([
```
ElDeveloper: This line would perhaps be a bit less confusing if instead of `bool` you used `any`. But this is rather minor.

PR #2132, opened 2017-05-12
Redbiom improving listing

antgonza:
Changes based on discussion during our 05/10/17 meeting. Basically, we are grouping by parent (trim, split, etc).

Note that there are different ways to do this: *(A)* changes in the db so it's "easy" to retrieve the default setting name from parents or any artifact, *(B)* do the matching in SQL or *(C)* python. *(A)* will need a lot of changes to the code and the db, and we will need B or C to make it happen as we will need to populate all current artifacts. Thus, decided that *(B)* or *(C)* were the option. Started with *(B)* and realized that all json operation in postgres (the processing parameters are stored as json) are not easy (or even possible) in any version below 9.4; in fact, greater of 9.5 will make life easier; our servers have 9.3. Even so, tried this way and the query was taking around 12 seconds from the 'infant' test (it originally took ~500ms) so discarded that version. Thus, implemented *(C)* and it's the one in this PR ('infant' test takes ~1s, not terrible).

Will deploy in the test environment in a few minutes.

### DIFF
--- a/qiita_pet/handlers/qiita_redbiom.py
+++ b/qiita_pet/handlers/qiita_redbiom.py
@@ -96,8 +96,7 @@ class RedbiomPublicSearch(BaseHandler):
                                     a.artifact_id
                             JOIN qiita.artifact_type at ON (
                                 at.artifact_type_id = a.artifact_type_id
-                                AND artifact_type = 'BIOM')
-                            ORDER BY artifact_id),
+                                AND artifact_type = 'BIOM')),
                          parent_query AS (
                             SELECT artifact_query.*,
                                 array_agg(parent_params) as parent_parameters
@@ -114,6 +113,7 @@ class RedbiomPublicSearch(BaseHandler):
                                 artifact_query.command_id,
                                 artifact_query.artifact_id)
                         SELECT * FROM parent_query
+                        ORDER BY parent_parameters, artifact_id
                         """
 
                         sql_params = """
--- a/qiita_pet/handlers/qiita_redbiom.py
+++ b/qiita_pet/handlers/qiita_redbiom.py
@@ -6,6 +6,7 @@
 # The full license is in the file LICENSE, distributed with this software.
 # -----------------------------------------------------------------------------
 
+from future.utils import viewitems
 from requests import ConnectionError
 import redbiom.summarize
 import redbiom.search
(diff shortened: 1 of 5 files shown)


qiita_pet/handlers/qiita_redbiom.py
```diff
@@ -83,42 +84,81 @@ def _redbiom_search(self, query, search_on, callback):
...
-                            name, command_id,
-                            (main_query.children).artifact_id AS artifact_id
-                        FROM main_query
-                        JOIN qiita.artifact a ON
-                            (main_query.children).artifact_id = a.artifact_id
-                        JOIN qiita.artifact_type at ON (
-                            at.artifact_type_id = a.artifact_type_id
-                            AND artifact_type = 'BIOM')
-                        ORDER BY artifact_id
+                            GROUP BY study_title, study_id, artifact_id),
```
wasade: would it be possible to get an english description of what this query does? inline comments within the query to describe the subqueries would also be helpful

qiita_pet/handlers/qiita_redbiom.py
```diff
@@ -83,42 +84,81 @@ def _redbiom_search(self, query, search_on, callback):
...
+                            # to have a huge growth in the near future
+                            qdbsc.TRN.add(sql_params)
+                            params = {pname: eval(params) for pname, params
+                                      in qdbsc.TRN.execute_fetchindex()}
+
+                            # now let's get the actual artifacts
+                            qdbsc.TRN.add(sql, [tuple(features)])
                             for row in qdbsc.TRN.execute_fetchindex():
-                                title, sid, samples, name, cid, aid = row
+                                title, sid, samples, name, cid, aid, pp = row
```
wasade: I dont understand how the select on `main_query` did not change but somehow `row` has an additional element now?

qiita_pet/handlers/qiita_redbiom.py
```diff
@@ -83,42 +84,81 @@ def _redbiom_search(self, query, search_on, callback):
...
+                                        commands[cid] = '%s - %s v%s' % (
+                                            c.name, c.software.name,
+                                            c.software.version)
+
+                                    # [-1] taking the last cause it's sorted by
+                                    #      the number of overlapping parameters
+                                    # [0] then taking the first element that is
+                                    # the name of the parameter set
+                                    ppc = sorted(
+                                        [[k, len(eval(pp) & v)]
```
wasade: what is being `eval`d?

qiita_pet/handlers/qiita_redbiom.py
```diff
@@ -83,42 +84,81 @@ def _redbiom_search(self, query, search_on, callback):
...
                                         c = qdb.software.Command(cid)
-                                        commands[cid] = {
-                                            'sfwn': c.software.name,
-                                            'sfv': c.software.version,
-                                            'cmdn': c.name
-                                        }
-                                    nr['command'] = commands[cid]['cmdn']
-                                    nr['software'] = commands[cid]['sfwn']
-                                    nr['version'] = commands[cid]['sfv']
+                                        commands[cid] = '%s - %s v%s' % (
```
wasade: should trim length be represented?

qiita_pet/handlers/qiita_redbiom.py
antgonza: added some information.

qiita_pet/handlers/qiita_redbiom.py
antgonza: the with is basically a select after another select, this allows us to make faster queries and "work" with json ... hopefully the explanation added helps.

qiita_pet/handlers/qiita_redbiom.py
antgonza: it has it, in the next block the processing parameters of the parent are added.

qiita_pet/handlers/qiita_redbiom.py
antgonza: psql return a set as a string so we are eval that ..

qiita_pet/static/js/qiita.js
```diff
@@ -301,17 +301,3 @@ function display_number_of_samples_added(num_samples) {
...
- *
- * Function to show the loading gif in a given div
- *
- * @param div_name string with the name of the div to populate with the loading gif
- *
- * This function replaces the content of the given div with the
- * gif to show that the section of page is loading
- *
- */
-function show_loading(div_name) {
```
josenavas: Why is this function removed from here?

josenavas: ok

PR #2136, opened 2017-05-22
Analysis refactor allow users change dflt params

josenavas:
Still working on it - issuing the PR so @antgonza can use the `mchoice` parameter type


josenavas:
Not sure why it is showing that many commits - anyways, the changes shown are mine. This PR allows the user to change the default parameter values in the analysis pipeline but not in the processing pipeline. The look&feel has also been improved by using bootstrap. Note that in the gif below, the commands are placeholders that I have on my DB so I can actually test the interface, so the parameters of those commands may not be correct.

processing

antgonza:
Is this ready to be reviewed? Not sure based on the initial comment ...

josenavas:
See the [comment above yours](https://github.com/biocore/qiita/pull/2136#issuecomment-303732993) ⬆️ 

Yes

qiita_pet/templates/artifact_ajax/processing_artifact.html
```diff
@@ -203,67 +213,140 @@
...
+
     var p_type = param_info[0];
     var allowed_types = param_info[1];
-    if (p_type == 'artifact') {
-      var sel = $('<select>').appendTo("#cmd-opts-div").attr('id', p_name).attr('name', p_name).addClass('required-parameter');
-      for(var key in sel_artifacts_info) {
-         if(allowed_types.indexOf(sel_artifacts_info[key].type) !== -1) {
-           sel.append($('<option>').attr('value', key).text(sel_artifacts_info[key].name));
-         }
+    var $inp;
```
antgonza: Is this needed? It seems like it gets overwrote as $inp = $('<select>'); or  $inp = $('<input>');

josenavas: This is the variable declaration, so the $inp variable is not global.

antgonza: get that but, is it better (I'm not sure, asking ...) to do var at the beginning of the 2 other lines?

josenavas: I'm not sure either which one is better, just I'm used to do this as a C programmer. If you feel strong about it I can change.

antgonza: If C, you need to add the data type 😄 , anyway, I think this makes sense: https://stackoverflow.com/a/11532092, so the problem is that it's undefined and if something happens to that code and the if/else is not there the next if will fail.

josenavas: Hahahaha, yeah I miss declaring types 😄 

I think the failure of the next if is desired, so we can catch that during development.

qiita_db/software.py
```diff
@@ -272,16 +274,25 @@ def create(cls, software, name, description, parameters, outputs=None,
...
+                dflt_val = dflt
+                if ptype.startswith('choice'):
+                    # In the choice case, the dflt value is a single string,
+                    # create a list with it the string on it to use the
+                    # issuperset call below
+                    dflt_val = [dflt_val]
+                else:
+                    # jsonize the list to store it in the DB
+                    dflt = dumps(dflt)
+                if not choices.issuperset(dflt_val):
```
wasade: should this check be done immediately after line 285?

qiita_pet/handlers/api_proxy/processing.py
```diff
@@ -62,7 +64,8 @@ def list_commands_handler_get_req(artifact_types):
     artifact_types = artifact_types.split(',')
     cmd_info = [
         {'id': cmd.id, 'command': cmd.name, 'output': cmd.outputs}
-        for cmd in Command.get_commands_by_input_type(artifact_types)]
+        for cmd in Command.get_commands_by_input_type(
+            artifact_types, exclude_analysis=exclude_analysis)]
```
wasade: I'm remain not jazzed about the arbitrary binary partitioning of commands but I think I'm out voted on this.

qiita_pet/templates/artifact_ajax/processing_artifact.html
```diff
@@ -203,67 +213,140 @@
...
+        $inp.attr('type', 'number').attr('step', 0.001);
+      }
+      else if (p_type == 'string') {
+        // For the float type, show an input of type text
+        $inp.attr('type', 'text');
+      }
+      else if (p_type == 'boolean') {
+        // For the boolean type, show an input of type checkbox
+        $inp.attr('type', 'checkbox');
+      }
```
wasade: can an `else` be added which excepts with a useful message?

qiita_pet/templates/artifact_ajax/processing_artifact.html
```diff
@@ -203,67 +213,140 @@
...
+    }
+
+    if (dflt_val !== undefined) {
```
wasade: an optional parameter must have a default value? that's unexpected

qiita_pet/templates/artifact_ajax/processing_artifact.html
```diff
@@ -292,33 +376,35 @@
       }
       sel_artifacts_info[node.id] = {'type': node.type, 'name': node.name}
     }
-    $.get('{% raw qiita_config.portal_dir %}/study/process/commands/', {artifact_types: types})
+    $.get('{% raw qiita_config.portal_dir %}/study/process/commands/', {artifact_types: types, exclude_analysis: !is_analysis_pipeline})
```
wasade: the use of the negative "exclude" makes this confusing, would it be better to refactor and describe it as `include_analysis`?

qiita_db/software.py
josenavas: Nope, line 291 makes sure that `dflt_val` is a list so this call works as expected.

qiita_pet/handlers/api_proxy/processing.py
josenavas: The "arbitrary" partitioning is matching the "upstream" and "downstream" commands that has been in Qiime for a long time. i.e. commands that you need to run to get a BIOM table, and commands that you run in a BIOM table to gain insight.

qiita_pet/templates/artifact_ajax/processing_artifact.html
josenavas: I've added a boostrapAlert which I think is the best way to "raise" this messages. This should never be reached and it should only be raised during development, but better safe than sorry!

qiita_pet/templates/artifact_ajax/processing_artifact.html
josenavas: All optional parameters have a default value - that's why they're optional (see the multiple available commands).

qiita_pet/templates/artifact_ajax/processing_artifact.html
josenavas: Done

wasade:
I can think of a large number of optional parameters with POSIX compliant
commands and which a default value does not make sense. Are you sure this
is correct behavior?

josenavas:
In such case we need to refactor the entire plugin structure. That is just the interface to what is done behind the scenes. The way an optional parameter is identified is if it has a default parameter or not.

josenavas:
Let me rephrase a bit more my last message. Although I do agree with you and potentially we need to revisit how that is done, I don't think it should block this PR. The point that you're raising is valid, but it is a bug (or feature :D) of the plugin system itself. The refactor of the analysis pipeline to use the plugin mechanism doesn't change the behavior of the plugin mechanism itself, so I would argue that opening an issue and fix later would be a better way of moving forward.

wasade:
There are examples in QIIME2 which are incompatible, so we're setting ourselves up for problems. For instance `--p-where` in `q2_feature_table.filter_samples` is optional and does not have a rational default. For a broader context, the `man` pages for the standard suite of Linux commands are rife with examples as well. However, I agree that we need progress here -- please create an issue. It is possible that we will not hit snags when wrapping external commands, but at the same time, it is very possible that we will. As such, please mark this as a recognized incompatibility with POSIX command argument structure in the issue. I'm inclined to suggest that this be marked as a priority to resolve sooner rather than later as it will only be harder to fix down the line if necessary.

josenavas:
Thanks @wasade - see https://github.com/biocore/qiita/issues/2140

josenavas:
I've marked the issue as priority high and it should be solved after the current work on the analysis refactor is completed.

wasade:
Thanks!

PR #2164, opened 2017-07-06
rm prep/sample constraints

antgonza:
This is what we discussed as a group to improve the delete speed.


antgonza:
gonna reopen against analysis-refactor

PR #2168, opened 2017-07-11
fix 55.py

antgonza:
While testing in the latest version of the DB the original code (which was also tested in the test env) failed due to lack of memory. These changes just make it use less memory in PgSQL.


ElDeveloper:
👍 thanks @antgonza!

PR #2170, opened 2017-07-17
Improving analysis listing

antgonza:


qiita_pet/templates/sitebase.html
```diff
@@ -54,6 +54,14 @@
     <link rel="shortcut icon" href="{% raw qiita_config.portal_dir %}/static/img/favicon.ico">
 
     <script>
+      $(window).resize(function () {
+        $('body').css('padding-top', parseInt($('#navigation-bar').css("height"))+10);
```
ElDeveloper: [From reading the jQuery docs](http://api.jquery.com/height/), I believe you want to use the `height` function directly i.e. `$('#navigation-bar').height()`, this way you don't have to worry about the units of this attribute, as it could return `500 px`, and you also don't need the `parseInt`.

qiita_pet/webserver.py
```diff
@@ -106,6 +106,7 @@ def __init__(self):
             (r"/analysis/description/(.*)/graph/", AnalysisGraphHandler),
             (r"/analysis/description/(.*)/jobs/", AnalysisJobsHandler),
             (r"/analysis/description/(.*)/", AnalysisDescriptionHandler),
+            (r"/analysis/sharing/", ShareAnalysisAJAX),
```
ElDeveloper: Minimal, but I think `sharing` should be `share`  ... ` ¯\_(ツ)_/¯`

qiita_pet/handlers/analysis_handlers/tests/test_sharing_handlers.py
```diff
@@ -0,0 +1,59 @@
...
+    def test_get(self):
+        a = Analysis(1)
+        u = User('shared@foo.bar')
+        self.assertEqual(a.shared_with, [u])
+
+        # deselecting
+        args = {'deselected': u.id, 'id': a.id}
+        response = self.get('/analysis/sharing/', args)
+        self.assertEqual(response.code, 200)
+        exp = {'users': [], 'links': ''}
```
ElDeveloper: When analyses are unshared the output data should be either empty or include a "completed" message. We were doing this for the portion of the REST API that @wasade and I worked on. Seemed clean and you don't really need an empty list of `users` or `links`. What do you think?

qiita_pet/templates/list_analyses.html
```diff
@@ -2,9 +2,15 @@
...
 <script type="text/javascript">
+
+
   $(document).ready(function(){
     {% if analyses %}
       $('#no-analyses').hide()
+      $('#analyses-table').dataTable({
+        "order": [[ 3, "desc" ]]
+      });
+      $('#analyses-table').DataTable();
```
ElDeveloper: I'm confused about this, but isn't this call to DataTable() overriding the settings set in the previous lines? Or do you need these two?

qiita_pet/templates/sitebase.html
```diff
@@ -54,6 +54,14 @@
...
+      });
+
+      $(window).load(function () {
+        $('body').css('padding-top', parseInt($('#navigation-bar').css("height"))+10);
```
ElDeveloper: Same as above.

qiita_pet/handlers/analysis_handlers/tests/test_sharing_handlers.py
antgonza: I think that's a fair point and perhaps worth it. Note that this will imply that we will need to have 2 calls for the same functionality as right now the call actually also changes the display so the new code will have to be: delete/add user, request current values. 

Now, based on (1) this code/mechanism is based on Artifact sharing, which means that we will need to change both to be complete and (2) that what I did is simply return these deleted lines from this branch + add some tests; I  also believe that this is out of the scope of this PR and opening an issue for further discussion is the best path forward.  See https://github.com/biocore/qiita/issues/2171.

qiita_pet/templates/list_analyses.html
antgonza: Funny enough it doesn't as we are not changing that value but I also realized that it will be better to sort by 2 (timestamp), changing.

qiita_pet/templates/sitebase.html
antgonza: didn't know, much cleaner, changing.

qiita_pet/webserver.py
antgonza: adding to the review as this also means changing the artifact sharing code

qiita_pet/handlers/analysis_handlers/tests/test_sharing_handlers.py
ElDeveloper: I agree, sounds good!

ElDeveloper:
Thanks @antgonza!

PR #2175, opened 2017-07-18
Improving analysis listing b

antgonza:


qiita_pet/templates/list_analyses.html
```diff
@@ -8,8 +8,78 @@
...
-        "order": [[ 2, "desc" ]]
+        "order": [[ 2, "desc" ]],
+        columnDefs: [
+          {"render": function ( data, type, row, meta ) {
+            data = JSON.parse(data);
+            var len = 0;
+            if (data !== null && data !== undefined){
+              len = data.length;
+             }
+            if (len != 0){
```
ElDeveloper: Minor, this could be short-circuited as: `if (data !== null && data !== undefined && data.length !=0)`

qiita_pet/handlers/study_handlers/tests/test_prep_template.py
```diff
@@ -59,6 +60,7 @@ def test_get(self):
 
 class TestPrepTemplateAJAXReadOnly(TestHandlerBase):
     def test_get(self):
+        sleep(1)
```
ElDeveloper: Why is this `sleep` needed?

qiita_pet/templates/list_analyses.html
antgonza: k

qiita_pet/handlers/study_handlers/tests/test_prep_template.py
antgonza: trying to make the test pass ... removing now

qiita_pet/templates/list_studies.html
```diff
@@ -204,10 +201,7 @@
         // render zero
         {"render": function ( data, type, row, meta ) {
           var len = 0;
-          if (data !== null && data !== undefined){
```
ElDeveloper: Great!! Also, we no longer need the line above `var len = 0;` 👍

qiita_pet/templates/list_studies.html
```diff
@@ -131,10 +131,7 @@
...
```
ElDeveloper: Same as below.

PR #2180, opened 2017-07-27
misc 2 and fix #2111

### LINKED ISSUE #2111: Can't remove all tags from a study
@gilsharon reported this bug. You can change, update, etc. but once you added a tag the study always requires a tag.

Note that this should be an easy fix as it's allowed by the backend but the issue comes from the JS, can't submit if nothing is in the tags block.

antgonza:
while testing the best way to delete within the new analysis branch we added the  enable/disable triggers but they are not needed any more (which also make the commands fail) so deleting + a small tag fix.


tkosciol:
@antgonza I don't really understand what this PR is trying to do. Maybe someone else would be more suitable to review?

PR #2182, opened 2017-07-29
fix #1526

### LINKED ISSUE #1526: too large SQL?
While trying to update the full AGP we hit an error due to the size of the SQL. Now, this could be seen as an issue from our side cause we are generating a large SQL or psycopg2 cause is the library that we use to connect and execute PgSQL queries. 

The error simply reports something that doesn't make sense, in this case a casting error, but if we limit the number of samples being updated/extended it works fine.

antgonza:


scripts/qiita
```diff
@@ -460,12 +460,16 @@ def start(port, master):
             # registering definition artifacts
             for s in sdefinition:
                 s.register_commands()
-                sleep(0.5)
+                # sleep is required as we don't want to saturate workers with
+                # lots of petitions
```
wasade: repetitions?

antgonza: really requests ... so gonna change, thanks!

Issue #2112, opened 2017-04-19
filter list by multiple tags


antgonza:
Suggestion by @gilsharon. Currently, we only can filter the list by one tag.

Issue #2114, opened 2017-04-20
Change the tag list on the study page list to be a multiple select2 dropdown


josenavas:
As seen below, the list is growing, which can be annoying once we get more tags. Using something like a [multiple selec2 dropdown]( https://select2.github.io/examples.html#select-boxes-with-labels
) can be useful. I'm unsure if this can be done directly on tag-it. 

![screen shot 2017-04-20 at 11 02 56 am](https://cloud.githubusercontent.com/assets/2501478/25245363/00cad8c6-25b9-11e7-9c5f-7a875d4c68c2.png)

antgonza:
closed via: https://github.com/biocore/qiita/pull/2183

PR #2183, opened 2017-08-01
improving tags

antgonza:
turns out that: `$('#shares-select').attr('data-current-id', id);` is actually required in`modify_sharing`, if not the share doesn't work AKA is broken. While doing this decided to also fix https://github.com/biocore/qiita/issues/2112 and https://github.com/biocore/qiita/issues/2114, see below.

tags


ElDeveloper:
👍 code looks good to me!

PR #2192, opened 2017-08-03
fix #2083

### LINKED ISSUE #2083: Allow to re-submit a given prep info to EBI
Currently the buttons disappears, but admins should be able to re-submit to EBI in order to update the metadata, for example.

antgonza:


ElDeveloper:
This looks good. One question, will the internal EBI code handle the resubmission as an update?

antgonza:
excellent question, yes, if you try to resubmit it will raise an error but if you "modify" it will do an update. in fact, this is why we want to add cause the button is not there and some times we need to update ...

PR #2194, opened 2017-08-03
fix #1521

### LINKED ISSUE #1521: Remove `filepath_table` from the `insert_filepaths` function signature
This table is always `qiita.filepath`, but that was not clear at the moment that the function was created. This parameter is always passed and always with the same value. We have a single table to hold all the file paths (and this was by design) so we can safely remove this parameter from the function signature and update the code accordingly.

antgonza:


qiita_db/util.py
```diff
@@ -579,9 +578,7 @@ def insert_filepaths(filepaths, obj_id, table, filepath_table,
         Id of the object calling the functions. Disregarded if move_files
         is False
     table : str
-        Table that holds the file data.
-    filepath_table : str
-        Table that holds the filepath information
+        Table that holds the file data
```
ElDeveloper: Minor, but you accidentally deleted a period at the end of this sentence. :neckbeard:

antgonza: in fact, it was on purpose cause no other line has a period and Atom highlighted it funny ... I can add back if you want.

ElDeveloper: 👍

Issue #2188, opened 2017-08-02
Remove processing options that are not possible to run


adswafford:
Example 1: select distance matrix, generate HTML summary, doesn't do anything and can't select parameter set
Example 2: alpha and beta options appear in the dropdown for non-rarefied biom artifacts
Example 3: taxa-plots appear for deblur artifacts

If we remove these options, it's possible that https://github.com/biocore/qiita/issues/2187 might not occur any more

antgonza:
This is an interesting one because it will mean that we make the analysis plugin know about the types we are dealing with, more below. In theory, we will not like to do this because of plugin/qiita independence. However, we are currently doing something like this to merge the bioms correctly. Gonna add the group-discussion tag so we can discuss. 

Ex1: this one I'm not sure why is happening but I realized that the command didn't have outputs so I'm gonna add a check for this in a PR.
Ex2: this was the intended behavior as this is a bug in Qiime2. Note that this has been solved in the latest version AKA it will be solved once we update but we need to coordinate how/when.
Ex2: same than the other ones, this is an issue with the processing and it will need a way to know the characteristics of the biom to display the command aka we need to discuss

antgonza:
Resolutions:
Ex1: @josenavas mentioned that perhaps the command was wrongly created in the plugin and will take a look. However, @antgonza did a PR with the fix and it's not super urgent
Ex2: we will do a deploy of qiime2 just after we test all the steps in the new qiita-test env
Ex3: this is a combination of showing less information to the user - just the error, adding a dictionary of possible errors, and fixing errors in other tools.

antgonza:
We need to discuss when we will update to the new Qiime2 ...

josenavas:
Include useful error message. Deploy the next release of Qiime 2 on the release candidate machine to test if we can upgrade the live system.

antgonza:
On 11/01/17 we agreed that this was solved.

PR #2200, opened 2017-08-04
partial fix -- 2188

antgonza:
While looking at #2188, I realized that we had old code that wasn't being used left behind so removing it and also adding the check for outputs described on that issue.


ElDeveloper:
👍

PR #2202, opened 2017-08-05
ENH: Move "Jobs using this data" button

ElDeveloper:
The button appeared at the bottom of the page, which makes it hard to
find this information when dealing with "per-sample FASTQ files".
With this commit, this button is moved to the top of the file download
list.


ElDeveloper:
Thanks!

PR #2203, opened 2017-08-08
Remove race condition - Fixes 2143

### LINKED ISSUE #2143: Possible race condition when completing jobs
If a job has more than one validator, sometimes the validators are set to "waiting" but the job never finish (i.e. deadlock). My guess is that the race condition is happening in [this call](https://github.com/biocore/qiita/blob/master/qiita_db/processing_job.py#L382). If the validators are executing that call at the same time, they will not see that the other job is already in the waiting status because the entire operation is executed inside a transaction, so both jobs end up waiting for the other job to complete.

Possible solutions that I can think of:
1. Instead of performing the counting by checking how many jobs are not in "waiting", put a specific counter with a lock in which serialization of operations is ensured by using a "[ROW EXCLUSIVE](https://www.postgresql.org/docs/9.1/static/explicit-locking.html))" lock.
2. Explicitly commit the change of status before performing that check. The amount of code changes may be higher than it may seem as there are multiple nested contexts that creates transactions, and enforcing a commit may break some other unexpected functionality.

I think (1) is the more correct and easy solution although it requires DB changes.

josenavas:
Fixes #2143 

Adds a new internal command: `release_validators`. The validators themselves no longer try to free up the parent job. The `release_validators` is a command that performs an active polling checking for the jobs to complete. It is automatically submitted to the queue after the validators have been submitted.

### DIFF
--- a/qiita_db/processing_job.py
+++ b/qiita_db/processing_job.py
@@ -421,8 +421,9 @@ class ProcessingJob(qdb.base.QiitaObject):
                     "Only artifact transformation and private jobs can "
                     "release validators")
 
-            # Check if all the validators are ready by checking that there is
-            # no validator processing job whose status is not waiting
+            # Check if all the validators are completed. Validator jobs can be
+            # in two states when completed: 'waiting' in case of success
+            # or 'error' otherwise
             sql = """SELECT COUNT(1)
                      FROM qiita.processing_job_validator pjv
                         JOIN qiita.processing_job pj ON
--- a/qiita_db/processing_job.py
+++ b/qiita_db/processing_job.py
@@ -653,6 +653,16 @@ class ProcessingJob(qdb.base.QiitaObject):
             for j in validator_jobs:
                 j.submit()
 
+            # Submit the job that will release all the validators
+            plugin = qdb.software.Software.from_name_and_version(
+                'Qiita', 'alpha')
+            cmd = plugin.get_command('release_validators')
+            params = qdb.software.Parameters.load(
+                cmd, values_dict={'job': self.id})
+            job = ProcessingJob.create(self.user, params)
+        # Doing the submission outside of the transaction
+        job.submit()
+
     def _set_validator_jobs(self, validator_jobs):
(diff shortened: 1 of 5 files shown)


antgonza:
@josenavas, just FYI, the PR actually has failed tests.

josenavas:
Ready for review!

qiita_db/processing_job.py
```diff
@@ -429,12 +430,49 @@ def release_validators(self):
                         JOIN qiita.processing_job_status USING
                             (processing_job_status_id)
                      WHERE pjv.processing_job_id = %s
-                        AND processing_job_status != %s"""
-            qdb.sql_connection.TRN.add(sql, [self.id, 'waiting'])
+                        AND processing_job_status NOT IN %s"""
+            sql_args = [self.id, ('waiting', 'error')]
```
antgonza: based on the description it sounds like here it should only be waiting, right? If this is incorrect, what happens with erred jobs?

qiita_db/test/test_processing_job.py
```diff
@@ -476,7 +483,8 @@ def test_complete_success(self):
 
         obsjobs = set(self._get_all_job_ids())
 
-        self.assertEqual(len(obsjobs), len(alljobs) + 1)
+        # The complete submits the release validators job
```
antgonza: Could you add an explanation of why + 2?

qiita_db/processing_job.py
josenavas: I will add a comment explaining this, but basically here we want to wait until all validators are completed. Validator jobs can be in two states when completed: 'waiting' -> success and 'error' -> error

qiita_db/test/test_processing_job.py
josenavas: Done

josenavas:
I deployed this in the test environment and executed ~10 deblur jobs. All of them succeeded and I haven't seen the race condition occurring again. I'm pretty confident that the issue has been solved because conceptually the race condition has been removed.

PR #2241, opened 2017-08-16
fix #2229

### LINKED ISSUE #2229: Information summary for preps should have values collapsed
This would be consistent with the "Sample Information" view

antgonza:


wasade:
Had to assess with just the code? But if it looks the same as the sample information then i think the issue is satisfied

PR #2243, opened 2017-08-17
fix #2216

### LINKED ISSUE #2216: Date and time information should be displayed when artifact is selected
The information about when an artifact was created should be shown in the summary of the artifact in the processing and analysis workflows. This information is already displayed in the view studies tablewhen selecting biom artifacts to add to analyses so this should be easy to do.

antgonza:
Now it looks like this:
screen shot 2017-08-17 at 12 22 26 pm


adswafford:
Yep, looks good to me. Thanks!

PR #2244, opened 2017-08-18
delete-root-analysis-artifact

antgonza:


josenavas:
If I understand correctly, if the user deletes the last artifact there will be an empty analysis (i.e. no artifacts) so the user can delete it. Did you test how the interface behaves with this case? I'm unsure if it will just go smooth or it will fail somewhere else...

antgonza:
Yes, that's correct and it works fine ... I guess you didn't tested when you wrote it :smile:

PR #2246, opened 2017-08-20
Fix 2210

antgonza:
Adds sample counts by prep in the listing. Deciding to go via PrepTemplate.keys() cause it and cache the values as it saves us some code/queries. 

Now, the study list looks like this:
study_list

The study/sample selected like this:
selected


josenavas:
Thanks @tanaes 🍺

tanaes:
I inevitably make a bunch of 'wait I don't understand why...' comments on
these PRs but then spend a bunch of time scrolling back and forth on the
diffs and then I'm like 'ahhh oh I see.' :P

PR #2247, opened 2017-08-21
Fixes 1880

### LINKED ISSUE #1880: The warnings in the sample template do not get reseted
The warnings are getting appended rather than reset.

josenavas:
Fixes #1880
Basically the problem was in this function:

https://github.com/biocore/qiita/blob/a369cad63827b92e89032c8e12d5865c61f73a5b/qiita_ware/dispatchable.py#L198-L204

The serial calls to extend and update was forcing a validation on the "extend" call first (which raised the warning message) and another on the update, which it corrected the values in the DB (but didn't remove the previous warning).

By creating a function that performs those 2 operations at the same time and issuing a single validation at the end of that function we only raise a warning at the end if there is a problem with the data in the DB at the end of the operation.


qiita_db/metadata_template/base_metadata_template.py
```diff
@@ -1168,15 +1165,16 @@ def update(self, md_template):
 
             # In order to speed up some computation, let's compare only the
             # common columns and rows. current_map.columns and
-            # current_map.index are supersets of new_map.columns and
-            # new_map.index, respectivelly, so this will not fail
-            current_map = current_map[new_map.columns].loc[new_map.index]
+            # current_map.index are supersets of md_template.columns and
+            # md_template.index, respectivelly, so this will not fail
+            current_map = current_map[
+                md_template.columns].loc[md_template.index]
```
tanaes: Is there a reason to do this way instead of ` current_map = current_map.loc[md_template.index, md_template.columns]`?

josenavas: I don't think so. I know that sometimes one option is creating a view while the other one returning an actual pandas DataFrame object, and that can raise Warning in some operations. I don't think it actually matters here, but I wanted to minimize code modifications (I'm actually just moving code around in this PR to be able to reuse code easily).

wasade:
:+1:

josenavas:
Thanks @wasade / @tanaes - able to merge? (I don't want to do a self high-five 😄 )

tanaes:
I approved, just had the question for learning purposes. Not sure @wasade
has finished review? Can merge if so.

josenavas:
No problem @tanaes - Yes, @wasade gave the 👍  so you can merge. Thanks!

PR #2255, opened 2017-08-28
replace printable for try/except utf-8

antgonza:


qiita_db/metadata_template/test/test_util.py
```diff
@@ -147,12 +147,23 @@ def test_load_template_to_dataframe_lowercase(self):
...
+        replace = EXP_SAMPLE_TEMPLATE.replace(
+            'Test Sample 2', u'Test Sample\x962')
+        qdb.metadata_template.util.load_template_to_dataframe(
+            StringIO(replace))
+        # setting back
+        replace = EXP_SAMPLE_TEMPLATE.replace(
+            u'Test Sample\x962', 'Test Sample 2')
+        qdb.metadata_template.util.load_template_to_dataframe(
+            StringIO(replace))
+
```
ElDeveloper: Unless I am missing something obvious, this test is missing a "test". Is "not erroring" the test? If so, are there any checks that can be done on the returned value from `load_template_to_dataframe`?

antgonza: The test is that it not raises an error, AKA that it can be done `test_load_template_to_dataframe_non_utf8_error` has the test for the raise.

ElDeveloper: Ah I see, thanks! 👍

PR #2256, opened 2017-08-29
Adding branch description to Contributing

josenavas:
Adding a description of the branches to the CONTRIBUTING.md file


antgonza:
:+1:

CONTRIBUTING.md
```diff
@@ -17,6 +17,14 @@ always take into consideration how these new features affect users and whether
 or not adding a new section or document to the documentation (found under the
 `doc` folder) would be useful.
 
+###Repository branch structure
```
ElDeveloper: I just noticed that the `###` is not being rendered as a title, see [here](https://github.com/josenavas/QiiTa/blob/73dca6fc78ca29a70725667ca916f7464618acd0/CONTRIBUTING.md). Seems to be a problem with a missing space between `###` and the first letter in the title.

CONTRIBUTING.md
```diff
@@ -17,6 +17,14 @@ always take into consideration how these new features affect users and whether
...
+
+The Qiita repository contains three branches:
+
+* `master`: This branch reflects the code deployed in our [main Qiita server](http://qiita.microbio.me).
+* `dev`: This branch is the active development branch. All new Pull Requests should be performed against this branch.
+* `release-candidate`: This branch is used to freeze the code from the `dev` branch, so we can deploy in our test servers and exercise the code extensively before deploying in our main system.
```
ElDeveloper: Do we want to make the test server URL public? Also, would it be worth noting the timeline?

ElDeveloper: By timeline I mean how long each branch is alive for, when are freezes happening, etc.

jdereus:
I think our test server(s) need to remain internal.  At the least not published.
Should be fire walled to only allow certain access.

As the test server code should only be slightly ahead of production, functionality should be fairly similar in most cases.

josenavas:
Thanks @ElDeveloper !

I've fixed the rendering error and I've added a comment about timelines. I agree with @jdereus on keeping the URL of the test systems internal.

ElDeveloper:
> I think our test server(s) need to remain internal.  At the least not published.
Should be fire walled to only allow certain access.

Sounds good.

Thanks for making the changes @josenavas

PR #2257, opened 2017-08-29
Fix 2212

### LINKED ISSUE #2212: Textbox should take focus when renaming artifacts
When you click "Edit" to rename and artifact, users have to click in the textbox instead of just being able to type. Can we just have the textbox take focus?

josenavas:
Fixes #2212 

The text input now gets focused when opening the modal:

focus


ElDeveloper:
Looks good! Although outside of the scope of this PR maybe also worth it to make `enter` submit the change?

Issue #250, opened 2014-08-26
Add python 3 back in to travis CI


adamrp:
Removed because biom was not python 3 compatible (see biocore/biom#535), which is breaking the qiita build when testing with python 3.

However, when biom becaomse python 3 compatible, we should add it back in here, too.

adswafford:
Haha, so it looks like the py3 upgrade has been in the work for 3 years now... Flagging for input so we can prioritize appropiately.

antgonza:
Closed in the qiita 4.19 release to be deployed on May 15, 2019.

PR #2261, opened 2017-08-31
upgrading software

antgonza:
- open_file completely changed from skbio 0.3.0 to 0.4.2, the main changes are due to supporting both python 2.7 and 3. After discussing with @josenavas we decided that it was easier to simply reuse the code and then address once we solve https://github.com/biocore/qiita/issues/250.
- After making that decision we realized that open_file was already within our code in qiita_ware so decided to move it to qiita_db
- flatten doesn't exist anymore in skbio, the suggestion is to use `list(chain(*`, which we are doing here
- updated test so we use ipython 3.2.0


josenavas:
Tests are expected to fail, and this is going to a different development branch, which is a WIP branch. The vast majority of this changes are just moving code around so I will merge as everything looks fine.

PR #2262, opened 2017-08-31
r_client to qiita

antgonza:
this one is built on top of https://github.com/biocore/qiita/pull/2261, so review/merge that one first.


josenavas:
@antgonza can you pull from upstream/remove-moi so those changes go away?

Thanks!

PR #2264, opened 2017-08-31
rm create raw data

antgonza:
when I searched for create_raw_data, I realized that we had some leftover code, which I'm also deleting here.


josenavas:
@antgonza  Can you pull from upstream/remove-moi?

antgonza:
done.

PR #2267, opened 2017-08-31
Transfer copy raw data

josenavas:
Depends on #2265 so review/merge that one first.

Transfer the copy of the raw data to the internal plugin system.


qiita_ware/test/test_private_plugin.py
```diff
@@ -0,0 +1,40 @@
...
+        params = Parameters.load(cmd, values_dict=values_dict)
+        job = ProcessingJob.create(user, params)
+        job._set_status('queued')
+        return job
+
+    def test_copy_artifact(self):
+        job = self._create_job('copy_artifact',
+                               {'artifact': 1, 'prep_template': 1})
+
+        private_task(job.id)
```
antgonza: is it possible to add a test for success?

PR #2274, opened 2017-09-06
Transfer update delete templates

josenavas:
Port all the remaining functionality from moi to the internal Qiita plugins.

Also updates the interface to reflect these changes.


qiita_db/test/test_commands.py
```diff
@@ -216,7 +214,7 @@ def tearDown(self):
     def test_load_parameters_from_cmd_error(self):
         with self.assertRaises(qdb.exceptions.QiitaDBUnknownIDError):
             qdb.commands.load_parameters_from_cmd(
-                "test", self.fp, 20)
+                "test", self.fp, 20000)
```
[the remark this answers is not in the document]
josenavas: Yes, the command 20 exists now.

qiita_ware/test/test_private_plugin.py
```diff
@@ -114,6 +151,226 @@ def test_create_sample_template_nonutf8(self):
...
+
+        job = self._create_job('delete_sample_or_column',
+                               {'obj_class': 'SampleTemplate', 'obj_id': 1,
+                                'sample_or_col': 'unknown', 'name': 'column'})
+        private_task(job.id)
+        self.assertEqual(job.status, 'error')
+        self.assertIn('Unknown value "unknown". Choose between "samples" '
+                      'and "columns"', job.log.msg)
+
+    def test_complete_job(self):
```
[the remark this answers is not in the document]
josenavas: Consistency with the rest of the tests in this class. A new class will trigger an unnecessary reset of the DB. If you feel strong on splitting in multiple tests I can do that.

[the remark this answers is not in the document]
antgonza: Nop, just wondering, could you add some text on the top of the test, explaining this?

qiita_db/test/test_commands.py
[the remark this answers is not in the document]
antgonza: Ha!

josenavas:
Changes made. Now the errors are shown like:
```
Error (log id: 3): Prep template 1 already has an artifact associated
```
where the number in the log id is another entry in the LogEntry table that contains the entire traceback and can be used by developers.

antgonza:
Thanks for the changes. Found one more while trying to delete a prep with artifacts: "Couldn't remove prep template: Cannot remove prep template 1 because it has an artifact associated with it". Note that trying to erase a sample info file is fine.

josenavas:
I don't understand what is the issue with that comment?

antgonza:
That the message is confusing: "Couldn't remove prep template: Cannot remove prep template 1 because it has an artifact associated with it" should only be "Cannot remove prep template 1 because it has an artifact associated with it"

josenavas:
oh, now makes sense :-) changed

qiita_db/processing_job.py
```diff
@@ -347,14 +325,19 @@ def submit(self):
...
+        with qdb.sql_connection.TRN:
+            status = self.status
+            if status not in {'in_construction', 'waiting'}:
+                raise qdb.exceptions.QiitaDBOperationNotPermittedError(
+                    "Can't submit job, not in 'in_construction' or "
+                    "'waiting' status. Current status: %s" % status)
+            self._set_status('queued')
+            # At this point we are going to involve other processes. We need
+            # to commit the changes to the DB or the other processes will not
+            # see this changes
```
ElDeveloper: this -> these

qiita_db/sql_connection.py
```diff
@@ -646,9 +646,12 @@ def _raise_execution_error(self, sql, sql_args, error):
         """
         self.rollback()
 
-        raise ValueError(
-            "Error running SQL: %s. MSG: %s\n" % (
-                errorcodes.lookup(error.pgcode), error.message))
+        try:
```
ElDeveloper: This block seems rather confusing, you `try` to find the error code, and check if that raises a `KeyError`, if it doesn't you show the error and the message, and if it does you just show the message. Are pgcodes not guaranteed to be in the errorcodes?

qiita_ware/private_plugin.py
```diff
@@ -145,26 +147,170 @@ def create_sample_template(job):
...
+        param_vals = job.parameters.values
+        study_id = param_vals['study']
+        fp = param_vals['template_fp']
+        with warnings.catch_warnings(record=True) as warns:
+            st = qdb.metadata_template.sample_template.SampleTemplate(study_id)
+            df = qdb.metadata_template.util.load_template_to_dataframe(fp)
+            st.extend_and_update(df)
+            remove(fp)
+
+            # Join all the warning messages into one.NOte that this info
```
ElDeveloper: one.NOte -> one. Note

qiita_db/sql_connection.py
```diff
@@ -937,3 +942,12 @@ def add_post_rollback_func(self, func, *args, **kwargs):
 
 # Singleton pattern, create the transaction for the entire system
 TRN = Transaction()
+
+
+def create_new_transacion():
```
ElDeveloper: transacion -> transaction

qiita_db/processing_job.py
josenavas: Done

qiita_db/sql_connection.py
josenavas: I found one case in which pgcodes was not in the error codes. It was mainly due to a bad usage of psycopg2 and the multiprocessing module. That is now fixed, but without this try/except I was unable to see the error message. Thus, in general they will be found, but just in case that in the future we find another issue, this will allow us to debug.

qiita_db/sql_connection.py
josenavas: Done

qiita_ware/private_plugin.py
josenavas: Done

josenavas:
Thanks @ElDeveloper - comments should be addressed now!

qiita_db/sql_connection.py
ElDeveloper: Wow, that's very strange!

PR #2282, opened 2017-09-13
Redbiom initial code

antgonza:
https://github.com/biocore/qiita/pull/2278 needs to be merged first.

Extra from adding redbiom:
- I rm qiita_pet/templates/analysis_waiting.html cause it's not used anywhere 
- show_loading was duplicated so just left one and made it work on all other pages


qiita_pet/handlers/qiita_redbiom.py
```diff
@@ -0,0 +1,117 @@
...
+                            query, 'value' if search_on == 'metadata' else
+                            'category'))
+                except ValueError:
+                    message = (
+                        'Not a valid search: "%s", your query is too small '
+                        '(too few letters), try a longer query' % query)
+
+                if message == '':
+                    config = redbiom.get_config()
+                    get = redbiom._requests.make_get(config)
```
wasade: this is a private module, it should not be necessary to call this method

qiita_pet/handlers/qiita_redbiom.py
```diff
@@ -0,0 +1,117 @@
...
+                        'Not a valid search: "%s", are you sure this is a '
+                        'valid metadata %s?' % (
+                            query, 'value' if search_on == 'metadata' else
+                            'category'))
+                except ValueError:
+                    message = (
+                        'Not a valid search: "%s", your query is too small '
+                        '(too few letters), try a longer query' % query)
+
+                if message == '':
```
wasade: can you describe in english what the goal of this block is? it looks like the intention is to establish a dict mapping a study, to the artifacts contained in that study that overlap with requested samples? if yes, wouldn't it be easier to just query the qiita API for that?

wasade: I guess I'm also a bit confused as to why the artifact IDs are necessary? I'm concerned with the subsequent code as it depends on the structure of redbiom IDs which we do not make guarantees about. If this is necessary (which I think can be handled by cross referencing with the Qiita API), can you please open an issue so that we can construct an API endpoint to retrieve the tags from a set of redbiom IDs?

qiita_pet/handlers/qiita_redbiom.py
```diff
@@ -0,0 +1,117 @@
...
+                                aid = rid.split('_', 1)[0]
+                                sid = i.split('.', 1)[0]
+                                study_artifacts[sid].append(aid)
+            elif search_on == 'feature':
+                query = [f for f in query.split(' ')]
+                for ctx in contexts:
+                    for idx in redbiom.util.ids_from(query, True, 'feature',
+                                                     ctx):
+                        aid, sid = idx.split('_', 1)
+                        sid = sid.split('.', 1)[0]
```
wasade: The sample IDs in qiita are not prefixed with the qiita study id?

qiita_pet/handlers/qiita_redbiom.py
```diff
@@ -0,0 +1,117 @@
...
+                        study_artifacts[sid].append(aid)
```
wasade: Okay, understanding this more. An API method should be exposed in redbiom which yields a the set of observed tags for a given iterable of redbiom IDs.

qiita_pet/handlers/qiita_redbiom.py
```diff
@@ -0,0 +1,110 @@
...
+                            study_artifacts[str(results[i]['study_id'])]))
+                else:
+                    message = "No samples where found! Try again ..."
+
+        callback((results, message))
+
+    @coroutine
+    @execute_as_transaction
+    def post(self, search):
+        search = self.get_argument('search', None)
```
josenavas: It looks like `search` and `search_on` are not really optional parameters.
I think the correct thing to do here is:
```python
search = self.get_argument('search')
search_on = self.get_argument('search_on')
```
and let tornado handle if the arguments are not passed. It will return a 400 error message which is the correct code given that the request is incorrect.

qiita_pet/static/js/qiita.js
```diff
@@ -115,6 +140,36 @@ function show_hide_process_list() {
   }
 }
 
+/*
```
josenavas: Documentation doesn't fit function signature. Can you update?

qiita_pet/static/js/qiita.js
```diff
@@ -115,6 +140,36 @@ function show_hide_process_list() {
...
+ * Note that we have a list of artifact ids cause the user can select one single
+ * artifact to add or all study artifacts
+ */
+function send_samples_to_analysis(button, aids) {
+  button.value = 'Adding';
+  button.disabled = true;
+  $(button).addClass("btn-info");
+  bootstrapAlert('We are adding ' + aids.length + ' artifact(s) to the analysis. This ' +
+                 'might take some time based on the number of samples on each artifact.', "warning", 10000);
+  $.get('/artifact/samples/', {ids:aids})
```
josenavas: Samples are added using a `get`? That is confusing... `get` should be use for read-only operations.

qiita_pet/templates/redbiom.html
```diff
@@ -0,0 +1,192 @@
...
+      columnDefs: [
+        // {type:'natural', targets:[2,6,7]},
+        {"targets": [ 2, 4 ], "visible": false},
+        // render zero
+        {"render": function ( data, type, row, meta ) {
+          if (data !== null && data !== undefined && data.length != 0){
+            {% if current_user is not None %}
+              return '<div class="container" style="max-width: 5em;">'+
+                       '<div class="row justify-content-md-center">' +
+                         '<div class="col-md-1 text-center details-control">&nbsp;</div>' +
```
josenavas: It looks like these two blocks are the same except for this line? You can then do:
```
              return '<div class="container" style="max-width: 5em;">'+
                       '<div class="row justify-content-md-center">' +
            {% if current_user is not None %}
                         '<div class="col-md-1 text-center details-control">&nbsp;</div>' +
            {% end %}
                         '<div class="col-md-1 text-center">' + data.length + '</div>' +
                       '</div>' +
                     '</div>';
```

qiita_pet/templates/redbiom.html
```diff
@@ -0,0 +1,192 @@
...
+      });
+    });
+  });
+</script>
+
+{%end%}
+
+{%block content%}
+  <small>
+    Redbiom only searches on public data and the data is updated nightly. Note that you will only be able to expand and add artifacts to analyses if you are signed into Qiita.
```
josenavas: Is this accurate? @wasade will it be possible to update the redis db every night?

wasade: We have not touched load mechanisms since June, and there is work to do there, but it should be feasible

qiita_pet/static/js/qiita.js
antgonza: that's not actually adding the samples, it's requesting the sample names from the artifacts to add. The addition is actually done in this line: `qiita_websocket.send('sel', data['data']);`

qiita_pet/templates/redbiom.html
antgonza: k

qiita_pet/templates/redbiom.html
antgonza: yes, that's a discussion for next week's meeting.

qiita_pet/static/js/qiita.js
josenavas: Thanks for the explanation!

PR #2286, opened 2017-09-14
fix #858

### LINKED ISSUE #858: Single study analysis shortcut
If we want qiita to replace qiime as a frontend, we need to be able to quickly do an analysis on a single study. If we have a way to jump from the study page directly to an analysis sample selection using only that study, it would speed up the analysis process.

antgonza:


qiita_pet/templates/artifact_ajax/artifact_summary.html
```diff
@@ -104,7 +107,11 @@
       <i id='summary-title'>{{name}}</i><i> (ID: {{artifact_id}}) Visibility: {{visibility}}</i>
       {% if editable %}
         <a class="btn btn-default btn-sm" data-toggle="modal" data-target="#update-artifact-name"><span class="glyphicon glyphicon-pencil"></span> Edit</a>
-        <a class="btn btn-default btn-sm" onclick="load_process_artifact_ui({{artifact_id}});"><span class="glyphicon glyphicon-play"></span> Process</a>
+        {% if artifact_type == 'BIOM' %}
+          <a class="btn btn-default btn-sm" onclick="send_samples_to_analysis(this, [{{artifact_id}}]);"> Add to analysis</a>
```
ElDeveloper: Will this display any indicator that the samples have been added to a study, like in the study list page?

antgonza: it uses the same code so it looks exactly the same!

ElDeveloper: 👍

Issue #2251, opened 2017-08-23
Add button to download per-sample EBI accession numbers


josenavas:

antgonza:
closed in https://github.com/biocore/qiita/pull/2287

PR #2287, opened 2017-09-14
fix #2258

### LINKED ISSUE #2258: admins don't have access to analysis
which makes it really hard to debug. The solution should be to add and if admin around here:
https://github.com/biocore/qiita/blob/master/qiita_pet/handlers/analysis_handlers/sharing_handlers.py#L52

antgonza:


antgonza:
by mistake I added the fixes for #2251 also here. I think this PR has an OK size so I'm gonna leave as this. 

BTW the study page looks like this:
study
and the prep:
prep

qiita_pet/handlers/download.py
```diff
@@ -222,3 +230,59 @@ def get(self, study_id):
...
+
+        text = "sample_name\texperiment_accession\n%s" % '\n'.join(
+            ["%s\t%s" % (k, v)
+             for k, v in viewitems(pt.ebi_experiment_accessions)])
+
+        self.set_header('Content-Description', 'text/csv')
+        self.set_header('Expires', '0')
+        self.set_header('Cache-Control', 'no-cache')
+        self.set_header('Content-Disposition', 'attachment; '
+                        'filename=ebi_sample_accessions_study_%s.tsv' % sid)
```
josenavas: Can you change the name of the file here so it's not the same as above? I would say something along the lines of `ebi_experiment_accessions_study_%s_prep_%s.tsv`

antgonza: good catch, I forgot to change that!

qiita_pet/handlers/download.py
```diff
@@ -222,3 +230,59 @@ def get(self, study_id):
         self.set_header('Content-Disposition',
                         'attachment; filename=%s' % zip_fn)
         self.finish()
+
+
```
ElDeveloper: These two handlers are doing pretty much the same, can this be factored in to a base class or in a helper function?

PR #2291, opened 2017-09-18
Fix 1293

antgonza:
This allows for full delete from the study, it will remove analyses, artifacts, prep templates, sample template and the study. This relies on the actual checks that each of those have to allow delete, which means is that it will only move forward if it can delete everything (not public, not submitted to EBI, etc). Before deleting the GUI will ask you to write the name of the study and reconfirm that you want to delete (2 steps). Note that I also wrote Study.analyses so we can easily get all analysis for a given study.


qiita_pet/templates/study_base.html
```diff
@@ -152,12 +152,7 @@
     else if(confirm("Are you sure you want to delete {% raw study_title_msg %}?")) {
       $.post('{% raw qiita_config.portal_dir %}/study/delete/', { study_id: {{study_info['study_id']}} })
         .done(function ( data ) {
-          if(data.status == "error") {
```
ElDeveloper: If there's an error with the deletion, the user will just see the page reload, would it reload to the same location? Meaning that the user wouldn't see any change?

qiita_db/study.py
```diff
@@ -998,6 +998,24 @@ def prep_templates(self, data_type=None):
             return [qdb.metadata_template.prep_template.PrepTemplate(ptid)
                     for ptid in qdb.sql_connection.TRN.execute_fetchflatten()]
 
+    def analyses(self):
+        """Get all analysis where samples from this study have been used
```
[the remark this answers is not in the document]
antgonza: k

qiita_pet/templates/study_base.html
antgonza: Excellent question! Well, the errors are now stored in a redis key as this is now gonna run as a job in the queue so the user will see the error on the reload.

PR #2294, opened 2017-09-18
Fix #2276

### LINKED ISSUE #2276: Trying to download a 'html summary dir' filepath type will fail
Because the download handler doesn't support directories ATM. It's possible to modify as it is done for the `DownloadStudyBIOMSHandler`

josenavas:
Fixes the download handler to support directories.

### DIFF
--- a/qiita_db/util.py
+++ b/qiita_db/util.py
@@ -845,6 +845,44 @@ def move_filepaths_to_upload_folder(study_id, filepaths):
         qdb.sql_connection.TRN.execute()
 
 
+def get_filepath_information(filepath_id):
+    """Gets the filepath information of filepath_id
+
+    Parameters
+    ----------
+    filepath_id : int
+        The filepath id
+
+    Returns
+    -------
+    dict
+        The filepath information
+    """
+    with qdb.sql_connection.TRN:
+        sql = """SELECT filepath_id, filepath, filepath_type, checksum,
+                        data_type, mountpoint, subdirectory, active,
+                        artifact_id
+                 FROM qiita.filepath
+                    JOIN qiita.filepath_type USING (filepath_type_id)
+                    JOIN qiita.data_directory USING (data_directory_id)
+                    LEFT JOIN qiita.artifact_filepath USING (filepath_id)
+                 WHERE filepath_id = %s"""
+        qdb.sql_connection.TRN.add(sql, [filepath_id])
+        res = dict(qdb.sql_connection.TRN.execute_fetchindex()[0])
+
+        def path_builder(db_dir, filepath, mountpoint, subdirectory, obj_id):
+            if subdirectory:
+                return join(db_dir, mountpoint, str(obj_id), filepath)
+            else:
+                return join(db_dir, mountpoint, filepath)
+
+        obj_id = res.pop('artifact_id')
+        res['fullpath'] = path_builder(get_db_files_base_dir(),
+                                       res['filepath'], res['mountpoint'],
+                                       res['subdirectory'], obj_id)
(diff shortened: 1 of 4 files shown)


qiita_pet/handlers/download.py
```diff
@@ -37,19 +38,42 @@ def get(self, filepath_id):
...
+
+            all_files = '\n'.join(
+                ["- %s /protected/%s %s" % (getsize(fp), sfp, n)
+                 for fp, sfp, n in to_download])
+
+            self.set_header('X-Archive-Files', 'zip')
+            self.write("%s\n" % all_files)
+            fname = '%s.zip' % fname
+        else:
+            # If we don't have nginx, write a file that indicates this
```
antgonza: A lot of this code is copy pastes from other classes, would you mind creating a new class from which these functions are inherit? Perhaps use the BaseHandlerDownload handler ...

josenavas:
@antgonza I factored out as much as code as I found possible.

antgonza:
thanks! Looks good to me except that `self.write("This installation of Qiita was not equipped with " ...` can be made into a single method, right? Currently is used in 2 different classes.

qiita_db/util.py
```diff
@@ -845,6 +845,44 @@ def move_filepaths_to_upload_folder(study_id, filepaths):
...
+                        artifact_id
+                 FROM qiita.filepath
+                    JOIN qiita.filepath_type USING (filepath_type_id)
+                    JOIN qiita.data_directory USING (data_directory_id)
+                    LEFT JOIN qiita.artifact_filepath USING (filepath_id)
+                 WHERE filepath_id = %s"""
+        qdb.sql_connection.TRN.add(sql, [filepath_id])
+        res = dict(qdb.sql_connection.TRN.execute_fetchindex()[0])
+
+        def path_builder(db_dir, filepath, mountpoint, subdirectory, obj_id):
```
wasade: Is a nested function necessary? It looks like it only adds complexity here...?

qiita_pet/handlers/download.py
```diff
@@ -77,6 +46,149 @@ def _generate_files(self, header_name, accessions, filename):
...
+            Path to the directory
+
+        Returns
+        -------
+        list of (str, str, str)
+            The path information needed by nginx for each file in the
+            directory
+        """
+        basedir = get_db_files_base_dir()
+        basedir_len = len(basedir) + 1
```
wasade: Is this, and its use below, basically the same as `os.dirname`? Same comment for line 89 and its use in the next function

qiita_pet/handlers/download.py
```diff
@@ -77,6 +46,149 @@ def _generate_files(self, header_name, accessions, filename):
...
+        to_download = []
+        for i, (fid, path, data_type) in enumerate(artifact.filepaths):
+            # ignore if tgz as they could create problems and the
+            # raw data is in the folder
+            if data_type == 'tgz':
+                continue
+            if data_type == 'directory':
```
wasade: Would it be better to ask the filesystem if the path is a directory?

qiita_db/util.py
josenavas: I moved it outside the function cause it is used in another function, to reduce code duplication.

qiita_pet/handlers/download.py
josenavas: No, this is used to remove a prefix from the path that can be formed by multiple dirs.

qiita_pet/handlers/download.py
josenavas: done

PR #2303, opened 2017-09-20
adds EBI submission of shotgun samples

antgonza:
This is still being tested in our test environment but in theory these are all the required changes. 

Now, I think this should go to the release candidate but for some reason, if I try to do the PR against that branch it raises a bunch of conflicts ... any ideas?


antgonza:
The submission is finally done and was successful. This is ready to go. Now, my guess is that this should be moved to the release-candidate branch, right? 

BTW the submission is still not in the main qiita, I'll start moving files now. Once the code and the files are in place I'll resubmit from the main qiita so we can have a full submission.

josenavas:
Yes - can you do the PR against release-candidate?

antgonza:
Is that good enough?

antgonza:
Done.

PR #2306, opened 2017-09-20
improving redbiom help and fixing bugs

antgonza:
@wasade, @josenavas, @tanaes. Now the help looks like this:
 
redbiom-help


josenavas:
@antgonza which is the base branch that you are creating your working branch? You should probably create it from the "release-candidate" branch, and you will not have the issue with all the commits showing up here.

antgonza:
that's what I did and if you see there are only 3 files changing ...

qiita_pet/handlers/qiita_redbiom.py
```diff
@@ -34,14 +34,14 @@ def _redbiom_metadata_search(self, query, contexts):
...
-                'Not a valid search: "%s", are you sure this is a '
-                'valid metadata value?' % query)
         except ValueError:
             message = (
                 'Not a valid search: "%s", your query is too small '
                 '(too few letters), try a longer query' % query)
+        except:
+            message = (
+                'Not a valid search: "%s", are you sure this is a '
+                'valid query value?' % query)
```
wasade: They probably won't be sure. What about: "The query ("%s") did not work and may be malformed. Please check the search help for more information on the queries."

qiita_pet/templates/redbiom.html
```diff
@@ -153,21 +158,83 @@
...
+      </li>
+      <li>
+        <b>Feature</b>:<br/>
+        The seach will be on all the features, in specific: OTU ids for close reference and exact sequences for deblur.
+        <br/><br/>
+        Examples:
+        <br/>
+        <ul>
+          <li>
+            Find all samples in which the GreenGenes feature 4479944 is found: "4479944"
```
wasade: GreenGenes -> Greengenes

qiita_pet/templates/redbiom.html
```diff
@@ -153,21 +158,83 @@
...
+          </li>
+        </ul>
+      </li>
+      <li><b>Taxon</b>:<br/>
+        The search will be only on closed reference and based on the taxonomies available. Note that you need to use the GreenGenes nomenclature.
```
wasade: Same here

"...nomenclature of a rank prefix and taxon (e.g., o__Clostridiales for order level Clostridiales)."

tanaes: I would also remind folks that there are dual underscores between rank and name. 

so maybe:
```
The search will be only on closed reference and based on the taxonomies available. 
Only exact matches are returned. Note that currently only the Greengenes taxonomy is 
searchable, and that it requires nomenclature of a rank prefix, two underscores, and then the
name (e.g., o__Clostridiales for order level Clostridiales).
```

qiita_pet/templates/redbiom.html
```diff
@@ -153,21 +158,83 @@
...
+            where the infants are under a certain number of days old:
+            "infant &amp; antibiotics where age_days < 30"
+          </li>
+          <li>
+            Find all samples only beloging to the EMP in which the ph in under 7:
+            "where ph < 7 &amp; emp_release1 == True"
+          </li>
+        </ul>
+      </li>
+      <li>
```
tanaes: So just to be clear, it's not possible to combine metadata searches with sequence searches? Like, I can find all ACGTTTGCCAGCAAAGACT in samples where ph < 7?

antgonza:
thanks @josenavas, @tanaes or @wasade able to merge?

qiita_pet/templates/redbiom.html
wasade: It's possible with redbiom. Remember the command line interface is intended for piping, so this specific functionality may not be exposed to the CLI but is not be a problem with the API. The high level approach is:

- fetch the set of samples containing feature X
- fetch the set of samples corresponding to metadata criteria Y
- assess the set intersection of X and Y

antgonza: right, this might be possible by adding a GUI where you can "connect" searches but not sure it it will be easier to do this via redbiom directly ...

PR #2327, opened 2017-10-09
fix #2214

### LINKED ISSUE #2214: Sample information and prep information templates pages should show update log
Users would like to be able to see a log of the filenames and timestamps for when they updated their metadata so they can tell when and which file they used to update it. Thi sis especially important when a user has not visited the study in some time.

antgonza:
Sample:
sample

Prep:
prep


qiita_pet/templates/study_ajax/prep_summary.html
```diff
@@ -472,9 +472,17 @@
 <div class="row">
   <div class="col-md-12">
     There are <b>{{num_samples}}</b> samples and <b>{{num_columns}}</b> columns in this preparation.</br>
+    {% if other_filepaths %}
+      <button class="btn btn-secondary btn-sm" data-toggle="collapse" data-target="#jobs-description">Show older files</button>
```
josenavas: Can you change `jobs-description` to something like `other-filepaths`? The name of the div will then be related to what is showing

qiita_pet/templates/study_ajax/sample_summary.html
```diff
@@ -115,13 +115,22 @@
     {% if download_id %}
       There are <b>{{num_samples}}</b> samples and <b>{{num_columns}}</b> columns in this study.
     {% end %}
+    {% if other_filepaths %}
+      <button class="btn btn-secondary btn-sm" data-toggle="collapse" data-target="#jobs-description">Show older files</button>
```
josenavas: Same here.

PR #2328, opened 2017-10-09
fix #2326

### LINKED ISSUE #2326: break out of frame on log in window
Currently, if you have 2 windows and in one you log out there is a chance that you will get a weird behavior cause we are not forcing/breaking from frames, see:
log-page

Probable solution: https://css-tricks.com/snippets/javascript/break-out-of-iframe/

antgonza:


antgonza:
Forgot to add an animation of the behavior, here it is:

wasade:
I don't understand this at all, so in the speak of @ElDeveloper, ¯\_(ツ)_/¯?

antgonza:
@wasade, thanks. Perhaps @ElDeveloper can review ...

PR #2329, opened 2017-10-10
fix #2209

### LINKED ISSUE #2209: Need a way to add all biom files from study at once
There should be a button that lets users add all bioms, and all bioms of the same type (16S closed ref vs deblur with Trim100, etc.) at the top of studies instead of making users click each button.

antgonza:
Decided to fix this with JS as it made the more sense to me. New graphical display is as follows:

Note how the "Add All" button, actually adds one by one so user can keep track of how many have been added and how many it has added.

### DIFF
--- a/qiita_pet/static/js/qiita.js
+++ b/qiita_pet/static/js/qiita.js
@@ -376,53 +376,118 @@ function show_alert(data) {
    $('#dflt-sel-info').css('color', 'rgb(0, 160, 0)');
 }
 
+function send_children_buttons(button, aids) {
+  button.disabled = true;
+
+  $.each(aids, function(idx, aid){
+    button.value = 'Adding ' + (idx + 1);
+    $('#send-button-'+aid).trigger("click");
+  });
+
+  button.value = 'Added';
+  $(button).removeClass("btn-info");
+}
+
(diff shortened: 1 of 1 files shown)


qiita_pet/static/js/qiita.js
```diff
@@ -376,55 +376,120 @@ function show_alert(data) {
...
-      proc_data_table += '</tr>';
+      processing_method[algorithm][data_type][params].push(info);
     }
   });
 
+  // creating rows
+  $.each(processing_method, function (pm, data_types) {
+    $.each(data_types, function (dt, parameters) {
+      $.each(parameters, function (params, artifacts) {
+        proc_data_table += '<tr>';
```
tanaes: So this is going to add a new row that condenses all preps from a study that have the same unique combination of data type and parameters, right? I.e. all the artifacts added when clicking the 'all' button are going to be deblur OR CR, but not a mix of both?

antgonza: That's correct but just to be clear, it's gonna group by processing method (cr, deblur, etc), data type (16S, 18S, etc) and parameters (cr against gg, silva, etc)

tanaes:
Cool that makes sense 👍

PR #2349, opened 2017-10-13
Fixes 2038

josenavas:


antgonza:
I'm actually surprised that there is no changes in the tests, are there even tests for this? If not, can you add?

josenavas:
@antgonza added the test

PR #2350, opened 2017-10-13
fixes #2245

### LINKED ISSUE #2245: Analysis pipeline won't show mapping file
When you create an analysis, the mapping file can't be downloaded from the analysis window. The only way to get it, is to go to the "See previous analyses" table.

josenavas:
screen shot 2017-10-13 at 8 59 18 am

### DIFF
--- a/qiita_db/analysis.py
+++ b/qiita_db/analysis.py
@@ -429,17 +429,17 @@ class Analysis(qdb.base.QiitaObject):
 
         Returns
         -------
-        tuple of (int, str) or None
-            filepath id and full filepath to the mapping file or None
+        int or None
+            The filepath id of the analysis mapping file or None
             if not generated
         """
-        fp = [(fp_id, fp)
-              for fp_id, fp, fp_type in qdb.util.retrieve_filepaths(
+        fp = [fp_id
+              for fp_id, _, fp_type in qdb.util.retrieve_filepaths(
                 "analysis_filepath", "analysis_id", self._id)
               if fp_type == 'plain_text']
 
         if fp:
-            # returning the actual tuple vs. an array
+            # returning the actual filepath id vs. an array
             return fp[0]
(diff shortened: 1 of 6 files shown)


antgonza:
Is the link to the mapping file still in the analysis list display?

josenavas:
yes

qiita_db/analysis.py
```diff
@@ -429,15 +429,17 @@ def mapping_file(self):
...
+            The filepath id of the analysis mapping file or None
+            if not generated
         """
-        fp = [fp for _, fp, fp_type in qdb.util.retrieve_filepaths(
-            "analysis_filepath", "analysis_id", self._id)
-            if fp_type == 'plain_text']
+        fp = [fp_id
+              for fp_id, _, fp_type in qdb.util.retrieve_filepaths(
+                "analysis_filepath", "analysis_id", self._id)
+              if fp_type == 'plain_text']
```
wasade: It's accurate that a mapping file is "plain_text" but should the system be able to differentiate those types with, say, biom summary? Out of context for this PR but seemed unexpected

josenavas: At this point only the mapping file is stored in the table "analysis_filepath". The BIOM-related files (and others) are specifically stored with the artifact, not directly to the analysis.

wasade: "At this point" suggests this could lead to bugs in the future if that changes?

josenavas: If the data model changes in the future then other sections beside this one will need to be updated - so I would say that it would depend on the nature of the data model changes.

antgonza: Discussing offline, @wasade and I agreed that perhaps the solution of this specific kind of conflicts will be to implement something like https://github.com/biocore/qiita/issues/2361

PR #2351, opened 2017-10-13
fix #2316

### LINKED ISSUE #2316: redbiom/qiita: sample number
currently we display the total amount of samples in all artifacts vs. the number of unique samples in the study.

antgonza:


qiita_pet/templates/redbiom.html
```diff
@@ -11,22 +11,17 @@
...
+        var api = this.api(), artifacts = 0, samples = 0;
 
-        // adding total features found
+        api.column(0).nodes().each(function (cell, i) {
+          var elements = $(cell).children().children().children();
+          var idx = elements.length - 1;
+          var text = $(elements[idx]).text();
+          elements = text.split(" | ");
+          artifacts += parseInt(elements[0].split(": ")[1]);
+          samples += parseInt(elements[1].split(": ")[1]);
```
wasade: what's actually being summed here? the number of samples in a study or artifacts?

antgonza: Excellent question. In the code below (lines 48-56) we count the number of artifacts and samples per study (row). Here we are building the footer to show the total so we could recount all studies/artifacts/samples or just loop over the totals (each row) and parse out the numbers to add. The format of each row is: 'Artifacts: ## |  Samples: ##'.

PR #2356, opened 2017-10-17
fix #2333

### LINKED ISSUE #2333: Bioms don't provide trim length information when selecting samples for analysis
Bioms should provide time length information so users can make informed decisions when adding .biom files to a study.
capture

antgonza:
This definitely wasn't an easy fix: @adswafford. 

Displaying the parent's processing name beside the artifacts was already implemented. However, the implementation assumed that each default_parameter_set was unique, which is not true (lots of 'Defaults') and the code was "getting confused" between them. This is solved in these code changes. I also made the search/match of parameters per command so that steps happens faster.

Additionally, we realized that for some reason (still looking into this) there is a difference between the trimming version in Qiita. The difference is that the length is being stored as string/int - this is the only command in the full database that has this problem (checked several times, this morning again). Thus, we need to discuss this during tomorrow's meeting.

### DIFF
--- a/qiita_db/util.py
+++ b/qiita_db/util.py
@@ -42,7 +42,6 @@ Methods
 
 from __future__ import division
 from future.builtins import zip
-from future.utils import viewitems
 from random import SystemRandom
 from string import ascii_letters, digits, punctuation
 from binascii import crc32
@@ -56,6 +55,7 @@ from datetime import datetime
 from itertools import chain
 from contextlib import contextmanager
 from future.builtins import bytes, str
+from collections import defaultdict
 import h5py
 
 from qiita_core.exceptions import IncompetentQiitaDeveloperError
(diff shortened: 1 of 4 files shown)


qiita_db/util.py
```diff
@@ -1506,13 +1509,15 @@ def get_artifacts_information(artifact_ids, only_biom=True):
...
             SELECT * FROM has_target_subfragment
-            ORDER BY command_id, data_type, artifact_id
+            ORDER BY cid, data_type, artifact_id
             """
 
-        sql_params = """
-            SELECT parameter_set_name, array_agg(parameter_set) AS param_set
-            FROM qiita.default_parameter_set
-            GROUP BY parameter_set_name"""
+        sql_params = """SELECT default_parameter_set_id, command_id,
```
josenavas: I think in this query the param_set value will always have a single value. In the GROUP BY clause you've included the "default_parameter_set_id" which is the primary key of the table "default_parameter_set". Thus, it is impossible for two rows in the table to have the same default_parameter_set_id.

antgonza: right and in the code below we are only getting the first value [0], changing.

PR #2366, opened 2017-10-20
fix #2125

### LINKED ISSUE #2125: Error: Unicode unequal comparison...
I receive the following error on prep 3050 of [study 11114](https://qiita.ucsd.edu/study/description/11114#) 

`Unicode unequal comparison failed to convert both arguments to Unicode - interpreting them as being unequal`

This is the prep-info file that caused this error.
[qiita_prep_info_16s.txt](https://github.com/biocore/qiita/files/969010/qiita_prep_info_16s.txt)

Edit: Also, about half of the samples were matched to `run_prefix` and placed in the right columns, but the second half of samples were not. IDK if this is related to the unicode error, or different.

Edit: In the interface, some samples are listed twice; once in the 'Available Files', then again in the other columns. I still don't know if this is related.

antgonza:
https://github.com/biocore/qiita/issues/2125 it's an interesting issue cause the problem is that many of the files match the run_prefix and thus the system can't decide which is which. I think the best solution is documentation and letting the user know how many files match the error. @colinbrislawn, what do you think?

The current solution looks like this:

### DIFF
--- a/qiita_pet/handlers/api_proxy/studies.py
+++ b/qiita_pet/handlers/api_proxy/studies.py
@@ -272,6 +272,7 @@ def study_files_get_req(user_id, study_id, prep_template_id, artifact_type):
     supp_file_types = supported_filepath_types(artifact_type)
     selected = []
     remaining = []
+    message = []
 
     uploaded = get_files_from_uploads_folders(study_id)
     pt = PrepTemplate(prep_template_id)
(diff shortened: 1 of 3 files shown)


colinbrislawn:
>providing the information to the user so they can fix it

I like it! This displays the problem to the users and implicitly shows how to fix it.

PR #2370, opened 2017-10-21
fix-1591

antgonza:
Minimized the code and added tests for the warning messages. Now, not sure if we want to get rid of the warning messages at all or just leave one of them.

### DIFF
--- a/qiita_db/metadata_template/util.py
+++ b/qiita_db/metadata_template/util.py
@@ -34,29 +34,28 @@ def prefix_sample_names_with_id(md_template, study_id):
     study_id : int
         The study to which the metadata belongs to
     """
-    # Get all the prefixes of the index, defined as any string before a '.'
-    prefixes = {idx.split('.', 1)[0] for idx in md_template.index}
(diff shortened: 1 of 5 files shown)


qiita_db/metadata_template/util.py
```diff
@@ -34,29 +34,28 @@ def prefix_sample_names_with_id(md_template, study_id):
...
+    # loop over the samples and prefix those that aren't prefixed
+    md_template['qiita_sample_name_with_id'] = pd.Series(
+        [idx if idx.split('.', 1)[0] == str(study_id)
+         else '%d.%s' % (study_id, idx)
+         for idx in md_template.index], index=md_template.index)
+
+    # get the rows that are gonna change
+    changes = len(md_template.index[
+        md_template['qiita_sample_name_with_id'] != md_template.index])
+    if changes == len(md_template.index):
```
josenavas: I think this warning message is not useful. If everything in the updated metadata file is correct this will still raise a warning, which is pretty annoying to have in a "good" study. The other one, however, I think it will be really useful for catching errors.

PR #2371, opened 2017-10-21
Fix 1812

antgonza:


antgonza:
This is ready for review: @josenavas, @ElDeveloper and @wasade. I think 2-3 reviews will be great. The main things to consider while reviewing:
1. The ProcessingJob.create has a new parameter force to force creation, as in some cases is fine to force creation even with duplicated parameters.
2. Be careful to check that it makes sense to use that flag in the tests, for example: must of our tests create duplicated jobs and @josenavas and I thought this was fine
3. Sample/Prep/Analysis workflows can still create duplicated jobs, thus adding there.

qiita_db/processing_job.py
```diff
@@ -128,14 +128,43 @@ def create(cls, user, parameters):
...
+                ["command_parameters->>'%s' = '%s'" % (k, v)
+                 for k, v in viewitems(parameters.values)
+                 if isinstance(v, Iterable) and not isinstance(v, str)
+                 for v in v])
+            sql = sql + ' AND ' if params else sql
+            qdb.sql_connection.TRN.add(sql + params, [command.id])
+            existing_jobs = qdb.sql_connection.TRN.execute_fetchindex()
+            if existing_jobs and not force:
+                raise ValueError(
+                    'Cannot create job because these jobs:\n%s' % '\n'.join(
```
josenavas: "Cannot create job because it is a duplicate of these jobs"

qiita_db/processing_job.py
```diff
@@ -128,14 +128,43 @@ def create(cls, user, parameters):
...
+
+            # check if a job with the same parameters already exists
+            sql = """SELECT processing_job_id, email, processing_job_status
+                     FROM qiita.processing_job
+                     LEFT JOIN qiita.processing_job_status
+                        USING (processing_job_status_id)
+                     WHERE command_id = %s AND processing_job_status IN (
+                        'success', 'waiting', 'running') """
+            params = " AND ".join(
+                ["command_parameters->>'%s' = '%s'" % (k, v)
```
ElDeveloper: This is one confusing double conditional list comprehension. `for v in v`?

qiita_db/processing_job.py
ElDeveloper: 👍 ... maybe worth including a bit more detail:

Cannot create job because the parameters are the same as jobs that are queued, running or already have succeeded: .....

qiita_db/processing_job.py
wasade: agree

qiita_db/test/test_processing_job.py
```diff
@@ -22,7 +22,7 @@
 from qiita_core.qiita_settings import qiita_config
 
 
-def _create_job():
+def _create_job(force=True):
```
wasade: If we default to `True`, are there situations in the codebase where this is `False`?

wasade: Or rather, this API has changed I believe since the existing functionality would be `force=False` if I understand the docs above. Will this change be a problem elsewhere?

wasade: ...or was the default behavior previously to force the execution?

josenavas:
@antgonza you've test failures here

antgonza:
Don't merge this as it still needs some work. The code works as expected but the GUI is not catching/working as expected.

antgonza:
Well, the "fix" was easier than expected, the error now is shown like this:

antgonza:
@wasade and @ElDeveloper, could you do another pass? Thanks!

qiita_db/processing_job.py
```diff
@@ -128,14 +128,56 @@ def create(cls, user, parameters):
...
+                        'success', 'waiting', 'running') """
+
+            # we need to use ILIKE because of booleans as they can be
+            # false or False
+            _format = "command_parameters->>'%s' ILIKE '%s'"
+            params = []
+            for k, v in viewitems(parameters.values):
+                # this is necessary in case we have an Iterable as a value
+                # but that is not unicode or string
+                if isinstance(v, Iterable) and not (isinstance(v, str) or
```
ElDeveloper: isinstance can take a tuple i.e. `isinstance(v, (str, unicode))`.

qiita_db/test/test_processing_job.py
```diff
@@ -256,6 +258,15 @@ def test_create(self):
         self.assertEqual(obs.step, None)
         self.assertTrue(obs in qdb.artifact.Artifact(1).jobs())
 
+    def test_create_duplicated(self):
+        job = _create_job()
+        job._set_status('running')
+        with self.assertRaisesRegexp(ValueError, 'Cannot create job because '
+                                     'the parameters are the same as jobs '
+                                     'that are queued, running or already '
+                                     'have succeeded:'):
```
ElDeveloper: How is this passing if this is not a regex, and there's no pattern to match here? Shouldn't there be a job id, or a regex indicating that this should match a job id?

qiita_db/processing_job.py
antgonza: k, changing

qiita_db/test/test_processing_job.py
antgonza: cause it behaves as an IN.

ElDeveloper: Got it!

PR #2372, opened 2017-10-22
Fix 2230

danteese:
Prevent large file downloads using modal dialog to confirm action. 
fix-2230


qiita_pet/templates/sitebase.html
```diff
@@ -570,6 +570,33 @@
...
+            <div class="modal-footer">
+              <button type="button" class="btn btn-danger" data-dismiss="modal">No</button>
+              <button type="button" class="btn btn-success" id="qiita_dwnld_accept">Yes</button>
+            </div>
+            <!-- href="{% raw qiita_config.portal_dir %}/release/download/public" -->
+          </div>
+        </div>
+      </div>
+
+      <script type="text/javascript">
```
josenavas: Can this piece of javascript be moved up with the rest of the JavaScript in this page? Thanks!

qiita_pet/test/test_download.py
```diff
@@ -161,21 +161,6 @@ def test_download_study(self):
         response = self.get('/download_study_bioms/1')
         self.assertEqual(response.code, 200)
         exp = (
-            '- 1256812 /protected/processed_data/1_study_1001_closed_'
```
josenavas: Why has this being deleted? I'd to fix this recently by adding this, so I'm surprised now that it needs to be removed.

danteese: Those lines had a conflict between the master and dev branches, which I fixed by deleting them. BTW the tests are passing.

PR #2375, opened 2017-10-25
Fix ILIKE parameters quoted params

antgonza:


qiita_db/processing_job.py
```diff
@@ -141,42 +141,57 @@ def create(cls, user, parameters, force=False):
...
+
+            if params:
+                # divided by 2 as it's key-value pairs
+                len_params = len(params)/2
+                sql = sql.format(' AND ' + ' AND '.join(
+                    ["command_parameters->>%s ILIKE %s"] * len_params))
+                params = [command.id] + params
+                TTRN.add(sql, params)
+            else:
+                TTRN.add(sql.format(""), [command.id])
```
josenavas: Do you mind adding a comment here on why there is a need for a call to format with an empty string? I had to read the 3 or 4 times until I realized why it was needed.

qiita_db/test/test_processing_job.py
```diff
@@ -948,5 +953,27 @@ def test_remove_error(self):
             tester.remove(tester.graph.edges()[0][0])
 
 
+@qiita_test_checker()
+class ProcessingJobDuplicated(TestCase):
```
josenavas: Is there a specific reason why you've created a different class here? That triggers a new reset of the DB, it will be great if we can avoid a reset of DB unless strictly necessary.

qiita_db/processing_job.py
antgonza: k, will add.

qiita_db/test/test_processing_job.py
antgonza: As we are setting all the jobs to error the other tests are failing. I think we have 2 options: A. reset back but it seemed ugly to me or B. have it as a different class so the reset happens, this seemed better to me

josenavas: Nah, it's okay - I just wanted to make sure that there was an actual reason to create the class.

PR #2377, opened 2017-10-27
Update redbiom.html

adswafford:


qiita_pet/templates/redbiom.html
```diff
@@ -208,18 +208,31 @@
         <ul>
           <li>
             Find all samples in which the word infant exists, as well as antibiotics,
-            where the infants are under a certain number of days old:
-            "infant &amp; antibiotics where age_days < 30"
+            where the infants are under a year old:
```
antgonza: Remember that this is HTML so it will not render as it's written. If you want a new line you will need to use `<br/>`, if you want to create a list you could do: 
```html
<ul>
   <li>element 1</li>
   <li>element 2</li>
</ul>
```
and please use `&amp;` for all the `&`

antgonza:
ping @adswafford

adswafford:
Still no luck at least on qiita-rc, but not sure if this is pointed at
redbiom-rc? Also, the message says the query is malformed?
[image: Inline image 1]

wasade:
Not sure, works for me against `redbiom-rc`:

```bash
$ redbiom search metadata "where ph < 7 and emp_release1 == 'True'" | wc -l
1176
```

adswafford:
Gotcha. I just tried to go to qiita-rc again now to test and it looks like
the system is down/updating, but we should check that it's working through
the GUI before we put it in the list of examples.

josenavas:
@adswafford we are currently updating the qiita-rc to do a dry upgrade, to ensure the success of tomorrow's main system update.

antgonza:
closing in favor of: https://github.com/biocore/qiita/pull/2382

PR #2386, opened 2017-11-04
Fix per sample fastq bug

antgonza:
During processing of the HMP2 metagenomic data I found an issue where it was possible to assign  the same run_prefix to multiple files.


antgonza:
This is ready for review.

Before:
before

After: 
after

qiita_pet/handlers/api_proxy/studies.py
```diff
@@ -291,9 +289,16 @@ def study_files_get_req(user_id, study_id, prep_template_id, artifact_type):
...
         # group files by prefix
-        sfiles = {p: [f for _, f in uploaded if f.startswith(p)]
-                  for p in prep_prefixes}
+        sfiles = defaultdict(list)
+        for p in prep_prefixes:
+            to_remove = []
+            for fid, f in uploaded:
+                if f.startswith(p):
+                    sfiles[p].append(f)
+                    to_remove.append((fid, f))
```
ElDeveloper: I think this can be changed to be:

```
uploaded.remove((fid, f))
```

And you can remove the list-comprehension below that's filtering the uploaded elements.

antgonza: Tried that originally, see previous failing push. The issue is that you are changing a looping variable, while you are looping on it.

ElDeveloper: Ah, I see, thanks!

PR #2397, opened 2017-11-22
Storing naming and merging scheme for the commands

josenavas:


qiita_db/handlers/plugin.py
```diff
@@ -101,9 +101,11 @@ def post(self, name, version):
             req_params = loads(self.get_argument('required_parameters'))
             opt_params = loads(self.get_argument('optional_parameters'))
 
-            for p_name, (p_type, dflt) in opt_params.items():
-                if p_type.startswith('mchoice'):
-                    opt_params[p_name] = [p_type, loads(dflt)]
+            for p_name, vals in opt_params.items():
+                if vals[0].startswith('mchoice'):
+                    opt_params[p_name] = [vals[0], loads(vals[1])]
+                    if len(vals) == 4:
```
antgonza: Just to be in the safe side, should this be `>= 4`?

antgonza: Just noticed that below you check for `== 4` and if greater you raise an error, this is also a fine behavior

PR #2404, opened 2017-11-27
add visibility to analysis

antgonza:
Here I'm adding the main background functionality to generate the 2 list for public and not public analysis. These changes are based on the artifacts and study code. Note that generate_analysis_list is not actually used anywhere but it has tests and the plan is to have another PR which uses it to display all analysis.


qiita_db/analysis.py
```diff
@@ -525,19 +525,70 @@ def can_be_publicized(self):
 
         Returns
         -------
-        bool
-            Whether the analysis can be publicized or not
+        bool, list
+            Whether the analysis can be publicized or not and the list of not
```
ElDeveloper: not public -> private ?

ElDeveloper: Also this bool and list should be two different return arguments in the docstring:

```
Returns
-------
bool
    foo bar baz
list
    foo fo foooooooo
```

qiita_db/analysis.py
```diff
@@ -525,19 +525,70 @@ def can_be_publicized(self):
...
+        """Returns if the analysis is public
+
+        Returns
+        -------
+        bool
+            If the analysis is public
+        """
+        with qdb.sql_connection.TRN:
+            # getting all root artifacts / command_id IS NULL
+            sql = """SELECT visibility
```
ElDeveloper: Minor, but this query could probably also use a `DISTINCT`. Though the only noticeable benefits would be if this table is returning a lot of visibilities.

qiita_db/support_files/patches/63.sql
```diff
@@ -1,12 +0,0 @@
--- Nov 22, 2017
```
ElDeveloper: Why was this patch removed and merged with 62.sql?

qiita_db/test/test_analysis.py
```diff
@@ -556,6 +556,21 @@ def test_add_file(self):
...
 
+    def test_is_public_make_public(self):
+        analysis = self._create_analyses_with_samples()
+        self.assertFalse(analysis.is_public)
+
+        # testing errors
+        with self.assertRaises(ValueError):
+            analysis.make_public()
+
+        # testing successfully makeing public
```
ElDeveloper: makeing -> making

qiita_db/util.py
```diff
@@ -1674,3 +1674,63 @@ def open_file(filepath_or, *args, **kwargs):
...
+
+def generate_analysis_list(analysis_ids, public_only=False):
+    """Get general analysis information
+
+    Parameters
+    ----------
+    analysis_ids : list of ints
+        The analysis ids to look for. Non-existing ids will be ignored
+    public_only : bool, optional
+        If true, return only public anaysis. Default: false.
```
ElDeveloper: anaysis -> analyses

qiita_db/util.py
```diff
@@ -1674,3 +1674,63 @@ def open_file(filepath_or, *args, **kwargs):
...
+
+    Returns
+    -------
+    list of dict
+        The list of studies and their information
+
+    Notes
```
ElDeveloper: Any notes? If not, can you remove this section?

qiita_db/support_files/patches/63.sql
antgonza: We decided that we will only have on patch per dev cycle.

ElDeveloper: Got it, 👍

PR #2405, opened 2017-11-28
Fix 2343

antgonza:
Review/merge first https://github.com/biocore/qiita/pull/2404. This connects the visibility changes in the Analysis with GUI and generates the 2 analysis lists. Looks like:
screen shot 2017-11-27 at 6 20 18 pm


qiita_pet/templates/list_analyses.html
```diff
@@ -22,83 +43,152 @@
...
-      </tr>
+<div class="row">
+  <div class="col-sm-12" id="user-studies-div">
+    <h3 class="gray-msg">Your Analyses</h3>
+    <table id="user-analyses-table" class="table table-bordered gray-msg">
+      <thead>
+        <tr>
+          <th>Artifacts</th>
+          <th>Analysis Name</th>
+          <th>Timestamp</th>
```
ElDeveloper: I believe this is the creation timestamp. Could it instead read "Creation Timestamp" or something like that?

qiita_db/util.py
```diff
@@ -1722,8 +1726,15 @@ def generate_analysis_list(analysis_ids, public_only=False):
...
             if mapping_files == [None]:
                 mapping_files = []
+            else:
+                mapping_files = [
+                    (mid, get_filepath_information(mid)['fullpath'])
+                    for mid in mapping_files]
             if artifacts == [None]:
                 artifacts = []
+            else:
+                # making sure they int so they don't brake the GUI
```
ElDeveloper: they int -> they are int (?)
brake -> break

ElDeveloper:
Thanks @antgonza!

PR #2407, opened 2017-11-28
button to make analysis public

antgonza:
Button to make it public:
screen shot 2017-11-28 at 1 19 31 pm

Error when can't delete:
screen shot 2017-11-28 at 1 19 39 pm

When success:
screen shot 2017-11-28 at 1 20 10 pm


josenavas:
Code looks good, but can you put the button next to the "Mapping file" button? Also, instead of "(Analysis is public)" can you just add "(public)" and "(private)" when it's not public?

antgonza:
Now, it looks like this:
screen shot 2017-11-28 at 1 50 30 pm

PR #2408, opened 2017-11-29
Improving processing network

josenavas:
This builds on top of @antgonza's previous improvement removing duplicated code.

Specifically, what I'm doing here is creating a single Vue component that generates the entire processing graph and takes care of all the steps involved on the creation of workflows and execute them. The good thing on having it as a Vue component is that it is setting it up for being able to automatically update the graph interface as jobs are being completed.

At this point, all the relevant code has been moved, and one can create workflows and submit them in the analysis pipeline.

TODOs:
- [x] Modify the processing (study) pipeline to use the new component (currently is completely broken)
- [x] Update the processing pipeline endpoints so they're normalized with the analysis pipeline ones (some of the answers use different nomenclature for the same data representation, which makes the generalization hard - normalizing this nomenclature it is essential for the common class to work).
- [x] Update the endpoints that provide the graph to include all jobs related with the current graph. At this point it only returns the completed ones, but it should return all of them (running, errored or in_construction) so in the graph we will have all the information regarding to the status of the jobs being executed.

This PR is already quite big, but it is mainly moving code around and adapting the function signatures. The automatic update of the graph interface will be out of scope for this PR, so that part would be easier to review. However, I would like to have the above 3 points completed before merging this, so we have a working system in "dev" (which currently is broken BTW).


qiita_pet/static/js/networkVue.js
```diff
@@ -0,0 +1,769 @@
...
+    $("#processing-network-div").hide();
+
+    $('#run-btn').on('click', function() { vm.runWorkflow(); });
+    $('#run-btn-div').hide();
+    // This call to udpate graph will take care of updating the jobs
+    // if the graph is not available
+    vm.updateGraph();
+    setInterval(function() {
+      // Only update if the graph has not been generated yet
+      if (vm.nodes.length == 0) {
```
antgonza: This is the interval the graph is gonna be refreshed, currently 5 seconds ... to fast/slow? should it be smarter? like only refresh the GUI if there are actual changes in the network?

antgonza:
Just saw your comments and now it makes sense about how little python changes are in this PR.

qiita_db/artifact.py
```diff
@@ -1115,23 +1119,98 @@ def descendants_with_jobs(self):
             sql = """SELECT *
```
antgonza: Can you list the columns that you are retrieving? I'm mainly concern about the order, which is assumed in the for loop

qiita_db/artifact.py
```diff
@@ -1115,23 +1119,98 @@ def descendants_with_jobs(self):
...
+            # The code above returns all the jobs that have been successfully
+            # executed. We need to add all the jobs that are in all the other
+            # status. Approach: Loop over all the artifacts and add all the
+            # jobs that have been attached to them.
+            visited = set()
+            queue = nodes.keys()
+            while queue:
+                current = queue.pop(0)
+                if current not in visited:
+                    visited.add(current)
```
antgonza: should this be the last line within this if? Just worry that something fails in the if/elifs and we still add the current.

qiita_db/artifact.py
```diff
@@ -1115,23 +1119,98 @@ def descendants_with_jobs(self):
...
+                        jstatus = n_obj.status
+                        # If the job is in success we don't need to do anything
+                        # else since it would've been added by the code above
+                        if jstatus != 'success':
+                            # Connect the job with his input artifacts, the
+                            # input artifacts may or may not exist yet, so we
+                            # need to check both the input_artifacts and the
+                            # pending properties
+                            for in_art in n_obj.input_artifacts:
+                                edge = (nodes[in_art.id], nodes[n_obj.id])
```
antgonza: These 3 lines are used over and over, perhaps add a _function for this?
```
edge = (nodes[in_node_id], nodes[n_obj.id])
if edge not in edges:
   edges.add(edge)
```

qiita_db/processing_job.py
```diff
@@ -1031,14 +1031,36 @@ def processing_job_worflow(self):
             The processing job workflow the job
         """
         with qdb.sql_connection.TRN:
-            sql = """SELECT processing_job_workflow_id
-                     FROM qiita.processing_job_workflow_root
-                     WHERE processing_job_id = %s"""
+            # Retrieve the workflow root jobs
+            sql = "SELECT * FROM qiita.get_processing_workflow_roots(%s)"
```
antgonza: could you add what you are "SELECTing"? Same concern than before.

qiita_pet/handlers/api_proxy/prep_template.py
```diff
@@ -624,37 +625,48 @@ def prep_template_graph_get_req(prep_id, user_id):
...
-    G = prep.artifact.descendants
-    node_ids = [id_ for id_, label in node_labels]
-    edge_list = [(n.id, m.id) for n, m in G.edges()
-                 if n.id in node_ids and m.id in node_ids]
 
-    return {'status': 'success',
-            'message': '',
-            'edge_list': edge_list,
-            'node_labels': node_labels}
+def prep_template_jobs_get_req(prep_id, user_id):
```
antgonza: Should this check for permission access? Also, currently, user_id is not used in the code

qiita_pet/handlers/analysis_handlers/base_handlers.py
```diff
@@ -125,36 +126,17 @@ def analyisis_graph_handler_get_request(analysis_id, user):
...
-            nodes.add((obj_type, otype, obj.id, name))
-
-        edges.update({(s[1].id, t[1].id) for s, t in g.edges()})
-
-    # Nodes and Edges are sets, but the set object can't be serialized using
-    # JSON. Transforming them to lists so when this is returned to the GUI
-    # over HTTP can be JSONized.
-    return {'edges': list(edges), 'nodes': list(nodes)}
+        if a.processing_parameters is None:
+            g = a.descendants_with_jobs
```
antgonza: My only concern with this approach is that you are generating the initial graph `a.descendants_with_jobs` and then formatting to JS `get_network_nodes_edges`. In other words, you are doing the same thing twice. What about generating the JS from within the artifact? This might safe loops and we can think it as a different `print` format for the artifact ...

qiita_db/artifact.py
josenavas: Done

qiita_db/artifact.py
josenavas: I've added an else that raises an error if `n_type`is not in `['artifact', 'job', 'type']`. That should never happen but better be safe than sorry.

qiita_db/artifact.py
josenavas: Done

qiita_db/processing_job.py
josenavas: Done

qiita_pet/handlers/analysis_handlers/base_handlers.py
josenavas: That would be coupling the representation of the data to the current interface. Note that `descendants_with_jobs` is returning a rich object (network.DiGraph) which we use in other parts of the code to access those objects. However, the JS representation of it is just 2 lists (nodes and edges) which will make it harder to do any traversing of the graph (i.e. we would need to re-implement those traversals that are already in networkx).

qiita_pet/handlers/api_proxy/prep_template.py
josenavas: Added - also a test for that case

qiita_pet/handlers/analysis_handlers/base_handlers.py
antgonza: After offline discussion, we will see general behavior on test server and reevaluate.

PR #2411, opened 2017-11-29
Fix comments from meeting in analysis list

antgonza:


antgonza:
Just for clarity, this fixes:
1. The sorting by parameter, remember that each of the examples below are activated via the command parameters
   - Without any parameters in the commands: 1-noparams
   - Adding deblur per file option: 2-deblur-filetype
   - Adding trim length: 3-trim
   - AGP: agp
2. The study list
   - The multiple mapping files was an actual bug (it was linking the tgz/biom that are not part of the new analysis)
   - Added the visibility column: analysis

PR #2413, opened 2017-12-04
Automatically refresh the graph as needed

josenavas:
The graph periodically checks the status of the jobs that need attention (i.e. the ones that are in waiting, queued, or running) and updates the graph as need.

In this example we see a job going from queued, to running, to error:

autoupdate

### DIFF
--- a/qiita_pet/static/js/networkVue.js
+++ b/qiita_pet/static/js/networkVue.js
@@ -55,7 +55,8 @@ Vue.component('processing-graph', {
           requests.push($.get(vm.portal + '/study/process/job/', {job_id: value}));
         });
 
-        // Reset the runningJobs list, since we are going to be reseting it here
+        // Reset the runningJobs list, since we are going to be adding
+        // the running jobs below
         vm.runningJobs = [];
 
         $.when.apply($, requests).then(function() {
@@ -77,7 +78,7 @@ Vue.component('processing-graph', {
             }
 
             if (jobNode['status'] !== jobStatus) {
-              // The status of the job changed. 
+              // The status of the job changed.
               // we decide what to do based on the new status.
               if (jobStatus === 'success' || jobStatus === 'error') {
                 // If the job succeeded or failed, we need to reset the entire graph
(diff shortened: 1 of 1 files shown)


qiita_pet/static/js/networkVue.js
```diff
@@ -44,6 +44,62 @@ Vue.component('processing-graph', {
...
+      var requests = [];
+      var jobId, jobStatus, jobNode;
+      var needsUpdate = false;
+
+      if (vm.runningJobs.length > 0) {
+        $.each(vm.runningJobs, function(index, value) {
+          requests.push($.get(vm.portal + '/study/process/job/', {job_id: value}));
+        });
+
+        // Reset the runningJobs list, since we are going to be reseting it here
```
antgonza: reset the xx, since we are gonna reset?

qiita_pet/static/js/networkVue.js
```diff
@@ -805,10 +866,20 @@ Vue.component('processing-graph', {
...
+        // Check for the initial poll - it only happens if a graph doesn't exist yet
+        if (vm.initialPoll) {
+          vm.updateGraph();
+        }
+
+        // Update the status of the jobs
+        vm.update_job_status();
       }
-    }, 5000);
+    }, 1000);
```
antgonza: Every second, gulp!

qiita_pet/static/js/networkVue.js
josenavas: Changed the doc

qiita_pet/static/js/networkVue.js
josenavas: Note that although this call is performed every second, the actual update only happens every 15 seconds. The every second is needed if you want the countdown to inform the user when the interface is going to be updated, as it can be seen in the GIF (look just above the network graph).

josenavas: Forgot to mention that in the GIF the counter is set to reset the interface every 5 seconds, but that was for testing, the actual code does it every 15 seconds.

PR #2426, opened 2017-12-09
addin skeleton for archive/observations/

antgonza:
This adds the minimum requirement for the new archiving feature. At this time, the response is just the same input as a dict with values of [].

### DIFF
--- a/qiita_db/handlers/archive.py
+++ b/qiita_db/handlers/archive.py
@@ -1,0 +1,28 @@
+# -----------------------------------------------------------------------------
+# Copyright (c) 2014--, The Qiita Development Team.
+#
+# Distributed under the terms of the BSD 3-clause License.
+#
+# The full license is in the file LICENSE, distributed with this software.
+# -----------------------------------------------------------------------------
+
+from .oauth2 import OauthBaseHandler, authenticate_oauth
+
+
+class APIArchiveObservations(OauthBaseHandler):
+    @authenticate_oauth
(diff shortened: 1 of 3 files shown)


qiita_db/handlers/archive.py
```diff
@@ -0,0 +1,28 @@
...
+#
+# The full license is in the file LICENSE, distributed with this software.
+# -----------------------------------------------------------------------------
+
+from .oauth2 import OauthBaseHandler, authenticate_oauth
+
+
+class APIArchiveObservations(OauthBaseHandler):
+    @authenticate_oauth
+    def post(self):
```
josenavas: Since this is retrieving information this should be a "get"

qiita_db/handlers/archive.py
```diff
@@ -0,0 +1,28 @@
...
+
+        Returns
+        -------
+        dict
+            The contents of the analysis keyed by sample id
+        """
+        # job_id = self.get_argument('job_id')
+        features = self.request.arguments['features']
+
+        # TODO: search on artifact
```
josenavas: artifact -> archive

qiita_db/handlers/archive.py
josenavas: After offline discussion it should be a POST. The requests can get large and there is a size limitation on the URL sent. Also, although it is possible to send the extra parameters on the body of the GET request, it is considered bad practice.

Issue #2185, opened 2017-08-02
Automatically refresh status of processing/analysis on the display page


adswafford:
Users should not have to manually refresh the page to see new artifacts in the processing or analysis networks

antgonza:
Let's keep a list so they are easier to track:
- [ ] Sample info file while deleting columns
- [ ] Sample info file while updating
- [ ] Prep info file while deleting columns
- [ ] Prep info file while updating
- [ ] Deleting analysis, originally reported in #2332
- [ ] ... add others by editing this post ...

antgonza:
This was solved in the previous release.

PR #2429, opened 2017-12-11
Make Sample Information Page reactive

josenavas:
The PR ended up being smaller than I expected (although I know is still pretty big). In this PR I modified the Sample Information Page so now it is reactive and automatically reloads the different sections of the page as needed.

Here is a GIF showing the changes - I'm not cheating the page is reloading alone xD.
stimprovements

While making these changes I found 2 bugs that I also fixed here: (1) when changing the visibility of an artifact the code was calling a function that no longer exists and (2) it turns out that in master, if you are accessing a public study which you are NOT the owner you're still presented with the buttons to delete samples (o.O) - this was pretty bad and unsure how long has been in the system (my guess is since forever), but I fixed it here. Also a GIF showing these changes:

otherbugfixesd


josenavas:
Forgot to mention that it fixes the sample information related items from this issue https://github.com/biocore/qiita/issues/2185

and also https://github.com/biocore/qiita/issues/2389

PR #2430, opened 2017-12-11
Fix JS error

antgonza:
@josenavas noticed an error that it's actually deployed in the main site but as it doesn't actually prevent any functionality we never caught it before. Also removing code that AFAIK we are not actually using.


qiita_pet/templates/study_base.html
```diff
@@ -20,22 +20,12 @@
   }
 
   function fill_artifact(aid) {
```
josenavas: Actually this entire function is not used anywhere (a quick grep for fill_artifact on the Qiita code only returned this occurrence). Can you remove the entire function? Thanks!

PR #2432, opened 2017-12-13
Fixes various small issues

josenavas:
The User job listing now ignores the hidden jobs
Interaction -> Navigation
The job page now shows the error message if there is one.


josenavas:
@ElDeveloper @antgonza are you available to merge?

PR #2437, opened 2017-12-15
Fixing adding command problem

josenavas:
Carolina and Stephanie found an issue in the qiita-rc system: if you tried to add a command to an artifact that has not been generated and the job generating it has been already submitted (i.e. no longer in construction), the system failed with a really weird error. Adding a command to those types of artifacts is not easy, there are a lot of assumptions perform through the code that rely on this behavior not happening, and removing all them is complex. For the time being, I've disabled the ability to add commands to those type of artifacts (you can still add commands if the workflow is in construction). Instead it shows a little information about that (there is not too much to show, but at least it doesn't fail and you see something on the screen). GIF with the resulting changes:

futureresult

### DIFF
--- a/qiita_pet/static/js/networkVue.js
+++ b/qiita_pet/static/js/networkVue.js
@@ -530,61 +530,81 @@ Vue.component('processing-graph', {
      *
      * Generates the GUI for selecting the commands to apply to the given artifacts
      *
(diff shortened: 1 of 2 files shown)


josenavas:
@antgonza @ElDeveloper available for a review?

PR #2446, opened 2017-12-22
Fixing download issue

josenavas:


wasade:
Seems worthy of a :+1:... is there a regression test that can be added?

josenavas:
We currently don't have JavaScript tests, probably they can be added later.

PR #2449, opened 2017-12-27
Adding archive

antgonza:
This adds the basic archiving code. @sjanssen2 and @josenavas, could you take a look so we can continue working on it? Thanks!


sjanssen2:
Hey @antgonza thanks for your efforts. I am travelling with my family right now who will get quite angry if I start staring into my laptop. I might find some time in two days

.travis.yml
```diff
@@ -3,20 +3,18 @@ language: python
 sudo: false
 env:
   global:
-    - PYTHON_VERSION=2.7
+    - PYTHON_VERSION=3.5
```
wasade: ?

qiita_db/archive.py
```diff
@@ -0,0 +1,178 @@
...
+        Parameters
+        ----------
+        artifact : qiita_db.artifact.Artifact
+            The artifact from which the features were generated
+        features : dict {str: str}
+            A dictionary of the features and the values to be stored
+
+        Raises
+        ------
+            ValueError
```
wasade: extra tab

qiita_db/archive.py
```diff
@@ -0,0 +1,178 @@
...
+                If the Artifact type is not BIOM
+                If the artifact doesn't have a biom filepath
+        """
+        with qdb.sql_connection.TRN:
```
wasade: seems like most of the logic contained can be done outside of the context?

qiita_db/archive.py
```diff
@@ -0,0 +1,178 @@
...
+            acmd = job.command
```
wasade: same?

qiita_db/archive.py
```diff
@@ -0,0 +1,178 @@
...
+        r"""Retrieves all features/values from the archive
+
+        Parameters
+        ----------
+        archive_merging_scheme : optional, str
+            The name of the archive_merging_scheme to retrieve
+
+        Notes
+        -----
+            If archive_merging_scheme is None it will return all
```
wasade: extra indent

qiita_db/archive.py
```diff
@@ -0,0 +1,178 @@
...
+            feature values
+        """
+        with qdb.sql_connection.TRN:
+            extras = []
```
wasade: same?

qiita_db/test/test_archive.py
```diff
@@ -0,0 +1,67 @@
...
+        # testing specific artifacts and parameters
+        for i in [4, 5, 8, 9]:
+            qdb.archive.Archive.insert_from_artifact(
+                qdb.artifact.Artifact(i), {
+                    'featureA%d' % i: '{valuesA: vA, int: 1}',
+                    'featureB%d' % i: '{valuesB: vB, float: 1.1}'})
+
+        # now let's tests that all the inserts happen as expected
+        exp = {
+            'featureA4': '{valuesA: vA, int: 1}',
```
wasade: are these supposed to be json? if so, they're not as the strings "vA" and "vB" are not quoted, nor are the keys

.travis.yml
antgonza: Long story short: there is a bug in conda that the Qiime2 guys reported and this is the best way to install everything. See that these changes are in master and are passing.

qiita_db/archive.py
antgonza: removing

qiita_db/archive.py
antgonza: possible but I think it's better to leave it like this as some of these steps are actually accessing the DB so this assures that everything it's in the same transaction.

qiita_db/archive.py
antgonza: same ... 😄

qiita_db/archive.py
antgonza: removing

qiita_db/archive.py
antgonza: ya' know, same 😉

qiita_db/test/test_archive.py
antgonza: well, they are strings but supposedly represent json, I can change for test if you feel strongly about it.

qiita_db/archive.py
wasade: oh. implicit operations remote resource operations? :/ okay.

qiita_db/archive.py
wasade: off topic but how does a developer know whether a given qiita method (or property) will issue a db call?

qiita_db/test/test_archive.py
wasade: if this is stuff generated from the code, then it certainly is not generating json. if this is supposed to be json, should the data contained in the insertion logic validate that the data are json?

qiita_db/archive.py
antgonza: I think the safest assumption is that any property/method from a Qiita object will issue a DB call ... as everything is store in there.

qiita_db/archive.py
```diff
@@ -0,0 +1,178 @@
...
+    -------
+    insert_from_biom
+
+    See Also
+    --------
+    qiita_db.QiitaObject
+    """
+
+    @classmethod
+    def _inserting_main_steps(cls, TRN, ms, features):
```
josenavas: Minor but there is no need to pass the TRN object by parameter. The system is using the singleton pattern, so using "with ...TRN:" should be good.

qiita_db/archive.py
```diff
@@ -0,0 +1,178 @@
...
+
+import qiita_db as qdb
+
+
+class Archive(qdb.base.QiitaObject):
+    r"""Extra information for any features stored in a BIOM Artifact
+
+    Methods
+    -------
+    insert_from_biom
```
josenavas: insert_from_artifact?

qiita_db/test/test_archive.py
```diff
@@ -0,0 +1,68 @@
...
+class ArchiveTest(TestCase):
+    def test_insert_from_biom_and_retrieve_feature_values(self):
+        # 1 - to test error as it's FASTQ
+        with self.assertRaises(ValueError) as err:
+            qdb.archive.Archive.insert_from_artifact(
+                qdb.artifact.Artifact(1), {})
+        self.assertEqual(
+            str(err.exception), 'To archive artifact must be BIOM but FASTQ')
+
+        # 7 - to test error due to not filepath biom
```
josenavas: This is testing that an artifact of type BIOM doesn't have a BIOM? that seems like it should never happen.

qiita_db/archive.py
antgonza: yup

qiita_db/archive.py
antgonza: thanks.

qiita_db/test/test_archive.py
antgonza: right, but I think an extra test never hurts ... anyway, want me to remove it?

PR #2453, opened 2018-01-04
fix master

antgonza:


tanaes:
This commit makes explicit some dependency versions for maintainability.

PR #2455, opened 2018-01-05
Dev

stephanieorch:
Finished help document. Please see that the references are set up correctly and the numbers are blue and link to the reference page (I'm unsure if the set up I used was correct and I am unable to check for it). Let me know if there are any issues so I can fix them.


antgonza:
Closing in favor of: https://github.com/biocore/qiita/pull/2458

PR #2460, opened 2018-01-12
finishing up archive

antgonza:
This intends to add the last functionality on the Qiita side to support archiving. The plan is that @sjanssen2 will use this to finish up the SEPP placement integration in qp-deblur and we will add changes as required. Then I will add test for the small sections that doesn't have one right now. 

Note that this will allow storing and retrieving all info but we need more work to actually build a tree during analysis, which will be the next step.


qiita_db/archive.py
```diff
@@ -77,6 +77,19 @@ def insert_from_artifact(cls, artifact, features):
 
             cls._inserting_main_steps(ms, features)
 
+    @classmethod
```
josenavas: I think this should be an instance method, since it should be applied to the archive.

sjanssen2:
before we merge: I need to explore how to store SEPP fragments placed in different reference phylogenies. I don't see yet how this information in incorporated into the merging scheme. Stay tuned

sjanssen2:
I think this PR for is ready to be merged, as it works great for the qp-deblur SEPP usecase (according PR will soon be pushed by me)

sjanssen2:
@ElDeveloper could you maybe take a look at this PR?

qiita_db/archive.py
```diff
@@ -83,10 +85,9 @@ def get_merging_scheme_from_job(cls, job):
 
         Parameters
         ----------
-        job : qiita_db.artifact.Artifact
-            The artifact from which the features were generated
-        features : dict {str: str}
-            A dictionary of the features and the values to be stored
+        job : str
+            The Qiita process job_id generating the artifact holding the
```
ElDeveloper: Looks like this job argument is not the ID but the `Job` object itself, right?

qiita_db/archive.py
```diff
@@ -149,6 +149,8 @@ def retrieve_feature_values(cls, archive_merging_scheme=None,
         ----------
         archive_merging_scheme : optional, str
             The name of the archive_merging_scheme to retrieve
+        features : list of str
```
ElDeveloper: Minor but can you add `features : list of str, optional`.

qiita_db/archive.py
```diff
@@ -179,3 +181,25 @@ def retrieve_feature_values(cls, archive_merging_scheme=None,
 
             return {k: v for k, v in
                     qdb.sql_connection.TRN.execute_fetchindex()}
+
+    def insert_features(self, merging_scheme, features):
```
ElDeveloper: A few questions, I'm having a little bit of trouble understanding this block of code. What are the different options for merging_scheme, and how does this relate back to an artifact? I'm not familiar with the Archive class and how it relates to the Artifact class and concept.

qiita_db/archive.py
antgonza: yup

qiita_db/archive.py
antgonza: Excellent question. This was an addition during the last release, where basically all software/plugin "decides" how their artifacts should be merged - mainly thought for bioms. This is done dynamically and based on parameters given by the plugins themself. The [main code](https://github.com/biocore/qiita/tree/master/qiita_db/software.py#L626) basically retrieves which parameters (for example, for trim the length, for close reference the reference, etc) should be accounted for when trying to merge files (meta-analysis) and we use the job/parameters that created the artifact we want to merge and the parameters of the parent (input of that job), for example: trim @ 150, close ref vs. GreenGenes. Now, here the artifact hasn't been created so we can't actually ask this directly so we query the job (processing params) and the input artifact. Note that this explanation is part of the plugins documentation; thus cc-ing, @stephanieorch and @adswafford .

qiita_db/archive.py
antgonza: yup

qiita_db/archive.py
ElDeveloper: Great, thanks for the very useful explanation. I assume that the merging schemes will be described in that documentation, so this is good to go 👍

PR #2468, opened 2018-01-15
fix #2360

### LINKED ISSUE #2360: Allow users to process any public study
Currently only the owner or admins can process those studies. This only applies to the processing not the analysis.

This should be implemented after: https://github.com/biocore/qiita/issues/1812

antgonza:
The solution of the issue is just in qiita_pet/templates/artifact_ajax/artifact_summary.html. However, while trying to find where to make that change I realized that all these files are not being used.


sjanssen2:
Good to me. But I have not checked the impact of deleting those 15 files and rely on unit-tests here.

PR #2469, opened 2018-01-15
fix #77

### LINKED ISSUE #77: Remove examples section from docstrings

antgonza:


sjanssen2:
what is the rational in removing those examples? Isn't it valuable information for the user of those functions? Are the exampled provided somewhere else?

antgonza:
Fair question. The example section of the docstrings are normally nice cause they show how to use it and are also tested during build. Now, in qiita we _cannot_ test them as there is no way to reset the DB automatically (which is pretty nice during development) and the engineering to do this is not easy, AFAIK. This caused that the minimal examples (added in the first versions of Qiita) are completely outdated and basically not used. Now, the new plugin documentation should alleviate some of this issues + the code is still documented - just no examples.

PR #2472, opened 2018-01-17
fix #1105

### LINKED ISSUE #1105: Empty cells in final column of metadata will lead to failed upload
Reports expecting X columns of metadata getting NNNNNN 9e.g. expecting 73, getting 537

antgonza:
This is an "interesting" issue because it's pretty hard to identify empty columns that have no header. Remember that the issue is that some of these columns/rows have spaces, which is what makes it hard. Anyway, the solution here is 2 part:
1. Check if some of the headers are empty/have empty values
2. Remove empty columns with headers <- this is not actually necessary but thought it was nice

### DIFF
--- a/qiita_db/metadata_template/test/test_util.py
+++ b/qiita_db/metadata_template/test/test_util.py
@@ -81,9 +81,26 @@ class TestUtil(TestCase):
         assert_frame_equal(obs, exp)
 
     def test_load_template_to_dataframe_duplicate_cols(self):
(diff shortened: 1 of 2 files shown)


sjanssen2:
Looks good to me.
However, I have one conceptional question left: what happens to columns with a valid (i.e. non empty) header but completely empty values?

antgonza:
They get drop in this line: `template.dropna(axis='columns', how='all', inplace=True)`

PR #2478, opened 2018-01-25
fix #2280

### LINKED ISSUE #2280: Attaching files to preparation allows beta, alpha, etc
Just noticed that when selecting the type of data that will be attached to a preparation, alpha diversity vectors, beta diversity results, and others are available options. See screencapture:

screen shot 2017-09-12 at 9 22 23 pm

antgonza:
Before:
before
After:
after

Note that this adds the functionality but we will need to update the qiita-spots: 
- [ ] qtp-biom, 
- [ ] qtp-visualization, 
- [ ] qtp-diversity, 
- [ ] qtp-target-gene & 
- [ ] qtp-template-cookiecutter

to actually use this new functionality.

### DIFF
--- a/qiita_pet/handlers/study_handlers/artifact.py
+++ b/qiita_pet/handlers/study_handlers/artifact.py
@@ -34,7 +34,8 @@ class NewArtifactHandler(BaseHandler):
     def get(self):
         study_id = self.get_argument("study_id")
         prep_id = self.get_argument("prep_template_id")
-        artifact_types = artifact_types_get_req()['types']
+        artifact_types = [(at, desc) for at, desc, _, _, raw in
+                          artifact_types_get_req()['types'] if raw]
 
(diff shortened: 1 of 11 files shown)


qiita_pet/handlers/study_handlers/artifact.py
```diff
@@ -34,7 +34,8 @@ class NewArtifactHandler(BaseHandler):
     def get(self):
         study_id = self.get_argument("study_id")
         prep_id = self.get_argument("prep_template_id")
-        artifact_types = artifact_types_get_req()['types']
+        artifact_types = [(at, desc) for at, desc, _, _, raw in
```
sjanssen2: is "raw" short for "is_user_uploadable", if so you might want to use it here explicitly to emphasize the filtering criterion

PR #2497, opened 2018-02-10
WIP: fix-2427 scripts/qiita_script

danteese:
This script automate the installation over Travis, useful to install in dev and master environments.


antgonza:
Thanks @danteese! Renamed to WIP as there are errors in the builds and I think we should have a couple of the qiita-spots using the script to know that it's working as we want. Suggest starting with https://github.com/qiita-spots/qp-deblur and https://github.com/qiita-spots/qtp-visualization. I'll be in Slack if you have any questions.

PR #2502, opened 2018-02-14
Final changes 02 18

antgonza:


antgonza:
Note that the changes are from @stephanieorch but I created the PR.

PR #2507, opened 2018-02-20
fix #1067

### LINKED ISSUE #1067: qiita_env make and database name conflict
I ran into a situation where a user named 'qiita' had a database named 'qiita' in a postgres server, and in my configuration file I had the the database setting set as `DATABASE = qiita`. When I tried to run `qiita_env make`, the command would fail saying something along the lines of 'database qiita already exists', so I decided to do `qiita_env drop`, however since this was a new system the command would fail saying that the "settings table" was not there.

After talking with @josenavas, the solution was to change the name of the database in the configuration file from `qiita` to :arrow_right:  `qiita_xxxxx` (though really any name would work).

The solution should be to either allow for the database to be dropped regardless of whether or not the settings table exists (which might be problematic), or to come up with a self-explanatory error message.

antgonza:
Fixing 1067. Note that decided to make it easier by checking if the database exists and the installation is test to overwrite.

### DIFF
--- a/qiita_db/environment_manager.py
+++ b/qiita_db/environment_manager.py
@@ -192,8 +192,9 @@ def make_environment(load_ontologies, download_reference, add_demo_user):
     try:
         admin_conn.execute('CREATE DATABASE %s' % qiita_config.database)
     except ValueError as error:
-        # if database exists, ignore if sett
-        if 'database "qiita_test" already exists' in error.message:
+        # if database exists ignore
+        msg = 'database "%s" already exists' % qiita_config.database
+        if msg in error.message:
             print("Database exits, let's make sure it's test")
             with qdb.sql_connection.TRN:
                 # Insert the settings values to the database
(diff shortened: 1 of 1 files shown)


qiita_db/environment_manager.py
```diff
@@ -191,30 +187,50 @@ def make_environment(load_ontologies, download_reference, add_demo_user):
...
     # Create the database
     print('Creating database')
+    create_settings_table = True
     admin_conn.autocommit = True
-    admin_conn.execute('CREATE DATABASE %s' % qiita_config.database)
+    try:
+        admin_conn.execute('CREATE DATABASE %s' % qiita_config.database)
+    except ValueError as error:
+        # if database exists, ignore if sett
+        if 'database "qiita_test" already exists' in error.message:
```
ElDeveloper: This will only work if the database is named qiita_test ... perhaps it should look for:

```python
if 'database "%s" already exists' % qiita_config.database in error.message:
    ...
```

Issue #1053, opened 2015-04-09
./test_data_studies/commands.sh does not fully populate db


squirrelo:
In #1039 we realized that the script is not fully populating the database. We need to have at least the `preprocessed_params` and `processed_params` tables populated, and hunt for any others that are not currently being populated.

ElDeveloper:
I'm trying to review #1039 but this problem keeps me from doing it i.e. the exception @josenavas reported, is showing for me as well. Is there a way to workaround this to be able to test the current functionality or should we just go with it as is? (broken, which is not terrible as it is not in master, but still).

squirrelo:
For now, you can manually add the reference to `reference` table and then add a row to `processed_params_sortmerna` and `processed_params_uclust`

ElDeveloper:
Any idea, why this wasn't happening before? It's kinda strange, I'm
creating a new environment using qiita_env make and still see the
problem.

squirrelo:
If you make a test database and log in as test@foo.bar you can test things using htat. Only one study, but it at least populates.

ElDeveloper:
Got it to work, I had to create a test environment and load my test data
there, more updates soon. Thanks for the help.

antgonza:
@ElDeveloper any updates? Is this still an issue?

josenavas:
The tables listed in the issue no longer exist - @ElDeveloper can you check if this is still an issue? If so, can you update the issue to reflect the specific problems that we may be receiving now?

ElDeveloper:
I will after tomorrow, need my installation as is for development purposes that should meet a deadline tomorrow morning.

ElDeveloper:
Sorry for the delay, I just checked and this seems to still be a problem, though likely not due to the same reasons, when I run:

```bash
qiita-env make --no-load-ontologies
./test_data_studies/commands.sh
```

And then I go to the study list page I get the following in the log file:
```
[E 160126 11:27:11 web:1228] Uncaught exception GET /study/search/?&user=demo@microbio.me&sEcho=446&query=&_=1453836431855 (::1)
[...]

PR #2514, opened 2018-02-22
fix #1999

### LINKED ISSUE #1999: Minimize CLI
After a group discussion we realized that the CLI is pretty outdated, mainly because we don't use it. Admins are the only ones that have access to it and we normally prefer to use (I)Python to do those tasks. We should reduce the amount of things we can do via the `qiita` command to:
- Add/remove system messages
- Start/stop the web workers
- Configure plugins
- (Re)submit artifacts to EBI

antgonza:
This removes all the not use/necessary commands from qiita. I also made sure that the test data is actually created fine and can be used in a meta-analysis.

### DIFF
--- a/qiita_db/commands.py
+++ b/qiita_db/commands.py
@@ -186,45 +186,6 @@ def load_prep_template_from_cmd(prep_temp_path, study_id, data_type):
         prep_temp, qdb.study.Study(study_id), data_type)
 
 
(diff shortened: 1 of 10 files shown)


antgonza:
BTW this is also addressing: https://github.com/biocore/qiita/issues/1053

PR #2515, opened 2018-02-23
fix #2512

### LINKED ISSUE #2512: Searching Studies
Many users use the "Filter studies by tags" search bar instead of the "Filter results by column data" bar. Can we switch the locations of these search bars or combine them so it can search for both?

antgonza:
The new distribution on the page will be:
screen shot 2018-02-23 at 9 51 11 am

cc: @stephanieorch, @tkosciol


antgonza:
After @ElDeveloper comments, it looks like this:
screen shot 2018-02-23 at 11 15 13 am

tkosciol:
looks good to me!
thanks for putting this together so quickly

ElDeveloper:
Looks great @antgonza!

PR #2522, opened 2018-03-08
Fix 2321

antgonza:


qiita_db/study.py
```diff
@@ -825,39 +831,58 @@ def ebi_submission_status(self):
...
         with qdb.sql_connection.TRN:
-            sql = """UPDATE qiita.{}
-                     SET ebi_submission_status = %s
-                     WHERE study_id = %s""".format(self._table)
-            qdb.sql_connection.TRN.add(sql, [value, self.id])
-            qdb.sql_connection.TRN.execute()
-
-    ebi_submission_status.__doc__.format(', '.join(_VALID_EBI_STATUS))
+            if self.ebi_study_accession:
+                status = 'submitted'
```
ElDeveloper: This method should return here, right? If there's an accession there's likely no need for any of the compute done down below.

qiita_ware/test/test_commands.py
```diff
@@ -145,14 +146,13 @@ def test_submit_EBI_parse_EBI_reply_failure(self):
...
     def test_full_submission(self):
-        artifact = self.generate_new_study_with_preprocessed_data()
-
-        # just making sure
-        self.assertEqual(artifact.study.ebi_submission_status, 'not submitted')
-
-        submit_EBI(artifact.id, 'VALIDATE', True, test=True)
-
-        self.assertEqual(artifact.study.ebi_submission_status, 'submitted')
+        ascp_pass = environ.get('ASPERA_SCP_PASS', '')
```
ElDeveloper: I think it would make sense to decorate this test with `@unitttest.skipIf` and check for the existence of the password:

https://docs.python.org/2/library/unittest.html#skipping-tests-and-expected-failures

qiita_ware/test/test_private_plugin.py
```diff
@@ -379,6 +387,46 @@ def test_complete_job(self):
         self.assertEqual(c_job.status, 'error')
         self.assertIn('No such file or directory', c_job.log.msg)
 
+    def test_submit_to_EBI(self):
+        # setting up test
+        from qiita_ware.test.test_ebi import FASTA_EXAMPLE
```
ElDeveloper: Any reason why this import statement is here and not at the top?

qiita_ware/test/test_private_plugin.py
```diff
@@ -379,6 +387,46 @@ def test_complete_job(self):
...
+
+        # submit job
+        job = self._create_job('submit_to_EBI', {
+            'artifact': artifact.id, 'submission_type': 'VALIDATE'})
+        job._set_status('in_construction')
+        job.submit()
+
+        # wait for the job to fail, and check that the status is submitting
+        checked_submitting = True
+        while (job.status != 'error'):
```
ElDeveloper: The parenthesis here are not necessary.

qiita_db/study.py
```diff
@@ -825,39 +831,58 @@ def ebi_submission_status(self):
...
+                        FROM qiita.study_artifact
+                        WHERE study_id = {0}) AND command_id = {1}""".format(
+                            self._id, cmd.id)
+            qdb.sql_connection.TRN.add(sql)
+            jobs = defaultdict(dict)
+            for info in qdb.sql_connection.TRN.execute_fetchindex():
+                jid, aid, js = info
+                jobs[js][aid] = jid
+
+            if 'queued' in jobs or 'running' in jobs:
```
ElDeveloper: Instead of putting the logic in python, it might be possible to use postgres to only search for the jobs that have a status of "error", and if the query is empty then you can assume that it's still submitting, as there's not an accession available.

Maybe it's more complicated, just thought I would share this idea.

qiita_db/study.py
antgonza: Nop, cause it could be that prep 1 was submitted and the study already has an accession but there is another prep being submitted which should and will overwrite this status.

qiita_db/study.py
antgonza: I think the only issue with this option is that we will still have to deal with those jobs in a success state.

qiita_ware/test/test_commands.py
antgonza: First time I hear of this, pretty cool, adding.

qiita_ware/test/test_private_plugin.py
antgonza: Cause I forgot! Moving to the top

qiita_ware/test/test_private_plugin.py
antgonza: K

qiita_db/study.py
ElDeveloper: Thanks for the explanation.

qiita_db/study.py
ElDeveloper: Ah, got it, thanks!

qiita_ware/test/test_commands.py
```diff
@@ -144,14 +145,12 @@ def test_submit_EBI_parse_EBI_reply_failure(self):
         with self.assertRaises(ComputeError):
             submit_EBI(ppd.id, 'VALIDATE', True)
 
+    @skipIf(environ.get('ASPERA_SCP_PASS', ''), 'skip: ascp not configured')
```
ElDeveloper: This means that the test will be skipped if the password is present. I think it would have to be:

```
@skipIf(environ.get('ASPERA_SCP_PASS', '') is None, 'skip: ascp not configured')
```

antgonza: Fixing, thanks!

PR #875, opened 2015-02-19
ENH: natural sorting on private/public study tables

adamrp:
Adds natural sorting plugin for dataTables, and applies natural sorting to private and public study tables.

Fixes #831

Replaces #862 (sorry, I couldn't git whisper a pull against @wasade's branch, so just created a new PR)

PR #2536, opened 2018-04-04
DOC: Cleanup of JavaScript libraries and licenses

### LINKED ISSUE #2535: List the JavaScript libraries being used in the system
It would be good to do an audit of the source and see what JavaScript libraries are being used.

ElDeveloper:
There were a few JavaScript files that were not used anywhere and we were missing a handful of license files.

One file that I didn't remove because although it isn't being referenced anywhere, the "type" is being used for initialization in a couple of places. @antgonza do you happen to know anything about this? This file was originally included in https://github.com/biocore/qiita/pull/875 but @wasade doesn't really recall anything.

Fixes #2535


antgonza:
Which file? Also, can we have a single page/md(?) with the list of all JS libraries?

ElDeveloper:
I meant jquery.dataTables.plugin.natural.js, sure, I’ll add that file.

Yoshiki Vázquez-Baeza

PR #2537, opened 2018-04-05
Fix aid str workflow

antgonza:


wasade:
:+1:

antgonza:
Just FYI, I was gonna issue another PR to fix the parameters in the commands in Qiita and realized that it will fit nicely within this PR so decided to add it here.

PR #2545, opened 2018-04-06
Fix 2505

antgonza:
Note that the "smaller" name version, algorithm_az, is just the name without any non a-zA-Z characters, which I think is a good start and devs can use to parse out/in whatever they need,


qiita_db/util.py
```diff
@@ -1665,23 +1670,37 @@ def get_artifacts_information(artifact_ids, only_biom=True):
...
+                platform = ''
+                target_gene = ''
+                if prep_template_id is not None:
                     if prep_template_id not in ps:
-                        ps[prep_template_id] = len(list(
-                            PT(prep_template_id).keys()))
-                    prep_samples = ps[prep_template_id]
+                        pt = PT(prep_template_id)
+                        categories = pt.categories()
+                        if 'platform' in categories:
```
ElDeveloper: I defer on this to @wasade, but I think instead of an empty string for platform and target_gene, it would likely be better to return "Not Available" or something along those lines.

wasade: Not sure, i thought it was "not available" lowercase. Best to ask Gail

antgonza: EBI is "null values: not applicable, missing: not collected, missing: not provided, missing: restricted access.". However, this is different and the current use case is redbiom, so whatever you prefer @wasade .

antgonza: After discussing offline with @wasade we are going with: 'not provided'

PR #2556, opened 2018-04-16
moved 2 analysis titles to make more sense

stephanieorch:
Removed a title so that there weren't duplicate titles. Moved a title so it was in a more appropriate place.


antgonza:
We want these changes for next deploy, thus moving to master via: https://github.com/biocore/qiita/pull/2557

PR #2587, opened 2018-05-17
Fix 2565

antgonza:
Simple change but during the process realized that we had an extra file that we are no longer using.


antgonza:
Forgot to add screenshots. 

Before:
before

After:
later

PR #2593, opened 2018-05-22
allow to hide job via GUI

antgonza:
Also realized that the delete of columns in the sample info file wasn't asking for confirmation from user.


antgonza:
Forgot to add screenshot!

delete-job

PR #2596, opened 2018-05-25
fixing bugs found during initial review

antgonza:
Found 3 things:
- By mistake left a hardcoded value, replacing for a variable vm.studyId
- The study list had an extra field:
screen shot 2018-05-25 at 11 45 37 am
- While testing the EBI submission caught that if we try modify an error will be raised.

### DIFF
--- a/qiita_pet/templates/list_studies.html
+++ b/qiita_pet/templates/list_studies.html
@@ -58,6 +58,7 @@ $(document).ready(function() {
   $('#user-studies-table').dataTable({
       "lengthMenu": [[5, 10, 50, -1], [5, 10, 50, "All"]],
       "deferRender": true,
+      "sDom": '<"top">rti<"bottom"p>',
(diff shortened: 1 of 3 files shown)


qiita_pet/templates/list_studies.html
```diff
@@ -58,6 +58,7 @@
   $('#user-studies-table').dataTable({
       "lengthMenu": [[5, 10, 50, -1], [5, 10, 50, "All"]],
       "deferRender": true,
+      "sDom": '<"top">rti<"bottom"p>',
```
tkosciol: I don't understand this `<"bottom"p>` syntax. but if this is ok, then the rest looks fine

antgonza: The syntax comes from: http://legacy.datatables.net/usage/options, but the screenshot shows the differences.

PR #2599, opened 2018-05-30
Add qiimp xlsx

antgonza:


qiita_db/util.py
```diff
@@ -1730,6 +1733,34 @@ def _get_filehandle(filepath_or, *args, **kwargs):
     if _is_string_or_bytes(filepath_or):
         if h5py.is_hdf5(filepath_or):
             fh, own_fh = h5py.File(filepath_or, *args, **kwargs), True
+        elif filepath_or.endswith('.xlsx'):
```
ElDeveloper: This is going to fail when `filepath_or` is not a string because it won't have the `endswith` method, right? 

I think the XLSX parsing should be moved elsewhere this function seems too generic as it should only open a file path or file handle. Is there any motivation to have this here instead of in the sample/prep information parsing?

antgonza: Right! I initially was gonna add the code in `load_template_to_dataframe` but realized that it really just calls this function (see, https://github.com/biocore/qiita/blob/master/qiita_db/metadata_template/util.py#L100) as it just calls the open_file method. BTW note that the same method is available to open hdf5, text, or iostring. Thus, I believe is fine to fail if not a string as the other code should be able to handle and parse correctly those errors. Thoughts?

ElDeveloper: Fair enough, the use of this function is rather limited (see these [search results](https://github.com/biocore/qiita/search?utf8=✓&q=open_file&type=)). It still seems too odd to me that a function that's intended to return a file handle will parse a file write it as a txt file and return that file handle.

antgonza: Well, the idea of parsing the xmlx is that it will be transformed to txt and then use the regular prep/sample code ... ¯\_(ツ)_/¯ ... I don't have a better place to move it but if you do, let me know.

ElDeveloper: I think it makes more sense to have this outside of this function. I see as two very different things, (1) opening a file and (2) parsing a file. And this function opens the file and shouldn't be parsing it. Is there any technical limitation to putting this in load_template_to_dataframe? Perhaps as a helper function somewhere in there?

qiita_db/metadata_template/test/test_util.py
```diff
@@ -69,6 +71,31 @@ def test_load_template_to_dataframe(self):
...
+        exp.index.name = 'sample_name'
+        assert_frame_equal(obs, exp)
+
+        # test loading an empty qiimp file
+        fp = join(mfp, 'empty_qiimp_wb.xlsx')
+        with self.assertRaises(qdb.exceptions.QiitaDBColumnError) as error:
+            qdb.metadata_template.util.load_template_to_dataframe(fp)
+        self.assertEqual(
+            str(error.exception), "The 'sample_name' column is missing from "
+            "your template, this file cannot be parsed.")
```
AmandaBirmingham: This isn't a show-stopper, but I am concerned because I think the auto-generated exception is inaccurate ... an empty QIIMP template *will* contain a sample_name column, as shown below, and the tsv produced from it will also have include column--actually, the tsv will contain ONLY a row of column names :)  I fear that an error message indicating the sample_name column isn't there when it actually is will mislead users/future maintainers when they try to figure out why their upload failed.
screen shot 2018-06-01 at 4 45 27 pm

antgonza: @AmandaBirmingham, good point. The reason this was happening was all columns are empty so all of them are dropped and the final dataframe was empty, and the first test is to check for validity is if sample_name column is there. Anyway, I added a new tests (empty) and now the error should make more sense.

PR #2607, opened 2018-06-11
Fix 2606

antgonza:
1. Making iframe larger
2. Adding 2 buttons:
    screen shot 2018-06-11 at 1 53 30 pm


antgonza:
For Qiita to have a link to open Qiimp in or out the iframe this is the easy solution; the hard one is to inject it in the Qiimp code (basically extremely hard). The easiest to achieve the desired behavior is to add the button in Qiimp (break/exit iframe). 

In other words, Qiita will only have one link to the iframe and Qiimp is the one that will need to add a button to exit the iframe. Thoughts?

adswafford:
Hmm, I think I'm missing something. I'm fine with the link, just want it as
a button on the page with the Qiimp iframe.

I think what you're describing is that if we want to carry the contents of
whatever the user has input into Qiimp within the iframe to a new window,
that will be hard right? I think that this is fine to hold off on this and
just have the link the dropdown item is currently pointed to instead
represented by a button on the Qiimp page.

Does that make sense? It's just clunky to have the link in the dropdown.

antgonza:
Now it looks like this:
screen shot 2018-07-06 at 9 38 21 am

PR #2617, opened 2018-07-05
fix #2601

### LINKED ISSUE #2601: color code legend for visual job tracking: feature request
in the point-and-click interface of queueing up an analysis, it would be helpful to have a little legend with a color code that explains what the different color circles mean.  thanks!

https://qiita.ucsd.edu/analysis/description/17522/

antgonza:
The new visualization is:
screen shot 2018-07-05 at 8 06 46 am

### DIFF
--- a/qiita_pet/static/js/networkVue.js
+++ b/qiita_pet/static/js/networkVue.js
@@ -76,7 +76,7 @@ Vue.component('processing-graph', {
                   '</div>' +
                   '<div class="row">' +
                     '<div class="col-md-12">' +
-                      '<b>Click on the graph to navigate through it. Click circles for more information. This graph will refresh in <span id="countdown-span"></span> seconds or reload <a href="#" id="refresh-now-link">now</a><br/><span id="sphere-explanation"></span></b>' +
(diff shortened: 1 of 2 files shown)


antgonza:
Good question! (... and not sure). I see two options:
1. ignore those 2 elements
2. add more information for each of them

Both options means special cases (if/else) in the for loop so no preference there but I I like 1 more, thoughts?

ElDeveloper:
Ignoring those would probably be confusing since the gray color of type and queued/waiting are so similar. Perhaps a better idea is to divide the legend in two rows, one for statuses and another one for types?

antgonza:
New version:
screen shot 2018-07-05 at 12 37 11 pm

qiita_pet/static/js/networkVue.js
```diff
@@ -1087,6 +1088,29 @@ Vue.component('processing-graph', {
       vm.update_job_status();
     });
 
+    var circle_statuses = [];
+    var circle_types = [];
+    for (circle_name in vm.colorScheme) {
```
ElDeveloper: `circle_name` -> `var circle_name`

PR #2619, opened 2018-07-05
add ignore parent for command merging scheme

antgonza:
For shogun/shotgun we should ignore all parent information, while for other processing methods we should not. To control this, we are adding a new column in the software_command table. 

cc: @tanaes


antgonza:
No, I cc-ed @tanaes as this is something we discussed it's needed for shotgun ...

PR #2620, opened 2018-07-06
adding get_validator_jobs

antgonza:


qiita_db/processing_job.py
```diff
@@ -931,6 +907,27 @@ def children(self):
             for jid in qdb.sql_connection.TRN.execute_fetchflatten():
                 yield ProcessingJob(jid)
 
+    @property
+    def get_validator_jobs(self):
```
ElDeveloper: The pythonic thing would be to name this property `validator_jobs`.

qiita_db/test/test_processing_job.py
```diff
@@ -492,7 +492,7 @@ def test_complete_success(self):
         job.complete(True, artifacts_data=artifacts_data)
         self._wait_for_job(job)
         # Retrieve the job that is performing the validation:
-        val_job = qdb.processing_job.ProcessingJob(job.step.rsplit(" ", 1)[-1])
+        val_job = list(job.get_validator_jobs)[-1]
```
ElDeveloper: Why is this checking only the last job?

qiita_db/processing_job.py
antgonza: K, will change.

qiita_db/test/test_processing_job.py
antgonza: Not sure, that's the test it was before ...

ElDeveloper: haha ok.

ElDeveloper:
Thanks!

PR #2636, opened 2018-07-29
adds UsersListDBHandler

antgonza:
adding a new endpoint for user name/email retrieval


qiita_db/handlers/tests/test_user.py
```diff
@@ -34,7 +34,32 @@ def test_get(self):
...
+        obs = self.get('/qiita_db/users/')
+        self.assertEqual(obs.code, 400)
+
+    def test_get(self):
+        obs = self.get('/qiita_db/users/',
+                       headers=self.header)
+        self.assertEqual(obs.code, 200)
+
+        obs = loads(obs.body)
+        self.assertEqual(obs.keys(), ['data'])
```
ElDeveloper: As far as I understand this test can just check for dictionary equality instead of checking for the keys and then the values i.e. replace line 53 to 62 for something like:

```python
exp = {'data': [....]}
self.assertEqual(obs, exp)
```

qiita_pet/support_files/doc/source/dev/rest.rst
```diff
@@ -43,6 +43,8 @@ the important parts of the values are described in the Parameters column.
 |PATCH   | ``/qiita_db/artifacts/<artifact_id>/``                                            | ``op: operation``, ``path: path``,      | Retrieves the artifact information                  | ArtifactHandler            |
 |        |                                                                                   | ``value: value``                        |                                                     |                            |
 +--------+-----------------------------------------------------------------------------------+-----------------------------------------+-----------------------------------------------------+----------------------------+
+|GET     | ``/qiita_db/users/``                                                              |                                         | Retrieves the users emails and names                | UsersListDBHandler         |
```
ElDeveloper: This is a tricky one, I think it should be `Retrieves the email and name of all the users`.

PR #2638, opened 2018-08-14
rm QiitaStudySearch

antgonza:
This rm QiitaStudySearch, which is not used and superseded by redbiom. While rm QiitaStudySearch I realized that we have the same test with different parameters that expected the same output cause those variables weren't used in the newer version of the code, an example (True/False and expecting the same output in `qiita_db/test/test_util.py`):
```
        obs_info = qdb.util.generate_study_list([1, 2, 3, 4], True)
        self.assertEqual(obs_info, exp_info)
         obs_info = qdb.util.generate_study_list([1, 2, 3, 4], False)
        self.assertEqual(obs_info, exp_info)
```


ElDeveloper:
👍

antgonza:
Thanks!

PR #2643, opened 2018-08-17
fix #2632

### LINKED ISSUE #2632: Under documents, the Qiimp link does not work
The Qiimp link under documents does not work.  This also affects the "next" button at the bottom of the Getting Started Guide.

antgonza:
:maxdepth: is basically how many levels you want to show the menu/element, no :maxdepth: means all; thus removing them.


antgonza:
Just for clarity, the error I'm fixing was an error that I introduced when I merged master and dev this morning ... clearly I clicked on the wrong side of the bar while reviewing.

PR #2644, opened 2018-08-20
Fix 2623

antgonza:
A video is better than a 1000 words. 
delete

Now, during the video I realized that the sample ids are shown twice, which I already fixed here.

### DIFF
--- a/qiita_db/metadata_template/base_metadata_template.py
+++ b/qiita_db/metadata_template/base_metadata_template.py
@@ -604,30 +604,33 @@ class MetadataTemplate(qdb.base.QiitaObject):
             qdb.sql_connection.TRN.add(sql)
             return qdb.sql_connection.TRN.execute_fetchflatten()
 
-    def _common_delete_sample_steps(self, sample_name):
+    def _common_delete_sample_steps(self, sample_names):
         r"""Executes the common delete sample steps
 
         Parameters
         ----------
-        sample_name : str
-            The sample name to be erased
+        sample_names : list of str
+            The sample names to be erased
 
         Raises
         ------
         QiitaDBUnknownIDError
-            If the `sample_name` doesn't exist
+            If the `sample_names` doesn't exist
         """
-        if sample_name not in self.keys():
-            raise qdb.exceptions.QiitaDBUnknownIDError(sample_name, self._id)
+        keys = self.keys()
+        missing = [sn for sn in sample_names if sn not in keys]
+        if missing:
+            raise qdb.exceptions.QiitaDBUnknownIDError(missing, self._id)
 
         with qdb.sql_connection.TRN:
-            sql = 'DELETE FROM qiita.{0} WHERE sample_id=%s'.format(
-                self._table_name(self._id))
-            qdb.sql_connection.TRN.add(sql, [sample_name])
+            for sn in sample_names:
+                sql = 'DELETE FROM qiita.{0} WHERE sample_id=%s'.format(
+                    self._table_name(self._id))
+                qdb.sql_connection.TRN.add(sql, [sn])
 
-            sql = "DELETE FROM qiita.{0} WHERE sample_id=%s AND {1}=%s".format(
-                self._table, self._id_column)
-            qdb.sql_connection.TRN.add(sql, [sample_name, self.id])
+                sql = "DELETE FROM qiita.{0} WHERE sample_id=%s AND {1}=%s".format(
+                    self._table, self._id_column)
+                qdb.sql_connection.TRN.add(sql, [sn, self.id])
 
             qdb.sql_connection.TRN.execute()
 
--- a/qiita_db/metadata_template/prep_template.py
+++ b/qiita_db/metadata_template/prep_template.py
@@ -689,13 +689,13 @@ class PrepTemplate(MetadataTemplate):
             is_submitted = qdb.sql_connection.TRN.execute_fetchlast()
         return is_submitted
 
-    def delete_sample(self, sample_name):
-        """Delete `sample_name` from prep information file
+    def delete_sample(self, sample_names):
+        """Delete `sample_names` from prep information file
 
         Parameters
         ----------
-        sample_name : str
-            The sample name to be deleted
+        sample_name : list of str
+            The sample names to be deleted
 
         Raises
         ------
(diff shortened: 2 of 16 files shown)


qiita_db/metadata_template/base_metadata_template.py
```diff
@@ -604,31 +604,32 @@ def metadata_headers(cls):
...
-        sample_name : str
-            The sample name to be erased
+        sample_names : list of str
+            The sample names to be erased
 
         Raises
         ------
         QiitaDBUnknownIDError
-            If the `sample_name` doesn't exist
+            If the `sample_names` doesn't exist
```
ElDeveloper: I think this should be "if any of the `sample_names` don't exist`"

qiita_db/metadata_template/base_metadata_template.py
```diff
@@ -604,31 +604,32 @@ def metadata_headers(cls):
...
         """
-        if sample_name not in self.keys():
-            raise qdb.exceptions.QiitaDBUnknownIDError(sample_name, self._id)
+        keys = self.keys()
+        missing = [sn for sn in sample_names if sn not in keys]
+        if missing:
+            raise qdb.exceptions.QiitaDBUnknownIDError(missing, self._id)
```
ElDeveloper: Looks like this exception class is not meant to accept a list of strings. Maybe using a different subclass or formatting the missing sample names in a different way. Though, even then the text doesn't make a lot of sense, see [this line here](https://github.com/biocore/qiita/blob/master/qiita_db/exceptions.py#L81).

qiita_db/metadata_template/base_metadata_template.py
```diff
@@ -604,31 +604,32 @@ def metadata_headers(cls):
...
-            sql = 'DELETE FROM qiita.{0} WHERE sample_id=%s'.format(
-                self._table_name(self._id))
-            qdb.sql_connection.TRN.add(sql, [sample_name])
-
-            sql = "DELETE FROM qiita.{0} WHERE sample_id=%s AND {1}=%s".format(
-                self._table, self._id_column)
-            qdb.sql_connection.TRN.add(sql, [sample_name, self.id])
-
+            base_sql = 'DELETE FROM qiita.{0} WHERE sample_id=%s'
+            sql1 = base_sql.format(self._table_name(self._id))
```
ElDeveloper: What do sql1 and sql2 exactly do? It seems like you are performing deletes on 2 tables at once. Maybe add some comments?

qiita_db/metadata_template/base_metadata_template.py
```diff
@@ -604,31 +604,32 @@ def metadata_headers(cls):
...
+            sql2 = '{0} AND {1}=%s'.format(
+                base_sql.format(self._table), self._id_column)
+            for sn in sample_names:
```
ElDeveloper: Any reason why instead of a for loop not doing something like `DELETE FROM qiita.xxxx WHERE sample_id in (....)`? Seems like it would turn N queries into 1 query.

qiita_db/metadata_template/prep_template.py
```diff
@@ -689,13 +689,13 @@ def is_submitted_to_ebi(self):
...
-    def delete_sample(self, sample_name):
-        """Delete `sample_name` from prep information file
+    def delete_sample(self, sample_names):
+        """Delete `sample_names` from prep information file
 
         Parameters
         ----------
-        sample_name : str
-            The sample name to be deleted
+        sample_name : list of str
```
ElDeveloper: sample_name -> sample_names

qiita_db/metadata_template/prep_template.py
```diff
@@ -689,13 +689,13 @@ def is_submitted_to_ebi(self):
             is_submitted = qdb.sql_connection.TRN.execute_fetchlast()
         return is_submitted
 
-    def delete_sample(self, sample_name):
-        """Delete `sample_name` from prep information file
+    def delete_sample(self, sample_names):
```
ElDeveloper: delete_sample -> delete_samples

qiita_db/metadata_template/sample_template.py
```diff
@@ -163,28 +162,28 @@ def columns_restrictions(self):
         """
         return qdb.metadata_template.constants.SAMPLE_TEMPLATE_COLUMNS
 
-    def delete_sample(self, sample_name):
-        """Delete `sample_name` from sample information file
+    def delete_sample(self, sample_names):
```
ElDeveloper: delete_sample -> delete_samples

qiita_db/metadata_template/sample_template.py
```diff
@@ -163,28 +162,28 @@ def columns_restrictions(self):
...
+        """Delete `sample_names` from sample information file
 
         Parameters
         ----------
-        sample_name : str
+        sample_names : array of strings
```
ElDeveloper: array -> list

qiita_pet/static/js/sampleTemplateVue.js
```diff
@@ -206,21 +206,32 @@ Vue.component('sample-template-page', {
...
-        data: {'op': 'remove', 'path': vm.studyId + '/samples/' + sample},
-        success: function(data) {
-          vm.rowId = rowId;
-          vm.rowType = 'sample';
-          vm.startJobCheckInterval(data['job']);
-        },
-        error: function (object, status, error_msg) {
-          bootstrapAlert("Error deleting sample: " + error_msg, "danger")
+      var total_samples = samples.length;
+      if (total_samples == 0){
```
ElDeveloper: == -> ===

qiita_pet/static/js/sampleTemplateVue.js
```diff
@@ -206,21 +206,32 @@ Vue.component('sample-template-page', {
...
+        alert('No samples selected!');
+      } else {
+        if (confirm('Are you sure you want to delete ' + total_samples + ' samples?')) {
+          var sample_names = [];
+          samples.each(function(){
```
ElDeveloper: Minor, but you could change this to use the map function so that you are not explicitly pushing each sample name into the sample_names object.

qiita_db/metadata_template/base_metadata_template.py
antgonza: Because the `many` option expects an array of arrays. In other words, we will need to change sample_names from: `['Sa', 'Sb', ...]` to `[['Sa'], ['Sb'], ...]` so it will not save us anything.

qiita_pet/static/js/sampleTemplateVue.js
antgonza: Not possible; well, not easy, cause samples it's a jquery object vs. an array ...

ElDeveloper: I see, thanks.

PR #2646, opened 2018-08-23
fix #2622

### LINKED ISSUE #2622: admin display to see active plugins and commands
Add a summary page for admins showing the active plugins and commands, this will be super helpful during deploys.

antgonza:
Screenshots:
- Menu
screen shot 2018-08-23 at 9 28 27 am

- Software Display
screen shot 2018-08-23 at 9 28 20 am

### DIFF
--- a/qiita_pet/handlers/software.py
+++ b/qiita_pet/handlers/software.py
@@ -1,0 +1,31 @@
+# -----------------------------------------------------------------------------
+# Copyright (c) 2014--, The Qiita Development Team.
+#
+# Distributed under the terms of the BSD 3-clause License.
+#
+# The full license is in the file LICENSE, distributed with this software.
+# -----------------------------------------------------------------------------
+
+from __future__ import division
+
+from tornado.gen import coroutine
+from tornado.web import HTTPError
+
+from qiita_core.util import execute_as_transaction
+from qiita_db.software import Software
+from .base_handlers import BaseHandler
+
+
+class SoftwareHandler(BaseHandler):
+    def check_access(self):
+        if self.current_user.level not in {'admin', 'dev'}:
+            raise HTTPError(405, reason="User %s doesn't have sufficient "
+                            "privileges to view error page" %
+                            self.current_user)
+
+    @coroutine
+    @execute_as_transaction
+    def get(self):
+        self.check_access()
+        software = Software.iter(False)
+        self.render("software.html", software=software)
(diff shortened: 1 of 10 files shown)


qiita_pet/handlers/software.py
```diff
@@ -0,0 +1,31 @@
...
+
+from qiita_core.util import execute_as_transaction
+from qiita_db.software import Software
+from .base_handlers import BaseHandler
+
+
+class SoftwareHandler(BaseHandler):
+    def check_access(self):
+        if self.current_user.level not in {'admin', 'dev'}:
+            raise HTTPError(405, reason="User %s doesn't have sufficient "
```
ElDeveloper: Minor, but this should be a 403 error.

qiita_pet/handlers/software.py
```diff
@@ -0,0 +1,31 @@
...
+                            "privileges to view error page" %
+                            self.current_user)
```
ElDeveloper: This should be `self.current_user.email` (I think), otherwise the message will format the default Python repr i.e. `<qiita_db.user.User object at 0x10cceb150>`.

qiita_pet/templates/software.html
```diff
@@ -0,0 +1,69 @@
...
+              </tbody>
+            </table>
+          </td>
+        </tr>
+      {% end %}
+      </tbody>
+    </table>
+  {% else %}
+      <div id="jumbotron" class="jumbotron">
+          <h1><span class="glyphicon glyphicon-thumbs-down"></span> There is no plugin software in this system </h1>
```
ElDeveloper: What about: `There are no plugins in this system`.

antgonza:
I think the information is important because a plugin can be active while a command can be not active ... cause we disable it or something went wrong ... so I rather keep it ...

ElDeveloper:
My bad, I meant to say that you should keep that information, just not as its own column. The table can become a single column if you put two the status and the name in the same column. That's what I was trying to refer to with my example above (I was just referring to the inner tables).

qiita_pet/handlers/software.py
antgonza: k, replacing everywhere ...

qiita_pet/handlers/software.py
antgonza: same ...

qiita_pet/templates/software.html
antgonza: k

antgonza:
Now, it looks like this:
screen shot 2018-08-28 at 5 44 20 pm

ElDeveloper:
Awesome, looks great! Tests seem to be failing though.

ElDeveloper:
Thanks @antgonza

Issue #2648, opened 2018-08-28
Incorrect formatting of data for PATCH requests


ElDeveloper:
Several of the PATCH requests in Qiita are sending data via the request's arguments, whereas the standard expects that the data be sent via the body of the request as JSON.

antgonza:
Thanks. Just to better classify this issue, what are the implications of this?

ElDeveloper:
One possible implication (which I found in #2649) was that the data 
could possibly be mangled.

antgonza:
Mangle how? Like it will not work?

PR #2649, opened 2018-08-28
Add specimen_id attribute to the study object

ElDeveloper:
This PR adds a new attribute `specimen_id_column` to the Study object. This attribute holds the name of the column that is used to match tubes (or specimens) to sample identifiers in Qiita. The user interface has been updated to include a drop-down menu that lets the user select from any of the categories in the sample information that have unique values.

I have also modified the SampleTemplate.delete_column to prevent the deletion of columns that are currently selected as a `specimen_id_column`.

You may notice that there are a few unrelated changes, this is because the functionality to handle PATCH requests in the Study object was limited to only operate on `tags`. This PR has extended that functionality to also operate on the `specimen_id_column`. While making these changes I noticed that the AJAX requests were being formatted incorrectly. The PATCH standard states that the contents be delivered via the body of the request in JSON format, not via the request's arguments. I only changed this for the requests pertaining to the study object. However, there's many other instances of this problem elsewhere in the code base. For now, I've opened https://github.com/biocore/qiita/issues/2648. 

The functionality itself can be seen here (note that only unique columns are selectable from the dropdown menu):

specimen-id

And when the user tries to delete that column:

screen shot 2018-08-28 at 3 49 26 pm

cc @AmandaBirmingham since this is related to https://github.com/jdereus/labman/issues/273


qiita_db/metadata_template/base_metadata_template.py
```diff
@@ -1516,3 +1527,17 @@ def validate(self, restriction_dict):
                 "columns:\n\t%s.\nSee the Templates tutorial for a description"
                 " of these fields." % ";\n\t".join(warning_msg),
                 qdb.exceptions.QiitaDBWarning)
+
+    def unique_columns(self):
```
antgonza: This is gonna be extremely slow for large info files. Examples from the real system:
```python
from qiita_db.metadata_template.sample_template import SampleTemplate

def unique_columns(st):
    df = st.to_dataframe()
    n = len(df)
    return {k for k, v in df.nunique().iteritems() if v == n}

In [22]: %timeit unique_columns(SampleTemplate(550)) # moving pics
1 loop, best of 3: 662 ms per loop

In [23]: %timeit unique_columns(SampleTemplate(10317)) # AGP
1 loop, best of 3: 1min 16s per loop
```

As you can imagine something blocking for 1min 16s is not desirable and more in the sample info page which is pretty common and we normally expect fasts replies. 

Options:
- Do everything in DB - preferred but not sure if much faster
- Do not display only unique values but all columns - and option but don't love it
- Store unique values in redis and change every time the sample id is replaced as part of the update - kind of cool to store all summaries as part of the update 
- Other?

qiita_db/metadata_template/base_metadata_template.py
```diff
@@ -1516,3 +1527,17 @@ def validate(self, restriction_dict):
...
+        Returns
+        -------
+        set of str
+            Returns a set of the columns that are unique in a template object.
+        """
+
+        df = self.to_dataframe()
+
+        n = len(df)
+        return {k for k, v in df.nunique().items() if v == n}
```
antgonza: items -> iteritems cause items doesn't exist in the pandas version we have in qiita

qiita_pet/handlers/api_proxy/studies.py
```diff
@@ -436,6 +437,15 @@ def study_tags_patch_request(user_id, study_id,
             message = study.update_tags(User(user_id), req_value)
             return {'status': 'success',
                     'message': message}
+        elif attribute == 'specimen_id_column':
+            try:
+                study.specimen_id_column = req_value
```
antgonza: Wondering if this step (as it has to verify that the column is unique and set the SQL values) should be a job vs. a simple set ...

qiita_db/test/test_study.py
```diff
@@ -875,6 +881,31 @@ def test_study_tags(self):
...
+
+    def test_specimen_id_column_no_sample_information(self):
+        empty = qdb.study.Study.create(
+            qdb.user.User('test@foo.bar'), "Fried duck microbiome",
+            self.info)
+        with self.assertRaises(qdb.exceptions.QiitaDBLookupError):
+            empty.specimen_id_column = 'foo'
+
+        # cleaning up the created study
+        qdb.study.Study.delete(empty._id)
```
wasade: For this to always clean up, it may be worth wrapping the assertion in a try/finally

qiita_pet/handlers/study_handlers/tests/test_base.py
```diff
@@ -67,24 +69,48 @@ def test_get(self):
...
         arguments = {'op': 'replace', 'path': '/tags',
-                     'value[]': "['testA', 'testB']"}
-        obs = self.patch('/study/tags/b', headers=self.header, data=arguments)
-        self.assertEqual(obs.code, 400)
+                     'value': ['testA', 'testB']}
+        obs = self.patch('/study/100000000000', headers=self.header,
+                         data=arguments, asjson=True)
+        self.assertEqual(json_decode(obs.body), {'status': 'error', 'message':
+                         'Study does not exist'})
+        self.assertEqual(obs.code, 200)
```
wasade: Should this be 404?

qiita_pet/handlers/api_proxy/studies.py
ElDeveloper: This is only checking one column, so I don't think it will take that much time. The request should also be asynchronous from the UI so it shouldn't block any interaction. Maybe I'm overlooking something?

qiita_pet/handlers/study_handlers/tests/test_base.py
ElDeveloper: Yes, I agree with you. Opened here: https://github.com/biocore/qiita/issues/2655

antgonza:
Code looks good, a few more questions/requests:
- Can you add explanation in the help or even beside the GUI selector of what specimen_id is? I can see users being confused about this and think is a requirement.
- Is it possible, via the GUI, to unselect specimen_id? I think this is desirable, in case a user sets it by mistake, and if this is possible, what are the implications for when it has been used for platting?

ElDeveloper:
> Can you add explanation in the help or even beside the GUI selector of what specimen_id is? I can see users being confused about this and think is a requirement.

Great idea, I've added a small `?` badge with a tooltip.

> Is it possible, via the GUI, to unselect specimen_id? I think this is desirable, in case a user sets it by mistake

Yes, now it is possible.

> ....  what are the implications for when it has been used for platting?

That should be fine since this category is mainly used for mapping the tube identifiers to the Qiita sample ids. Hence if this category ceased to exist (or was deselected) the plate viewer would instead show qiita sample identifiers.

--- 

For reference here's how the UI currently looks. Things to note (1) the question mark badge and the tooltip, (2) the new default in the select i.e. `None (not available)`, and (3) how this information can be updated to either a column or to None.

specimen-id-part2

antgonza:
Thanks! Did you notice that the button beside the selector is not even, it gets cut, I guess this depends on the size of the window and the solution is to make sure the number of "bootstrap" columns is correct, can you fix?

ElDeveloper:
Good catch, fixed. Looks like I had incorrectly set the column width of the button.

antgonza:
Sorry, just realized that: https://github.com/jdereus/labman/issues/175; so either we decide on a name and use it here or we remove the name from the tooltip ...

ElDeveloper:
I would prefer removing the name. What about "Only for use with external LIMS systems".

ElDeveloper:
Thanks for the feedback!

PR #2658, opened 2018-09-01
Fixes #2398

### LINKED ISSUE #2398: UTF-8 chars in title break analysis
First detected by @sjanssen2. 

Basically, if your title has a UTF8 char, the analysis breaks during validation (my guess is that this will be fine for Q2).

charles-cowart:
Strips UTF-8 characters from study titles
UTF-8 characters that are not also printable ASCII characters are stripped
from new study titles, as well as existing studies flagged 'public'. This is
because study titles are passed as metadata to new analyses, and some analysis
packages may not work well with them.


antgonza:
A few comments so we can review:
- Flake8 errors, please fix
- There are some commented out lines, please remove if not important
- We try to only have one patch for release, could you add your code to the end of 66.sql and 66.py? Note that you might need to pull the latest dev ...

qiita_db/support_files/patches/python_patches/66.py
```diff
@@ -0,0 +1,14 @@
...
+# characters, it becomes important to remove them from study titles, as
+# they are used as metadata/identifiers when creating new analyses.
+from qiita_db.study import Study
+from re import sub
+
+studies = Study.get_by_status('public')
+
+for study in studies:
+    title = study.title
+    new_title = sub(r'[^\x20-\x7E]+', '', title)
```
antgonza: Perhaps to speed up, it should only replace if the title changed ...

qiita_pet/templates/edit_study.html
```diff
@@ -156,8 +160,11 @@ <h3>
     {% if form_item.label.text == 'Environmental Packages' %}
         {% set kwargs['size'] = len(form_item.choices) %}
         {% set additional_info = 'You can select multiple entries by control-clicking (mac: command-clicking)' %}
-    {% elif form_item.label.text == 'Principan Investigator'%}
+    {% elif form_item.label.text == 'Principal Investigator'%}
```
antgonza: thanks!

qiita_db/support_files/patches/python_patches/66.py
charles-cowart: doh!

charles-cowart: done

qiita_db/support_files/patches/python_patches/66.py
```diff
@@ -1,15 +1,26 @@
...
 from qiita_db.util import convert_to_id
+from qiita_db.study import Study
+from re import sub
 
+# August 31, 2018
+# Strip any UTF-8 characters that are not also printable ASCII characters
+# from study titles. As some analysis packages cannot interpret UTF-8
+# characters, it becomes important to remove them from study titles, as
+# they are used as metadata/identifiers when creating new analyses.
+studies = Study.get_by_status('public')
```
antgonza: Just realized that this will only loop over the public studies and we need to check _all_ studies ...

charles-cowart: Looking through the methods, I haven't yet seen a way to enumerate through all of the studies, regardless of status (Study.get_all() or similar). It looks like the source of truth for the set of valid status strings is ultimately the qiita.visibility table, which lists the following:
   awaiting_approval
   sandbox
   private
   public

A nested loop to process each status in turn isn't horrible, but this might be a good time to write a get_all(), if it's not already available.

Another alternative would be to simply perform the same action in SQL directly. It looks like the only state we're changing through the API is study_title in the qiita.study table. The update statement below has already been tested, and could be placed in 66.sql:

update qiita.study set study_title = REGEXP_REPLACE(study_title,'[^\x20-\x7E]+','','g');

What do you think?

antgonza: Good point, perhaps we should open an issue about making sure that all objects have an __iter__ method. Anyway, we normally do:
```
studies = Study.get_by_status('private').union(
    Study.get_by_status('public')).union(Study.get_by_status('sandbox'))
```

antgonza:
Thanks @mortonjt

qiita_db/support_files/patches/python_patches/66.py
```diff
@@ -1,15 +1,30 @@
...
 from json import loads, dumps
-
 from qiita_db.sql_connection import TRN
 from qiita_db.software import Software, Command
 from qiita_db.exceptions import (QiitaDBError, QiitaDBDuplicateError)
 from qiita_db.util import convert_to_id
+from qiita_db.study import Study
+from re import sub
 
+# August 31, 2018
```
ElDeveloper: This patch needs to be moved to its own file.

antgonza: I can do this in my PR as I'm fixing conflicts; however, in the past we have decided to have only one patch per release cycle ... thus; the idea of having in the same patch

Issue #2579, opened 2018-05-03
Error: Code 500! should give incorrect password error instead


stephanieorch:
A user just got this error:

![error](https://user-images.githubusercontent.com/31080457/39585537-e6013fe2-4ea9-11e8-86c5-0eeb9281bf02.png)

When Antonio looked at, the error should be saying something along the lines of IncorrectPasswordError("Bad password given!") instead.

stephanieorch:
Along with this: when you create a password with special characters you get a message saying your password isn't verified rather than getting a "special characters can't be used" error.

charles-cowart:
Just to confirm, the cause of the above screenshot is a user entering a password with special characters in it? Or is the special characters case in addition to whatever caused the above screenshot?

stephanieorch:
When the user tried to make an account, he received the screenshotted error. To fix this, he just had to input a different password that didn't contain special characters. 
So: Replate "ERROR: CODE 500!" with something like "Error: Passwords cannot contain special characters".

antgonza:
Just to confirm, the current html doesn't show the password requirements/restrictions and the error doesn't tell you about this and give you the chance to try again ...

stephanieorch:
Yes

charles-cowart:
Just to reiterate, the 500 error is caused because the server side is expecting valid inputs for password & confirm_password; error-handling appears to be performed entirely on the client-side. What I've done is add a custom handler (as in, not a JQuery rule) to validate that the password only contains printable ASCII characters (no UTF-8).

This should prevent the user from ever seeing the above 500 error; instead the user will need to correct their input before they can submit their query. If we'd still like to add server-side parameter checking for this option, I would recommend changing the status code from 500 to 400 (bad parameter) or 422 (unprocessable entity).

![password](https://user-images.githubusercontent.com/42684307/45114295-7755de80-b101-11e8-8055-dc82c24f9bba.gif)

antgonza:
Fixed: https://github.com/biocore/qiita/pull/2659

PR #2659, opened 2018-09-04
Fixes #1802 #2579

### LINKED ISSUE #1802: reset password errors are not shown in GUI
When users try to reset their password and the password is not valid the error is not shown to the user and only a 500 error is shown.

charles-cowart:
Password validation extended to allow only printable ASCII characters
A custom validator was written to check whether a password string contains any UTF-8 characters that are not also printable ASCII characters and return false. if true.

### DIFF
--- a/qiita_pet/static/js/password_validation.js
+++ b/qiita_pet/static/js/password_validation.js
@@ -6,29 +6,39 @@ function dualpass_validator() {
     $(this).popover('show');
   });
 
+  function validatePasswordIsASCIIOnly(value)
+  {
+    if(value.match(/^[\x20-\x7E]{1,}$/)) {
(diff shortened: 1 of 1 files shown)


antgonza:
Wondering if you were also planning to address https://github.com/biocore/qiita/issues/2579 ? Note that the main goal of that issue is that the user gets a message letting them know that the password is not valid ...

antgonza:
Just to be clear, I'm not sure the code here actually solves the problems describe on those issues: mainly makes the user aware of what's wrong in an informative way vs the 500 error code. @charles-cowart, could you use LICEcap to create a small video of the behavior with these changes to facilitate review? Thanks!

PR #2660, opened 2018-09-05
Scp

antgonza:


qiita_ware/commands.py
```diff
@@ -21,6 +25,144 @@
...
+
+    directory = p_url.path
+    valid_files = _list_valid_files(ssh, directory)
+    file_paths = [join(directory, f) for f in valid_files]
+
+    # step 2: download files
+    scheme = p_url.scheme
+    # note that scp/sftp's code seems similar but the local_path/localpath
+    # variable is different within the for loop
+    if scheme == 'scp':
```
ElDeveloper: *From my previous review*

I think this can be broken down to be a bit simpler, for example:

```python
     if scheme == 'scp':
        client = SCPClient(ssh.get_transport())
     elif scheme == 'sftp':
        client = ssh.open_sftp()

     for f in file_paths:
         download = partial(client.get,
                            local_path=join(destination, basename(f)))
         download(f)
```

Also it should throw an exception if the scheme is not supported.

qiita_ware/private_plugin.py
```diff
@@ -377,6 +378,45 @@ def delete_analysis(job):
...
+    Parameters
+    ----------
+    job : qiita_db.processing_job.ProcessingJob
+        The processing job performing the task
+    """
+    with qdb.sql_connection.TRN:
+        url = job.parameters.values['url']
+        private_key = job.parameters.values['private_key']
+        try:
+            list_remote(url, private_key)
```
ElDeveloper: *From my previous review*:

What is happening to the output of this function, it seems like it is not being utilized anywhere and that this job does nothing(?)

qiita_ware/test/test_commands.py
```diff
@@ -31,6 +34,65 @@
...
+
+    def test_download_scp(self):
+        """Tests remote file listing using private key and scp"""
+        download_remote('scp://localhost:'+self.remote_dir_path,
+                        self.test_ssh_key, self.temp_local_dir)
+        remote_filename_list = self._get_valid_files(self.remote_dir_path)
+        self.assertEqual(remote_filename_list, self.exp_files)
+
+    def test_download_sftp(self):
+        """Tests remote file listing using private key and sftp"""
```
ElDeveloper: Although this was my previous comment:

> should be Tests remote file downloading ...

I think we stopped adding docstrings to test cases so I would suggest removing them altogether.

qiita_ware/commands.py
antgonza: Sorry, forgot to mention this.
1. The split is not really possible because, as the note above says, scp and sftp use a different variable local_path vs localpath; and couldn't figure out how to make it play nice with your above come - note that most of the pushes were due errors on these lines.
2. The exception is gonna be thrown a few lines above, when: `ssh = _ssh_session(p_url, private_key)`

qiita_ware/private_plugin.py
antgonza: ... also forgot to add a message ... this will store the information in redis so the GUI can retrieve it ... I'm working on this PR and should be ready pretty soon.

qiita_ware/test/test_commands.py
antgonza: K

qiita_ware/private_plugin.py
ElDeveloper: I see, so the redis support will be added in the next PR?

antgonza: yup, yup ...

qiita_ware/commands.py
ElDeveloper: I see, thanks for pointing that out. Subtle differences in the api 😑

qiita_ware/commands.py
```diff
@@ -21,6 +25,144 @@
...
+        scp = SCPClient(ssh.get_transport())
+        for f in file_paths:
+            download = partial(
+                scp.get, local_path=join(destination, basename(f)))
+            download(f)
+    elif scheme == 'sftp':
+        sftp = ssh.open_sftp()
+        for f in file_paths:
+            download = partial(
+                sftp.get, localpath=join(destination, basename(f)))
```
ElDeveloper: This for loop is missing a `download` call right? Since the tests are passing I assume this is not being tested?

antgonza:
Thanks!!

PR #2662, opened 2018-09-06
Adding SCP GUI

antgonza:
First review/merge: https://github.com/biocore/qiita/pull/2658 because I already pulled those changes to avoid future SQL patch conflicts.

See functionality, just testing listing cause download works the same way but takes much longer cause it downloads the files:
ssh-download

Note that I had to add study_id to the list_remote_files method so we could store in redis the results.


antgonza:
Thanks for review @charles-cowart, what about merging and then I add that comment in the next pr as I need to fix conflicts on that one too ...

charles-cowart:
Sure - you'd like me to do the merge then?

antgonza:
👍

PR #2664, opened 2018-09-10
Plugin outdate version

antgonza:
First review/merge https://github.com/biocore/qiita/pull/2662.

Notes (sending to qiita-admin for further discussion):
1. The deprecation flags are shown if the plugin is not active so this info is not stored in the DB. This means that during start if a plugin doesn't start correctly the artifacts will be marked as outdated until they get started properly. This can only happen during restart.
2. If an analysis was created with an old version of qiime2, it will be marked as deprecated - which is not 100% accurate but perhaps fine 
3. This doesn't affect the loading of artifacts/context in Redbiom but we will need to figure out there how to deal with deprecation and how often we will need to update things
4. For the artifact listing (how we create analysis) decided to not display deprecated artifacts as this will make the displaying nicer (less artifacts) but the other option is to display it and add something to show that it's deprecated. Note that you can still add the deprecated artifacts via the study page.

Some examples:
- How deprecated artifacts will look in an analysis:
deprecated-analysis

- Deprecated artifacts in study page:
deprecated-study

- Artifact listing now:
study-artifact-before
vs. before: 
study-artifact-now

### DIFF
--- a/qiita_pet/static/js/qiita.js
+++ b/qiita_pet/static/js/qiita.js
@@ -314,17 +314,20 @@ function format_biom_rows(data, row, for_study_list = true, samples = null) {
 
   // grouping by processing_method, data_type and parameters
   $.each(data, function (idx, info) {
-    if (typeof info !== 'string' && !(info instanceof String)) {
-      var algorithm = info.algorithm;
-      if (!(algorithm in processing_method)) {
-        processing_method[algorithm] = {};
-      }
+    // ignore the artifacts that were generated with software that is not active
+    if (info['active']) {
+      if (typeof info !== 'string' && !(info instanceof String)) {
+        var algorithm = info.algorithm;
+        if (!(algorithm in processing_method)) {
+          processing_method[algorithm] = {};
+        }
 
-      var data_type = info.data_type + ' (' + info.target_subfragment.join(', ') + ')';
-      if (!(data_type in processing_method[algorithm])) {
-        processing_method[algorithm][data_type] = [];
+        var data_type = info.data_type + ' (' + info.target_subfragment.join(', ') + ')';
+        if (!(data_type in processing_method[algorithm])) {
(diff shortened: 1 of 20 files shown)


qiita_pet/static/js/qiita.js
```diff
@@ -314,17 +314,20 @@ function format_biom_rows(data, row, for_study_list = true, samples = null) {
...
+        var algorithm = info.algorithm;
+        if (!(algorithm in processing_method)) {
+          processing_method[algorithm] = {};
+        }
 
-      var data_type = info.data_type + ' (' + info.target_subfragment.join(', ') + ')';
-      if (!(data_type in processing_method[algorithm])) {
-        processing_method[algorithm][data_type] = [];
+        var data_type = info.data_type + ' (' + info.target_subfragment.join(', ') + ')';
+        if (!(data_type in processing_method[algorithm])) {
```
ElDeveloper: Is this checking that data_type is a substring of processing_method.algorithm?

antgonza: yup, basically this is grouping the artifacts per data_type and it doesn't exits, it creates it ... a classic:
```
myvar = {}
for x in y:
   if x not in myvar:
      myvar[x] = init
  .....
```

antgonza:
Note that the changes are basically adding a new column to the qiita database, which solves the 2 issues @ElDeveloper raised. Also decided to add this new info into the admin summary and this is how a deprecated software looks:
screen shot 2018-09-12 at 8 23 12 am

qiita_pet/static/js/qiita.js
ElDeveloper: I see, thanks!

ElDeveloper:
Thanks @antgonza, this looks great! Lets just wait for tests to finish.

PR #2667, opened 2018-09-13
Qsub scripts

antgonza:


antgonza:
This is adding code that is already in production so we know it works. Also, is just so we can have it in the repo as reference. It was added by @wasade but had some flake8 errors, which I fixed here. Thus, merging.

PR #2668, opened 2018-09-13
Dev 2553

charles-cowart:
Template classes now manage their own class-level lists of forbidden words. Incoming data passed to Template.extend(), Template.update(), or other functions that internally call Template._clean_validate_template() will now validate column names against an appropriate list.

I removed a method from util.py that was especially specific to Template._clean_validate_template(); it was only used by this method, and a large component of its functionality was to aggregate exceptional conditions so as to pass one large 'meta-exception' back to the UI for consumption. This seemed very specific to the needs of this method, so I integrated the functionality into MetadataTemplate.


antgonza:
Changing this pull request to dev ...

qiita_db/metadata_template/base_metadata_template.py
```diff
@@ -50,6 +50,11 @@
 from qiita_core.exceptions import IncompetentQiitaDeveloperError
 import qiita_db as qdb
 
+if PY3:
```
antgonza: We only support python 2.7 in qiita so `PY3` will never be true; however, in the future we will like to move to python3 ... anyway, IMOO this should be removed.

qiita_db/metadata_template/base_metadata_template.py
```diff
@@ -513,7 +523,12 @@ def _clean_validate_template(cls, md_template, study_id,
         # In the database, all the column headers are lowercase
         md_template.columns = [c.lower() for c in md_template.columns]
 
-        # Droping/Ignoring internal generated colums
+        # Droping/Ignoring internal generated columns:
+        # A legacy feature of _clean_validate_template() is that it silently
+        # removes these columns before validating. It would be better to make
```
antgonza: Could you expand why will it be better to make it optional? Is this for testing? Do you think is worth adding an issue about this?

qiita_db/metadata_template/prep_template.py
```diff
@@ -91,6 +91,13 @@ class PrepTemplate(MetadataTemplate):
     _id_column = "prep_template_id"
     _sample_cls = PrepSample
     _filepath_table = 'prep_template_filepath'
+    # This has been explicitly redefined here for clarity, even though
```
antgonza: I'll argue that this is the cool thing about objects, you define it in parent and it's inherit in the children so no need for redefinition. Anyway, I think the actual implementation should be that:
- MetadataTemplate: _forbidden_words = {} or None, {} will be easier
- Leave _forbidden_words in SampleTemplate and PrepTemplate as they are

qiita_db/metadata_template/test/test_base_metadata_template.py
```diff
@@ -67,6 +67,60 @@ def test_clean_validate_template(self):
         with self.assertRaises(IncompetentQiitaDeveloperError):
             MT._clean_validate_template(None, 1)
 
+    def test_identify_forbidden_words(self):
+        """_identify_forbidden_words_in_column_names returns forbidden words
```
antgonza: Sometime ago we decided to not add docstrings to tests, could you remove? Applies to all methods below ...

qiita_db/metadata_template/base_metadata_template.py
```diff
@@ -513,7 +523,12 @@ def _clean_validate_template(cls, md_template, study_id,
...
+        # this optional. These two keywords are also in forbidden words lists.
```
antgonza: Adding here so is in the same block but this is in response to your comment in the tests (*). The reason we have this silent removal is for usability; the normal path is that users add their info files (before they were called templates) without these cols, qiita adds them, the user downloads the new file, adds/modifies columns, uploads. In the past, before these lines, the code will raise an error and will ask the user to remove them, which is less than ideal ...

qiita_db/metadata_template/test/test_prep_template.py
```diff
@@ -735,6 +735,65 @@ def test_clean_validate_template(self):
...
         assert_frame_equal(obs, exp)
 
+    def test_clean_validate_template_no_forbidden_words1(self):
+        PT = qdb.metadata_template.prep_template.PrepTemplate
+        self.metadata.rename(columns={'center_name': 'sampleid'}, inplace=True)
+        with self.assertRaises(qdb.exceptions.QiitaDBColumnError):
+            PT._clean_validate_template(self.metadata, 2)
+
+    '''
+    Commenting these two out for now. There is legacy code in
```
antgonza: (*) see comments above. BTW I'm OK with removing/replacing these tests as long as we have some to make sure that the automatic drop happens ...

qiita_db/metadata_template/test/test_sample_template.py
```diff
@@ -22,6 +22,8 @@
 from qiita_core.exceptions import IncompetentQiitaDeveloperError
 import qiita_db as qdb
 
+from qiita_db.metadata_template.constants import SAMPLE_TEMPLATE_COLUMNS
```
antgonza: IMOO kind of odd that we are importing all `qiita_db as qdb` and then importing this. If you want to avoid some long strings what about in the `setUp` so something like `self.STC = qdb.metadata_template.constants.SAMPLE_TEMPLATE_COLUMNS` and then just call `self.STC`?

qiita_db/metadata_template/test/test_sample_template.py
```diff
@@ -825,6 +829,61 @@ def test_clean_validate_template_no_invalid_chars(self):
...
+            ST._clean_validate_template(self.metadata, 2)
+
+    def test_clean_validate_template_no_forbidden_words(self):
+        ST = qdb.metadata_template.sample_template.SampleTemplate
+        self.metadata.rename(columns={'taxon_id': 'sampleid'}, inplace=True)
+        with self.assertRaises(qdb.exceptions.QiitaDBColumnError):
+            ST._clean_validate_template(self.metadata, 2)
+
+    '''
+    Commenting these two out for now. There is legacy code in
```
antgonza: (*) Same comments than above.

qiita_db/metadata_template/base_metadata_template.py
charles-cowart: Agreed, will do.

qiita_db/metadata_template/test/test_base_metadata_template.py
charles-cowart: Sure thing.

qiita_db/metadata_template/test/test_sample_template.py
charles-cowart: Agreed, and it yes, it was to avoid some really long identifier names. How about using this top-level alias, instead of making it a class variable?

STC = qdb.metadata_template.constants.SAMPLE_TEMPLATE_COLUMNS

antgonza: will that make it a global variable? I always _feel funny_ about them ...

qiita_db/metadata_template/prep_template.py
charles-cowart: Agreed, however there are tests in test_base_metadata_template.py to test the forbidden words method that wouldn't mean much with set = {}. I can move them to one or both of the sub-class tests, if we're okay with testing parent functionality in children classes. 

Just to round out the discussion, do you think in general, we'll be redefining the set of forbidden words every time we create a new Template? Or do you think in general, new Templates will need to inherit a set of basic forbidden words common to every Template? I ask because they're currently defined as sets, and might be better off as lists, so that they can be appended to.

antgonza: I think each template will have their own rules but there is a chance that some of them will overlap ... now, something interesting is that the main method `_clean_validate_template` has cls._check_subclass() and these auxiliary methods do not. I guess we should add ...

qiita_db/metadata_template/base_metadata_template.py
charles-cowart: I agree, making it optional might be a little much, if our intention is to always have it on. As you said, it is good for usability. My concern is that it be made explicitly known to the caller, because it is silent behavior that did catch me off guard; it might just be as simple as writing it into the method docs.

I would only make a separate issue out of it, if we think it might be better to move that usability behavior a level above this method, so that we can better write tests to this method.

antgonza: Good point, add as a Note in the docstring ...

qiita_db/metadata_template/test/test_prep_template.py
charles-cowart: Depending on whether it's better to move the silent drop a level above the _clean_validate_template, I'd say yes, we could drop these tests, and write ones to test the automatic drop instead.

qiita_db/metadata_template/test/test_sample_template.py
charles-cowart: I believe in Python, it is global only to the module, and we can prevent it from being imported with the others by prefixing with '__'. The standard way to create a global variable in Python is to declare the variable in a module, and import that module into every file.

qiita_db/metadata_template/base_metadata_template.py
charles-cowart: Gotcha. Removed comment, and added behavior description in the docstring, per the conversation above.

qiita_db/metadata_template/prep_template.py
charles-cowart: Added to _identify_forbidden_words_in_column_names() but not to _identify_pgsql_reserved_words_in_column_names() and _identify_column_names_with_invalid_characters(), since they don't require anything child_specific, and I won't have to move the tests out to one of the child classes.

antgonza: 👍

qiita_db/metadata_template/test/test_prep_template.py
charles-cowart: Omitting automatic drop test for now; may need to shuffle tests around in clean_validate_template first. That dovetails with another issue I found earlier that I want to make an issue in GitHub.

charles-cowart:
Confirmed there are tests for duplicate columns and sample names in both the PrepTemplate and SampleTemplate tests.

qiita_db/metadata_template/base_metadata_template.py
```diff
@@ -484,12 +489,15 @@ def _clean_validate_template(cls, md_template, study_id,
...
-            Cleaned copy of the input md_template
+            Cleaned deep-copy of the input md_template:
+                Removes 'qiita_study_id' and 'qiita_prep_id' columns,
+                if present.
 
         Raises
         ------
         QiitaDBColumnError
-            If the sample names in md_template contains invalid names
+            If the column names in md_template contains invalid characers,
```
antgonza: characers -> characters

qiita_db/metadata_template/test/test_base_metadata_template.py
```diff
@@ -67,6 +67,33 @@ def test_clean_validate_template(self):
         with self.assertRaises(IncompetentQiitaDeveloperError):
             MT._clean_validate_template(None, 1)
 
+    def test_identify_pgsql_reserved_words(self):
+        MT = qdb.metadata_template.base_metadata_template.MetadataTemplate
+        results = MT._identify_pgsql_reserved_words_in_column_names([
+            'select',
+            'column',
+            'just_fine1'])
+        self.assertTrue(set(results), {'column', 'select'})
```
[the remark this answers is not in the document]
charles-cowart: Agreed.

qiita_db/metadata_template/test/test_base_metadata_template.py
```diff
@@ -67,6 +67,33 @@ def test_clean_validate_template(self):
...
+            'tax on',
+            'bla.',
+            '.',
+            'sampleid',
+            'sample_id',
+            '{',
+            'this|is',
+            '4column',
+            'just_fine2'])
+        self.assertTrue(set(results) == {'tax on',
```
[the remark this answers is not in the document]
charles-cowart: This one should be correct. We are taking the set of results, which are unordered, and making a statement that those results should be equal to the right-hand-side. We assert that this statement should be true.

qiita_db/metadata_template/test/test_sample_template.py
```diff
@@ -825,6 +830,71 @@ def test_clean_validate_template_no_invalid_chars(self):
...
+        results = ST._identify_forbidden_words_in_column_names([
+            'just_fine3',
+            'sampleid',
+            'alice',
+            'linkerprimersequence',
+            'bob',
+            'qiita_study_id',
+            'qiita_prep_id',
+            'eve'])
+        self.assertTrue(set(results), {'qiita_prep_id',
```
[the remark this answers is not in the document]
charles-cowart: Agreed.

qiita_db/metadata_template/test/test_base_metadata_template.py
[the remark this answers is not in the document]
charles-cowart: Changed to assertEquals, with == operator added as well.

qiita_db/metadata_template/test/test_sample_template.py
```diff
@@ -825,6 +830,71 @@ def test_clean_validate_template_no_invalid_chars(self):
...
+                                       'qiita_study_id',
+                                       'sampleid'})
+
+    def test_silent_drop(self):
+        ST = qdb.metadata_template.sample_template.SampleTemplate
+        self.metadata.rename(columns={'taxon_id': 'qiita_prep_id'},
+                             inplace=True)
+        results = list(ST._clean_validate_template(
```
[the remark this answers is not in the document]
charles-cowart: Looks nice! Will do.

qiita_db/metadata_template/test/test_base_metadata_template.py
```diff
@@ -67,6 +67,33 @@ def test_clean_validate_template(self):
         with self.assertRaises(IncompetentQiitaDeveloperError):
             MT._clean_validate_template(None, 1)
 
+    def test_identify_pgsql_reserved_words(self):
+        MT = qdb.metadata_template.base_metadata_template.MetadataTemplate
+        results = MT._identify_pgsql_reserved_words_in_column_names([
+            'select',
+            'column',
+            'just_fine1'])
+        self.assertTrue(set(results) == {'column', 'select'})
```
antgonza: sent via PM but also to have it here, this can be: self.assertItemsEqual(a, b)

qiita_db/metadata_template/base_metadata_template.py
charles-cowart: thanks for noticing - fixed!

qiita_db/metadata_template/test/test_base_metadata_template.py
charles-cowart: Done!

PR #2676, opened 2018-09-19
fix scp instructions

antgonza:
now it looks like this:
screen shot 2018-09-19 at 3 51 42 pm


qiita_pet/templates/upload.html
```diff
@@ -193,6 +193,7 @@ <h4 class="panel-title">
                   <li>Currenlty we only support ftp and scp.</li>
                   <li>Prepare study files by storing them in one folder on your fileserver.</li>
                   <li>Generate a new key by running: <code>ssh-keygen -t rsa -C "ssh test key" -f ~/.ssh/qiita-key -P ""</code> in your remote server.</li>
+                  <li>Allow access using the new key: <code>cat ~/.ssh/qiita-key >> ~/.ssh/authorized_keys</code>. Note that erasing that line will remove access.</li>
```
tkosciol: actually, my mistake and it should read `cat ~/.ssh/qiita-key.pub`, right?

PR #2694, opened 2018-10-05
Additional Processing Command added to Software_Command

charles-cowart:
Adds an 'addtl_processing_cmd' property to Command, to obtain additional processing steps needed to merge BIOMs.


qiita_db/test/test_analysis.py
```diff
@@ -91,16 +96,55 @@ def _create_analyses_with_samples(self, user='demo@microbio.me',
         """
         user = qdb.user.User(user)
         dflt_analysis = user.default_analysis
+        aid = 4
+        if addtl_processing_cmd:
+            cmd_id = qdb.artifact.Artifact(aid).processing_parameters.command
```
antgonza: note that this is cmd, and the line below is cmd_id

qiita_db/test/test_analysis.py
```diff
@@ -91,16 +96,55 @@ def _create_analyses_with_samples(self, user='demo@microbio.me',
...
-                 '1.SKM9.640192', '1.SKM4.640180']})
+            {aid: ['1.SKB8.640193', '1.SKD8.640184', '1.SKB7.640196',
+                   '1.SKM9.640192', '1.SKM4.640180']})
         new = qdb.analysis.Analysis.create(
             user, "newAnalysis", "A New Analysis", from_default=True,
             merge_duplicated_sample_ids=merge)
 
         self._wait_for_jobs(new)
+
+        with qdb.sql_connection.TRN:
```
antgonza: you need to add the same if than above to only enter this block when needed ...

qiita_db/test/test_analysis.py
charles-cowart: I just broke it into two lines to pass flake8. In the next line, cmd_id is reassigned to cmd_id.id. I just didn't want to create an additional variable. It's probably bad form, now that I look at it.

qiita_db/software.py
```diff
@@ -27,6 +27,18 @@ class Command(qdb.base.QiitaObject):
 
     Attributes
     ----------
+    active
```
antgonza: Thanks for updating.

qiita_db/analysis.py
```diff
@@ -869,6 +876,8 @@ def build_files(self, merge_duplicated_sample_ids):
                         dup_samples = dup_samples | s
 
             self._build_mapping_file(samples, rename_dup_samples)
+            # TODO: addtl_processing_cmd needs to be passed to here and work
```
antgonza: All these comments seem a bit too much for me but I guess, please confirm, that make the code clearer for you, if that's the case, let's leave them. However, could you remove the here section, I guess it's implied when the comment is in that line. For example:
```
# TODO: addtl_processing_cmd needs to be passed to here and work 
# here.
```
can be: `# TODO: addtl_processing_cmd needs to be passed`
or  `# save final biom here` -> `# save final biom`

qiita_db/support_files/patches/67.sql
```diff
@@ -0,0 +1,2 @@
+alter table qiita.software_command add addtl_processing_cmd varchar;
```
antgonza: Could you follow the style of previous patches? Like the date and the goal of the changes are missing, thanks.

qiita_db/support_files/qiita-db.dbs
```diff
@@ -1854,4 +1854,4 @@ ALTER TABLE oauth_software ADD CONSTRAINT fk_oauth_software FOREIGN KEY ( client
 ]]></string>
 		</script>
```
antgonza: This is no right, correct? There should have been changes in this file, cause we added a new column. Could you double check?

qiita_db/support_files/qiita-db.html
```diff
@@ -5722,6 +5723,12 @@
 		<td> bool   DEFO false </td>
 		<td>  </td>
 	</tr>
+	<tr>
+		<td>*</td>
```
antgonza: The patch has a comment, could you make sure that you also add it in the GUI or remove it from both?

qiita_db/test/test_analysis.py
```diff
@@ -100,6 +105,61 @@ def _create_analyses_with_samples(self, user='demo@microbio.me',
...
         return new
+        """
+        user = qdb.user.User(user)
+        aid = 4
+        dflt_analysis = user.default_analysis
+        dflt_analysis.add_samples(
+            {aid: ['1.SKB8.640193', '1.SKD8.640184', '1.SKB7.640196',
+                   '1.SKM9.640192', '1.SKM4.640180']})
+
+        if addtl_processing_cmd:
```
antgonza: Just understood the problem with this section, we change the command and then revert so in praxis we never changed. Based on this, I would say leave this method as it was and do the injection and correction in the test method. In other words undo the changes in `test_build_files_addtl_processing_cmd `, move this block and `_post_create_analysis_cleanup` (which is already there) within `test_build_files_addtl_processing_cmd`.

scripts/qiita-recover-jobs
```diff
@@ -16,7 +16,7 @@ from time import sleep
 SLEEP_TIME = 2
 
 
-def _submit_jobs(jids_to_recover, recover_type):
+def Qsubmit_jobs(jids_to_recover, recover_type):
```
antgonza: If we are fixing Click, do we really need these changes? Either way, this script does _not_ use click so should not be affected, right? If that's the case, please revert.

qiita_db/software.py
```diff
@@ -445,6 +466,22 @@ def name(self):
             qdb.sql_connection.TRN.add(sql, [self.id])
             return qdb.sql_connection.TRN.execute_fetchlast()
 
+    @property
+    def addtl_processing_cmd(self):
```
antgonza: This method needs tests!!!!

qiita_db/analysis.py
charles-cowart: Not a problem. I can remove them.

qiita_db/support_files/patches/67.sql
charles-cowart: Not a problem, thanks for pointing it out!

scripts/qiita-recover-jobs
charles-cowart: I don't think it needs it. Changing functions and methods starting with '_' to starting with 'Q' didn't change anything. I'll revert it.

qiita_db/support_files/qiita-db.dbs
charles-cowart: Just added a line. Just to clarify, this is an internal format right? I don't see a schema reference at the top.

qiita_db/test/test_analysis.py
charles-cowart: No problem. Once I did that, it became obvious I should rewrite the test as well.

qiita_db/software.py
charles-cowart: thanks for spotting it! done!

scripts/qiita
```diff
@@ -26,7 +26,7 @@ from psycopg2 import OperationalError
 import qiita_db as qdb
 from qiita_core.qiita_settings import qiita_config, r_client
 from qiita_ware.ebi import EBISubmission
-from qiita_ware.commands import submit_EBI as _submit_EBI
+from qiita_ware.commands import submit_EBI as Qsubmit_EBI
```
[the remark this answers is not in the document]
charles-cowart: I believe that once we fixed the Click version to 6.7, we don't need to make any further changes.  Yeah, I'm still picking up Click; at the time I just assumed Click looked at everything inside the script.

scripts/qiita-env
```diff
@@ -11,7 +11,7 @@
 import click
 
 import qiita_db as qdb
-from qiita_core.environment_manager import test as _test, TEST_RUNNERS
+from qiita_core.environment_manager import test as Qtest, TEST_RUNNERS
```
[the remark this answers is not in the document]
charles-cowart: My apologies, Qtest slipped through the revert.

PR #2698, opened 2018-10-10
Post Processing Merged BIOMs extended

charles-cowart:
analysis.build_files will now run an arbitrary number of post-processes defined
in qiita.software_command.post_processing_cmd column. It's assumed that each
post-process is a python script, with a defined Conda environment. It's also
currently assumed that each post-process overwrites the merged BIOM file. A
step removed from calling plugins, but getting closer.


qiita_db/analysis.py
```diff
@@ -958,7 +975,35 @@ def _build_biom_tables(self, grouped_samples, rename_dup_samples=False):
...
                         f, "Generated by Qiita, analysis id: %d, info: %s" % (
                             self._id, label))
-                biom_files.append((data_type, biom_fp))
+
+                # post_processing_cmds are a list of commands to run on the
+                # final BIOM. The order of operations is list-order. Each
+                # element of the list is a dictionary containing the Conda env
+                # to use, the script to run, and a dictionary of parameters.
+                if post_processing_cmds:
+                    print(type(post_processing_cmds))
```
antgonza: rm print

qiita_db/analysis.py
```diff
@@ -958,7 +975,35 @@ def _build_biom_tables(self, grouped_samples, rename_dup_samples=False):
...
+                    for cmd in post_processing_cmds:
+                        # for now, assume BIOM file is the last parameter
+                        # and assume biom_fp is overwritten
+                        params = ' '.join(["%s=%s" % (k, v) for k, v in
+                                          cmd['script_params'].items()])
+                        params = "%s %s" % (params, biom_fp)
+                        # if environment is successfully activated,
+                        # run script with parameters
+                        p = Popen(["source activate %s && python %s %s" % (
```
antgonza: This is given by the software itself and should not be hard-coded; for example, we have environments in conda and in virtualenv. In other words, you just need cmd['script_env'] and cmd['script_path'] separated by semicolon, nothing else.

qiita_db/analysis.py
```diff
@@ -958,7 +975,35 @@ def _build_biom_tables(self, grouped_samples, rename_dup_samples=False):
...
+                                  cmd['script_env'],
+                                  cmd['script_path'],
+                                  params)], shell=True, stdout=PIPE)
+                        p.wait()
+                        # consider processing p.communicate, returning it up
+                        # the stack, annotating post_processing_cmds, etc.
+                        print(p.communicate())
```
antgonza: rm print

qiita_db/test/test_software.py
```diff
@@ -128,29 +130,39 @@ def test_name(self):
...
         self.assertEqual(qdb.software.Command(2).name, "Split libraries")
 
-    def test_addtl_processing_cmd(self):
+    def test_post_processing_cmd(self):
         # initial test
-        self.assertEqual(qdb.software.Command(1).addtl_processing_cmd, None)
+        self.assertEqual(qdb.software.Command(1).post_processing_cmd, None)
+
+        results = {}
+        results['script_env'] = 'qiita'
```
antgonza: this should be: deactivate; source activate qiita

qiita_db/test/test_analysis.py
```diff
@@ -515,36 +516,50 @@ def test_build_files(self):
...
-    def test_build_files_addtl_processing_cmd(self):
+    def test_build_files_post_processing_cmd(self):
         tmp = qdb.artifact.Artifact(4).processing_parameters.command
         cmd_id = tmp.id
 
         # set a known artifact's additional processing command
         # to a known value. Then test for it.
+        # qiita_db/worker.py will work w/py2.7 & 3.6 envs.
+        results = {}
+        results['script_env'] = 'qiita'
```
antgonza: this should be: deactivate; source activate qiita

qiita_db/worker.py
```diff
@@ -0,0 +1,7 @@
+import sys
```
antgonza: This should be in the scripts folder so it's added automatically or in the qiita_db/test/support_files/

qiita_db/analysis.py
```diff
@@ -958,7 +983,78 @@ def _build_biom_tables(self, grouped_samples, rename_dup_samples=False):
...
+                        # --fp_fragments=<path_to_fragments_file>
+                        # assume output dir is passed as:
+                        # --output_dir=<path_to_output_dir>
+
+                        # concatenate any other parameters into a string
+                        params = ' '.join(["%s=%s" % (k, v) for k, v in
+                                          cmd['script_params'].items()])
+
+                        # append fragments file and output dir parameters
+                        params = "%s --fp_fragments=%s --output_dir=%s" %\
```
antgonza: Can you change the name of the parameter from `--fp_fragments=` to `--fp_archive` so it's more generic; at the end of the day this can be fragments or anything else the plugin stores in the database/archive

qiita_db/analysis.py
```diff
@@ -958,7 +983,78 @@ def _build_biom_tables(self, grouped_samples, rename_dup_samples=False):
...
+                    archives = {f: loads(archives[f])
+                                 for f, plc
+                                 in archives.items()
+                                 if plc != ''}
+
+                    # since biom_fp uses base_fp as its location, assume it's
+                    # suitable for other files as well.
+                    output_dir = base_fp
+
+                    fp_archive = '%s/archive.json' % output_dir
```
antgonza: sorry for not catching this before: you should use join to create this path and also use the analysis id to be safe something like: `fp_archive = join(output_dir, 'archive_%d.json' % self.id)`

qiita_db/analysis.py
charles-cowart: No problem - changes already made to qiita and qp_deblur. One thing though, the command line util that strips fragments from the biom that aren't in the tree takes --fp_phylogeny and --fp_biom as options; do you think they might be too specific as well?

Originally, I thought a chain of commands would each in turn iterate over the same biom file, but since the second command will need a tree and a biom file as input, the database entry will need to be updated with the fp_s to the data, or the second command will need to assume the name and location of the inputs.

antgonza: I think those names internally in deblur are fine but you are right about those here. Also all the comments that mention phylogeny vs. extra file for analysis ...

qiita_db/analysis.py
charles-cowart: Ah gotcha, platform independence path creation.

qiita_db/analysis.py
```diff
@@ -30,6 +30,10 @@
 from qiita_core.exceptions import IncompetentQiitaDeveloperError
 from qiita_core.qiita_settings import qiita_config
 import qiita_db as qdb
+from qiita_db.archive import Archive
```
antgonza: Sorry one more, if we have `import qiita_db as qdb` we don't need `from qiita_db.archive import Archive`; simply do qdb.archive.Archive

charles-cowart: No problem.

Issue #2679, opened 2018-09-23
allow non owners/admins to download the raw data


antgonza:
One of the original designs in Qiita was to not allow to download the full raw data to not owners because it's really common for users to share runs and a lot of times the other samples in that raw data was private. Anyway, more and more we are getting requests for not owners to retrieve the raw data. 

I think we should make this an option for the owners to change. If they select to share, anyone could download. This will mean to open the "download all raw data" and add buttons to download the files in individual raw data.

cc: @wasade

PR #2714, opened 2018-10-25
add public_raw_download

antgonza:
This add the new column public_raw_download to the study table and the interactions with the GUI. ~Note that to solve https://github.com/biocore/qiita/issues/2679 we still need to fix the RawDownload handler, which will happen in a follow up PR.~

This is how it looks:
public_raw_download

### DIFF
--- a/qiita_db/study.py
+++ b/qiita_db/study.py
@@ -487,12 +487,42 @@ class Study(qdb.base.QiitaObject):
         Parameters
         ----------
         title : str
-            The new study title
+            The study title
         """
         with qdb.sql_connection.TRN:
             sql = """UPDATE qiita.{0} SET study_title = %s
                      WHERE study_id = %s""".format(self._table)
             qdb.sql_connection.TRN.add(sql, [title, self._id])
+            return qdb.sql_connection.TRN.execute()
+
+    @property
+    def public_raw_download(self):
+        """Returns the public_raw_download of the study
+
+        Returns
+        -------
+        str
+            public_raw_download of study
+        """
+        with qdb.sql_connection.TRN:
+            sql = """SELECT public_raw_download FROM qiita.{0}
+                     WHERE study_id = %s""".format(self._table)
+            qdb.sql_connection.TRN.add(sql, [self._id])
+            return qdb.sql_connection.TRN.execute_fetchlast()
+
+    @public_raw_download.setter
+    def public_raw_download(self, public_raw_download):
+        """Sets the public_raw_download of the study
+
+        Parameters
+        ----------
+        public_raw_download : bool
+            The study public_raw_download
+        """
+        with qdb.sql_connection.TRN:
+            sql = """UPDATE qiita.{0} SET public_raw_download = %s
+                     WHERE study_id = %s""".format(self._table)
+            qdb.sql_connection.TRN.add(sql, [public_raw_download, self._id])
             return qdb.sql_connection.TRN.execute()
 
     @property
--- a/qiita_db/study.py
+++ b/qiita_db/study.py
@@ -497,7 +497,7 @@ class Study(qdb.base.QiitaObject):
 
     @property
     def public_raw_download(self):
-        """Returns the public_raw_download of the study
+        """Returns if the study's raw data is available for download
 
         Returns
         -------
@@ -512,7 +512,7 @@ class Study(qdb.base.QiitaObject):
 
     @public_raw_download.setter
     def public_raw_download(self, public_raw_download):
-        """Sets the public_raw_download of the study
+        """Sets if the study's raw data is available for download
 
         Parameters
         ----------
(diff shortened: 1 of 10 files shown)


antgonza:
Turns out that the number of changes to allow public raw data download was much smaller than expected so just pushed to this PR.

qiita_db/study.py
```diff
@@ -487,14 +487,44 @@ def title(self, title):
...
+
+    @public_raw_download.setter
+    def public_raw_download(self, public_raw_download):
+        """Sets the public_raw_download of the study
+
+        Parameters
+        ----------
+        public_raw_download : bool
+            The study public_raw_download
+        """
```
charles-cowart: Returns is missing in documentation

qiita_pet/handlers/api_proxy/studies.py
```diff
@@ -111,7 +112,7 @@ def study_get_req(study_id, user_id):
     # Study.has_access no_public=True, will return True only if the user_id is
     # the owner of the study or if the study is shared with the user_id
     study_info['has_access_to_raw_data'] = study.has_access(
```
charles-cowart: The above comment is no longer true, right? Now, if the data is considered public, it will also return True.

qiita_db/study.py
charles-cowart: Returns is missing in documentation

qiita_pet/handlers/api_proxy/studies.py
charles-cowart: Study.has_access no_public=True, will return True only if the user_id is
the owner of the study or if the study is shared with the user_id

This comment should be appended to include study.public_raw_download.

qiita_db/study.py
antgonza: This is a setter ... doesn't have a return

qiita_pet/handlers/api_proxy/studies.py
antgonza: k

qiita_pet/handlers/api_proxy/tests/test_studies.py
```diff
@@ -106,45 +107,33 @@ def test_study_get_req(self):
...
-                'study_alias': 'FCM',
-                'study_id': new_study.id,
-                'most_recent_contact': datetime(2015, 5, 19, 16, 11),
-                'publication_doi': [],
-                'publication_pid': [],
-                'num_samples': 25,
-                'study_title': 'Some New Study',
-                'number_samples_collected': 27},
-            'editable': True}
+        exp = {'status': 'success', 'study_info': {
```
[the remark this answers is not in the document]
antgonza: I _really_ hate that formatting ... I rather not ...

qiita_pet/templates/study_ajax/base_info.html
```diff
@@ -114,6 +142,10 @@
       <b>Shared With:</b> <span id="shared_html_{{study_info['study_id']}}"></span><br/>
       <b>Samples:</b> {{study_info['num_samples']}}<br />
       <b>EBI:</b> {% raw ebi_info %}<br />
+      <b>Public raw data:</b> {{study_info['public_raw_download']}}
```
ElDeveloper: I feel like this controller is a bit awkward. What about making the text be "Allow Qiita users to download raw data files:", and making the button+true/false a checkbox that is preset to checked/unchecked depending on the status.

qiita_db/study.py
```diff
@@ -487,14 +487,44 @@ def title(self, title):
...
         """
         with qdb.sql_connection.TRN:
             sql = """UPDATE qiita.{0} SET study_title = %s
                      WHERE study_id = %s""".format(self._table)
             qdb.sql_connection.TRN.add(sql, [title, self._id])
             return qdb.sql_connection.TRN.execute()
 
+    @property
+    def public_raw_download(self):
+        """Returns the public_raw_download of the study
```
ElDeveloper: c

qiita_db/study.py
```diff
@@ -487,14 +487,44 @@ def title(self, title):
...
+        """
+        with qdb.sql_connection.TRN:
+            sql = """SELECT public_raw_download FROM qiita.{0}
+                     WHERE study_id = %s""".format(self._table)
+            qdb.sql_connection.TRN.add(sql, [self._id])
+            return qdb.sql_connection.TRN.execute_fetchlast()
+
+    @public_raw_download.setter
+    def public_raw_download(self, public_raw_download):
+        """Sets the public_raw_download of the study
```
ElDeveloper: More appropriately `Sets if the study's raw data is available for download`.

qiita_pet/handlers/api_proxy/tests/test_studies.py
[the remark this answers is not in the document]
charles-cowart: :) Fair enough. :)

qiita_db/study.py
charles-cowart: d'oh!

PR #2715, opened 2018-10-25
fix #2681

### LINKED ISSUE #2681: upload is blocked when there is a banner message
I kind of like this but happy to fix

antgonza:
Before fix:
before

After fix:
after

### DIFF
--- a/qiita_pet/templates/upload.html
+++ b/qiita_pet/templates/upload.html
@@ -96,7 +96,7 @@
 
 {% block content %}
 
-{% if (level != 'info' or 'Retrieving remote files:' not in message) or not message %}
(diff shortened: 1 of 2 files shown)


qiita_pet/templates/upload.html
```diff
@@ -96,7 +96,7 @@
 
 {% block content %}
 
-{% if (level != 'info' and 'Retrieving remote files:' not in message) or not message %}
+{% if (level != 'info' or 'Retrieving remote files:' not in message) or not message %}
```
charles-cowart: Does "level != 'info' or 'Retrieving remote files:' not in message" still need to be in parentheses, if they are now all OR conditions?

PR #2718, opened 2018-10-27
EBI 10M max

antgonza:
Built on top of https://github.com/biocore/qiita/pull/2717, so please review that one first.


antgonza:
This should be ready for review.

PR #2730, opened 2018-11-12
adding Safari

antgonza:

PR #2731, opened 2018-11-12
Interaction with /qiita_db/authenticate was off

wasade:
I think this change renders correct -- did this via Github. The structure necessary to authenticate is:

```{'grant_type': 'client', 'client_secret': str, 'client_id': str}```

...whereas the prior parameters suggested I think `{'grant_type': ['client', 'password'], 'client_id': str}`.


antgonza:
thanks @wasade! I have move this change to the other open PR: https://github.com/biocore/qiita/pull/2730/files.

PR #2735, opened 2018-11-14
some extra fixes for rc

antgonza:
This adds:
- sorting study listing by study id
- the number of studies shown in the list
- the ebi large study trimming new options
- some general code that I'm using now for job recovery and stats


scripts/qiita-recover-jobs
```diff
@@ -68,16 +70,86 @@ def _get_jids_to_recover(recover_type):
...
 
 
+def _parse_queue_values(d):
+    max_mem = 0
+    max_pmem = 0
+    max_vmem = 0
+    max_wt = timedelta(hours=0, minutes=0, seconds=0)
+    d = d.split(',')
+    for dd in d:
+        if dd.startswith('mem'):
```
charles-cowart: These ifs should really be elifs, if that is your intended logic; it will make a difference when someone else is looking at it and trying to decide whether to expect the value to match with one conditional or multiple ones.

antgonza: 👍

PR #2736, opened 2018-11-16
WIP [do not review]: converting info files to key-value pairs

antgonza:


qiita_db/metadata_template/base_metadata_template.py
```diff
@@ -597,19 +597,18 @@ def _common_creation_steps(cls, md_template, obj_id):
 
             # Create table with custom columns
```
charles-cowart: Comment should be changed to, "Create table and populate with metadata key/value pairs" or similar.

qiita_db/support_files/patches/python_patches/51.py
```diff
@@ -90,7 +90,8 @@ def transform_date(value):
             # converting to datetime
             for col in columns:
                 st_df[col] = st_df[col].apply(transform_date)
-            st.update(st_df)
+            # SHOULD WE FIX?
```
charles-cowart: Since this is a schema change, I think it's all right to break with convention and remove/inactivate this.

PR #2742, opened 2018-11-20
hot-fix-111918

antgonza:


qiita_pet/static/js/networkVue.js
```diff
@@ -1045,7 +1045,11 @@ Vue.component('processing-graph', {
...
               // No graph is available - execute the callback
               $('#network-header-div').hide();
-              $("#processing-job-div").html("<h3>Error generating the analysis:</h3><h5>" + jobErrors + '</h5>');
+              if (!vm.isAnalysisPipeline) {
+                vm.noInitJobsCallback('processing-job-div', jobErrors);
+              } else {
+                $("#processing-job-div").html("<h3>Error generating the analysis:</h3><h5>" + jobErrors + '</h5>');
+              }
             } else {
               // A graph is available, update de current graph
```
charles-cowart: de -> the

qiita_pet/templates/sitebase.html
```diff
@@ -57,6 +57,7 @@
         'Chrome': {'version': 49, 'download': 'https://www.google.com/chrome/browser/'},
         'Firefox': {'version': 48, 'download': 'https://support.mozilla.org/en-US/kb/update-firefox-latest-version'},
         'Safari': {'version': 12, 'download': 'http://support.apple.com/en-us/HT6104'},
+        'Chrome Mobile': {'version': 69, 'download': 'https://www.google.com/chrome/browser/'},
```
charles-cowart: Both Chrome-related links are the same. Just to confirm, it's the site that intelligently figures out what to offer? 'Chrome Mobile' is just a needed synonym for 'Chrome' for Browser tags, etc? It also looks like one is 69 and one is 49.

antgonza: afaik, yup, and the numbers are right ...

charles-cowart: ok. it sounds like everything is as it should be.

PR #2743, opened 2018-11-28
Info files from str to jsonb

antgonza:


qiita_db/support_files/patches/68.sql
```diff
@@ -0,0 +1,102 @@
...
+    LOOP
+      EXECUTE 'INSERT INTO ' || dyn_table || ' (sample_id, sample_values) VALUES (''' || sid || ''',  row_to_json((SELECT r FROM (SELECT * from qiita.' || dyn_table_bk || ' where sample_id = ''' || sid || ''') r)));';
+    END LOOP;
+
+    -- adding index
+    EXECUTE 'ALTER TABLE ' || dyn_table || ' ADD CONSTRAINT pk_jsonb_' || dyn_t || ' PRIMARY KEY ( sample_id );';
+  END LOOP;
+END $do$;
+
+-- Dropping all the _bk tables
```
charles-cowart: Confirmed DROP TABLES will free space on COMMIT, provided there are no other connections using the table as well. The OS should reclaim the space in time, although it may not be immediately after performing the operation; hence, an exceedingly large number of _bk tables being created and dropped may impact system resources, fyi.

qiita_db/support_files/patches/python_patches/45.py
```diff
@@ -8,8 +8,9 @@
 
 from future.utils import viewitems
 
-from qiita_db.metadata_template.sample_template import SampleTemplate
-from qiita_db.metadata_template.prep_template import PrepTemplate
+# TODOJSONB: uncomment
```
charles-cowart: TODO present.

qiita_db/metadata_template/base_metadata_template.py
```diff
@@ -705,9 +729,23 @@ def delete_column(self, column_name):
                     column_name)
 
         with qdb.sql_connection.TRN:
-            sql = 'ALTER TABLE qiita.%s%d DROP COLUMN %s' % (
-                self._table_prefix, self._id, column_name)
-            qdb.sql_connection.TRN.add(sql)
+            table_name = 'qiita.{0}{1}'.format(self._table_prefix, self._id)
+            # deleting from all samples
+            sql = """UPDATE {0}
```
charles-cowart: If I read this right, we might want to reiterate the '-' operator is a JSONB operation?

qiita_db/metadata_template/base_metadata_template.py
```diff
@@ -793,56 +831,53 @@ def _common_extend_steps(self, md_template):
...
-                    sql = """UPDATE qiita.{0}
-                             SET {1}
-                             WHERE sample_id=%s""".format(table_name,
-                                                          ",".join(set_str))
-                    for v in values:
-                        qdb.sql_connection.TRN.add(sql, v)
+                    for sid, df in md_filtered.iterrows():
+                        values = dict(df.iteritems())
+                        sql = """UPDATE qiita.{0}
+                                 SET sample_values = sample_values || %s
```
charles-cowart: Similarly, if I read this correctly, it might be good to remind the reader that sample_values is a JSONB column, or maybe we can include a URL reference to JSONB. I can look for one.

qiita_db/metadata_template/test/test_prep_template.py
```diff
@@ -280,6 +280,8 @@ def test_delitem(self):
 @qiita_test_checker()
 class TestPrepTemplate(TestCase):
     def setUp(self):
+        self.QCN = \
```
charles-cowart: self.QCN is a state variable used in base and derived classes to store the name of the column where all column names are stored. We may need this in a comment, because it's not apparent what it's for, if we first see it here.

qiita_db/metadata_template/test/test_sample_template.py
```diff
@@ -1011,7 +1011,7 @@ def test_create(self):
                 'taxon_id': '9606',
                 'scientific_name': 'homo sapiens'}}
         for s_id in exp_sample_ids:
-            self.assertEqual(st[s_id]._to_dict(), exp_dict[s_id])
+            self.assertDictEqual(st[s_id]._to_dict(), exp_dict[s_id])
```
charles-cowart: In part, this change is because JSONB dicts are unordered, right?

qiita_db/support_files/populate_test_db.sql
```diff
@@ -535,6 +508,17 @@ INSERT INTO qiita.filepath (filepath, filepath_type_id, checksum, checksum_algor
 INSERT INTO qiita.prep_template_filepath VALUES (1, 20), (1, 21);
 
 
+-- Insrting the BIOM artifact filepath
```
charles-cowart: Insrting -> Inserting

qiita_db/metadata_template/base_metadata_template.py
charles-cowart: The JSONB-using queries look slick! After going through the code, I noticed some were easier to identify as JSONB-using (those that use ->>), while others weren't (Those that used the '-' operator). I'd like to include a link to a JSONB reference at the top of the files that use them, or else comment the queries that specifically use them, to make it bonehead easy to identify.

antgonza: same, adding but this is like asking what does && or || in an if ...

qiita_db/metadata_template/test/test_prep_template.py
antgonza: added some explanation

qiita_db/metadata_template/test/test_sample_template.py
antgonza: That's correct!

qiita_db/environment_manager.py
```diff
@@ -398,19 +402,23 @@ def patch(patches_dir=PATCHES_DIR, verbose=False, test=False):
...
-            py_patch_filename = basename(py_patch_fp)
+        # patch 43.sql is when we started testing patches, then in patch
+        # 68.sql is when we transitioned to jsonb for the info files; let's do
+        # this in it's own transition
+        if sql_patch_filename == '68.sql' and test:
+            with qdb.sql_connection.TRN:
+                _populate_test_db()
 
+        with qdb.sql_connection.TRN:
             with open(sql_patch_fp, 'U') as patch_file:
```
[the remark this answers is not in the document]
charles-cowart: Doh! Good point! :)

qiita_db/meta_util.py
```diff
@@ -316,18 +316,17 @@ def get_lat_longs():
...
+                    AND table_schema = 'qiita'"""
         qdb.sql_connection.TRN.add(sql, [tuple(portal_table_ids)])
 
-        sql = [('SELECT CAST(latitude AS FLOAT), '
-                '       CAST(longitude AS FLOAT) '
-                'FROM qiita.%s '
-                'WHERE isnumeric(latitude) AND isnumeric(longitude) '
-                "AND latitude <> 'NaN' "
-                "AND longitude <> 'NaN' " % s)
+        sql_query = """
```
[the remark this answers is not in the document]
charles-cowart: Good point. :)

qiita_db/environment_manager.py
```diff
@@ -382,6 +382,10 @@ def patch(patches_dir=PATCHES_DIR, verbose=False, test=False):
     Pulls the current patch from the settings table and applies all subsequent
     patches found in the patches directory.
     """
+    # we are going to open and close 2 main transactions; this is a required
+    # change since patch 68.sql where we transition to jsonb for all info
+    # files. The 2 main transitions are: (1) get the current settings,
+    # (2) each patch in their independent trasaction
```
charles-cowart: trasaction -> transaction

qiita_db/environment_manager.py
```diff
@@ -398,19 +402,23 @@ def patch(patches_dir=PATCHES_DIR, verbose=False, test=False):
...
 
-            py_patch_fp = corresponding_py_patch(
-                splitext(basename(sql_patch_fp))[0] + '.py')
-            py_patch_filename = basename(py_patch_fp)
+        # patch 43.sql is when we started testing patches, then in patch
+        # 68.sql is when we transitioned to jsonb for the info files; let's do
+        # this in its own transition
+        if sql_patch_filename == '68.sql' and test:
+            with qdb.sql_connection.TRN:
+                _populate_test_db()
```
wasade: would this execute on production as well?

qiita_db/metadata_template/base_metadata_template.py
```diff
@@ -198,15 +208,14 @@ def _to_dict(self):
...
+                     WHERE sample_id=%s""".format(self._dynamic_table)
             qdb.sql_connection.TRN.add(sql, [self._id])
-            d = dict(qdb.sql_connection.TRN.execute_fetchindex()[0])
 
-            # Remove the sample_id, is not part of the metadata
-            del d['sample_id']
+            result = qdb.sql_connection.TRN.execute_fetchindex()
 
-            return d
+            return result[0]['sample_values']
```
wasade: Why isn't a `loads` needed here when it is on like 198?

qiita_db/metadata_template/base_metadata_template.py
```diff
@@ -597,19 +611,20 @@ def _common_creation_steps(cls, md_template, obj_id):
...
             qdb.sql_connection.TRN.add(sql)
 
-            values = [list(md_template[h]) for h in headers]
-            values.insert(0, sample_ids)
-            values = [list(v) for v in zip(*values)]
-            sql = """INSERT INTO qiita.{0} (sample_id, {1})
-                     VALUES (%s, {2})""".format(
-                table_name, ", ".join(headers),
-                ', '.join(["%s"] * len(headers)))
+            values = '{"columns": %s}' % dumps(md_template.columns.tolist())
```
wasade: what about: `values = dumps({"columns": md_template.columns.tolist()})`?

qiita_db/metadata_template/base_metadata_template.py
```diff
@@ -793,56 +834,54 @@ def _common_extend_steps(self, md_template):
...
                 new_cols = sorted(new_cols)
 
-                sql_alter = """ALTER TABLE qiita.{0} ADD COLUMN {1} {2}"""
-                for category in new_cols:
-                    qdb.sql_connection.TRN.add(
-                        sql_alter.format(table_name, category, 'varchar'))
+                cols = self.categories()
+                cols.extend(new_cols)
+
+                values = '{"columns": %s}' % dumps(cols)
```
wasade: same here, safer to rely fully on the encoder instead of manually creating json

qiita_db/metadata_template/base_metadata_template.py
```diff
@@ -793,56 +834,54 @@ def _common_extend_steps(self, md_template):
...
-                values = [list(md_filtered[h]) for h in md_filtered]
-                values.insert(0, new_samples)
-                values = [list(v) for v in zip(*values)]
-                sql = """INSERT INTO qiita.{0} (sample_id, {1})
-                         VALUES (%s, {2})""".format(
-                    table_name, ", ".join(headers),
-                    ', '.join(["%s"] * len(headers)))
+                # inserting new samples to the info file
+                values = [(k, df.to_json())
+                          for k, df in md_filtered.iterrows()]
```
wasade: df -> row?

qiita_db/metadata_template/base_metadata_template.py
```diff
@@ -1105,20 +1144,27 @@ def to_dataframe(self):
         """
         with qdb.sql_connection.TRN:
             # Retrieve all the information from the database
-            sql = "SELECT * FROM qiita.{0}".format(self._table_name(self._id))
+            cols = self.categories()
+            sql = """SELECT sample_id, sample_values
+                     FROM qiita.{0}
+                     WHERE sample_id != '{1}'""".format(
+                        self._table_name(self._id), QIITA_COLUMN_NAME)
```
wasade: `QIITA_COLUMN_NAME` is a `sample_id`?

qiita_db/metadata_template/base_metadata_template.py
```diff
@@ -1105,20 +1144,27 @@ def to_dataframe(self):
...
-            cols = sorted(meta[0].keys())
-            meta_matrix = [[r[c] for c in cols] for r in meta]
-            df = pd.DataFrame(meta_matrix, columns=cols, dtype=str)
+            # we could have used dict() and then use pandas from_dict but
+            # that takes forever so it's much faster to simply parse here and
+            # load via pandas.DataFrame
+            data = []
+            for r in qdb.sql_connection.TRN.execute_fetchindex():
+                v = [r[0]]
+                v.extend([r[1][c] if c in r[1] else None for c in cols])
```
wasade: I have no idea what this is doing

qiita_db/metadata_template/base_metadata_template.py
```diff
@@ -1155,18 +1201,26 @@ def get_filepaths(self):
...
         Returns
         -------
         cols : list
-            The static and dynamic category fields
-
+            The category fields
         """
-        cols = qdb.util.get_table_cols(self._table_name(self._id))
-        cols.remove("sample_id")
+        with qdb.sql_connection.TRN:
```
wasade: this all seems pretty similar to ` _get_categories`, right?

qiita_db/metadata_template/base_metadata_template.py
```diff
@@ -1255,47 +1309,26 @@ def _update(self, md_template):
...
-                    """.format(self._table_name(self._id), sql_eq_cols,
-                               single_value, sql_cols)
-            for sample in samples_to_update:
-                sample_vals = [md_template[col][sample]
-                               for col in cols_to_update]
-                sample_vals.insert(0, sample)
-                qdb.sql_connection.TRN.add(sql, sample_vals)
+            new_columns = []
+            for sid, df in to_update.groupby(level=0):
+                values = {k[1]: v for k, v in df.to_dict()['to'].iteritems()}
```
wasade: is "to" a special key? i don't really understand what's going on here. Why not just take a row or a column and cast that to `dict`?

qiita_db/metadata_template/base_metadata_template.py
```diff
@@ -1255,47 +1309,26 @@ def _update(self, md_template):
...
+                new_columns.extend(values.keys())
+                sql = """UPDATE qiita.{0}
+                         SET sample_values = sample_values || %s
+                         WHERE sample_id = %s""".format(
+                            self._table_name(self._id))
+                qdb.sql_connection.TRN.add(sql, [dumps(values), sid])
+
+            new_columns = list(set(new_columns).union(set(self.categories())))
+            table_name = self._table_name(self.id)
+            values = '{"columns": %s}' % dumps(new_columns)
```
wasade: same same. may be better to decompose this into a helper method since this pattern has come up a bunch

qiita_db/metadata_template/base_metadata_template.py
```diff
@@ -1397,9 +1430,12 @@ def get_category(self, category):
             If category is not part of the template
         """
         with qdb.sql_connection.TRN:
-            qdb.util.check_table_cols([category], self._table_name(self._id))
-            sql = 'SELECT sample_id, {0} FROM qiita.{1}'.format(
-                category, self._table_name(self._id))
+            if category not in self.categories():
+                raise qdb.exceptions.QiitaDBColumnError(category)
+            sql = """SELECT sample_id, sample_values->>'{0}' as {0}
```
wasade: pretty similar to `__getitem__`?

qiita_db/metadata_template/test/test_sample_template.py
```diff
@@ -1343,10 +1343,13 @@ def test_update_numpy(self):
 
         sql = "SELECT * FROM qiita.sample_{0}".format(st.id)
         obs = self.conn_handler.execute_fetchall(sql)
-        exp = [['%s.Sample1' % self.new_study.id, 'false',
-                '2015-09-01 00:00:00'],
-               ['%s.Sample2' % self.new_study.id, 'true',
-                '2015-09-01 00:00:00']]
+        exp = [
+            ['%s.Sample2' % self.new_study.id, {
+                'bool_col': 'true', 'date_col': '2015-09-01 00:00:00'}],
```
wasade: We want to be careful here:

```python
In [11]: json.loads('{"foo": true}')
Out[11]: {'foo': True}

In [12]: json.loads('{"foo": "true"}')
Out[12]: {'foo': 'true'}
```

qiita_db/util.py
```diff
@@ -1651,22 +1649,22 @@ def get_artifacts_information(artifact_ids, only_biom=True):
                     'deprecated': cmd.software.deprecated}
 
             # now let's get the actual artifacts
-            ts = {}
+            ts = {None: []}
```
wasade: `None` for a key is weird, why is this needed?

qiita_db/metadata_template/base_metadata_template.py
antgonza: Not really, my guess is that the confusion is how github displays this block. Basically, to is created in the line above: `to_update = pd.DataFrame({'to': changed_to}, index=changed.index)`

qiita_db/metadata_template/base_metadata_template.py
antgonza: 👍

qiita_db/metadata_template/base_metadata_template.py
antgonza: Yes, but again Sample vs SampleTemplate issue ...

qiita_db/metadata_template/test/test_sample_template.py
antgonza: Good point! For our peace of mind, we always load the info via pandas as str (dtype=str) and use the df.to_json

qiita_db/util.py
antgonza: ts is a cache (prep id : target subfragment) so we don't have to query multiple times the target subfragment for a prep info file. Now, some artifacts (like analysis) do not have a prep info file; thus the None. It's possible to do and if/else but this was easier when first written, let me know ...

qiita_db/metadata_template/base_metadata_template.py
wasade: What i think is going on here:

- the group is cast to a dict
- the key "to" is selected from this dict, which corresponds to the column in the group
- the key/value pairs in the Series representing the column are iterated
- a new dict is created using something. I don't know what that something is. It seems like the index is a heirarchical one in which case `k[1]` is pulling out some field, but that doesn't make sense given that the index value on line 1315 is described as `sid` so I'm assuming that's a sample ID. In which case, `k[1]` would be the character at index 1 in the sample ID string

If the above is correct, than I remain confused :) If the above is not correct, would it be possible to restructure the code, and/or include a comment describing precisely what this block of code does?

qiita_db/metadata_template/base_metadata_template.py
wasade: ah okay

qiita_db/metadata_template/test/test_sample_template.py
wasade: There was a lot of use of `json.dumps` which is going to likely have a similar issue. I don't see how the use of pandas protects us from implicit behavior of the json encoders / decoders

qiita_db/util.py
wasade: If `None` corresponds to analyses which do not have a prep (or it doesn't make sense to have one), would it be possible to include a comment as such denoting the interpretation of the key?

qiita_db/metadata_template/base_metadata_template.py
wasade: Isn't the `c in r[1]` a list lookup?

it would really help to break this up and not use a dict comprehension, or to break up the magic numbers and assign variable names

qiita_db/metadata_template/test/test_sample_template.py
antgonza: pandas dtype=str load (or should load) everything as string; thus, pandas will not try to infer datatypes. From my own experience, tools will use the dtype of pandas to infer the datatype of the column. Thus, if in pandas everything is str, the other tools will use as the type.

```
In [1]: import pandas as pd

In [2]: pd.DataFrame.from_dict({'val': {'row': 5}}).dtypes
Out[2]: 
val    int64
dtype: object

In [3]: pd.DataFrame.from_dict({'val': {'row': 5}}, dtype=str).dtypes
Out[3]: 
val    object
dtype: object
```

qiita_db/metadata_template/base_metadata_template.py
antgonza: After your comment I realized that this could be simplified much more, hope that's the case.

PR #2746, opened 2018-12-04
Port appears twice for localhost

charles-cowart:
Minor fix for bug that shows https://localhost:21174:21174 during start,
which may confuse student developers.

### DIFF
--- a/scripts/qiita
+++ b/scripts/qiita
@@ -28,7 +28,7 @@ from qiita_core.qiita_settings import qiita_config, r_client
 from qiita_ware.ebi import EBISubmission
 from qiita_ware.commands import submit_EBI as _submit_EBI
 
-import re
+from re import search
 
 
 try:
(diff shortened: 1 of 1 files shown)


scripts/qiita
```diff
@@ -28,6 +28,8 @@ from qiita_core.qiita_settings import qiita_config, r_client
 from qiita_ware.ebi import EBISubmission
 from qiita_ware.commands import submit_EBI as _submit_EBI
 
+import re
```
antgonza: what about `from re import search` and then just use `search`?

scripts/qiita
```diff
@@ -306,7 +308,14 @@ def start(port, master):
         else:
             raise
     base_url = qiita_config.base_url
-    click.echo("Qiita started on %s:%d @ %s" % (base_url, port, ctime()))
+
+    if re.search(':\d+$', base_url):
+        tmp_str = "Qiita started on %s @ %s" % (base_url, ctime())
```
antgonza: We normally try to avoid using the data type in the variable name, for example we try to _not_:
```
my_list = ['1']
var_str = 'dfasdf'
a_set = set([1, 1 ,1, 2])
```

PR #2749, opened 2018-12-06
Sample generate files improvements

antgonza:
~https://github.com/biocore/qiita/pull/2743 should be reviewed/merged first~

### DIFF
--- a/qiita_db/metadata_template/base_metadata_template.py
+++ b/qiita_db/metadata_template/base_metadata_template.py
@@ -1360,7 +1360,9 @@ class MetadataTemplate(qdb.base.QiitaObject):
             # 4  XX.Sample3  physical_specimen_location  new location
             to_update.reset_index(inplace=True)
             new_columns = []
+            samples_updated = []
             for sid, df in to_update.groupby('sample_name'):
+                samples_updated.append(sid)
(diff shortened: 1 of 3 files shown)


qiita_db/metadata_template/base_metadata_template.py
```diff
@@ -1315,7 +1331,7 @@ def _update(self, md_template):
                     "There are no differences between the data stored in the "
                     "DB and the new data provided",
                     qdb.exceptions.QiitaDBWarning)
-                return
+                return set([]), set([])
```
charles-cowart: Is it better to return two empty sets vs None, or (None, None), throw an Exception, etc?

qiita_db/metadata_template/base_metadata_template.py
```diff
@@ -1396,9 +1416,9 @@ def update(self, md_template):
             # Clean and validate the metadata template given
             new_map = self._clean_validate_template(
                 md_template, self.study_id, current_columns=self.categories())
-            self._update(new_map)
+            samples, columns = self._update(new_map)
             self.validate(self.columns_restrictions)
```
charles-cowart: Feels like it'd be more readable to test for not None or similar here, instead of having to go down into validate(), but I defer.

PR #2752, opened 2018-12-07
fix #2734

### LINKED ISSUE #2734: Deleting multiple artefacts to reprocess requires delete back to linked data artefact
It is really tedious and time-consuming to delete all the generated artefacts back down the tree.
An option to delete back to the linked data would be helpful when reprocessing

antgonza:


qiita_db/artifact.py
```diff
@@ -512,21 +512,30 @@ def delete(cls, artifact_id):
...
                     artifact_id, "it is public")
 
-            # Check if this artifact has any children
-            if instance.children:
-                raise qdb.exceptions.QiitaDBArtifactDeletionError(
-                    artifact_id,
-                    "it has children: %s"
-                    % ', '.join([str(c.id) for c in instance.children]))
+            # delete children if children exists
+            for c in sorted(instance.children):
```
charles-cowart: I wonder, if deleting individual children in sequence will take time, if we should have a version of delete that can accept a list of children and process the deletes on the back-end 'in bulk'. What do you think?

qiita_db/artifact.py
```diff
@@ -512,21 +512,30 @@ def delete(cls, artifact_id):
...
+                try:
+                    instance.delete(c.id)
+                except qdb.exceptions.QiitaDBArtifactDeletionError as e:
+                    msg = 'because children %s' % str(e)
```
charles-cowart: 'because children %s' seems to potentially be lacking a verb, as far as error messages go

qiita_db/artifact.py
charles-cowart: IMHO, this could be made an issue for later...

qiita_db/artifact.py
charles-cowart: ...this might need to be clarified before approving.

PR #2771, opened 2018-12-21
Sanitize 'Add Person' name

charles-cowart:
In Create/Edit Study page, there is an option to add PIs and staff using
pull-down menus. New PIs and staff are added to the system using a
sub-form. This sub-form has a field for the person's name. This field
cannot have non-printable ASCII characters, as the PI name becomes part
of system fields downstream in Qiita. This will not correct names for
existing persons entered, only sanitize names for future persons.

### DIFF
--- a/qiita_pet/templates/edit_study.html
+++ b/qiita_pet/templates/edit_study.html
@@ -30,6 +30,7 @@ $(document).ready(function() {
     $( "#{{creation_form.lab_person.id}}" ).chosen({
         allow_single_deselect: true
       });
+
     $("#study_title").on("keyup", function() {
       var title = $(this).val();
       // removing any duplicated whitespaces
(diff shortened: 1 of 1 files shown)


qiita_pet/templates/edit_study.html
```diff
@@ -39,6 +40,17 @@
       // removing white spaces from the front of the text
       $(this).val(title.trimLeft());
     });
+
+    $("#new_person_name").on("keyup", function() {
+      var charlie  = $(this).val();
```
antgonza: mhhh ... not sure if I agree with the name of the variable ...

charles-cowart: doh! My apologies, it was guaranteed to be a unique name; I forgot to make it a more appropriate one. :)

charles-cowart:
No problem. I'm not sure how much space is available in the add new person window, but I'll add something. Patch for the DB forthcoming.

PR #2775, opened 2018-12-27
fix #2741

### LINKED ISSUE #2741: Add study id to stats map
When you go on Qiita and select "stats" under "More info" it brings up a map of the world. If you zoom out of this map and keep zooming out, it doesn't stop when the entire world is shown. Instead, the continents keep repeating....

antgonza:
Example:


charles-cowart:
Is the map in the screen capture above maximally zoomed out? Still, it appears that the map is non-repeating.

PR #2776, opened 2018-12-29
fix #2747

### LINKED ISSUE #2747: Color deprecated analysis artifacts black

antgonza:


antgonza:
Forgot to add the examples, here they are (see network color and message):
- Regular artifact: regular
- Outdated artifact: outdated
- Deprecated artifact: deprecated

qiita_pet/templates/artifact_ajax/artifact_summary.html
```diff
@@ -111,10 +111,16 @@
...
   <div class='col-md-12'>
     <h4>
-      {% if processing_info and not processing_info['software_active'] %}
-        <div class="alert alert-warning" role="alert">
-          Warning, the software that generated this artifact is deprecated as a new version is available.
-        </div>
+      {% if processing_info %}
+        {% if processing_info['software_deprecated'] %}
+          <div class="alert alert-danger" role="alert">
+            Danger, the software that generated this artifact was produced by a software version with a known bug and the results are wrong, please re-run with the newer version.
```
charles-cowart: Danger, the software that generated this artifact was produced by a software version with a known bug and the results are wrong, please re-run with the newer version. -> Warning, the software that generated this artifact has a known bug, and these results are incorrect. Please re-run to produce new results with the latest version of the software.

charles-cowart:
@ElDeveloper That's a good point, it wasn't clear to me that there should be a color difference between 'deprecated' and 'deprecated with bug'. However, I can see that the latter is darker. It seems like in either case, you'd want to re-run, so they should really be the same color, or else there should be two different colors in the legend as well.

antgonza:
RE: different colors: sounds like the suggestion is either making them really different or similar. I like really different. What about orange (with bug), and grey?

RE: spanning different categories: this is not new, see current; one option is to change shapes, could you open an issue if you feel strong about it?
screen shot 2019-01-03 at 6 29 29 am

ElDeveloper:
I prefer black for "deprecated" as it is very easy to spot in a graph. That being said, my main point was that an artifact that was "deprecated without a bug" should not be that much different from a "normal" artifact. My suggestion is that artifacts that have been "deprecated without a bug" should be a very light gray.

As for the shapes, I have opened this issue #2778

charles-cowart:
I agree it'd be nice if deprecated without bug could be just a shade difference from a normal artifact. However, I think a light gray artifact might be easily missed when the graph might contain numerous other colors.

PR #2777, opened 2019-01-01
fix #2106, fix #2713

### LINKED ISSUE #2106: Public dump should differentiate platform, target, etc
The public dump does not differentiate between platform, target gene or target subfragment. These details are not accessible across all studies, but across many. This is important in order for a user to programmatically know whether study X can reasonably be merged with study Y. In the absence of this information, a user may, for instance, mistakenly combine Deblur'd ITS data with Deblur'd 16S data.

### LINKED ISSUE #2713: Add software version to bulk release
Hi,
I am crawling through the rich data treasure of qiita in terms of the "public BIOM and metadata files". Many studies now have deprecated Deblur tables. I wonder how I can tell apart the deprecated ones from the current ones?
Thanks
Stefan

antgonza:
To fix those issues we first merged the code for merging scheme so we can use it in the biom/maps generation + then added the target gene and platform info.

### DIFF
--- a/qiita_db/meta_util.py
+++ b/qiita_db/meta_util.py
@@ -359,38 +359,66 @@ def generate_biom_and_metadata_release(study_status='public'):
             if a.processing_parameters is None:
                 continue
 
-            cmd_name = a.processing_parameters.command.name
+            processing_params = a.processing_parameters
+            cmd_name = processing_params.command.name
+            ms = processing_params.command.merging_scheme
+            software = processing_params.command.software
+            software = '%s v%s' % (software.name, software.version)
 
             # this loop is necessary as in theory an artifact can be
             # generated from multiple prep info files
-            human_cmd = []
+            afps = [fp for _, fp, _ in a.filepaths if fp.endswith('biom')]
+            merging_schemes = []
+            parent_softwares = []
             for p in a.parents:
-                pp = p.processing_parameters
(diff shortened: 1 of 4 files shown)


qiita_db/meta_util.py
```diff
@@ -359,38 +359,66 @@ def generate_biom_and_metadata_release(study_status='public'):
...
-                #          human readable name)
                 for pt in a.prep_templates:
+                    categories = pt.categories()
+                    platform = ''
+                    target_gene = ''
+                    if 'platform' in categories:
+                        platform = ', '.join(
+                            set(pt.get_category('platform').values()))
+                    if 'target_gene' in categories:
+                        target_gene = ', '.join(
```
charles-cowart: I feel like usually it's better to pass the list of values, instead of concatenating them into a string prematurely.

antgonza: I _think_ this is the right place to concatenate but perhaps I'm missing something, why do you consider that here is premature?

charles-cowart: My mistake; I reviewed the whole method and there's nothing wrong.

charles-cowart:
@antgonza Would you like to resolve the conflicts GitHub is showing? My guess is that the code from fix#2106 supersedes the code from dev, but I'd like to confirm before merging.

antgonza:
@charles-cowart, ready for another pass ...

PR #2781, opened 2019-01-09
Fix 2778

antgonza:
- Regular artifact: screen shot 2019-01-09 at 3 03 37 pm

- Outdated: screen shot 2019-01-09 at 3 04 59 pm

- Deprecated: screen shot 2019-01-09 at 3 05 26 pm

- Text box: screen shot 2019-01-09 at 3 12 26 pm

### DIFF
--- a/qiita_pet/static/js/networkVue.js
+++ b/qiita_pet/static/js/networkVue.js
@@ -187,7 +187,9 @@ Vue.component('processing-graph', {
                 } else {
                   // In this case the job changed to either 'running', 'queued' or 'waiting'. In
                   // this case, we just need to update the internal values of the nodes and the colors
-                  jobNode.color = vm.colorScheme[jobStatus];
+                  var node_info = vm.colorScheme[jobStatus]
+                  jobNode.color = node_info;
+                  jobNode.shape = node_info['shape'];
                   jobNode.status = jobStatus;
                   vm.nodes_ds.update(jobNode);
                 }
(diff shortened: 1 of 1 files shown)


ElDeveloper:
@antgonza, thanks for putting this together. I just noticed that maybe we should make the types a different shape too so that the legend becomes something like:

Job status (circles): success | running | error ....
Artifact status (triangles): normal | outdated | deprecated
Data types (diamonds): ..(?)...

Can a data type change its status? Or is it always "normal".

antgonza:
It's always normal and basically is a "promised" artifact ... in other words, when a job is created (in construction) it already knows its outputs, represented as "type"

antgonza:
Now, the banner looks like this:
screen shot 2019-01-10 at 3 53 37 am

qiita_pet/static/js/networkVue.js
```diff
@@ -1123,11 +1128,11 @@ Vue.component('processing-graph', {
     }
     var full_text = '<table style="border-spacing: 3px;border-collapse: separate;">' +
       '<tr>' +
-        '<td><small>Circle status:</small></td>' +
+        '<td><small>Job status (circles):</small></td>' +
         '<td>' + circle_statuses.join('') + '</td>' +
       '</tr>' +
       '<tr>' +
-        '<td><small>Circle types:</small>' +
+        '<td><small>Artifact status (triangle):</small>' +
```
ElDeveloper:
```suggestion
        '<td><small>Artifact status (triangles):</small>' +
```

ElDeveloper:
I see, it's a bit confusing but I guess we can just keep it for now. Correct me if I am wrong but the legend should have "in_construction" and "deleting" under the Artifact Status section, right?

ElDeveloper:
Oh OK, fair enough. Then I only have one outstanding comment.

PR #2789, opened 2019-01-15
Minor fix

charles-cowart:
This fix was left out of the last PR.

### DIFF
--- a/qiita_db/processing_job.py
+++ b/qiita_db/processing_job.py
@@ -86,12 +86,18 @@ class Watcher(Process):
     def _element_extract(self, snippet, list_of_elements,
                          list_of_optional_elements):
         results = {}
+        missing_elements = []
+
         for element in list_of_elements:
             value = search('<%s>(.*?)</%s>' % (element, element), snippet)
             if value:
                 results[element] = value.group(1)
             else:
-                raise AssertionError("%s element is missing!" % element)
+                missing_elements.append(element)
+
+        if missing_elements:
+            raise AssertionError("The following elements were not found: %s"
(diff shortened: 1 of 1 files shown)


qiita_db/processing_job.py
```diff
@@ -160,9 +166,7 @@ def run(self):
 
                         # determine if anything has changed since last poll
                         if results['Job_Id'] in self.processes:
-                            tmp = cmp(self.processes[results['Job_Id']],
-                                      results)
-                            if 0 != tmp:
+                            if self.processes[results['Job_Id'] != results:
```
ElDeveloper: This is missing one closing bracket.

qiita_db/processing_job.py
```diff
@@ -86,12 +86,18 @@ def __init__(self):
...
             value = search('<%s>(.*?)</%s>' % (element, element), snippet)
             if value:
                 results[element] = value.group(1)
             else:
-                raise AssertionError("%s element is missing!" % element)
+                missing_elements.append(element)
+
+        if missing_elements:
+            raise AssertionError("The following elements were not found: %s"
+                                 % str(missing_elements))
```
ElDeveloper: Can you format the list as: `', '.join(missing_elements)` and remove the `str` cast.

qiita_db/processing_job.py
charles-cowart: thanks!

qiita_db/processing_job.py
charles-cowart: You got it.

PR #2792, opened 2019-01-17
split the bk table removal from patch 69 and 70

antgonza:


qiita_db/support_files/patches/71.sql
```diff
@@ -4,15 +4,18 @@
 -- 4/4 This is the continuation of the patching that started in 68.sql, let's
 -- remove all the temp (_bk) tables we created
 
--- Dropping all the _bk tables
+-- On January 17, 2019 while releasing the release candidate; we realized that
+-- we were hitting the max memory allocated for the machine so we needed to
+-- split this in to 2
```
charles-cowart: in to 2 -> into two.

qiita_db/support_files/patches/71.sql
```diff
@@ -4,15 +4,18 @@
...
+-- Dropping all the _bk tables 1/2
 DO $do$
 DECLARE
     dyn_table varchar;
 BEGIN
   FOR dyn_table IN
     SELECT DISTINCT table_name
     FROM information_schema.columns
-    WHERE table_name LIKE '%_bk'
+    WHERE table_name LIKE '%_bk' LIMIT 2100
```
charles-cowart: Will the hardcoded 2100 limit need to be updated in the future, as the number of tables grow? We might need a comment here whether this will be an issue in the future.

antgonza: patches are only ran once, so it should not be a problem in the future ...

PR #2796, opened 2019-01-22
Fix qsub hanging error

charles-cowart:
It appears that qsub can hang when not given a config file, or perhaps a
parameter is in error; it assumes that more information will be provided
on stdin, hence it will wait forever. If this happens within Tornado, it
will cause Qiita workers to become unresponsive.

Python3 allows one to call .communicate() with a timeout. However, this
feature is not available in Python2. As a temporary measure, Qopen has
been created to wrap Popen() calls in a different Thread, and the Qopen
object can be subsequently joined with a timeout.

### DIFF
--- a/qiita_db/processing_job.py
+++ b/qiita_db/processing_job.py
@@ -21,6 +21,8 @@ from subprocess import Popen, PIPE
 from time import sleep
 from uuid import UUID
 from os.path import join
+from threading import Thread
+from qiita_core.exceptions import IncompetentQiitaDeveloperError
 
 
 class Watcher(Process):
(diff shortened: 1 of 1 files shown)


antgonza:
Should this change be added (code replaced) in _all_ instances of `stdout, stderr = proc.communicate()`?

charles-cowart:
I thought about that, but I thought this method changed things the least. We know that qsub can give us behavior we didn't know about, but on the whole we know that most of the stuff we call seems like they've been well-behaved.

Also, there are still backend things calling ProcessingJob._system_call() directly, that will take longer than a few seconds, but how long who knows. We could make it a TODO? To be honest, I think it would be better to migrate to Python3 and replace Qopen() with a proper communicate(timeout=n) call.

PR #2797, opened 2019-01-23
fix download button for public and shared raw data

antgonza:


qiita_db/test/test_meta_util.py
```diff
@@ -68,8 +68,18 @@ def test_validate_filepath_access_by_user(self):
         # Now shared has access to public study files
         self._set_artifact_public()
         for i in [1, 2, 3, 4, 5, 9, 12, 17, 18, 19, 20, 21]:
-            self.assertTrue(qdb.meta_util.validate_filepath_access_by_user(
-                user, i))
+            obs = qdb.meta_util.validate_filepath_access_by_user(user, i)
+            if i < 3:
```
charles-cowart: A comment on what 3 means would be good.

PR #2803, opened 2019-01-26
P3.6 items equal

antgonza:
In this PR:
- re-add line so matplotlib works fine
- self.assertItemsEqual -> self.assertCountEqual
- fix all qiita_core
- fix all processing_job

Note that the biggest differences in networkx versions are:
- before when calling edges/nodes/etc it used to return a dict and now returns a tuple
- all the lists are of the type vs. just a list, for example: nodes, returns NodeView(()) vs []

### DIFF
--- a/qiita_core/tests/test_configuration_manager.py
+++ b/qiita_core/tests/test_configuration_manager.py
@@ -138,7 +138,7 @@ class ConfigurationManagerTests(TestCase):
 
             obs_warns = [str(w.message) for w in warns]
             exp_warns = ['Random cookie secret generated.']
-            self.assertItemsEqual(obs_warns, exp_warns)
+            self.assertCountEqual(obs_warns, exp_warns)
 
         self.assertNotEqual(obs.cookie_secret, "SECRET")
         # Test default base_data_dir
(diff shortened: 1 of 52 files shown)


qiita_pet/webserver.py
```diff
@@ -239,6 +239,7 @@ def __init__(self):
 
         # 404 PAGE MUST BE LAST IN THIS LIST!
         handlers.append((r".*", NoPageHandler))
+        handlers.append((r"/.*", NoPageHandler))
```
charles-cowart: This line is new, not sure what's supposed to be for. Can you add a comment please?

charles-cowart:
52 files is a lot, but most of the changes are straightforward enough (method name change, slight parameter syntax change, etc.). There is one line where I'd like your comment, just to know what it's doing, and to confirm that it should still be there.

qiita_core/tests/test_configuration_manager.py
```diff
@@ -30,7 +30,7 @@ def setUp(self):
         environ['QIITA_CONFIG_FP'] = self.conf_fp
 
         self.conf = ConfigParser()
-        with open(self.conf_fp, 'U') as f:
+        with open(self.conf_fp, newline=None) as f:
```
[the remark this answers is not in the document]
antgonza: you are right but agree about being explicit ...

qiita_pet/webserver.py
antgonza: right, this was a debugging line, removing ....

antgonza:
Thanks for review, I have addressed your comments ...

PR #2814, opened 2019-02-20
py3.6 qiita-ware

antgonza:


.travis.yml
```diff
@@ -22,9 +22,7 @@ before_install:
 install:
   # install a few of the dependencies that pip would otherwise try to install
   # when intalling scikit-bio
-  - travis_retry conda create -q --yes -n qiita python=3.6 pip nose flake8
-    pyzmq networkx pyparsing natsort mock future libgfortran seaborn nltk pandas
-    matplotlib scipy numpy h5py
+  - travis_retry conda create -q --yes -n qiita python=3.6 pip libgfortran numpy
```
charles-cowart: It's okay to remove these other dependencies? nltk and pandas are being picked up somewhere else?

.travis.yml
```diff
@@ -33,7 +31,7 @@ install:
   - make
   - ./webdis &
   - popd
-  - travis_retry pip install . --process-dependency-links
+  - travis_retry pip install . --no-binary redbiom
```
charles-cowart: As I understand it, this forces redbiom to be installed from source, rather than precompiled binary? I take it this is to resolve one or more dependency issues?

qiita_db/test/test_study.py
```diff
@@ -866,7 +866,7 @@ def test_study_tags(self):
 
         # testing that insertion went fine
         obs = qdb.study.Study.get_tags()
-        exp = {'user': ['this is my tag', 'I want GOLD!!'],
+        exp = {'user': ['I want GOLD!!', 'this is my tag'],
```
charles-cowart: I want GOLD? :)

qiita_ware/ebi.py
```diff
@@ -704,8 +704,10 @@ def write_xml_file(self, element, fp):
         fp : str
             The filepath to which the XML will be written
         """
-        makedirs(self.xml_dir)
-        ET.ElementTree(element).write(fp, encoding='UTF-8')
+        if not exists(self.xml_dir):
```
charles-cowart: I have a util function in qiita_db that wraps this called create_nested_path(). I think this is fine for now. I'll make a note to migrate that functionality so it's accessible in both locations and replace this line.

qiita_ware/ebi.py
```diff
@@ -1060,6 +1062,7 @@ def _generate_demultiplexed_fastq_demux(self, mtime):
                 raise EBISubmissionError(error_msg)
             for s, i in to_per_sample_ascii(demux_fh,
                                             self.prep_template.keys()):
+                s = s.decode('ascii')
```
charles-cowart: Any chance there will be characters in the data that could make this fail or produce /x000 type encodings that would cause problems downstream?

.travis.yml
charles-cowart: confirmed dependencies have been moved to setup.py.

.travis.yml
antgonza: Well, turns out that the wheel redbiom binary is broken and the quickest way forward is to install from source ...

qiita_db/test/test_study.py
antgonza: Yes, yah know, fun with tests!

qiita_ware/ebi.py
antgonza: Sounds good!!

qiita_ware/ebi.py
antgonza: Anything is possible but IMOO almost impossible as these are sample names and these are fields that are really restricted so we _should_ be fine.

.travis.yml
charles-cowart: sounds good then. Compilation for redbiom itself won't add appreciable compilation time. If versioning is 'latest' that could introduce variability over time, but no more so than using the latest binary.

qiita_ware/ebi.py
charles-cowart: Sounds reasonable.

qiita_db/test/test_study.py
charles-cowart: Fair enough. :)

PR #2817, opened 2019-02-22
python3.6 qiita_pet

antgonza:


qiita_pet/handlers/api_proxy/tests/test_studies.py
```diff
@@ -347,8 +347,11 @@ def test_study_files_get_req_multiple(self):
             'remaining': ['uploaded_file.txt'], 'message': '',
             'file_types': [
                 ('raw_forward_seqs', True,
-                 ['test_2.R1.fastq.gz', 'test_1.R1.fastq.gz']),
+                 sorted(['test_2.R1.fastq.gz', 'test_1.R1.fastq.gz'])),
```
charles-cowart: Would it be better to simply reorder this two element list manually, rather than sort it programmatically?

charles-cowart: Two additional occurrences below.

charles-cowart: I mean it's nice to imply that this list needs to be sorted before acting on it, even if/when the contents change.

qiita_pet/test/test_download.py
```diff
@@ -317,7 +320,10 @@ def test_download(self):
                "1.SKD8.640184\tERX0000001\n1.SKM5.640177\tERX0000005\n"
                "1.SKM7.640188\tERX0000010\n1.SKD7.640191\tERX0000021")
         self.assertEqual(response.code, 200)
-        self.assertRegexpMatches(response.body, exp)
+        # testing as lists so we ignore order
```
charles-cowart: Would it be better to sort the lists so they could be compared for equality? There's a test earlier where all of the strings are the same, but this test data appears to have multiple IDs.

charles-cowart:
Approving as-is, as this is on the 3.7 Migration branch. A few of the comments I think are worth looking into, however.

qiita_db/metadata_template/test/test_prep_template.py
```diff
@@ -1204,7 +1204,7 @@ def test_qiime_map_fp(self):
         pt = qdb.metadata_template.prep_template.PrepTemplate(1)
         exp = join(qdb.util.get_mountpoint('templates')[0][1],
                    '1_prep_1_qiime_[0-9]*-[0-9]*.txt')
-        self.assertRegexpMatches(pt.qiime_map_fp, exp)
+        self.assertRegex(pt.qiime_map_fp, exp)
```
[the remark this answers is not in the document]
antgonza: yup, assertRegexpMatches has been superseded by assertRegex

qiita_pet/handlers/api_proxy/tests/test_prep_template.py
```diff
@@ -435,21 +435,20 @@ def test_prep_template_post_req(self):
...
-                     'library_construction_protocol, platform.'),
-                    ('See the Templates tutorial for a description of these '
-                     'fields.')],
+                   'Both a converter and dtype were specified for column '
+                   'sample_name - only the converter will be used', 'Some '
+                   'functionality will be disabled due to missing columns:',
+                   '\tEBI submission disabled: center_name, '
+                   'experiment_design_description, instrument_model, '
+                   'library_construction_protocol, platform;',
+                   '\tDemultiplexing disabled.: barcode;', '\tDemultiplexing '
```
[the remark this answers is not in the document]
antgonza: Right, this should be part of the refactor of the warnings messages, IMOO out of the scope of this PR.

qiita_pet/handlers/api_proxy/tests/test_studies.py
antgonza: I rather be certain that we always order based on the machine the tests are running on ...

qiita_pet/test/test_download.py
antgonza: Possible to change but not sure what we will gain from this ...

qiita_pet/handlers/api_proxy/tests/test_prep_template.py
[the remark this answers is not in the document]
charles-cowart: ok

qiita_pet/handlers/api_proxy/tests/test_studies.py
charles-cowart: ok

qiita_pet/test/test_download.py
charles-cowart: Just that before, our test was 'self.assertRegexpMatches(response.body, exp)', and now we're just asserting that the number of elements in obs and exp is equal: self.assertCountEqual(obs, exp).

charles-cowart: As Antonio pointed out, assertCountEqual() does in fact test the elements of both lists, regardless of order: 
https://www.google.com/url?q=https://docs.python.org/3/library/unittest.html%23unittest.TestCase.assertCountEqual&sa=D&source=hangouts&ust=1550959180059000&usg=AFQjCNHmkIF8ZDeLHZRdII_u_Cwd0ffBHQ

Hence, this is not an issue.

PR #2821, opened 2019-02-25
fix #2558

### LINKED ISSUE #2558: include processing information in analysis
Right now it's possible to infer `qiita_study_id` and `qiita_prep_id` from analysis mapping file, but for people who merged multiple studies which were done using a variety of processing options, it's not possible to look up how exactly the input BIOM artifact has been processed.

This information is displayed before the analysis is created so it must be recorded somewhere. Exposing it to users later would be useful.

antgonza:
This is how an analysis looks now:
analysis-study-artifact-listing

This PR also fixes the artifact selection display, before:
screen shot 2019-02-25 at 10 30 26 am
now:
screen shot 2019-02-25 at 10 40 03 am


antgonza:
That's correct.

PR #2822, opened 2019-02-26
fix #2487

### LINKED ISSUE #2487: Filter view studies by preparation type
This came up on the help account. A user noted that we cannot filter studies by preparation type. In this specific usecase, the user wanted to find all of the studies in which we had a shotgun preparation.

antgonza:


antgonza:
Forgot to add a screenshot:
screen shot 2019-02-26 at 11 46 31 am

qiita_db/util.py
```diff
@@ -1331,9 +1331,20 @@ def generate_study_list(user, visibility):
...
-                              ORDER BY artifact_id)
+      including their software deprecated value and their prep info file data
+      type values
+            (SELECT array_agg(row_to_json(
+                (artifact_id, qs.deprecated, (
+                    SELECT array_agg(DISTINCT data_type)
+                    FROM qiita.prep_template
+                    LEFT JOIN qiita.data_type USING (data_type_id)
+                    WHERE artifact_id IN (
+                        SELECT *
```
charles-cowart: It might be better to explicitly define what we want out of this table, rather than using wildcard (*), but then our tables don't exactly change often.

PR #2833, opened 2019-02-28
WIP: fix #2774

### LINKED ISSUE #2774: md5sum for artifact files
I've downloaded a biom table a while ago, published a paper with these data and today someone requested information on how to reproduce results. I forgot to rename the suggested filename back than to include the artifact ID :-/ And the raw biom file also does not hold this information - at least not when using `biom summarize-table`. Thus, I have to compare md5 hash values. 
Could qiita somehow show those hash values for the file to download. That would save me downloading several files to computing and compare the md5s.

antgonza:
This is how the GUI looks now:
screen shot 2019-02-28 at 3 57 20 pm


antgonza:
superseded by https://github.com/biocore/qiita/pull/2878

PR #2841, opened 2019-03-19
User creation status msg

charles-cowart:
After a new user is created, the user is given a green status message informing them that their account was creating successfully, and to check their mailbox for a verification email.


qiita_pet/handlers/auth_handlers.py
```diff
@@ -66,7 +66,18 @@ def post(self):
                 self.redirect(u"%s/?level=danger&message=%s"
                               % (qiita_config.portal_dir, url_escape(msg)))
                 return
-            self.redirect(u"%s/" % qiita_config.portal_dir)
+
+            msg = ("<H1><p>User Successfully Created</H1>Your Qiita account "
```
antgonza: - [ ] Can you make sure to use all lower case or all upper case (I prefer lower) for the html? You have H1 and p
- [ ] Can you make sure that you open and close all <p>? See: `<H1><p>User Successfully Created</H1>`
- [ ] I think H1 is too large, what about H3? This is up to you.

charles-cowart: Thanks! Tried h2 as well, felt it was too in the middle, so I switched to h3.

qiita_pet/handlers/auth_handlers.py
```diff
@@ -66,7 +66,18 @@ def post(self):
                 self.redirect(u"%s/?level=danger&message=%s"
                               % (qiita_config.portal_dir, url_escape(msg)))
                 return
-            self.redirect(u"%s/" % qiita_config.portal_dir)
+
+            msg = ("<h3>User Successfully Created</h3>Your Qiita account "
+                   "has been successfully created. An email has been sent to "
+                   "the email address you provided. This email contains "
+                   "instructions on how to activate your account.</p>"
```
antgonza: This `</p>` doesn't have an opening.

charles-cowart: Good catch!

PR #2844, opened 2019-03-20
adding nginx/supervisord

antgonza:


qiita_core/support_files/config_test.cfg
```diff
@@ -23,7 +23,7 @@ LOG_DIR =
 REQUIRE_APPROVAL = True
 
 # Base URL: DO NOT ADD TRAILING SLASH
-BASE_URL = https://localhost:21174
+BASE_URL = https://localhost:8383
```
ElDeveloper: Are there any documents that need to change based on this modification? If I understand correctly, new installations that do not use nginx will have to go to localhost:8383 to use Qiita, is that right?

antgonza: Thanks for the quick review. That is right if they use the exact same configuration that the one that is in the example and being used in Travis; however, in local machines using 80 works fine - at least in my laptop. Now, I checked all the documentation and install instructions and updated the places that needed updating. However, happy to add something else, if you consider it proper; just let me know.

.travis.yml
```diff
@@ -31,6 +31,9 @@ install:
   - ./webdis &
   - popd
   - travis_retry pip install . --no-binary redbiom
+  # staring nginx
```
charles-cowart: staring -> starting

.travis.yml
```diff
@@ -31,6 +31,9 @@ install:
...
+  - mkdir /home/travis/miniconda3/envs/qiita/var/run/nginx/
```
charles-cowart: Not sure why are we putting the NGINX directory under Qiita's Conda environment. Is that just for document root purposes?

qiita_core/support_files/config_test.cfg
charles-cowart: I might be missing something - why are we shifting from port 21174 to 8383? It might be good to put a comment in on that.

qiita_pet/nginx_example.conf
```diff
@@ -19,8 +42,7 @@ http {
 
           # CHANGE ME: This should match the WORKING_DIR in your qiita
           # config. E.g.,
-          # alias /Users/username/qiita/qiita_db/support_files/test_data/working_dir/;
-          alias ;
+          alias /Users/username/qiita/qiita_db/support_files/test_data/working_dir/;
```
charles-cowart: This line was previously commented out as part of the above comment, while the live setting was 'alias ;'. Just making sure this wasn't a temp change that needs to be changed back.

.travis.yml
antgonza: k

.travis.yml
antgonza: Good question, note that this due to how conda adds nginx and the folders it expects to have to create logs/etc; will add it to the instructions

qiita_core/support_files/config_test.cfg
antgonza: Adding some lines to clarify this; the reason in nginx and the multiple workers ...

qiita_pet/nginx_example.conf
antgonza: Yup, before we were not actually using this for anything but now we do and we can't have empty alias ...

PR #2849, opened 2019-03-25
rm hardcoded biom configuration

antgonza:


.travis.yml
```diff
@@ -78,6 +78,11 @@ script:
   - qiita pet webserver --no-build-docs start &
   - sleep 5
   - QIITA_PID=$!
+  - source deactivate
```
charles-cowart: source doesn't need to be changed to conda at this point?

antgonza: I guess no, cause if that was the case the builds will fail and they seem fine on [that line](https://travis-ci.org/biocore/qiita/jobs/511147785#L814).

charles-cowart: thought so too, but thought I'd ask.

PR #2857, opened 2019-04-02
add qza as a filetype

antgonza:


qiita_db/support_files/patches/74.sql
```diff
@@ -0,0 +1,3 @@
+-- Apr 2nd, 2019
+-- Add a new filepath type
+INSERT INTO qiita.filepath_type (filepath_type) VALUES ('qza');
```
charles-cowart: Can you comment on the meaning of qza?

antgonza: I thought it was common knowledge: qiime zip archive ...

charles-cowart: Gotcha. I'll make a note of it.

PR #2869, opened 2019-04-13
fix #2526, fix #2656

### LINKED ISSUE #2526: Problematic visual display when viewing errors
Additional characters are visible in display when viewing the errors from the prep info page - from processing graphical interface.  Screenshot attached.
screenshot_error_deblur_visual interface

### LINKED ISSUE #2656: Test data Artifact 7 is missing file paths
When we (Yoshiki, myself) access the file paths property of article 7, it is an empty list. This leads to failures in the meta-analysis workflow.

```
Error on job 2e104cff-bb0e-4099-8dd4-a44d7fc18677:]
File "/Users/ccowart/qiita/qiita_ware/private_plugin.py", line 412, in private_taskTASK_DICT[task_name](job)
File "/Users/ccowart/qiita/qiita_ware/private_plugin.py", line 37, in build_analysis_filesbiom_files = analysis.build_files(merge_duplicated_sample_ids)
File "/Users/ccowart/qiita/qiita_db/analysis.py", line 873, in build_filesgrouped_samples, rename_dup_samples)
File "/Users/ccowart/qiita/qiita_db/analysis.py", line 905, in _build_biom_tables% aid)

RuntimeError: Artifact 7 does not have a biom table associated
-- | --
```

antgonza:


qiita_pet/static/js/networkVue.js
```diff
@@ -396,7 +396,8 @@ Vue.component('processing-graph', {
         }
 
         if (data['job_status'] === 'error' && data['job_error'] !== null) {
-          rowsContent.push(['Error message:', data['job_error']]);
+          // based on https://stackoverflow.com/a/14129989
+          rowsContent.push(['Error message:', $('<div>').text(data['job_error']).html()]);
```
ElDeveloper: This fix seems to be related to the problem display error messages (as noted in the issue). Would you mind explaining what's going on here? It seems to me that the incorrectly displayed characters are "\n" not any of the HTML tags, so it's not clear to me how this would help improve the display.

antgonza: Some errors will add a bunch of html to the message, that html gets embedded into the main page, which causes the buttons and other elements to look weird; the solution: make sure the errors are encoded/displayed correctly.

ElDeveloper: Got it, thanks for the explanation!

antgonza:
Thanks

PR #2871, opened 2019-04-17
fix #1570

### LINKED ISSUE #1570: errors when raw files have spaces in their name
If a prep template with spaces in it's name is uploaded, split libraries is going to fail. Not sure of the consequences of doing the same in a sample template.

antgonza:
Note that #1570 is a pretty old issue but it still had some validity, here are the 2 cases described there: 
- adding sample/prep info files with spaces: it works fine without any changes
- adding raw files with spaces: we already supported this in python but not in JS cause it was missing `[]`

Additionally, while debugging, found that we needed a couple of changes to better handle errors and properly show them to users so they can report the issue back to us

### DIFF
--- a/qiita_pet/handlers/api_proxy/artifact.py
+++ b/qiita_pet/handlers/api_proxy/artifact.py
@@ -22,6 +22,7 @@ from qiita_db.util import (
     get_mountpoint, get_visibilities, get_artifacts_information)
 from qiita_db.software import Command, Parameters, Software
 from qiita_db.processing_job import ProcessingJob
+from qiita_db.exceptions import QiitaDBError
 
 PREP_TEMPLATE_KEY_FORMAT = 'prep_template_%s'
 
@@ -215,7 +216,7 @@ def artifact_post_req(user_id, filepaths, artifact_type, name,
         # activated so there is no Validate for the given artifact_type
         try:
             command = Command.get_validator(artifact_type)
-        except Exception as e:
+        except QiitaDBError as e:
             return {'status': 'error', 'message': str(e)}
         job = ProcessingJob.create(
             user,
--- a/qiita_pet/handlers/api_proxy/artifact.py
+++ b/qiita_pet/handlers/api_proxy/artifact.py
@@ -211,9 +211,12 @@ def artifact_post_req(user_id, filepaths, artifact_type, name,
             return {'status': 'error',
                     'message': "Can't create artifact, no files provided."}
 
-        # TODO: It might be good to have an API call that wraps to a lower
-        # level method that creates this job.
-        command = Command.get_validator(artifact_type)
+        # This try/except will catch the case when the plugins are not
+        # activated so there is no Validate for the given artifact_type
+        try:
+            command = Command.get_validator(artifact_type)
+        except Exception as e:
+            return {'status': 'error', 'message': str(e)}
         job = ProcessingJob.create(
             user,
             Parameters.load(command, values_dict={
--- a/qiita_pet/templates/study_ajax/artifact_file_selector.html
+++ b/qiita_pet/templates/study_ajax/artifact_file_selector.html
@@ -129,7 +129,7 @@
     $(".fileListParam").each( function(i, cont) {
       var files = [];
       $("#"+cont.id+"-list").children().each( function(j, c_cont) {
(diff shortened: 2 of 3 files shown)


qiita_pet/templates/study_ajax/artifact_file_selector.html
```diff
@@ -245,7 +245,7 @@
   </div>
   {% for ft, req, files in file_types %}
     <div class="col-md-3" id="{{ft}}-div">
-      <input type="hidden" name="{{ft}}" id="{{ft}}" value="" class="fileListParam">
+      <input type="hidden" name="{{ft}}[]" id="{{ft}}" value="" class="fileListParam">
```
ElDeveloper: What is the [] used for? Is it to make the name of the input element different?

qiita_pet/handlers/api_proxy/artifact.py
```diff
@@ -211,9 +211,12 @@ def artifact_post_req(user_id, filepaths, artifact_type, name,
             return {'status': 'error',
                     'message': "Can't create artifact, no files provided."}
 
-        # TODO: It might be good to have an API call that wraps to a lower
-        # level method that creates this job.
-        command = Command.get_validator(artifact_type)
+        # This try/except will catch the case when the plugins are not
+        # activated so there is no Validate for the given artifact_type
```
ElDeveloper: If you want to catch the missing validator I think using [hasattr](https://docs.python.org/3/library/functions.html#hasattr) would likely be better i.e.

```python
if not hasattr(Command, 'attribute'):
     return {'status': 'error', 'message': 'The plugin xxxx has not been initialized'}
```

antgonza: I don't think that's gonna work as Command.get_validator will raise an error:
```python
In [4]: Command.get_validator('bla')                                                                                                                                               
IndexError                                Traceback (most recent call last)
~/svn_programs/qiita/qiita_db/software.py in get_validator(cls, artifact_type)
    166             try:
--> 167                 res = qdb.sql_connection.TRN.execute_fetchlast()
    168             except IndexError:

~/svn_programs/qiita/qiita_db/sql_connection.py in wrapper(self, *args, **kwargs)
     49                 "invoked within the context manager.")
---> 50         return func(self, *args, **kwargs)
     51     return wrapper

~/svn_programs/qiita/qiita_db/sql_connection.py in execute_fetchlast(self)
    333         """
--> 334         return self.execute()[-1][0][0]
    335 

During handling of the above exception, another exception occurred:

QiitaDBError                              Traceback (most recent call last)
<ipython-input-4-73541800be9d> in <module>
----> 1 Command.get_validator('bla')

~/svn_programs/qiita/qiita_db/software.py in get_validator(cls, artifact_type)
    169                 raise qdb.exceptions.QiitaDBError(
    170                     "There is no command to generate the Validate for "
--> 171                     "artifact type '%s'" % artifact_type)
    172 
    173             return cls(res)

QiitaDBError: There is no command to generate the Validate for artifact type 'bla'
```

qiita_pet/templates/study_ajax/artifact_file_selector.html
antgonza: `[]` will make it an array vs as a string, and missing it is why files with spaces will fail cause JS tries to parse it incorrectly

qiita_pet/handlers/api_proxy/artifact.py
ElDeveloper: I see, thanks. Should we instead be catching the QiitaDBError instead? Also maybe explaining what could be a reasonable workaround? Alternatively if you think there's many more unexpected exceptions that might come from this block of code, then let's just leave as is.

qiita_pet/templates/study_ajax/artifact_file_selector.html
ElDeveloper: Thanks!

qiita_pet/handlers/api_proxy/artifact.py
antgonza: Good point! Replacing ...

ElDeveloper: 👍

PR #2873, opened 2019-04-17
fix #2867

### LINKED ISSUE #2867: qiita_db tests could use some comments for magic numbers
In qiita_db/test/test_meta_util.py and the like, there are SQL statements such as, "UPDATE qiita.artifact SET visibility_id=3". It would be nice to have comments for the visibility_ids, what they stand for, etc.

antgonza:


antgonza:
Forgot to add the screenshots:

- Before: current

- New: new

PR #2874, opened 2019-04-17
fix #2221

### LINKED ISSUE #2221: Fix bugs and labels on prep information template addition page
When adding a new 16S prep template, the most common thing that happens in Qiita, we want users to click on the option of 'Other' for  the Investigation type, and when they do, they click Amplicon Sequencing from a menu that contains an extra blank space and does not indicate 16S in any way. 16S amplicon sequencing should be an option from the first dropdown, and perhaps the subtype could be which V or something rather than the current options. Screenshots below.

image

image

antgonza:
After discussing #2221 with @adswafford we realized the only outstanding issue was that Other (only on prep creation) had 2 empty entries; the reason is cause we were forcing adding an empty one, while there was one already in the DB (we were duplicating).


antgonza:
Ping @adswafford

PR #2878, opened 2019-04-19
Fix 2774

antgonza:


antgonza:
Forgot screenshot: 
checksum_size

antgonza:
Some highlights of this PR:
- fixed #2774, which discovered a few things and the reason there are more fixes here
- retrieve_filepaths used to return a list now returns a dict; this change was cause originally I kept returning a list and realized that it was crazy to keep track of what element was what, see: https://github.com/biocore/qiita/pull/2833
- had to fix all tests so the previous changed worked
- some tests were using artifact 1 as the initial artifact for tests, which in some cases deleted or altered its files; decided to always create a new artifact for these cases so the files are not changed, so the checksum/size do not change
- purge_files_from_filesystem was one of these tests/methods and also found a bug on how the legacy files were being dealt with; which is something we saw the first time we ran it on main qiita
- compute_checksum had a bug which was only present in python2 (which is how all the checksums are currently calculated) so decided to rewrite it to be python2/3 safe and recalculate all the checksums during the patch

qiita_db/environment_manager.py
```diff
@@ -436,5 +436,6 @@ def patch(patches_dir=PATCHES_DIR, verbose=False, test=False):
         # for the test Study (1) so alot of the tests actually expect this.
```
charles-cowart: Outside the scope of this change, but if you can change 'alot' -> 'a lot', that would be awesome ty.

qiita_db/util.py
```diff
@@ -422,17 +420,17 @@ def compute_checksum(path):
...
     for fp in filepaths:
-        with open(fp, 'rb', newline=None) as f:
-            # Go line by line so we don't need to load the entire file
-            for line in f:
-                if crc is None:
-                    crc = crc32(line)
-                else:
-                    crc = crc32(line, crc)
-    # We need the & 0xffffffff in order to get the same numeric value across
+        with open(fp, 'rb') as f:
```
charles-cowart: Just to confirm, removing 'newline=None' from the open call was intended?

qiita_pet/handlers/artifact_handlers/base_handlers.py
```diff
@@ -8,6 +8,7 @@
 
 from os.path import basename, relpath
 from json import dumps
+from humanize import naturalsize
```
charles-cowart: Are you sure it's all right to use humanize? I looks like it last had a release just before 2015:

https://pypi.org/project/humanize/

qiita_db/environment_manager.py
antgonza: k

qiita_db/util.py
antgonza: Yup ... it's automatic.

qiita_pet/handlers/artifact_handlers/base_handlers.py
antgonza: Yup, we have been using it for a while; note that it has been part of the setup.py for a while ...

antgonza:
Thanks!!

charles-cowart:
no problem! Sorry for the delay

PR #2879, opened 2019-04-21
fix #2174

### LINKED ISSUE #2174: download artifact name
If the user put the effort into the processing graph to rename artifacts, those names should be used when downloading them as pre-filled name recommendations in the file selector window.

antgonza:


qiita_db/util.py
```diff
@@ -983,6 +985,36 @@ def filepath_id_to_rel_path(filepath_id):
...
+    -----
+    This helper function is intended to be used with the download handler so
+    we can prepend downloads with the artifact id; thus, we will only look for
+    filepath ids in qiita.analysis_filepath and qiita.artifact_filepath as
+    search in qiita.reference, qiita.prep_template_filepath and
+    qiita.sample_template_filepath will make the naming redundat (those already
+    have the study_id in their filename)
+    """
+    with qdb.sql_connection.TRN:
+        sql = """
```
sjanssen2: what if - for strange reasons - DB tables don't exist? Are those errors handled appropriately?

qiita_pet/handlers/download.py
```diff
@@ -185,6 +186,9 @@ def get(self, filepath_id):
             self.set_header('Content-Type', 'application/octet-stream')
             self.set_header('Content-Transfer-Encoding', 'binary')
             self.set_header('X-Accel-Redirect', '/protected/' + relpath)
+            aid = filepath_id_to_object_id(fid)
+            if aid is not None:
+                fname = '%d_%s' % (aid, fname)
```
sjanssen2: do we intend to split aid and fname somewhere else according to the `_` char? If so, should be allow further `_` in `fname`?

qiita_db/util.py
antgonza: Yup, handled in the TRN object and if that happens we'll have larger and scarier issues ... :)

qiita_pet/handlers/download.py
antgonza: No, we don't; this is just so users have the artifact id prepended in the name, [as you suggested](https://github.com/biocore/qiita/issues/2174#issuecomment-398388959) ...

antgonza:
Thanks!

PR #2881, opened 2019-04-26
adding info file restrictions

antgonza:
This adds the new set of restrictions we discussed. Note that these are the backend changes so we can build the GUI on top. As we are adding new restrictions, a lot of tests were affected so multiple changes.


qiita_pet/test/test_download.py
```diff
@@ -146,27 +146,27 @@ def test_download_study(self):
             '- 1256812 /protected/processed_data/1_study_1001_closed_'
             'reference_otu_table.biom processed_data/1_study_1001_closed_'
             'reference_otu_table.biom\n'
-            '- 36615 /protected/templates/1_prep_1_qiime_[0-9]*-'
+            '- [0-9]* /protected/templates/1_prep_1_qiime_[0-9]*-'
```
charles-cowart: [0-9]+ or \d+ might be better choices if you know that a number must always be there.
\d{5} if you know that the number must always be five digits.

PR #2883, opened 2019-04-29
fix #2433

### LINKED ISSUE #2433: Add upload metadata or prep info from computer to update information dropdowns
To improve usability, users should have the option to select a file from their computer to upload and update the information rather than having to go to the 'Upload Files' page to add them first and then go select from the dropdown.

antgonza:
Example: Update sample info file via the upload folder, then use the direct upload. Note that we are testing that the jobs are created and the errors retrieved correctly; thus, the error messages on the top.

### DIFF
--- a/qiita_pet/handlers/study_handlers/sample_template.py
+++ b/qiita_pet/handlers/study_handlers/sample_template.py
@@ -283,9 +283,9 @@ class SampleTemplateHandler(BaseHandler):
         study_id = int(self.get_argument('study_id'))
         filepath = self.get_argument('filepath')
         data_type = self.get_argument('data_type')
-        direct_upload = self.get_argument('data_type', False)
+        direct_upload = self.get_argument('direct_upload', False)
 
-        if direct_upload:
+        if direct_upload and direct_upload == 'true':
             direct_upload = True
             fd, filepath = mkstemp(suffix='.txt')
             close(fd)
--- a/qiita_pet/handlers/study_handlers/sample_template.py
+++ b/qiita_pet/handlers/study_handlers/sample_template.py
@@ -6,8 +6,10 @@
 # The full license is in the file LICENSE, distributed with this software.
 # -----------------------------------------------------------------------------
 
+from os import close
 from os.path import basename
 from json import loads, dumps
+from tempfile import mkstemp
 
 from tornado.web import authenticated, HTTPError
 
@@ -65,7 +67,7 @@ def sample_template_checks(study_id, user, check_exists=False):
 
 
 def sample_template_handler_post_request(study_id, user, filepath,
-                                         data_type=None):
+                                         data_type=None, direct_upload=False):
     """Creates a new sample template
 
     Parameters
@@ -79,6 +81,9 @@ def sample_template_handler_post_request(study_id, user, filepath,
     data_type: str, optional
         If filepath is a QIIME mapping file, the data type of the prep
         information file
+    direct_upload: boolean, optional
+        If filepath is a direct upload; if False we need to process the
+        filepath as part of the study upload folder
 
     Returns
     -------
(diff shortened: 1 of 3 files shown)


qiita_pet/handlers/study_handlers/sample_template.py
```diff
@@ -277,9 +283,19 @@ def post(self):
         study_id = int(self.get_argument('study_id'))
         filepath = self.get_argument('filepath')
         data_type = self.get_argument('data_type')
+        direct_upload = self.get_argument('data_type', False)
```
ElDeveloper: Is this correct? This line is querying for `data_type` but is assigning a boolean value *and* the previous line is also querying for `data_type`.

qiita_pet/handlers/study_handlers/sample_template.py
```diff
@@ -277,9 +283,19 @@ def post(self):
...
+
+        if direct_upload:
+            direct_upload = True
```
ElDeveloper: Is `directd_upload` a boolean value? This seems odd since you check this value and then assign it to True. Would you mind explaining what's going on here?

qiita_pet/static/js/sampleTemplateVue.js
```diff
@@ -160,17 +160,45 @@ Vue.component('sample-template-page', {
...
-        },
-        error: function (object, status, error_msg) {
-          bootstrapAlert("Error updating sample template: " + error_msg, "danger")
-        }
-      });
+      var file = $('#st-direct-upload')[0].files[0];
+
+      // We can have 2 scenarios: direct file upload or path from upload
+      // folder. The former is sent via PATCH using defauls, the latter
+      // via POST using "processData: false, contentType: false"
```
ElDeveloper: Why would the contentType be a boolean value? It seems non-standard/unexpected.

qiita_pet/static/js/sampleTemplateVue.js
```diff
@@ -435,6 +463,26 @@ Vue.component('sample-template-page', {
         // Add the select to update the sample information
         $row = $('<div>').attr('id', 'update-st-div').addClass('row form-group').appendTo($tab);
         $('<label>').addClass('col-sm-2 col-form-label').append('Update sample information:').appendTo($row);
+
+        // Add the direct upload field
+        $('<label>')
+          .addClass('btn btn-default')
+            .append('Direct upload file <small>(< 2M)</small>')
```
ElDeveloper: 2M -> 2MB

qiita_pet/static/js/sampleTemplateVue.js
```diff
@@ -435,6 +463,26 @@ Vue.component('sample-template-page', {
...
+            .appendTo($row)
+            .append('<input type="file" style="display: none;" id="st-direct-upload">');
+        $('#st-direct-upload').on('change', function() {
+          if (this.files.length != 1) {
+            alert('You can only upload one file.')
+            return false;
+          }
+          if (this.files[0].size > 2097152) {
+            alert('You can only upload files smaller than 2M; please use the regular uploader.');
```
ElDeveloper: 2M -> 2MB. A newcomer might wonder what is the regular uploader, what about linking to it?

qiita_pet/handlers/study_handlers/sample_template.py
antgonza: direct_upload is a b'true' so just making sure it's a boolean; changing for clarity.

qiita_pet/static/js/sampleTemplateVue.js
antgonza: good question and to be honest not sure, from: https://stackoverflow.com/a/5976031; adding the ref to the code

qiita_pet/static/js/sampleTemplateVue.js
antgonza: good point, note that afaik a basic alert doesn't have clickable section; anyway, re-writing to make clearer.

qiita_pet/static/js/sampleTemplateVue.js
```diff
@@ -435,6 +464,26 @@ Vue.component('sample-template-page', {
...
+            .append('Direct upload file <small>(< 2MB)</small>')
+            .appendTo($row)
+            .append('<input type="file" style="display: none;" id="st-direct-upload">');
+        $('#st-direct-upload').on('change', function() {
+          if (this.files.length != 1) {
+            alert('You can only upload one file.')
+            return false;
+          }
+          if (this.files[0].size > 2097152) {
+            alert('You can only upload files smaller than 2MB; please use "Upload Files" button on the left.');
```
ElDeveloper: What about something like this:

```suggestion
            alert('You can only upload files smaller than 2MB. For larger files please use the "Upload Files" button on the left.');
```

antgonza:
@ElDeveloper, let me know if there is anything missing here; thanks!

PR #2889, opened 2019-05-03
adding patch 74 helper

antgonza:


qiita_db/support_files/patches/python_patches/74.py
```diff
@@ -9,18 +10,30 @@
...
+                     names=['filepath_id', 'checksum', 'fp_size'])
+    cache = df.to_dict('index')
+
+for fid in fids:
+    if fid not in cache:
+        finfo = get_filepath_information(fid)
+        try:
+            size = getsize(finfo['fullpath'])
+        except FileNotFoundError:
+            size = 0
```
charles-cowart: Technically you could merge the innards of both tries.

Issue #410, opened 2014-10-15
Lookup table for *_status and user level


wasade:
This field currently takes a varchar with the unenforced vocabulary: `('not submitted', 'submitting', 'success', 'failed')`

antgonza:
Even so that I can see why it will be nice to have a lookup table, I believe the current way is nice as it allows us to store the failure in the text so we can display to the user. However, I don't have strong feelings about it as long as we can do the same with the new implementation. 

What do you guys think?

josenavas:
In order to have a lookup table for this we need a normalized way of storing the errors across the system (i.e. using the Log table). Probably will be easier to do after the artifact changes. My vote goes to move this to alpha 0.4

antgonza:
This is basically a duplicate of https://github.com/biocore/qiita/issues/292 but here he have more info. Additionally, note that we would need to do the same for **the user level status**

antgonza:
AFAIK, this can be closed.

PR #2892, opened 2019-05-07
fixing some issues

antgonza:


qiita_db/util.py
```diff
@@ -1486,17 +1488,16 @@ def generate_study_list(user, visibility):
 
                 # cleaning aids_with_deprecation
                 info['artifact_biom_ids'] = []
-                info['preparation_data_types'] = []
                 if info['aids_with_deprecation'] is not None:
                     for x in info['aids_with_deprecation']:
-                        # f1-2 are the default names given by pgsql
+                        # f1-3 are the default names given by pgsql
```
charles-cowart: Just to confirm, f1-3 I assume means (f1, f2, f3). You've increased the range by 1, but the conditional below is still testing only for not x['f2']. This is still okay, right?

qiita_pet/handlers/study_handlers/artifact.py
```diff
@@ -51,9 +51,16 @@ def post(self):
...
         # Request the rest of the arguments, which will be the files
-        files = {arg: self.get_argument(arg) for arg in self.request.arguments
-                 if arg not in ['name', 'prep-template-id', 'artifact-type',
-                                'import-artifact']}
+        files = dict()
+        for arg in self.request.arguments:
+            if arg not in ['name', 'prep-template-id', 'artifact-type',
+                           'import-artifact']:
+                arg_name = arg
+                # removing ending [], in case they exists, necessary for JS
```
charles-cowart: I believe exists -> exist

qiita_db/artifact.py
```diff
@@ -740,7 +740,7 @@ def visibility(self, value):
 
             # then let's check that the sample/prep info files have the correct
             # restrictions
-            if value != 'sandboxed' and study is not None:
+            if value != 'sandbox' and study is not None:
```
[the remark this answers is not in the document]
antgonza: Yup, there is an open issue: https://github.com/biocore/qiita/issues/410

qiita_db/util.py
antgonza: Yeah, good catch. Originally we had 3 but the comment only mentioned 2, then changed the code to only use 2 and forgot to bring the comment back ... changing ...

Issue #2133, opened 2017-05-16
Provide .md5 checksum of files of dynamic nginx zip files


wasade:
A user requested this over the help email account so that they could verify the integrity of their large file download.

antgonza:
This is an interesting question because for biom/fna, raw files, etc. is an on the fly download ... so not sure how this will happen ... any ideas?

wasade:
We compute once and store on the file system in proximity to the files being downloaded? On the fly is quite expensive

antgonza:
It's quite cheap to generate as nginx takes care of everything. Let's discuss during a future Qiita dev-meeting.

antgonza:
Let's check if nginx has something for this, cause if not ... 😢

antgonza:
Just realized that this is a duplicate of https://github.com/biocore/qiita/issues/2920; closing this one as the other one has an explanation of how it currently works.

PR #2896, opened 2019-05-12
adding text about downloads

antgonza:


antgonza:
BTW the new link looks like this:
new

qiita_pet/support_files/doc/source/faq.rst
```diff
@@ -90,10 +90,26 @@ A few more instructions: for the example above the workflow should be:
...
-How to solve unzip errors?
---------------------------
-
-When downloading large zip files within Qiita there is a change that you will get
+How to solve download or unzip errors?
+--------------------------------------
+
+Dealing with large files might be daunting but, in general, following these
+instructions should make things easier. First, make sure that you have enough space
+for the zip download file; for the American Gut, you should have at least close to 700Gb
```
wasade: Would it be possible to, near the download link, just show the size of the zip file to download? The size estimate for the AGP will be incorrect pretty soon.

qiita_pet/support_files/doc/source/faq.rst
```diff
@@ -90,10 +90,26 @@ A few more instructions: for the example above the workflow should be:
...
+but having close to a Tera (1000Gb) should be better as you will need to uncompress
+the zip file. Second, make sure that your computer has all the sleep settings turned off;
+for example, in a Mac, got to System Preferences, Energy Saver, Power Adapter and unselect
+the option of "Put hard disks to sleep when possible"; don't forget to save the settings.
+Third, download the file but point to the storage that you want to save your file in; using
+Chrome, right click on the download button and select "Save Link As ..."; and select the
+location where you have enough space (see point 1). Fourth, wait for the download to finish,
+around 13hrs in an OK USA Internet connection for the AGP but this will depend on your Internet
```
wasade: That's roughly 14MB/s which is a pretty nice connection, and many ISPs have total download limits per month. I'd omit this sentence... the browser should provide a download time estimate anyway?

qiita_pet/support_files/doc/source/faq.rst
```diff
@@ -90,10 +90,26 @@ A few more instructions: for the example above the workflow should be:
...
+service. Finally, unzip de file with a newer version of zip.
```
wasade: de -> the

What is a newer version of zip?

wasade: Oh, I see below. Might want to mention "(see below)". Note that this is likely limited to versions of OSX. unzip 6.0.0 on 10.13 reports a compilation date of 2017, and the unzip 6.0.0 build on barnacle reports a compilation date of 2010

qiita_pet/support_files/doc/source/faq.rst
antgonza: Just noticed that the full uncompressed + the zip is close to 2T; will change. BTW estimating the size of the zip is actually pretty hard to do accurately for the merged zip files but we could do ballpark estimates; let's discuss in the Wednesday qiita meeting. BTW there is an issue about this: https://github.com/biocore/qiita/issues/2133

qiita_pet/support_files/doc/source/faq.rst
antgonza: Good point, will remove as arguably 14M is "basic" based https://www.speedtest.net/reports/ at least on the countries listed there ... but as always depends compared to what ...

qiita_pet/support_files/doc/source/faq.rst
antgonza: ha! de! thanks for the info ...

qiita_pet/support_files/doc/source/faq.rst
```diff
@@ -90,10 +90,26 @@ A few more instructions: for the example above the workflow should be:
...
-How to solve unzip errors?
---------------------------
-
-When downloading large zip files within Qiita there is a change that you will get
+How to solve download or unzip errors?
+--------------------------------------
+
+Dealing with large files might be daunting but, in general, following these
+instructions should make things easier. First, make sure that you have enough space
+for the zip download file; for the American Gut, you should have at least close to 900Gb
```
wasade: This statement will be inaccurate soon. I recommend not listing a size here. 

I did a test against study 2136, and Chrome immediately saw the download size, does the size info just get cached or something?

antgonza: Got it! I thought you meant that we should display beside the button the size, which is the difficult part; if having it in Chrome is sufficient, that's great! Note that the way it works is that python will create the list of files; nginx will take that info, create the zip (and the size reported in Chrome); and Chrome will do the rest. 

Anyway, what about replacing for: if you are unsure of the size required click on the button and your browser will show an estimate size of the download.

PR #2915, opened 2019-07-03
admins will get an email for non analysis jobs changing status to running/success/error

antgonza:
Also fixing how we are reading the job log files as we now have extra fields ...


jdereus:
What is to be gained here?  Are these user specific emails?  Every single non-analysis job going through qiita? This seems like a lot of email traffic for not much benefit.  Please feel free to change my mind.

PR #2917, opened 2019-07-09
download_public -> public_download

antgonza:
After testing this new functionality; realized that:
1. public_download is a better name than download_public
2. 405 will not show the reason in the browser which will be misleading; thus changing to 200

### DIFF
--- a/qiita_pet/handlers/base_handlers.py
+++ b/qiita_pet/handlers/base_handlers.py
@@ -45,9 +45,6 @@ class BaseHandler(RequestHandler):
                 # Any issue with this check leaves default as not admin
                 pass
 
-        # render error page
-        self.render('error.html', status_code=status_code, is_admin=is_admin)
-
         # log the error
         from traceback import format_exception
         exc_info = kwargs["exc_info"]
@@ -59,7 +56,12 @@ class BaseHandler(RequestHandler):
         request_info = ''.join(["<strong>%s</strong>: %s\n" %
                                (k, req_dict[k]) for k in
                                 req_dict.keys() if k != 'files'])
-        error = exc_info[1]
+        error = str(exc_info[1]).split(':', 1)[1]
+
+        # render error page
+        self.render('error.html', status_code=status_code, is_admin=is_admin,
+                    error=error)
+
         LogEntry.create(
             'Runtime',
             'ERROR:\n%s\nTRACE:\n%s\nHTTP INFO:\n%s\n' %
--- a/qiita_pet/test/test_download.py
+++ b/qiita_pet/test/test_download.py
@@ -387,31 +387,31 @@ class TestDownloadPublicHandler(TestHandlerBase):
 
     def test_download(self):
         # check failures
-        response = self.get('/download_public/')
-        self.assertEqual(response.code, 405)
-        self.assertEqual(response.reason, 'You need to specify both data (the '
-                         'data type you want to download - raw/biom) and '
-                         'study_id')
+        response = self.get('/public_download/')
+        self.assertEqual(response.code, 200)
+        self.assertEqual(response.body.decode('ascii'), 'You need to specify '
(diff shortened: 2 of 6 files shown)


qiita_pet/test/test_download.py
```diff
@@ -387,31 +387,31 @@ def tearDown(self):
 
     def test_download(self):
         # check failures
-        response = self.get('/download_public/')
-        self.assertEqual(response.code, 405)
-        self.assertEqual(response.reason, 'You need to specify both data (the '
-                         'data type you want to download - raw/biom) and '
-                         'study_id')
+        response = self.get('/public_download/')
+        self.assertEqual(response.code, 200)
```
charles-cowart: @antgonza I'm not sure about this change. Are these meant to be REST endpoints? If so, then they really do need to return the proper status code, as well as the appropriate message string.

antgonza: I'm actually not sure; my concern is that when a user goes to: https://qiita-rc.ucsd.edu/download_public/ in their browser this is what they see:
Screen Shot 2019-07-09 at 12 07 49 PM but if they use wget:
```
$ wget --no-check-certificate "https://qiita-rc.ucsd.edu/download_public/"
--2019-07-09 12:08:52--  https://qiita-rc.ucsd.edu/download_public/
Resolving qiita-rc.ucsd.edu (qiita-rc.ucsd.edu)... 132.249.211.60
Connecting to qiita-rc.ucsd.edu (qiita-rc.ucsd.edu)|132.249.211.60|:443... connected.
WARNING: cannot verify qiita-rc.ucsd.edu's certificate, issued by ‘emailAddress=jdereus@ucsd.edu,CN=qiita-rc.ucsd.edu,O=UCSD,L=La Jollaq,ST=California,C=US’:
  Self-signed certificate encountered.
HTTP request sent, awaiting response... 405 You need to specify both data (the data type you want to download - raw/biom) and study_id
2019-07-09 12:08:52 ERROR 405: You need to specify both data (the data type you want to download - raw/biom) and study_id.
```

Thus, thought that changing to 200 so the user can see the error as this will minimize enquiries.

antgonza:
Now, after the latest changes, the error in the website are:

Screen Shot 2019-07-09 at 12 54 41 PM

Screen Shot 2019-07-09 at 12 54 35 PM

and wget:
```
--2019-07-09 13:01:52--  https://qiita-rc.ucsd.edu/download_public/
Resolving qiita-rc.ucsd.edu (qiita-rc.ucsd.edu)... 132.249.211.60
Connecting to qiita-rc.ucsd.edu (qiita-rc.ucsd.edu)|132.249.211.60|:443... connected.
WARNING: cannot verify qiita-rc.ucsd.edu's certificate, issued by ‘emailAddress=jdereus@ucsd.edu,CN=qiita-rc.ucsd.edu,O=UCSD,L=La Jollaq,ST=California,C=US’:
  Self-signed certificate encountered.
HTTP request sent, awaiting response... 405 You need to specify both data (the data type you want to download - raw/biom) and study_id
2019-07-09 13:01:52 ERROR 405: You need to specify both data (the data type you want to download - raw/biom) and study_id.
```

qiita_pet/handlers/base_handlers.py
```diff
@@ -59,7 +56,12 @@ def write_error(self, status_code, **kwargs):
         request_info = ''.join(["<strong>%s</strong>: %s\n" %
                                (k, req_dict[k]) for k in
                                 req_dict.keys() if k != 'files'])
-        error = exc_info[1]
+        error = str(exc_info[1]).split(':', 1)[1]
```
charles-cowart: Just to confirm, these array elements, the split on ':', etc. will work in all cases?

antgonza: All cases I have tested ... cause the error gets prepended: `Error X: bla`.

PR #2928, opened 2019-07-25
BUG: Count samples by artifact visibility

wasade:
A study can have multiple visibilities, and samples are counted per visibility per study. As a result, samples in studies that have multiple visibilities were counted multiple times.

I did not create a regression test as I don't _believe_ the test data have a study where there is a prep that has multiple visibilities. If there is one, I'd be happy to use if for a regression test. If there isn't one, and if a regression test is important here, I may need some guidance on how best to create a mock study with artifacts spanning visibilities.

### DIFF
--- a/qiita_db/meta_util.py
+++ b/qiita_db/meta_util.py
@@ -184,12 +184,16 @@ def update_redis_stats():
     number_of_samples = {}
     ebi_samples_prep = {}
     num_samples_ebi = 0
+
     for k, sts in viewitems(studies):
         number_of_samples[k] = 0
         for s in sts:
             st = s.sample_template
             if st is not None:
-                number_of_samples[k] += len(list(st.keys()))
+                for p in s.prep_templates():
+                    if p.artifact is not None and p.artifact.visibility == k:
+                        n = len(list(p.keys()))
+                        number_of_samples[k] += n
 
             ebi_samples_prep_count = 0
             for pt in s.prep_templates():


qiita_db/meta_util.py
```diff
@@ -184,12 +184,16 @@ def update_redis_stats():
     number_of_samples = {}
```
antgonza: Adding here cause this is the highest (smallest line number) github will allow me to add a message; studies is missing the status: 'awaiting_approval'; all the statuses are in 'qiita.visibility'. We could ignore, as we have been doing until now, add via qiita.visibility or just add to the dict in line 179.

qiita_db/meta_util.py
```diff
@@ -184,12 +184,16 @@ def update_redis_stats():
...
     ebi_samples_prep = {}
     num_samples_ebi = 0
+
     for k, sts in viewitems(studies):
         number_of_samples[k] = 0
         for s in sts:
             st = s.sample_template
             if st is not None:
-                number_of_samples[k] += len(list(st.keys()))
+                for p in s.prep_templates():
```
antgonza: Note that this will ignore all studies/samples without preps/artifacts; which I think it's fine but just FYI.

wasade: Correct, but I also think it's fine. I wasn't sure how to handle the case where there isn't an artifact because visibility is denoted at the artifact, not the template level, right?

qiita_db/meta_util.py
wasade: Shouldn't an artifact retain its prior state until approved?

wasade:
The failure is real and represents a possible logical flaw. The test study has two preps with the same samples represented twice so the count is 54 not 27. This is a valid scenario for resequenced samples, or samples with multiple preparations.

Are the sample counts on the stats page supposed to be the sum of the number of physical specimens or the sum of the number of unique preparations of the specimens?

rob-knight:
I think they should be physical specimens rather than counting different amplicons, shotgun preps etc multiple times.

On Jul 26, 2019, at 12:18 AM, Daniel McDonald <notifications@github.com<mailto:notifications@github.com>> wrote:


The failure is real and represents a possible logical flaw. The test study has two preps with the same samples represented twice so the count is 54 not 27. This is a valid scenario for resequenced samples, or samples with multiple preparations.

Are the sample counts on the stats page supposed to be the sum of the number of physical specimens or the sum of the number of unique preparations of the specimens?

qiita_db/meta_util.py
antgonza: Well, the artifact state is awaiting_approval and I _think_ to accomplish what you suggest we will need another table where we have all sandbox artifacts that are awaiting for approval. Not sure if worth the dev time to accomplish this. As a reminder, when an artifact is created its status is sandbox, then awaiting_approval, then private and finally public.

qiita_db/meta_util.py
antgonza: Good question!! Historically, status was given per study, then we moved to do it by artifact, then we realized that admins will need to click on each artifact to make all artifacts private/public so we decided to do this per prep. However, the status is kept by artifact and just propagated to all artifacts in a given preparation.

qiita_db/meta_util.py
```diff
@@ -184,12 +184,16 @@ def update_redis_stats():
...
+                    if p.artifact is not None and p.artifact.visibility == k:
+                        n = len(list(p.keys()))
```
antgonza: I _think_ the easiest to accomplish the goal of having the most accurate count is to count each status per study via this prep loop and then just add those counts to the total count ... something like this:
```python
if st is not None:
    prep_totals = {'public': 0, 'sandbox': 0, 'private': 0, 'awaiting_approval': 0}
    for p in s.prep_templates():
           if p.artifact is not None and p.artifact is not None:
                prep_totals[p.artifact.visibility] += len(list(p.keys()))
   ... 
```

wasade: Unfortunately this would still over count when a sample is on multiple preps

wasade:
Thanks, @rob-knight and @antgonza. 

If counting primary specimen, which I agree is ideal, then I believe we have an issue with conflicting visibilities. Since state is controlled at the artifact level, it is possible for a physical specimen to be associated to a public artifact and a sandbox one. In that situation, we cannot readily determine whether to count the sample as public or sandbox. One possible solution is establishing an order of precedence over visibility for these counts.

This is not a hypotheical situation as I believe this corresponds to @adswafford's situation discussed in our email thread.

qiita_db/meta_util.py
wasade: Does awaiting apptoval only ever mean sandbox -> private?

antgonza: Yup, an artifact to become private needs to be approved ...

rob-knight:
Precedence is we count the sample as the most “public” version of it that’s available ie “submitted to ebi” > “public” > “private” > “sandbox”

I wonder if this is a good time to consider implementing BioSample?

On Jul 26, 2019, at 10:38 AM, Daniel McDonald <notifications@github.com<mailto:notifications@github.com>> wrote:


Thanks, @rob-knight<https://github.com/rob-knight> and @antgonza<https://github.com/antgonza>.

If counting primary specimen, which I agree is ideal, then I believe we have an issue with conflicting visibilities. Since state is controlled at the artifact level, it is possible for a physical specimen to be associated to a public artifact and a sandbox one. In that situation, we cannot readily determine whether to count the sample as public or sandbox. One possible solution is establishing an order of precedence over visibility for these counts.

This is not a hypotheical situation as I believe this corresponds to @adswafford<https://github.com/adswafford>'s situation discussed in our email thread.

PR #2937, opened 2019-08-07
Update glossary

JTFouquier:
Fix scientific definitions for clarity.


JTFouquier:
I will be out of town for a few weeks, so if there is an issue with the PR, feel free to copy the file and push it via another route. Thanks! :)

PR #2939, opened 2019-08-08
Switch to Globus Auth authentication

lukaszlacinski:
The pull request replaces local authentication with the Globus Auth IdP. It still uses the same User model which contains username, name, password, affiliation, address, phone but the information is obtained from Globus Auth. Globus Auth provides a sub (OIDC subject which is a unique UUID in this case), preferred_username, name, organization, email. The following Globus information is mapped to the Qiita User fields:
preferred_username -> username,
name -> name,
organization -> affiliation.
The GlobusOAuth2LoginHandler also exchanges an OAuth2 authorization code to access/refresh tokens that can be used to submit file transfers, etc.

### DIFF
--- a/qiita_pet/handlers/auth_handlers.py
+++ b/qiita_pet/handlers/auth_handlers.py
@@ -6,6 +6,7 @@
 # The full license is in the file LICENSE, distributed with this software.
 # -----------------------------------------------------------------------------
 
+import secrets
 from tornado.escape import url_escape, json_encode
 
 from qiita_pet.handlers.base_handlers import BaseHandler
@@ -214,12 +215,10 @@ class GlobusOAuth2LoginHandler(BaseHandler, GlobusOAuth2Mixin):
 
             # Check if the user exists. If it does not, create the user
             username = user_info.get("preferred_username")
-            password = 'RviwFvie83!#'
+            password = secrets.token_urlsafe(16)
             info = {
-                # "sub": user_info.get("sub"),
                 "name": user_info.get("name"),
                 "affiliation": user_info.get("organization")
-                # "email": user_info.get("email")
             }
             try:
                 User.create(username, password, info)
--- a/qiita_pet/handlers/auth_handlers.py
+++ b/qiita_pet/handlers/auth_handlers.py
@@ -200,10 +200,7 @@ class GlobusOAuth2LoginHandler(BaseHandler, GlobusOAuth2Mixin):
         if self.get_argument("code", False):
             nextpage = self.get_argument("next", None)
             if nextpage is None:
-                if "auth/" not in self.request.headers["Referer"]:
-                    nextpage = self.request.headers["Referer"]
-                else:
-                    nextpage = "%s/" % qiita_config.portal.portal_dir
+                nextpage = "%s/" % qiita_config.portal_dir
             # Exchange the code to access/refresh tokens
             tokens = await self.get_tokens(
                 key=qiita_config.globus_client_key,
--- a/qiita_pet/handlers/auth_handlers.py
+++ b/qiita_pet/handlers/auth_handlers.py
@@ -9,6 +9,7 @@
 from tornado.escape import url_escape, json_encode
 
 from qiita_pet.handlers.base_handlers import BaseHandler
+from qiita_pet.handlers.globus import GlobusOAuth2Mixin
 from qiita_core.qiita_settings import qiita_config, r_client
 from qiita_core.util import execute_as_transaction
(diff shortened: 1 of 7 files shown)


qiita_pet/handlers/auth_handlers.py
```diff
@@ -175,3 +176,61 @@ class AuthLogoutHandler(BaseHandler):
...
+            tokens = await self.get_tokens(
+                key=qiita_config.globus_client_key,
+                secret=qiita_config.globus_client_secret,
+                redirect_uri=qiita_config.globus_redirect_uri,
+                code=self.get_argument("code"))
+            # Get userinfo from Globus Auth
+            user_info = await self.get_user_info(tokens["access_token"])
+
+            # Check if the user exists. If it does not, create the user
+            username = user_info.get("preferred_username")
```
antgonza: Are you sure is fine to leave this info here? Also, when we create a new user, will this be in the Qiita system or Globus? I think it should be in the latter, right?

qiita_pet/handlers/auth_handlers.py
```diff
@@ -175,3 +176,61 @@ class AuthLogoutHandler(BaseHandler):
...
+            password = 'RviwFvie83!#'
+            info = {
+                #"sub": user_info.get("sub"),
```
antgonza: If this comments are not needed, could you remove them? Thank you!

qiita_pet/handlers/auth_handlers.py
```diff
@@ -175,3 +176,61 @@ class AuthLogoutHandler(BaseHandler):
...
+                "name": user_info.get("name"),
+                "affiliation": user_info.get("organization")
+                #"email": user_info.get("email")
+            }
+            try:
```
antgonza: Just to confirm, the idea is that we catch this error so we don't need to check if the user already exists in the DB, correct?

qiita_pet/handlers/auth_handlers.py
```diff
@@ -175,3 +176,61 @@ class AuthLogoutHandler(BaseHandler):
...
+                created = User.create(username, password, info)
+            except QiitaDBDuplicateError:
+                pass
+            # Set current  user
+            self.set_current_user(username)
+
+            # Update tokens
```
antgonza: If these comments are not needed, could you remove them? Thanks.

qiita_pet/handlers/auth_handlers.py
lukaszlacinski: I removed the comment. Users do not create Qiita accounts. A new local Qiita account is created automatically, when a user logs in to Qiita via Globus Auth for the first time. A local Qiita account is a tornado/Qiita version of a copy of a Globus account.

qiita_pet/handlers/auth_handlers.py
lukaszlacinski: Yes. The try-except is in case the user already exists in the DB.

qiita_pet/handlers/auth_handlers.py
lukaszlacinski: Done.

qiita_pet/handlers/auth_handlers.py
lukaszlacinski: Done.

lukaszlacinski:
> * Once you login, you end in the page below; not sure why but if I go back to the main page I can see that I'm logged in: Screen Shot 2019-08-08 at 2 26 18 PM

It has bee fixed.

> * The login doesn't show my UCSD email but the globus ID (I think), is that expected? Screen Shot 2019-08-08 at 2 26 31 PM

Qiita uses email addresses as usernames. Globus does not. In effect, you see your Globus username, you used to log into Globus as a Qiita username, instead of your email address you were asked to provide when you were creating your Globus account.

ElDeveloper:
@lukaszlacinski apologies for the radio silence, but we were wondering if instead of replacing the current authentication system with Globus' we could allow users to link their existing Qiita accounts with a Globus account. This way existing users can augment their existing capabilities. In addition users who don't need to use Globus wouldn't have to create a Globus account.

rpwagner:
@ElDeveloper this may be possible. Ideally we'd be able to set it up so new users use Globus, but existing users can do the linking. Do you want to set up a call and have a short chat about how it would work? I'm back at UCSD so I can only ask Lukasz or others in Globus if they have the time or capacity to work on this. If we can define the use case that makes the request clearer.

ElDeveloper:
Very exciting, sure let's discuss this in more detail.

PR #2944, opened 2019-09-03
minor fixes to qiita-rc

antgonza:


antgonza:
These are minor fixes that were tested in qiita-rc; thus merging.

PR #2945, opened 2019-09-04
improve public page

antgonza:
Forgot to add a screenshot of the download button:
Screen Shot 2019-09-04 at 8 34 59 AM

### DIFF
--- a/qiita_pet/handlers/download.py
+++ b/qiita_pet/handlers/download.py
@@ -365,7 +365,7 @@ class DownloadPublicHandler(BaseHandlerDownload):
             else:
                 public_raw_download = study.public_raw_download
                 if study.status != 'public':
-                    raise HTTPError(422, reason='Study is not public. If this '
+                    raise HTTPError(404, reason='Study is not public. If this '
                                     'is a mistake contact: '
                                     'qiita.help@gmail.com')
                 elif data == 'raw' and not public_raw_download:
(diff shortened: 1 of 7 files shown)


qiita_pet/handlers/download.py
```diff
@@ -393,3 +394,38 @@ def get(self):
...
+        if artifact_id is None:
+            raise HTTPError(422, reason='You need to specify an artifact id')
+        else:
+            try:
+                artifact = Artifact(artifact_id)
+            except QiitaDBUnknownIDError:
+                raise HTTPError(422, reason='Artifact does not exist')
+            else:
+                if artifact.visibility != 'public':
+                    raise HTTPError(422, reason='Artifact is not public. If '
```
wasade: Should this be a 404 to avoid leaking information about non public data?

antgonza: I don't have a strong preference but thought that after the last 2 releases where we are making things more public it will make more sense to add these kind of errors/warnings so if another user gets it and it is a mistake it can reported to us. A good example is that a user publishes one of these ids but forgets to make public.

wasade: But they would report the issue with a 404 too right? We [respond](https://github.com/biocore/qiita/blob/master/qiita_pet/handlers/study_handlers/base.py#L30) with a 404 when a user attempts to access a study description that they don't have access too.

PR #2956, opened 2019-10-28
fix #2948

### LINKED ISSUE #2948: f.readline() error: utf-8' codec can't decode byte
In the last few days we have had 2 file infos that have failed with the error below, in both cases the problem was not UTF-8 chars characters. The issue is that the system doesn't report this but just fails.

```
[E 190917 13:30:47 web:1670] Uncaught exception POST /study/description/sample_template/ (127.0.0.1)
    HTTPServerRequest(protocol='https', host='qiita.ucsd.edu', method='POST', uri='/study/description/sample_template/', version='HTTP/1.1', remote_ip='127.0.0.1')
      File "/home/qiita/miniconda3/envs/qiita/lib/python3.6/site-packages/tornado/web.py", line 1590, in _execute
      File "/home/qiita/miniconda3/envs/qiita/lib/python3.6/site-packages/tornado/web.py", line 3006, in wrapper
      File "/home/qiita/qiita/qiita_pet/handlers/study_handlers/sample_template.py", line 298, in post
        direct_upload=direct_upload))
      File "/home/qiita/qiita/qiita_pet/handlers/study_handlers/sample_template.py", line 108, in sample_template_handler_post_request
        is_mapping_file = looks_like_qiime_mapping_file(filepath)
      File "/home/qiita/qiita/qiita_db/metadata_template/util.py", line 255, in looks_like_qiime_mapping_file
        first_line = f.readline()
      File "/home/qiita/miniconda3/envs/qiita/lib/python3.6/codecs.py", line 321, in decode
        (result, consumed) = self._buffer_decode(data, self.errors, final)
    UnicodeDecodeError: 'utf-8' codec can't decode byte 0xa1 in position 1383: invalid start byte
[D 190917 13:30:47 cmd:722] Popen(['git', 'cat-file', '--batch-check'], cwd=/home/qiita/qiita, universal_newlines=False, shell=None)
```

antgonza:


qiita_db/metadata_template/test/test_util.py
```diff
@@ -336,6 +343,15 @@ def test_get_qiime2_reserved_words(self):
     "Fasting_mouse_I.D._607\n"
 )
 
+QIIME_TUTORIAL_MAP_SUBSET_UNICODE = (
```
wasade: The variable suggests this is unicode but the this is a byte string?

qiita_db/metadata_template/util.py
```diff
@@ -251,13 +251,13 @@ def looks_like_qiime_mapping_file(fp):
...
     first_line = None
-    with qdb.util.open_file(fp, newline=None) as f:
+    with qdb.util.open_file(fp, newline=None, errors='replace') as f:
         first_line = f.readline()
     if not first_line:
         return False
 
     first_col = first_line.split()[0]
-    return first_col == '#SampleID'
+    return first_col in ('#SampleID', b'#SampleID')
```
wasade: why not either always open in raw, or cast the value on line 255 to bytes? Technically, I think this is a qiime1 mapping file too, right?

antgonza: Good point, will try to address this. Yup, qiime1 mapping file but the issue comes cause the users are in Windows with some weird Excel configuration that added some unicode that wasn't parsable.

antgonza: My issue was on creating and then reading the file. Decided to simplify things by just having the file in the repo and reading it as it's done in other tests ...

PR #2958, opened 2019-10-28
fix #2930

### LINKED ISSUE #2930: Convert ` and similar characters to ' on search page.
It appears some users are attempting to query, and using ` or similar characters to encapsulate strings, etc., instead of '. This will cause a warning, etc.

Anticipating some or all of these instances should be ', we can convert them in the javascript as they are typed.
unnamed

antgonza:


antgonza:
Forgot to add example. Note how we copy paste a ` and it's changed to a ' 
redbiom_quote

PR #2962, opened 2019-12-04
Non public artifact download

dhakim87:
This branch allows creation of shareable download links for private artifacts:

A logged in user with permission to download a private artifact posts to /private_download/generate/artifact/
This saves a jwt to the database and responds with a url

That user, or anyone with the generated url can then download the artifact.  Revocation is handled through expiration of the token or removal of access permissions of the user signed in the jwt.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,15 @@
 # Qiita changelog
 
+Version 112019
+--------------
+
+* Added PacBio_SMRT to the list of platform and PacBio RS, PacBio RS II, Sequel, Sequel II as valid instrument models
+* Improved downloads for public data (BIOM, raw data, etc.; see https://qiita.ucsd.edu/static/doc/html/downloading.html)
+* Added the possibility of just downloading sample or preparation information files
+* During a meta-analysis if `sample_name` and `sample_id`/`sample-id` are present, Qiita will drop the columns `sample_id`/`sample-id` to avoid issues with QIIME2
+* The redbiom webpage only accepts ' or " for escaping strings
+* Added qiime2.2019.10 to the system; which updated these plugins: qp-qiime2, qtp-biom, qtp-diversity, qtp-visualization
+
 Version 092019
 --------------
 
--- a/qiita_db/download_link.py
+++ b/qiita_db/download_link.py
@@ -117,7 +117,8 @@ class DownloadLink(qdb.base.QiitaObject):
 
         Returns
         -------
-        A jwt
+        str
+            A JSON web token
 
         """
 
--- a/qiita_db/download_link.py
+++ b/qiita_db/download_link.py
@@ -6,8 +6,6 @@
 # The full license is in the file LICENSE, distributed with this software.
 # -----------------------------------------------------------------------------
 
-from __future__ import division
-
 import qiita_db as qdb
 
 from jose import jwt as jose_jwt
--- a/qiita_db/download_link.py
+++ b/qiita_db/download_link.py
@@ -61,7 +61,7 @@ class DownloadLink(qdb.base.QiitaObject):
             sql = """INSERT INTO qiita.{0} (jti, jwt, exp)
             VALUES (%s, %s, %s) RETURNING jti""".format(cls._table)
             qdb.sql_connection.TRN.add(sql, [jti, jwt, exp])
-            return qdb.sql_connection.TRN.execute_fetchlast()
+            qdb.sql_connection.TRN.execute()
 
     @classmethod
     def delete(cls, jti):
@@ -75,7 +75,7 @@ class DownloadLink(qdb.base.QiitaObject):
         with qdb.sql_connection.TRN:
             sql = """DELETE FROM qiita.{0} WHERE jti=%s""".format(cls._table)
             qdb.sql_connection.TRN.add(sql, [jti])
-            return qdb.sql_connection.TRN.execute_fetchlast()
+            qdb.sql_connection.TRN.execute()
 
     @classmethod
     def exists(cls, jti):
@@ -91,7 +91,7 @@ class DownloadLink(qdb.base.QiitaObject):
             sql = """SELECT COUNT(jti) FROM qiita.{0}
                      WHERE jti=%s""".format(cls._table)
             qdb.sql_connection.TRN.add(sql, [jti])
-            return qdb.sql_connection.TRN.execute_fetchlast()
+            return qdb.sql_connection.TRN.execute_fetchlast() == 1
 
     @classmethod
     def delete_expired(cls):
--- a/qiita_db/download_link.py
+++ b/qiita_db/download_link.py
@@ -111,7 +111,7 @@ class DownloadLink(qdb.base.QiitaObject):
         with qdb.sql_connection.TRN:
             sql = """DELETE FROM qiita.{0} WHERE exp<%s""".format(cls._table)
             qdb.sql_connection.TRN.add(sql, [now])
-            return qdb.sql_connection.TRN.execute_fetchlast()
+            qdb.sql_connection.TRN.execute()
 
     @classmethod
     def get(cls, jti):
--- a/qiita_db/download_link.py
+++ b/qiita_db/download_link.py
@@ -86,7 +86,6 @@ class DownloadLink(qdb.base.QiitaObject):
         -------
         bool
             True if link exists else false
-
         """
 
         with qdb.sql_connection.TRN:
--- a/qiita_db/download_link.py
+++ b/qiita_db/download_link.py
@@ -97,13 +97,7 @@ class DownloadLink(qdb.base.QiitaObject):
 
     @classmethod
     def delete_expired(cls):
-        r"""Deletes all expired download links
-
-        Returns
-        -------
-        None
-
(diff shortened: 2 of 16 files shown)


antgonza:
An example of how this works:

qiita_db/download_link.py
```diff
@@ -0,0 +1,130 @@
...
+                                   algorithms='HS256')
+        jti = jwt_data["jti"]
+        exp = datetime.utcfromtimestamp(jwt_data["exp"] / 1000)
+
+        with qdb.sql_connection.TRN:
+            if cls.exists(jti):
+                raise qdb.exceptions.QiitaDBDuplicateError(
+                    "JTI Already Exists")
+
+            # Insert study into database
```
ElDeveloper:
```suggestion
            # insert token into database
```

qiita_db/download_link.py
```diff
@@ -0,0 +1,130 @@
...
+        Parameters
+        ----------
+        jti : object
+            The jwt token identifier
+
+        """
+        with qdb.sql_connection.TRN:
+            sql = """DELETE FROM qiita.{0} WHERE jti=%s""".format(cls._table)
+            qdb.sql_connection.TRN.add(sql, [jti])
+            return qdb.sql_connection.TRN.execute_fetchlast()
```
ElDeveloper: What's the return value here?

qiita_pet/static/js/qiita.js
```diff
@@ -398,3 +398,24 @@ function format_biom_rows(data, row, for_study_list = true, samples = null) {
...
+        title: response.url,
+        href: response.url
+      });
+      $('#privateDownloadText').text('Link will expire in 7 days');
+      $('#privateDownloadText').append('<br/>')
+      $('#privateDownloadText').append(newLink)
+      $('#privateDownloadLink').collapse('show')
+  },
+  error: function(resp){
+    $('#downloadLinkText').text('Failed to Generate Download Link');
```
ElDeveloper: Is there any additional information we could give the users for why a download might have failed?

qiita_pet/test/test_download.py
```diff
@@ -561,5 +564,33 @@ def test_download_sample_information(self):
         self.assertEqual(response.reason, 'You need to specify an artifact id')
 
 
+class TestDownloadPrivateArtifactHandler(TestHandlerBase):
+    def setUp(self):
+        super(TestDownloadPrivateArtifactHandler, self).setUp()
+
+    def tearDown(self):
+        super(TestDownloadPrivateArtifactHandler, self).tearDown()
```
ElDeveloper: Are these calls necessary? I would imagine that if there's no changes to be made to setup/teardown then the test runner can use the super class's methods.

qiita_pet/test/test_download.py
```diff
@@ -561,5 +564,33 @@ def test_download_sample_information(self):
...
+
+    def test_download(self):
+        # Stupidly, you can't post None, you must post an empty byte array
```
ElDeveloper:
```suggestion
        # you can't post None, you must post an empty byte array
```

qiita_pet/handlers/qiita_redbiom.py
```diff
@@ -49,13 +49,12 @@ def _redbiom_metadata_search(self, query, contexts):
             study_artifacts = defaultdict(lambda: defaultdict(list))
             for ctx in contexts:
                 # redbiom.fetch.data_from_samples returns a biom, which we
-                # will ignore, and a dict
+                # will ignore, and a dict: {sample_id_in_table: original_id}
```
ElDeveloper: These changes shouldn't be here, right?

qiita_db/download_link.py
dhakim87: I admittedly have no idea how sql_connection.TRN is supposed to be used.  In an ideal world, I would return whether the cursor's row count was 1, indicating that a row was deleted.  I think I'll settle for returning nothing and switch it to a TRN.execute()

qiita_pet/static/js/qiita.js
dhakim87: Nothing friendly I don't think.  A failure here indicates that the AJAX request to DownloadPrivateArtifactHandler.post failed out.  That could mean: Qiita is down, The user session id has expired so their authentication failed, The user is not (no longer?) able to view the chosen artifact according to check_artifact_access, various server side or network issues - no postgres server access etc.  

Basically the only thing they can possible do is refresh the page, and the only thing that might change is that they are forced to login again or their access to that artifact disappears out from under them.

qiita_pet/test/test_download.py
dhakim87: I don't know.  I'm fairly certain they don't hurt anything at least.  This is the pattern used by every class in the test_download file.

qiita_pet/handlers/qiita_redbiom.py
dhakim87: Agree, this is separate from my changes.

CHANGELOG.md
```diff
@@ -1,5 +1,15 @@
 # Qiita changelog
 
+Version 112019
```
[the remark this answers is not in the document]
dhakim87: I'm not sure I understand.  Is 112019 the latest or not?  I see 112019 as the current top of the changelog in biocore/qiita/master.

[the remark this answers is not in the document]
ElDeveloper: What I mean is that since this block is in the latest version of master, the diff shouldn't show these lines as new additions.

I suspect either the `dev` branch in `biocore/qiita` is not up to date (with master), or that your branch is somehow out of date.

qiita_pet/static/js/qiita.js
ElDeveloper: I see, this should be fine then. Thanks for the explanation.

qiita_pet/test/test_download.py
ElDeveloper: If these are not necessary then I would remove them. Looks like only a few of sub-classes are doing additional work after calling super. The other classes probably copy-pasted the code? @antgonza do you know if this is necessary?

qiita_pet/handlers/qiita_redbiom.py
ElDeveloper: Thanks for confirming, probably related to the changelog.

qiita_pet/test/test_download.py
antgonza: No idea, what about removing from here and check if everything works as expected?

CHANGELOG.md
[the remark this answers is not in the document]
dhakim87: Both this and the redbiom changes, as well as several other changes that I suspect snuck in as well are authored in commit 9c2e0476641befcac6a8a00d1f3d975dd482e2b4.  I'm not sure why this commit is considered part of this pull request.  As far as I can tell, this commit has already been applied to the biocore/qiita/dev branch.  Presumably this means I busted the git history while I was cleaning up the commit log for the PR...

qiita_pet/test/test_download.py
ElDeveloper: That sounds like a good idea to me, make sure to count the same number of tests that are being executed, and that you can run the test suite a couple of times (to ensure there's no side effects).

CHANGELOG.md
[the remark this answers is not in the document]
ElDeveloper: Hmmm, maybe try git pull upstream dev. Otherwise that should be fine.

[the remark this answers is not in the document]
dhakim87: git pull upstream dev says there are no changes.  

I suspect this is the problem mentioned in: https://stackoverflow.com/questions/16306012/github-pull-request-showing-commits-that-are-already-in-target-branch

After the flake8 changes, which took a lot of commits and pushes to resolve, I squashed a series of commits down.  I suspect the github diff procedure is unable to recognize that commit 9c2e047 was applied to both the dev branch and the ST_DH_artifact_download branch, but from what I've read, we can fix by either reopening the pull request or just leaving it be as git will handle the merge correctly.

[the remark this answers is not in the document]
dhakim87: Possibly this is the same issue as: https://stackoverflow.com/questions/16306012/github-pull-request-showing-commits-that-are-already-in-target-branch

Seems like it happens after a rebase, which I did to squash the commits after making the flake8 changes.  Could be fixed by reopening the pull request, or just ignoring it and letting git handle it, since git seems to handle the diff correctly even though github can't.

qiita_pet/test/test_download.py
dhakim87: Removed them, 12 tests before, 12 tests after, no apparent side effects.  Looks like they're unnecessary.

ElDeveloper: Wonderful, thanks for looking into this!

CHANGELOG.md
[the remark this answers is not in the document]
ElDeveloper: Sounds good, lets leave this aside then.

antgonza:
Thank you @dhakim87 and @ElDeveloper !

ElDeveloper:
Thanks @dhakim87 and @antgonza

PR #2963, opened 2019-12-13
add private download info to docs

antgonza:
Looks like this: 
Screen Shot 2019-12-13 at 8 48 46 AM

### DIFF
--- a/qiita_pet/support_files/doc/source/downloading.rst
+++ b/qiita_pet/support_files/doc/source/downloading.rst
@@ -74,6 +74,16 @@ replace artifact-id):
 
 - https://qiita.ucsd.edu/public_artifact_download/?artifact_id=artifact-id
 
+Access a non-public artifacts without login
+-------------------------------------------
+
+**no log required / wget or curl friendly**
+
+Note that this is only possible for non-public and first you need to generate a unique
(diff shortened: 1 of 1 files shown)


qiita_pet/support_files/doc/source/downloading.rst
```diff
@@ -74,6 +74,16 @@ replace artifact-id):
 
 - https://qiita.ucsd.edu/public_artifact_download/?artifact_id=artifact-id
 
+Access a non-public artifacts without login
```
ElDeveloper:
```suggestion
Access non-public artifacts without a login
```

qiita_pet/support_files/doc/source/downloading.rst
```diff
@@ -74,6 +74,16 @@ replace artifact-id):
...
+-------------------------------------------
+
+**no log required / wget or curl friendly**
+
+Note that this is only possible for non-public and first you need to generate a unique
```
ElDeveloper: Looks good, but a little convoluted what about something like:

```
This feature is currently only available for non-public artifacts.
To create a link you must be the owner of the study that contains
the artifact. To generate the link ... 
```

I would also change link to URL but that's up to you.

PR #2966, opened 2020-01-02
fix #2959

### LINKED ISSUE #2959: Using the update sample information button yields an error
If you upload a sample information file using the "Direct upload" button in the sample information page in a study that already has sample information, you will get an error:

```
Error (log id: 242137): The 'SampleTemplate' object with attributes (id: 12802) already exists
```

If you use the uploads page and then click on the "update" button then everything works as expected.

antgonza:
Turns out that the direct upload never really worked because the JS assumed that it was only being used to create new Sample Informations while the Python code assumed that it was used during update. This PR fixes both; here an example of creating and updating the sample information file via the regular upload-folder and the direct-upload.


qiita_pet/handlers/study_handlers/sample_template.py
```diff
@@ -203,11 +206,16 @@ def sample_template_handler_patch_request(user, req_op, req_path,
                 raise HTTPError(400, reason="Value is required when updating "
                                 "sample information")
 
-            # Check if the file exists
-            fp_rsp = check_fp(study_id, req_value)
-            if fp_rsp['status'] != 'success':
-                raise HTTPError(404, reason='Filepath not found')
-            filepath = fp_rsp['file']
+            if direct_upload:
+                # We can assume that the file exist as it was generated by
```
wasade: I don't believe this is safe. If the file is going onto a NFS mount, then just because it's uploaded does not mean all hosts mounting the share can see the file.

qiita_pet/handlers/study_handlers/sample_template.py
```diff
@@ -303,9 +312,19 @@ def patch(self):
         req_path = self.get_argument('path')
         req_value = self.get_argument('value', None)
         req_from = self.get_argument('from', None)
+        direct_upload = self.get_argument('direct_upload', False)
+
+        if direct_upload and direct_upload == 'true':
+            direct_upload = True
+
+            fd, req_value = mkstemp(suffix='.txt')
```
wasade: The pattern here is unexpected, why get a handle and then close it? What about:

```python
with tempfile.NamedTemporaryFile(suffix='.txt') as fp:
     fp.write(self.request.files['value'][0]['body'])
```

This approach should not require the decode as the file is opened in binary.

qiita_pet/handlers/study_handlers/sample_template.py
wasade: Also, this method isn't implied to be private or encapsulated, so this is really more more in hope that the caller has taken care of this...

qiita_pet/static/js/sampleTemplateVue.js
```diff
@@ -60,6 +60,7 @@ Vue.component('sample-template-page', {
               '</div>' +
             '</div>',
   props: ['portal', 'study-id'],
+  maxDirectUploadSize: 2097152,
```
wasade: 2MB? Any reason to keep it so limited?

qiita_pet/handlers/study_handlers/sample_template.py
antgonza: Good point, didn't consider originally. However, in Qiita this is safe as the direct-upload folder uses mkstemp, which uses TMPDIR, which is shared in the system.

qiita_pet/handlers/study_handlers/sample_template.py
antgonza: Good to know, will replace; thanks.

qiita_pet/static/js/sampleTemplateVue.js
antgonza: Comes from original issue discussion on the file sizes: https://github.com/biocore/qiita/issues/2433; I can raise if you think is fine/safe; do you have a # suggestion?

qiita_pet/handlers/study_handlers/sample_template.py
wasade: Is `TMPDIR` assured to be a local filesystem?

qiita_pet/static/js/sampleTemplateVue.js
wasade: At least 10MB given EBI's limit?

qiita_pet/handlers/study_handlers/sample_template.py
antgonza: No, it's a mount folder, just like the upload folder for any study. I can add an explicit check (if exists) and raise an error but it will look pretty similar to the error the regular open will raise.

wasade: may be nice to be explicit, and to catch a pathological situation as early as possible though?

antgonza: Sure ... will add an error saying to contact qiita.help immediately ...

PR #2970, opened 2020-01-15
improve QIIME2 help info

antgonza:
Now looks like this:
Screen Shot 2020-01-15 at 11 45 11 AM

@nbokulich, could you confirm that this looks fine?


nbokulich:
LGTM thanks @antgonza , maybe just tweak this:
"Thus, we suggest posting your question" --> "Thus, we suggest posting general-interest questions"

antgonza:
This was reviewed by the core qiime2-devs and the text was agreed via email; thus merging.

Issue #2967, opened 2020-01-02
Add a deprecation flag to preparations


antgonza:
Adding a simple flag in the database to mark preparations as deprecated will allow users/owners to mark all the artifacts from that preparation as deprecated.

PR #2986, opened 2020-03-17
fix #2952

### LINKED ISSUE #2952: Add a simple "notes for data users" field for a given study
As brought up in a meeting this morning. This could include information relevant to users interested in using data from that study, preferably in a brief couple of sentences/bullet points.

I know there's the "Abstract" field, but in my experience that is usually just the straight-up abstract from a writeup, and I've literally never seen it used in this way. The proposed field could just be a simple string that'd be shown towards the top of the Study Information (and maybe Sample Information?) pages.

Some immediately obvious example things that a "Notes for data users" field could be used for:

- For the AGP (study ID 10317), mention the bloom filter (maybe even link to the paper on correcting for fecal blooms) and explain that we strongly recommend applying this to the data due to the vast majority of the samples being shipped at room-temperature

- For a study where some samples were incorrectly collected or processed somehow (as was brought up as an example problematic scenario in this morning's meeting), mention which samples are bad (or refer to the sample mapping/metadata info, e.g. "the samples where `is_sample_terrible` is `YEAH` should be filtered out of downstream analyses")

- For a study with metadata column names that are hard to interpret in isolation, give explanations for what these columns mean (e.g. "the `is_a_cool_dude` metadata column is `YES` for samples where the host's name is `Antonio`, and `PROBABLY NOT` for all other examples")

I think esp. the bloom filter example would make this sort of feature valuable.

antgonza:


antgonza:
Ready for review! @fedarko, could you take a look too? 

Notes:
- There are 12 files that are changing but the changes are super minor
- The SQL already includes changes for https://github.com/biocore/qiita/issues/2967 so we can simplify that PR and avoid having to regenerate all db files
- The notes (Analytical Notes) will be added in the Study creation/edition. This will allow only owners and admins to add them, with minimal changes to the code:
add-notes

- Studies with no analytical notes will be display as normal:
no-notes

- Studies with notes will have an extra space in their summary:
notes

fedarko:
Thanks @antgonza! From a UI standpoint, at least, this looks good to me. One minor suggestion would be maybe adding an example in the notes section description -- for example, changing it to

> Any relevant information about the samples or the processing that other users should be aware of **(e.g. problematic samples, explaining certain metadata columns, ...)**

The main case I'm thinking of here is someone rushing through the process of creating/editing a study who might skip over this section without these sorts of guiding examples.

antgonza:
Good idea, I'll add it once @ElDeveloper takes a look as I'm almost certain we will need some other changes ...

qiita_db/study.py
```diff
@@ -515,6 +515,36 @@ def title(self, title):
...
+        """
+        with qdb.sql_connection.TRN:
+            sql = """SELECT notes FROM qiita.{0}
+                     WHERE study_id = %s""".format(self._table)
+            qdb.sql_connection.TRN.add(sql, [self._id])
+            return qdb.sql_connection.TRN.execute_fetchlast()
+
+    @notes.setter
+    def notes(self, notes):
+        """Sets the title of the study
```
ElDeveloper:
```suggestion
        """Sets the notes of the study
```

antgonza:
@ElDeveloper good idea, this change is fine if we _just_ display the saved text as markdown; like in my latests push.

The create/edit area hasn't changed (except for the changes requested by @fedarko):
create:edit

but the display now uses markdown:
render

qiita_pet/handlers/study_handlers/edit_handlers.py
```diff
@@ -63,6 +63,10 @@ class StudyEditorForm(Form):
                                          coerce=lambda x: x)
 
     lab_person = SelectField('Lab Person', coerce=lambda x: x)
+    notes = TextAreaField('Analytical Notes', description=(
+        'Any relevant information about the samples or the processing that '
+        'other users should be aware of (e.g. problematic samples, '
+        'explaining certain metadata columns, etc) - accepts markdown'))
```
ElDeveloper:
```suggestion
        'explaining certain metadata columns, etc) - renders as markdown'))
```

PR #2987, opened 2020-03-19
Fix 2967

antgonza:
GUI interaction:

How a non-owner/admin will see the deprecation message:
deprecated

Note that this will not prevent adding the artifacts to an analyses but now the analyses will have an auto-added new column: `qiita_prep_deprecated`.

@justinshaffer, could you take a look too on the functionality?


antgonza:
Well, to have the study list not show artifacts that were deprecated via the prep information was much harder than expected. Mainly because there was no artifact/preparation table. 

Note that all performance testing was done in qiita-rc. Anyway, I initially tried to avoid having to create a new table by using a pgsql function but the retrieval time incremented by ~5 times (*):
```python
In [13]: %timeit x = generate_study_list_new(User('antoniog@ucsd.edu'), 'public')                                                                                                                                                                                                                 
20.1 s ± 39.6 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

However, adding the new table actually slightly improves retrieval time:
```python
In [7]: %timeit x = generate_study_list(User('antoniog@ucsd.edu'), 'public')                                                                                                                                                                                                                      
4.03 s ± 11.9 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

In [26]: %timeit x = generate_study_list_new(User('antoniog@ucsd.edu'), 'public')                                                                                                                                                                                                                 
3.86 s ± 14.1 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

qiita_pet/templates/study_ajax/prep_summary.html
```diff
@@ -133,6 +133,41 @@
     }
   }
 
+
+  /*
+   *
+   * Updates deprecated flag in the prep information <fieldset>
+   *
+   * @param prep_id the prep template id
+   *
```
ElDeveloper: looks like `new_value` is missing here.

qiita_pet/templates/study_ajax/prep_summary.html
```diff
@@ -133,6 +133,41 @@
...
+   */
+  function deprecate_preparation(prep_id, new_value) {
+    var status = 'deprecate';
+    console.log(new_value)
+    if (!new_value){
+      status = 'DE-deprecate'
+    }
+    if(confirm("Are you sure you want to " + status + " this preparation?")) {
+      if(confirm("Are you _really_ sure you want to " + status + " this preparation?")) {
```
ElDeveloper: Two "alerts" seems like a little too much in my opinion. Is there a strong motivation for this?

qiita_pet/templates/study_ajax/prep_summary.html
```diff
@@ -396,6 +436,12 @@ <h4>
         <a class="btn btn-default" href="{% raw qiita_config.portal_dir %}/download_ebi_accessions/experiments/{{prep_id}}"><span class="glyphicon glyphicon-download-alt"></span> EBI experiment accessions</a>
       {% end %}
       {% if editable %}
+        <br/>
+        {% if deprecated %}
+          <a class="btn btn-warning" onclick="deprecate_preparation({{prep_id}}, false);"><span class="glyphicon glyphicon-pushpin"></span> Remove Deprecate</a>
```
ElDeveloper:
```suggestion
          <a class="btn btn-warning" onclick="deprecate_preparation({{prep_id}}, false);"><span class="glyphicon glyphicon-pushpin"></span> Remove Deprecation</a>
```

qiita_db/util.py
```diff
@@ -1520,17 +1499,25 @@ def generate_study_list(user, visibility):
...
-                info['artifact_biom_ids'] = []
-                if info['aids_with_deprecation'] is not None:
-                    for x in info['aids_with_deprecation']:
-                        # f1-2 are the default names given by pgsql
-                        if not x['f2']:
-                            info['artifact_biom_ids'].append(x['f1'])
-                del info['aids_with_deprecation']
-
-                if info['preparation_data_types'] is None:
-                    info['preparation_data_types'] = []
```
ElDeveloper: Is this no longer needed?

qiita_db/metadata_template/prep_template.py
```diff
@@ -480,6 +480,29 @@ def study_id(self):
             qdb.sql_connection.TRN.add(sql, [self.id])
             return qdb.sql_connection.TRN.execute_fetchlast()
 
+    @property
+    def deprecated(self):
+        with qdb.sql_connection.TRN:
+            sql = """SELECT deprecated FROM qiita.prep_template
+                     WHERE {0} = %s""".format(self._id_column)
+            qdb.sql_connection.TRN.add(sql, [self._id])
+            return qdb.sql_connection.TRN.execute_fetchlast()
```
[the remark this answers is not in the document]
antgonza: bool:
```python
In [1]: from qiita_db.metadata_template.prep_template import PrepTemplate                                                                                                                                                                                                                         

In [2]: PrepTemplate(1).deprecated                                                                                                                                                                                                                                                                
Out[2]: False

In [3]: type(PrepTemplate(1).deprecated)                                                                                                                                                                                                                                                          
Out[3]: bool
```

qiita_db/util.py
antgonza: This is correct, now the preparation data types is retrieved together with the artifact info, lines 1502 and 1511 below.

qiita_pet/templates/study_ajax/prep_summary.html
antgonza: Just that in sensitive things we always add two but in this case it's pretty easy to correct so I will leave only one.

qiita_db/metadata_template/prep_template.py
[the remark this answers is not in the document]
ElDeveloper: Thanks!

qiita_pet/templates/study_ajax/prep_summary.html
ElDeveloper: My bad, if two confirmations are widespread in Qiita then I rather be consistent than go for a personal preference. Thanks eitherway! 👍

antgonza:
Nah, I agree that 2 confirmations were too much and normally the double confirmation is for those things that can't be undone (like delete a full study) ... and here a user will just need to click again, so 2 confirmations were a lot ... thank you for checking.

ElDeveloper:
Thanks @antgonza!

PR #2992, opened 2020-04-13
fix EBI-ENA submission with already submitted samples

antgonza:
This PR:
1. fixes the EBI-ENA submissions that contain samples that were submitted before
2. updated the size of the submission that is now allowed (10M, before 8.5M)

The big change is that before we could "update" the sample information and now it's not allowed, so basically we need to skip any sample that has been previously submitted from the sample information - note that it is still added in the experiment.

### DIFF
--- a/qiita_ware/commands.py
+++ b/qiita_ware/commands.py
@@ -207,7 +207,7 @@ def submit_EBI(artifact_id, action, send, test=False, test_size=False):
                  ebi_submission.submission_xml_fp]
     total_size = sum([stat(tr).st_size for tr in to_review if tr is not None])
     # note that the max for EBI is 10M but let's play it safe
-    max_size = 8.5e+6 if not test_size else 6000
+    max_size = 10e+6 if not test_size else 5000
     if total_size > max_size:
         LogEntry.create(
             'Runtime', 'The submission: %d is larger than allowed (%d), will '
@@ -228,10 +228,12 @@ def submit_EBI(artifact_id, action, send, test=False, test_size=False):
         cols_to_drop = cols_to_drop - {'taxon_id', 'scientific_name',
                                        'description'}
         all_samples = ebi_submission.sample_template.ebi_sample_accessions
-        samples = {k: all_samples[k] for k in ebi_submission.samples}
-        ebi_submission.write_xml_file(
-            ebi_submission.generate_sample_xml(samples, cols_to_drop),
-            ebi_submission.sample_xml_fp)
+        samples = {k: all_samples[k] for k in ebi_submission.samples
+                   if all_samples[k] is None}
+        if samples:
+            ebi_submission.write_xml_file(
+                ebi_submission.generate_sample_xml(samples, cols_to_drop),
+                ebi_submission.sample_xml_fp)
 
         # now let's recalculate the size to make sure it's fine
         new_total_size = sum([stat(tr).st_size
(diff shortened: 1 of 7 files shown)


qiita_ware/test/test_ebi.py
```diff
@@ -586,13 +587,11 @@ def test_generate_xml_files(self):
         e.generate_demultiplexed_fastq()
         self.assertIsNone(e.run_xml_fp)
         self.assertIsNone(e.experiment_xml_fp)
-        self.assertIsNone(e.sample_xml_fp)
```
ElDeveloper, line 589: From this PR I can't see why this attribute is no longer tested. Would you mind explaining?

antgonza: Will re-add, thank you for catching ...

qiita_ware/test/test_ebi.py
```diff
@@ -608,7 +609,6 @@ def test_generate_xml_files(self):
         e.generate_xml_files()
         self.assertIsNotNone(e.run_xml_fp)
         self.assertIsNotNone(e.experiment_xml_fp)
-        self.assertIsNone(e.sample_xml_fp)
```
antgonza, line 611: ... forgot this one, will wait for tests and readd ...

qiita_ware/commands.py
```diff
@@ -228,10 +228,12 @@ def submit_EBI(artifact_id, action, send, test=False, test_size=False):
         cols_to_drop = cols_to_drop - {'taxon_id', 'scientific_name',
                                        'description'}
         all_samples = ebi_submission.sample_template.ebi_sample_accessions
-        samples = {k: all_samples[k] for k in ebi_submission.samples}
-        ebi_submission.write_xml_file(
-            ebi_submission.generate_sample_xml(samples, cols_to_drop),
-            ebi_submission.sample_xml_fp)
+        samples = {k: all_samples[k] for k in ebi_submission.samples
+                   if all_samples[k] is None}
```
ElDeveloper, line 232: There's one thing that threw me off when re-reading this PR. The argument `samples` for `generate_sample_xml` is described as a list in the docstring. Should this instead be:

```suggestion
        samples = [k for k in ebi_submission.samples if all_samples[k] is None]
```

Or should the docstring be corrected?

qiita_ware/ebi.py
```diff
@@ -411,14 +411,18 @@ def generate_sample_xml(self, samples=None, ignore_columns=None):
...
 
-            if self._ebi_sample_accessions[sample_name] is None:
-                sample = ET.SubElement(sample_set, 'SAMPLE', {
-                    'alias': self._get_sample_alias(sample_name),
-                    'center_name': qiita_config.ebi_center_name}
-                )
+            sample_accession = self._ebi_sample_accessions[sample_name]
+            if self.action in ('ADD', 'VALIDATE'):
+                if sample_accession is not None:
+                    continue
```
ElDeveloper, line 417: Should we raise a warning for samples that were already submitted? So that this gets logged somewhere, or would that information not be useful at all? I'm thinking it might be helpful in the context of admin/help tasks.

antgonza: It has never came up and IMOO I don't think it will be useful here. Note that we display this information (samples in multiple preps) in the GUI in the sample/prep tab.

qiita_ware/commands.py
antgonza: Good catch, I'll change. Note that it works cause:
```python
>>> for x in {'a': 1, 'b': 2}:
...    print(x)
... 
a
b
```

ElDeveloper:
Thanks @antgonza !

PR #2993, opened 2020-04-24
fix #2773

### LINKED ISSUE #2773: nightly redbiom updates
We have a message at the top of the page that says that redbiom updates nightly, which is not accurate, need to remove until this is true.

antgonza:
Originally #2773 was created as the qiita redbiom page displayed the incorrect information of when redbiom was deployed. This PR adds that information to the page; thus it should close that issue.

### DIFF
--- a/qiita_pet/handlers/qiita_redbiom.py
+++ b/qiita_pet/handlers/qiita_redbiom.py
@@ -13,8 +13,10 @@ import redbiom.search
 import redbiom._requests
 import redbiom.util
 import redbiom.fetch
+import redbiom.admin
 from tornado.gen import coroutine, Task
 from tornado.web import HTTPError
+from requests.exceptions import HTTPError as rHTTPError
 
 from qiita_core.util import execute_as_transaction
 from qiita_db.util import generate_study_list_without_artifacts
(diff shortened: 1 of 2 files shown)


qiita_pet/handlers/qiita_redbiom.py
```diff
@@ -27,9 +29,18 @@ class RedbiomPublicSearch(BaseHandler):
     def get(self, search):
         # making sure that if someone from a portal forces entry to this URI
         # we go to the main portal
+        try:
+            timestamps = redbiom.admin.get_timestamps()
+        except (rHTTPError):
+            timestamps = []
+
+        if timestamps:
+            latests_release = timestamps[0]
```
ElDeveloper, line 38: This is minor but the variable name should be `latest_releases` right? If you agree then make sure to also update the template.

antgonza: Happy to change but note that is only taking one by `[0]`

ElDeveloper: Gotcha, then `latest_release` should work well.

ElDeveloper:
Thanks @antgonza!

PR #2994, opened 2020-04-27
fix #2902

### LINKED ISSUE #2902: The "deleting" state of an artifact is not picked up on reload
If a large preparation artifact is deleted, it will show the "red" state on the navigation window. If the user changes pages, and then comes back to the preparation, the "red" state is no longer present, and no indication to the user that the artifact is deleting is obvious. 

As a result, a user may attempt to delete the artifact multiple times which is presently allowed by the system.

antgonza:
Before this patch we didn't keep track of deleting jobs and this was the behavior:
2902_before

The main issues were:
- The artifact id was not being parse as string so JS couldn't be updated
- The internal jobs were not accounted as part of the artifacts, which mean that they weren't able to be track
- We didn't have a method to query if an artifact was being deleted; because it was not easy to track
- Once we have those things we can use in the GUI to reflect the delete status

### DIFF
--- a/qiita_db/processing_job.py
+++ b/qiita_db/processing_job.py
@@ -591,6 +591,9 @@ class ProcessingJob(qdb.base.QiitaObject):
                         TTRN.add(sql, [artifact_info, job_id])
                     else:
                         pending[artifact_info[0]][pname] = artifact_info[1]
+                elif pname == 'artifact':
+                    TTRN.add(sql, [parameters.values[pname], job_id])
+
             if pending:
                 sql = """UPDATE qiita.processing_job
                          SET pending = %s
--- a/qiita_db/test/test_artifact.py
+++ b/qiita_db/test/test_artifact.py
@@ -14,7 +14,6 @@ from os.path import exists, join, basename
 from shutil import copyfile
 from functools import partial
 from json import dumps
-from time import sleep
 
 import pandas as pd
 import networkx as nx
@@ -22,6 +21,7 @@ from biom import example_table as et
 from biom.util import biom_open
 
 from qiita_core.util import qiita_test_checker
+from qiita_core.testing import wait_for_processing_job
 import qiita_db as qdb
 
 
@@ -1148,8 +1148,7 @@ class ArtifactTests(TestCase):
             qdb.user.User('test@foo.bar'), params, True)
         job.submit()
         # let's wait for job
-        while job.status in ('queued', 'running'):
-            sleep(1)
+        wait_for_processing_job(job.id)
 
         with self.assertRaises(qdb.exceptions.QiitaDBUnknownIDError):
             qdb.artifact.Artifact(test.id)
--- a/qiita_db/test/test_artifact.py
+++ b/qiita_db/test/test_artifact.py
@@ -14,6 +14,7 @@ from os.path import exists, join, basename
 from shutil import copyfile
 from functools import partial
 from json import dumps
+from time import sleep
 
 import pandas as pd
 import networkx as nx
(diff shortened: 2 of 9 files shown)


qiita_db/processing_job.py
```diff
@@ -591,6 +591,9 @@ def create(cls, user, parameters, force=False):
                         TTRN.add(sql, [artifact_info, job_id])
                     else:
                         pending[artifact_info[0]][pname] = artifact_info[1]
+                elif pname == 'artifact':
```
ElDeveloper, line 594: Would you mind describing what's going on here?

qiita_db/test/test_artifact.py
```diff
@@ -1114,6 +1115,45 @@ def test_delete_with_jobs(self):
...
+        job._set_error('Killed by admin')
+        self.assertIsNone(test.being_deleted_by)
+
+        # now, let's actually remove
+        job = qdb.processing_job.ProcessingJob.create(
+            qdb.user.User('test@foo.bar'), params, True)
+        job.submit()
+        # let's wait for job
+        while job.status in ('queued', 'running'):
+            sleep(1)
```
ElDeveloper, line 1152: I was looking around the repo and found `qiita_core.testing.wait_for_processing_job`. I think that would be able to do this.

qiita_pet/handlers/api_proxy/prep_template.py
```diff
@@ -660,11 +661,14 @@ def prep_template_graph_get_req(prep_id, user_id):
     G = artifact.descendants_with_jobs
 
     nodes, edges, wf_id = get_network_nodes_edges(G, full_access)
+    artifacts_being_deleted = [a[2] for a in nodes if a[0] == 'artifact' and
```
ElDeveloper, line 664: Can you add a short description of what `a[0]` and `a[2]` are?

qiita_pet/handlers/artifact_handlers/base_handlers.py
```diff
@@ -390,25 +391,33 @@ def artifact_post_req(user, artifact_id):
     """
     artifact_id = int(artifact_id)
     artifact = Artifact(artifact_id)
-    check_artifact_access(user, artifact)
```
ElDeveloper, line 393: Any reason why the access doesn't need to be checked if the artifact is being deleted?

qiita_pet/util.py
```diff
@@ -188,6 +188,9 @@ def get_network_nodes_edges(graph, full_access, nodes=None, edges=None):
     # n[1] is the object
     for n in graph.nodes():
         if n[0] == 'job':
+            # ignoring internal Jobs
```
ElDeveloper, line 191: Does this need a new test case?

qiita_db/processing_job.py
antgonza: Sure, the `create_command` method in patch `58.py` adds the `delete_artifact`  and other Qiita internal commands; however, the artifact_id is in the parameter "artifact" vs. the parameter description (*); thus, if we want to add these commands to the database we need this special case. 

(*) all regular jobs, except internal Qiita, have their parameters in this format: parameter_name: [parameter_type, value]; so normally, we just check that parameter_type is `artifact` but for internal jobs we need to check the actual parameter.

qiita_pet/handlers/artifact_handlers/base_handlers.py
antgonza: Thank you for catching this!!! It was an extra delete!

qiita_pet/util.py
antgonza: This if is needed as a consequence of now adding the internal jobs to the database; my guess is that originally we didn't have it cause we didn't store them. In other words, the test for these method would need to change if we didn't have this if (as now the jobs will be returned here) AKA it's being tested here.

qiita_pet/handlers/artifact_handlers/base_handlers.py
antgonza: Just realized that it was under the new if and now is duplicated; will rm the duplicated line

PR #3003, opened 2020-05-31
rm future

antgonza:


qiita_db/test/test_util.py
```diff
@@ -1172,13 +1172,7 @@ def test_BytesIO(self):
         with qdb.util.open_file(f) as fh:
             self.assertTrue(fh is f)
 
-    def test_hdf5IO(self):
```
antgonza, line 1175: Not sure why we need this test but it seems like a duplicate of the one below ...

ElDeveloper: Looks like this test is different from 1175, since here the test is checking for the identity of the filehandle, whereas the other test checks that the returned object is an instance of `File`. Maybe worth adding the assertion in 1178 to the test below.

ElDeveloper:
Ooops, looks like my local repo was very outdated. Future was already removed from the installation instructions and from the travis config.

antgonza:
AFAIK we shouldn't have any issues with missing files/templates but hard to confirm a negative ... 

BTW I didn't use any of those tools; however, just tried to install drop2 and ran into some weird errors in the installation and wasn't sure how to run pyupgrade. 

I'll update the test.

ElDeveloper:
Thanks @antgonza!

PR #3006, opened 2020-06-18
adding scripts/qiita-auto-processing

antgonza:


scripts/qiita-auto-processing
```diff
@@ -0,0 +1,271 @@
...
+for pipeline in full_pipelines:
+    # Step 2. From the steps generate the list of commands to add to the
+    #         workflow
+    commands = []
+    for step in pipeline['steps']:
+        plugin = _get_plugin(step['plugin'], step['version'])
+        cmds = [c for c in plugin.commands if c.name == step['cmd_name']]
+        if len(cmds) != 1:
+            raise ValueError('There is more than one command with this '
+                             'defenition %s' % str(step))
```
ElDeveloper, line 116:
```suggestion
                             'definition %s' % str(step))
```

scripts/qiita-auto-processing
```diff
@@ -0,0 +1,271 @@
...
+        commands.append({
+            'command': cmd,
+            'command-name': cmd.name,
+            'previous-step': step['previous-step'],
+            'parent_artifact_name': step['parent_artifact_name'],
+            'input_name': step['input_name'],
+            'parameters': parameters})
+
+    # Step 3. Find all preparations/artifacts that we can add the pipeline
+    artifacs_all = [a for study in Study.iter()
```
ElDeveloper, line 133: Might be easier to do a find/replace of artifacs_all -> artifacts_all.

scripts/qiita-auto-processing
```diff
@@ -0,0 +1,271 @@
...
+                            satisfied = False
+                        continue
+                    prep_value = list(map(str.lower, set(
+                        pt.get_category(req['column']).values())))
+                    if req['equal'] and prep_value != req['value']:
+                        satisfied = False
+                        continue
+                    elif not req['equal'] and prep_value == req['value']:
+                        satisfied = False
+                        continue
```
ElDeveloper, line 188: Any chance the code here could be put in a function that gets called for both cases?

antgonza:
Good point; I'm not sure if it's gonna work but I added the script to travis to see what happens ... anyway, this is code is so specific for the main site and full deploy that it will take a full machine to test ...

antgonza:
@ElDeveloper, the code could be tested (https://travis-ci.com/github/biocore/qiita/jobs/351694131#L2726) but we would need to change the user and the pipeline or install some plugins; thoughts?

ElDeveloper:
Got it. Then perhaps the easiest would be to include this script in the flake8 checks (I think it isn't yet) for a static analysis. If we think we'll be using this more and more in production, then I think testing the individual steps as functions (trying to abstract out the db) might be the best way.

ElDeveloper:
All checks seem tot be passing now, so let's go ahead and merge this for now. Once the library code is more fledged out we should be able to test that with more ease.

PR #3007, opened 2020-06-23
improve delete analysis

antgonza:
When deleting an analysis a user could find this error `Error (log id: 3): The object with ID '12' does not exists in table 'artifact'`, when the analysis had this _valid_ shape:
```
        #           8 -> 9 -> 10 -> 12 -> 14
        #                  -> 11 -> 13
```

### DIFF
--- a/qiita_ware/private_plugin.py
+++ b/qiita_ware/private_plugin.py
@@ -291,6 +291,14 @@ def delete_sample_or_column(job):
         job._set_status('success')
 
 
+def _delete_analysis_artifacts(analysis):
+    aids = [a.id for a in analysis.artifacts]
+    aids.sort(reverse=True)
+    for aid in aids:
+        qdb.artifact.Artifact.delete(aid)
+    qdb.analysis.Analysis.delete(analysis.id)
+
+
 def delete_study(job):
     """Deletes a full study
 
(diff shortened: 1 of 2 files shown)


qiita_ware/test/test_private_plugin.py
```diff
@@ -413,7 +415,41 @@ def test_delete_study_empty_study(self):
 @qiita_test_checker()
 class TestPrivatePluginDeleteAnalysis(BaseTestPrivatePlugin):
     def test_delete_analysis(self):
-        # as samples have been submitted to EBI, this will fail
+        # adding extra filepaths to make sure the delete works as expected, we
+        # basically want 8 -> 9 -> 10 -> 12 -> 14
+        #                       -> 11 -> 13
+        fd, fp10 = mkstemp(suffix='_table.biom')
```
ElDeveloper, line 421: Are these being added to a list of files to delete? If not, then maybe would be better to use TemporaryFile instead:

From the docs:
> Unlike TemporaryFile(), the user of mkstemp() is responsible for deleting the temporary file when done with it.

qiita_ware/private_plugin.py
```diff
@@ -291,6 +291,14 @@ def delete_sample_or_column(job):
         job._set_status('success')
 
 
+def _delete_analysis_artifacts(analysis):
+    aids = [a.id for a in analysis.artifacts]
```
ElDeveloper, line 295: The lines of code that this function is replacing also check for `if a.processing_parameters is None` and also delete `a.descendants.nodes()`. Why are these no longer relevant? Did something else change in the code ever since that was written?

antgonza: Yup, no needed anymore; I'm actually not sure why we decided to delete that way - it makes no sense to me today.

ElDeveloper: Gotcha, thanks!

qiita_ware/test/test_private_plugin.py
ElDeveloper: Also, looks like `fd` needs to explicitly be closed. I seemed to recall that, and looking at other portions of code in this file seem to confirm it.

ElDeveloper:
Thanks @antgonza!

Issue #3005, opened 2020-06-10
Add MTX to processing-recommendations


antgonza:
The recommendations are at the bottom of the page but the initial list doesn't have MTX: https://qiita.ucsd.edu/static/doc/html/processingdata/processing-recommendations.html

PR #3008, opened 2020-06-23
improve job.external_id

antgonza:
We currently store the external (queue job id or process local id) in the DB; this PR displays them in the GUI and improves how it gets set.


scripts/qiita
```diff
@@ -319,18 +319,18 @@ def start(port, master):
...
-                # Qiita recognizes the external ID from Watcher, and has
-                # mapped it to a Qiita ID. Now, create a Job object to
-                # register the job's changed status in Qiita.
-                job_obj = qdb.processing_job.ProcessingJob(qid)
-                job_obj.complete(job_state, error=job_error)
-            else:
+            try:
+                qjob = qdb.processing_job.ProcessingJob.by_external_id(
+                    msg['Job_Id'])
+            except Exception:
```
ElDeveloper, line 325: What exceptions are expected here? From the code in `by_external_id` it seems like there aren't any exceptions raised, right?

antgonza: Excellent question! Note that I was trying to be overly cautions. Anyway, the original `by_external_id` returned the job id and if it didn't existed it returned empty string; then inside the `if` it will instantiate the job. Thus, I thought it will be better to instantiate the job inside the new `by_external_id`, which can raise an issue if the job-id doesn't exist - this is actually yielded by the class instantiation. Should I limit to just `QiitaDBUnknownIDError`?

ElDeveloper: Thanks for the explanation, I think limiting to `QiitaDBUnknownIDError` would be best.

antgonza:
@ElDeveloper, I _think_ I fixed this branch and is ready to go. Note that the changes now show also the changes from dev (#3005 ) ...

ElDeveloper:
Thanks @antgonza

PR #3009, opened 2020-06-23
add disable to run button

antgonza:
Once in a while the system will have a delay when submitting new workflows and users can push the "Run" button more than once. This PR disables the Run button once the user hits it.

### DIFF
--- a/qiita_pet/static/js/networkVue.js
+++ b/qiita_pet/static/js/networkVue.js
@@ -349,6 +349,8 @@ Vue.component('processing-graph', {
         vm.updateGraph();
       })
         .fail(function(object, status, error_msg) {
+          $('#run-btn').attr('disabled', false);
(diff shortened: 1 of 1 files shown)


antgonza:
Forgo to add example: run-once

qiita_pet/static/js/networkVue.js
```diff
@@ -1120,7 +1120,11 @@ Vue.component('processing-graph', {
     show_loading('processing-network-div');
     $("#processing-network-div").hide();
 
-    $('#run-btn').on('click', function() { vm.runWorkflow(); });
+    $('#run-btn').on('click', function() {
+      $('#run-btn').attr('disabled', true);
```
ElDeveloper, line 1124: If there's an error with the submission request, then runWorkflow is going to catch that error in the `fail` method, right (line 351)? If so, can you remove the disabled class and reset the innerHTML for this button there?

ElDeveloper:
Thanks @antgonza

PR #3013, opened 2020-06-26
Mem allocation formula

antgonza:
Depends on https://github.com/biocore/qiita/pull/3012.

### DIFF
--- a/qiita_db/processing_job.py
+++ b/qiita_db/processing_job.py
@@ -22,7 +22,6 @@ from uuid import UUID
 from os.path import join
 from threading import Thread
 from humanize import naturalsize
-from numbers import Number
 
 from qiita_core.exceptions import IncompetentQiitaDeveloperError
 from qiita_core.qiita_settings import qiita_config
--- a/qiita_db/processing_job.py
+++ b/qiita_db/processing_job.py
@@ -22,6 +22,7 @@ from uuid import UUID
 from os.path import join
 from threading import Thread
 from humanize import naturalsize
+from numbers import Number
 
 from qiita_core.exceptions import IncompetentQiitaDeveloperError
 from qiita_core.qiita_settings import qiita_config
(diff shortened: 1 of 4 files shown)


qiita_db/test/test_processing_job.py
```diff
@@ -741,13 +741,13 @@ def test_hide(self):
     def test_shape(self):
         jids = {
             # Split libraries FASTQ
-            '6d368e16-2242-4cf8-87b4-a5dc40bb890b': (27, 31),
+            '6d368e16-2242-4cf8-87b4-a5dc40bb890b': (27, 31, 116),
             # Pick closed-reference OTUs
-            '80bf25f3-5f1d-4e10-9369-315e4244f6d5': (27, 31),
+            '80bf25f3-5f1d-4e10-9369-315e4244f6d5': (27, 31, 0),
```
ElDeveloper, line 746: Why is the input size zero? Is it because the FASTA files associated with this artifact are empty?

qiita_db/processing_job.py
```diff
@@ -488,9 +490,34 @@ def get_resource_allocation_info(self):
...
-            # [0] sending one element as execute_fetchflatten returns
-            # an array
-            return result[0]
+            allocation = result[0]
+
+            if ('{samples}' in allocation or '{columns}' in allocation or
+                    '{input_size}' in allocation):
+                samples, columns, input_size = self.shape
+                parts = []
+                for part in allocation.split():
```
ElDeveloper, line 499: What is the split being done on? Are we expecting multiple lines per allocation?

qiita_db/processing_job.py
```diff
@@ -488,9 +490,34 @@ def get_resource_allocation_info(self):
...
+                            '{input_size}' in part):
+                        variable, value = part.split('=')
+                        error_msg = ('Obvious incorrect allocation. Please '
+                                     'contact qiita.help@gmail.com')
+                        if (('{samples}' in value and samples is None) or
+                                ('{columns}' in value and columns is None) or
+                                ('{input_size}' in value and input_size is
+                                 None)):
+                            self._set_error(error_msg)
+                            return 'Not valid'
```
ElDeveloper, line 510: Would you mind elaborating a bit on what is happening here, I don't think I can follow. Adding some comments might help.

qiita_db/processing_job.py
antgonza: by default is done by space ...

qiita_db/processing_job.py
antgonza: Sure  ...

qiita_db/test/test_processing_job.py
antgonza: Yup, they are empty ...

qiita_db/processing_job.py
ElDeveloper: Forgot about that.

qiita_db/processing_job.py
```diff
@@ -488,9 +490,40 @@ def get_resource_allocation_info(self):
...
+                        value = naturalsize(eval(value.format(
+                            samples=samples, columns=columns,
+                            input_size=input_size)), gnu=True, format='%.0f')
+
```
ElDeveloper, line 520: Comments helped a lot, thanks! Should we check that the allocation is always >0? Or for that matter that the result of eval is a positive integer?

antgonza: adding ...

PR #3014, opened 2020-06-29
ENA library strategy

antgonza:
This adds two main things:
- After some discussions with ENA we realized that we were still using the study_type attribute which """isn't used very extensively anymore across INSDC. It is considered 'legacy' metadata and the controlled vocabulary list of valid terms has not be expanded in some time.""", and we weren't adding the preparation level attribute: `<LIBRARY_STRATEGY>AMPLICON</LIBRARY_STRATEGY>`. Thus removing study_type and adding library strategy. 
- Turns out that is `Ion_Torrent` and not `Ion Torrent`; notice the underscore vs space ...


qiita_ware/ebi.py
```diff
@@ -502,6 +489,11 @@ def generate_experiment_xml(self, samples=None):
 
         samples = samples if samples is not None else self.samples.keys()
 
+        if self.investigation_type == 'Other':
+            library_strategy = self.new_investigation_type
```
ElDeveloper, line 493: Can we add a test that shows this investigation type in one of the XMLs?

qiita_ware/test/test_ebi.py
```diff
@@ -527,6 +530,25 @@ def test_generate_experiment_xml(self):
         obs = ET.tostring(submission.generate_experiment_xml())
         self.assertEqual(obs.decode('ascii'), exp)
 
+        # changing investigation_type to tests user defined terms, first let's
+        # create a new term
+        new_term = 'Ultimate Term'
```
ElDeveloper, line 535: 👍

qiita_ware/test/test_ebi.py
```diff
@@ -527,6 +530,25 @@ def test_generate_experiment_xml(self):
         obs = ET.tostring(submission.generate_experiment_xml())
         self.assertEqual(obs.decode('ascii'), exp)
 
+        # changing investigation_type to tests user defined terms, first let's
```
ElDeveloper, line 533:
```suggestion
        # changing investigation_type to test user defined terms, first let's
```

PR #3019, opened 2020-07-07
adding GUI access to users for software

antgonza:
This allows users to access the list of available software/plugins and their details:
Screen Shot 2020-07-07 at 9 19 21 AM


ElDeveloper:
Thanks @antgonza

PR #3020, opened 2020-07-07
Fixing minor issues

antgonza:


qiita_db/processing_job.py
```diff
@@ -1007,75 +1007,74 @@ def release(self):
 
     def release_validators(self):
         """Allows all the validator job spawned by this job to complete"""
-        with qdb.sql_connection.TRN:
```
antgonza, line 1010: The main change here is moving the `with qdb.sql_connection.TRN:` to line 1074 as there is the only place actually needed. 

This change was needed as the updates to the self.step (line 1030) was not being reflected in the GUI until everything was done due to transaction encapsulation - AKA useless.

qiita_ware/private_plugin.py
```diff
@@ -75,10 +75,9 @@ def release_validators(job):
     job : qiita_db.processing_job.ProcessingJob
         The processing job with the information of the parent job
     """
-    with qdb.sql_connection.TRN:
```
antgonza, line 78: Same than above: rm `with qdb.sql_connection.TRN`.

antgonza:
Note that this passed tests but the PR wasn't updated, not sure why ...

qiita_pet/static/js/networkVue.js
```diff
@@ -349,10 +349,11 @@ Vue.component('processing-graph', {
         vm.updateGraph();
       })
         .fail(function(object, status, error_msg) {
-          $('#run-btn').attr('disabled', false);
```
ElDeveloper, line 352: Shouldn't this be returned only if the submission fails?

antgonza: Nop, it should be reenabled either way as the button is reused on the same page to submit multiple times. In fact, that was the issue: if a user added and ran a workflow the button stayed disabled and with the "Submitting!" text.

ElDeveloper: Got it, thanks!

ElDeveloper:
Looks like github is not updating but the build finished in travis.

PR #3023, opened 2020-08-03
inject study_type in EBI and improvements to current automatic processing pipeline

antgonza:
…

### DIFF
--- a/qiita_ware/ebi.py
+++ b/qiita_ware/ebi.py
@@ -356,6 +356,11 @@ class EBISubmission(object):
         study_title = ET.SubElement(descriptor, 'STUDY_TITLE')
         study_title.text = escape(clean_whitespace(self.study_title))
 
+        # study type is depricated and not displayed anywhere on EBI-ENA;
+        # however it's required for submission so just injecting with Other
+        ET.SubElement(
+            descriptor, 'STUDY_TYPE', {'existing_study_type': 'Other'})
+
         study_abstract = ET.SubElement(descriptor, 'STUDY_ABSTRACT')
         study_abstract.text = clean_whitespace(escape(self.study_abstract))
 
--- a/qiita_ware/ebi.py
+++ b/qiita_ware/ebi.py
@@ -356,7 +356,7 @@ class EBISubmission(object):
         study_title = ET.SubElement(descriptor, 'STUDY_TITLE')
         study_title.text = escape(clean_whitespace(self.study_title))
 
-        # study type is depricated and not displayed anywhere on EBI-ENA;
+        # study type is deprecated and not displayed anywhere on EBI-ENA;
         # however it's required for submission so just injecting with Other
         ET.SubElement(
             descriptor, 'STUDY_TYPE', {'existing_study_type': 'Other'})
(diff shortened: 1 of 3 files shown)


qiita_ware/ebi.py
```diff
@@ -356,6 +356,11 @@ def generate_study_xml(self):
         study_title = ET.SubElement(descriptor, 'STUDY_TITLE')
         study_title.text = escape(clean_whitespace(self.study_title))
 
+        # study type is depricated and not displayed anywhere on EBI-ENA;
```
ElDeveloper, line 359:
```suggestion
        # study type is deprecated and not displayed anywhere on EBI-ENA;
```

scripts/qiita-auto-processing
```diff
@@ -161,7 +179,9 @@ for pipeline in full_pipelines:
                      if c['previous-step'] == commands[0]['command-name']]
 
     # Step 3. Find all preparations/artifacts that we can add the pipeline
-    artifacts_all = [a for study in Study.iter()
+    #         ... as a first pass we will only process study 10317 (AGP) ...
+    # artifacts_all = [a for study in Study.iter()
+    artifacts_all = [a for study in [Study(10317)]
```
ElDeveloper, line 184: Is this OK to use for deploy in the main system?

scripts/qiita-auto-processing
```diff
@@ -194,23 +217,33 @@ for pipeline in full_pipelines:
     #          of Step 4 but for debugging it makes sense to separate
     artifact_process = []
     children_compliant = []
+    cmd = commands[0]
     for a in artifacts_compliant:
-        cmd = commands[0]
         # getting all jobs, includen hiddens, in case the job failed
         jobs = a.jobs(cmd=cmd['command'], show_hidden=True)
-        params = [j.parameters.values for j in jobs]
+        params = [{k: str(v) for k, v in j.parameters.values.items()
```
ElDeveloper, line 224: Why is the string conversion needed here?

scripts/qiita-auto-processing
```diff
@@ -194,23 +217,33 @@ for pipeline in full_pipelines:
...
             else:
                 for c in a.children:
-                    if c.processing_parameters.values == p:
-                        children_compliant.append(c)
+                    cpp = c.processing_parameters
+                    if cpp.command.name == cmd['command-name']:
+                        cparams = {
+                            k: str(v)
+                            for k, v in cpp.values.items()
+                            if k not in cmd['ignore_parameters']}
```
ElDeveloper, line 244: This list comprehension pattern is repeated a couple times, any chance this can be put in a single function, or something like that?

scripts/qiita-auto-processing
antgonza: yup, the idea is that in the main deploy we will _only_ process AGP automatically; once we test multiple times we can add to full qiita.

scripts/qiita-auto-processing
antgonza: Excellent question; for some reason (for old jobs) the parameters are not string and the parameters here are defined as strings; the easiest is to make sure that everything is string

scripts/qiita-auto-processing
ElDeveloper: Thanks!

ElDeveloper:
Thanks @antgonza

PR #3026, opened 2020-08-26
some general fixes/additions for next release

antgonza:


qiita_db/test/test_processing_job.py
```diff
@@ -457,13 +461,18 @@ def test_complete_success(self):
                                             'artifact_type': 'BIOM'}}
         job = _create_job()
         job._set_status('running')
+
+        # here we can test that job.release_validator_job hasn't been created
+        # yet so it has to be None
+        self.assertIsNone(job.release_validator_job)
```
ElDeveloper, line 467: Can you add another test for the case where the return value is not None?

antgonza: Ha! I knew I was forgetting something ... thank you!

ElDeveloper:
Great, thanks @antgonza!

PR #3028, opened 2020-09-03
WIP: fix #2868

### LINKED ISSUE #2868: Make tests more DRY
There are tests, other code in qiita_db/test/test_meta_util.py, etc. that are of the form:

With TRN:
    sql = <something>
    results = execute(sql)
    return results

or similar. This could be wrapped in a util function that reduces the amount of code repeats.

antgonza:


antgonza:
This is ready for review. There are lots of changes but there is only one main change [here](https://github.com/biocore/qiita/pull/3028/files#diff-29a37d36652014d33c798df45a644b46R496), which is then propagated to all the rest of the changes to make the tests more DRY.

qiita_db/test/test_meta_util.py
```diff
@@ -467,10 +461,8 @@ def test_generate_biom_and_metadata_release(self):
         self.assertEqual(txt_obs, txt_exp)
 
         # returning configuration
-        with qdb.sql_connection.TRN:
-            qdb.sql_connection.TRN.add(
-                "UPDATE settings SET base_data_dir = '%s'" % obdr)
-            bdr = qdb.sql_connection.TRN.execute()
+        qdb.sql_connection.encapsulated_query(
+            "UPDATE settings SET base_data_dir = '%s'" % obdr)
```
ElDeveloper, line 465: Shouldn't obdr be an argument to encapsulated_query?

qiita_db/test/test_meta_util.py
```diff
@@ -34,17 +34,13 @@ def tearDown(self):
 
     def _set_artifact_private(self):
         id_status = qdb.util.convert_to_id('private', 'visibility')
-        with qdb.sql_connection.TRN:
-            qdb.sql_connection.TRN.add(
-                "UPDATE qiita.artifact SET visibility_id = %d" % id_status)
-            qdb.sql_connection.TRN.execute()
+        qdb.sql_connection.encapsulated_query(
+            "UPDATE qiita.artifact SET visibility_id = %d" % id_status)
```
ElDeveloper, line 38: shouldn't id_status be an argument to encapsulated_query?

... there's a few other instances here. If I'm missing something then feel free to disregard these messages.

qiita_db/test/test_study.py
```diff
@@ -204,10 +204,8 @@ def _change_processed_data_status(self, new_status):
         # Change the status of the studies by changing the status of their
         # artifacts
         id_status = qdb.util.convert_to_id(new_status, 'visibility')
-        with qdb.sql_connection.TRN:
-            qdb.sql_connection.TRN.add(
-                "UPDATE qiita.artifact SET visibility_id = %s", (id_status,))
-            qdb.sql_connection.TRN.execute()
+        qdb.sql_connection.encapsulated_query(
+            "UPDATE qiita.artifact SET visibility_id = %s", (id_status,))
```
ElDeveloper, line 208: Classic one-element tuple, maybe worthwhile converting it into a list? 

```suggestion
            "UPDATE qiita.artifact SET visibility_id = %s", [id_status])
```

Should be fine to disregard of course, I suggest using a list since that's what's mostly used in this PR.

qiita_db/sql_connection.py
```diff
@@ -493,6 +493,24 @@ def add_post_rollback_func(self, func, *args, **kwargs):
 TRNADMIN = Transaction(admin=True)
 
 
+def encapsulated_query(sql, parameters=None):
```
ElDeveloper, line 496: Name is fine but I think it would be nice to make it clear that queries are executed as transactions? Up to your discretion. What about `execute_as_transaction`.

ElDeveloper (approved):
Looks good, just a couple of questions.

qiita_db/sql_connection.py
antgonza: Yeah, selected `single_query` first, then changed it to `encapsulted_query` but still not happy. `execute_as_transaction` is a decorator we hace; thus, gonna change to `perform_as_transaction`.

qiita_db/test/test_meta_util.py
antgonza: Yeah, we normally use the parameters as arguments when adding a parameter to have safe data type conversions in the query; however, if a developer is sure about the data type you can do this. Note that this is done based on the developer and didn't change any of the code to keep it simple (and cause it was out of the scope of this issue/PR).

antgonza (commented):
Thank you @ElDeveloper

qiita_db/test/test_meta_util.py
ElDeveloper: Alright, that sounds fair enough.

qiita_db/sql_connection.py
ElDeveloper: 👍

ElDeveloper:
Thanks @antgonza would you mind fixing the conflicts?

antgonza:
Done, didn't see them before ... thank you!

ElDeveloper:
Thanks!

PR #3040, opened 2020-11-05
adding job submit ENVIRONMENT and minor improvement to runWorkflow

antgonza:
adding job submit ENVIRONMENT and minor improvement to runWorkflow


qiita_db/processing_job.py
```diff
@@ -880,12 +880,30 @@ def submit(self, parent_job_id=None, dependent_jobs_list=None):
...
+                self._set_status('running')
+                qdb.sql_connection.TRN.commit()
+
+            create_nested_path(job_dir)
+            cmd = (f'{plugin_env_script}; {plugin_start_script} '
+                   f'{url} {self.id} {job_dir}')
+            stdout, stderr, return_value = _system_call(cmd)
+            if return_value != 0 or stderr != '':
+                self._set_error(stderr)
+            job_id = stdout
```
ElDeveloper, line 900: Does this string need to be stripped for new lines or something like that?

ElDeveloper (approved):
Just one question/comment.

qiita_db/processing_job.py
antgonza: Excellent question! IMOO no. Basically, the idea is to display whatever information the process creating the job returns.

ElDeveloper: Got it, should be used to display to the end-user not to decode a return value. Makes sense.

PR #3041, opened 2020-11-05
Fix 3036

antgonza:
As reference this is how it looks now when an analysis creation fails:
failure

and when it succeeds:
success


ElDeveloper:
Thanks @antgonza

Issue #2920, opened 2019-07-12
improve zip/archiving of download files


antgonza:
A user reported seeing CRC unzip values errors/warnings; for example: "extracting: BIOM/57457/all.biom      bad CRC f6b2a86b  (should be 38903659)".

Background: one of the largest changes from last release was that now we are storing the file size and it's CRC32 checksum in the database; this checksum is added to mod_zip/nginx so it can be used for file download resume. To be clear, the CRC of the file (f6b2a86b) is the value stored in the DB, displayed in the Qiita pages; and the value you get in a laptop once you download the file via the crc32 command AKA this value is the correct one. Also note, the downloaded zip contains the correct files; even if zip complains, the files are _fine_. Thus, there is something "odd" with mod_zip/nginx.

After further investigation @ElDeveloper  and I realized that:
- mod_zip doesn't actually compress the files (it creates a zip file without compression) and it's part of their design: https://github.com/evanmiller/mod_zip/issues/23
- the checksums added to the zip file are incorrect

PR #3042, opened 2020-11-05
Fix 2920

antgonza:
This basically stops adding the CRC information while downloading files, which originally was added to allow download resumes, but turns out that it doesn't work (see #2920) and only creates problems with the downloads, as described here: https://qiita.ucsd.edu/static/doc/html/faq.html#how-to-solve-download-or-unzip-errors. Thus, this should remove those problems. 

Note that the this should at least minimize those questions.


ElDeveloper (approved):
This looks good, but wouldn't this still throw an error because whatever checksum is checked will be different from `'-'`? Otherwise looks good.

antgonza:
No checksum is denoted by '-', [from the tutorial](https://github.com/evanmiller/mod_zip#usage): 
The CRC-32 is optional. Put "-" if you don't know the CRC-32; note that in this case mod_zip will disable support for the Range header.

ElDeveloper:
Thanks for the explanation!

PR #3043, opened 2020-11-06
rm create qiime mapping file

antgonza:


qiita_db/handlers/prep_template.py
```diff
@@ -66,21 +66,25 @@ def get(self, prep_id):
...
             response = {
                 'data_type': pt.data_type(),
                 'artifact': artifact,
                 'investigation_type': pt.investigation_type,
-                'study': pt.study_id,
+                'study': sid,
                 'status': pt.status,
-                'qiime-map': pt.qiime_map_fp,
+                # get_filepaths returns an ordered list of [filepath_id,
+                # filepath] and we want the latests
```
ElDeveloper, line 85:
```suggestion
                # filepath] and we want the last pair
```

qiita_db/analysis.py
```diff
@@ -1040,13 +1040,16 @@ def _build_mapping_file(self, samples, rename_dup_samples=False):
         with qdb.sql_connection.TRN:
             all_ids = set()
             to_concat = []
+            sample_infos = dict()
```
ElDeveloper, line 1043: Unrelated to this line, but would it be a good time to rename this method to `_build_sample_metadata`?

qiita_pet/handlers/download.py
```diff
@@ -119,13 +119,15 @@ def _list_artifact_files_nginx(self, artifact):
                     (x['fp'], x['fp'], '-', str(x['fp_size'])))
 
         for pt in artifact.prep_templates:
-            qmf = pt.qiime_map_fp
-            if qmf is not None:
-                sqmf = qmf
-                if qmf.startswith(basedir):
-                    sqmf = qmf[basedir_len:]
+            # the latests prep template file is always the first [1] tuple and
```
ElDeveloper, line 122:
```suggestion
            # the latest prep template file is always the first [0] tuple and
```

qiita_pet/templates/study_ajax/prep_summary.html
```diff
@@ -429,7 +429,9 @@ <h4>
       <span id="prep-name-span">{{name}}</span> - ID {{prep_id}} ({{data_type}})
       <a class="btn btn-default" data-toggle="modal" data-target="#update-prep-name"><span class="glyphicon glyphicon-pencil"></span> Edit name</a>
       <a class="btn btn-default" href="{% raw qiita_config.portal_dir %}/download/{{download_prep_id}}"><span class="glyphicon glyphicon-download-alt"></span> Prep info</a>
-      <a class="btn btn-default" href="{% raw qiita_config.portal_dir %}/download/{{download_qiime_id}}"><span class="glyphicon glyphicon-download-alt"></span> QIIME map</a>
+      {% if download_qiime_id is not None %}
```
ElDeveloper, line 432: This is for backwards compatibility, right? This means that unless a QIIME mapping file has already been created then we won't show this button.

ElDeveloper (changes requested):
Looks good. One question, should this include changes to the documentation? Reflecting these changes?

qiita_db/analysis.py
antgonza: Well, in practice it is still merging sample/prep info files so building mapping files ...

qiita_pet/templates/study_ajax/prep_summary.html
antgonza: That's 100% correct!

antgonza:
I'm planning to add to the change log; AFAIK there is no information about the qiime-map in any of the documentation.

ElDeveloper:
Thanks @antgonza!

PR #3046, opened 2020-11-11
Study autoloaded

antgonza:
Changes to support a study flag to mark as automatically loaded. In general, it should always be false, once we support autoloading we will use this flag.


adswafford (approved):
Not familiar with the code for Qiita, but from what I can tell it looks like everything is set as expected for adding the autoloaded flag for setting and getting later.

qiita_db/study.py
```diff
@@ -481,6 +483,34 @@ def insert_tags(cls, user, tags):
...
 
 # --- Attributes ---
+    @property
+    def autoloaded(self):
+        """Returns if the study was autoloaded
+
+        Returns
+        -------
+        bool
+            If the study was loaded or not
```
ElDeveloper, line 493:
```suggestion
            If the study was autoloaded or not
```

qiita_db/study.py
```diff
@@ -481,6 +483,34 @@ def insert_tags(cls, user, tags):
...
+        """
+        with qdb.sql_connection.TRN:
+            sql = """SELECT autoloaded FROM qiita.{0}
+                     WHERE study_id = %s""".format(self._table)
+            qdb.sql_connection.TRN.add(sql, [self._id])
+            return qdb.sql_connection.TRN.execute_fetchlast()
+
+    @autoloaded.setter
+    def autoloaded(self, value):
+        """Sets the title of the study
```
ElDeveloper, line 503:
```suggestion
        """Sets the autoloaded status of the study
```

qiita_db/study.py
```diff
@@ -481,6 +483,34 @@ def insert_tags(cls, user, tags):
...
+
+        Parameters
+        ----------
+        value : bool
+            Value to set the autoloaded of this study
```
ElDeveloper, line 508:
```suggestion
            Whether the study was autoloaded
```

ElDeveloper (changes requested):
Just a few small changes. Quick question, is this attribute going to be visible as a metadata column?

antgonza:
Thank you for review; that's a good question! I don't think we should display it but it will be good to limit functionality for those studies we didn't submit ...

ElDeveloper:
Yes, I think I was thinking mostly in terms of a meta-analysis to easily check for systematic effects.

PR #3056, opened 2020-12-02
rm key

antgonza:


qiita_ware/commands.py
```diff
@@ -120,6 +120,11 @@ def list_remote(URL, private_key):
     ssh = _ssh_session(p_url, private_key)
     valid_files = _list_valid_files(ssh, directory)
     ssh.close()
+
+    # for security, remove key
+    if exists(private_key):
```
wasade, line 125: If the key is removed for a file listing, does this mean the user needs to upload the key each time?

ElDeveloper (approved):
The last commit still needs to run in Travis.

PR #3061, opened 2021-01-04
Fix 3057

### LINKED ISSUE #3057: Add a download option for the sample/preparation summary table
Add a button in this page: Screen Shot 2020-12-07 at 9 21 20 AM
to download that table in tab separated format. Basically, this will allow for a quick review of what samples are in which preparations.

antgonza:
1. Fixes #3057 by simply generating a tsv file from the already loaded data; the new button looks like this:
sample_prep_summary

2. It also adds the possibility of downloading the sample information file for a given prep (AKA download the sample information for only those samples in the prep); the new button looks like this:
samples_from_prep

### DIFF
--- a/qiita_db/metadata_template/base_metadata_template.py
+++ b/qiita_db/metadata_template/base_metadata_template.py
@@ -1157,22 +1157,30 @@ class MetadataTemplate(qdb.base.QiitaObject):
             df.to_csv(fp, index_label='sample_name', na_rep="", sep='\t',
                       encoding='utf-8')
 
-    def _common_to_dataframe_steps(self):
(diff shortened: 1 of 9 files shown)


ElDeveloper:
This looks good @antgonza, thanks. My only suggestion would be to change the text in `Sample Info` to emphasize that this only includes the samples in the current prep. Something like `Sample info (in this prep)`, not sure how to make this concise and clear. Another option would be a small warning or even an alert? 

Other than that it looks like there's conflicts in this branch.

ElDeveloper:
nvm about the conflicts, I just fixed those :+1:

PR #3062, opened 2021-01-04
fix #2751

### LINKED ISSUE #2751: Tentative improvements to sample/prep info files
During the jsonb conversion we noticed a couple of things that could be nice to improve in the info files:
- [x] make sample/prep categories() a property 
- [x] there is an "interesting" combination/duplication-of-code of Sample (one at the time) and SampleTemplate (multiple); for example categories and _get_categories

antgonza:
After review, the only code that is kind of duplicated is the code between BaseSample._get_categories and MetadataTemplate.categories; which we are merging in a single helper function in this PR + made MetadataTemplate.categories a property.


ElDeveloper:
Thanks @antgonza

PR #3066, opened 2021-01-25
adding creation and modification timestamps to prep_templates

antgonza:


qiita_db/environment_manager.py
```diff
@@ -431,6 +431,7 @@ def patch(patches_dir=PATCHES_DIR, verbose=False, test=False):
         # for the test Study (1) so a lot of the tests actually expect this.
         # Now, trying to regenerate directly in the populate_test_db might
         # require too many dev hours so the easiest is just do it here
-        # UPDATE 02/27/19: moving to 74.sql as we added the file sizes
-        if test and sql_patch_filename == '74.sql':
+        # UPDATE 01/25/2021: moving to 81.sql as we timestamps to prep
+        #                    info files
```
ElDeveloper, line 435:
```suggestion
        # UPDATE 01/25/2021: moving to 81.sql as we added timestamps to
        #                    prep info files
```

qiita_db/metadata_template/prep_template.py
```diff
@@ -507,11 +506,16 @@ def generate_files(self, samples=None, columns=None):
...
             _id, fp = qdb.util.get_mountpoint('templates')[0]
+            # update timestamp in the DB first
+            qdb.sql_connection.TRN.add(
+                """UPDATE qiita.prep_template
+                   SET modification_timestamp = CURRENT_TIMESTAMP
+                   WHERE prep_template_id = %s""", [self._id])
+            ctime = self.modification_timestamp
             fp = join(fp, '%d_prep_%d_%s.txt' % (self.study_id, self._id,
-                      strftime("%Y%m%d-%H%M%S")))
+                      ctime.strftime("%Y%m%d-%H%M%S")))
```
ElDeveloper, line 516: Should we query the `modification_timestamp` value so there's always a 1:1 match between the filenames and the database entries? Otherwise it seems like those two might be out of sync.

qiita_db/metadata_template/prep_template.py
```diff
@@ -666,3 +670,35 @@ def to_dataframe(self, add_ebi_accessions=False):
                 lambda sid: accessions[sid])
 
         return df
+
+    @property
+    def creation_timestamp(self):
```
ElDeveloper, line 675: Isn't there a missing call to `creation_timestamp` where the object itself is created? And if so, should this property have a setter?

ElDeveloper (changes requested):
Looks good just a question and a couple of comments.

qiita_db/metadata_template/prep_template.py
antgonza: That's right and that's what is doing AFAIK. We first insert the current time in the DB, then we retrieve the modification_timestamp (`ctime = self.modification_timestamp`) and we use ctime as the timestamp for the filename.

qiita_db/metadata_template/prep_template.py
antgonza: Good question, as the default in the database is CURRENT_TIME, when the prep is inserted for the first time (creation) it will automatically add the CURRENT_TIMESTAMP so no need to actually update it (setter).

ElDeveloper: Oh! Gotcha, thanks for clarifying. 😮

qiita_db/metadata_template/prep_template.py
ElDeveloper: Apologies, reading ctime made me think of the `ctime` function - my bad!

ElDeveloper (approved):

ElDeveloper:
Thanks @antgonza

PR #3067, opened 2021-01-26
minor improvements

antgonza:
Two minor improvements:
1. Improve performance of `_common_to_dataframe_steps` ~17 seconds for AGP:
```
In [50]: %timeit SampleTemplate(10317).to_dataframe()                                                                                                                                                                                                                                       
1min 8s ± 224 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

In [51]: %timeit ST(10317).to_dataframe()                                                                                                                                                                                                                                                   
51.8 s ± 259 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```
2. Turns out that jobs actually append the node where the job was submitted so that list comprehension will miss jobs submitted in the workers.

### DIFF
--- a/qiita_db/metadata_template/base_metadata_template.py
+++ b/qiita_db/metadata_template/base_metadata_template.py
@@ -1172,7 +1172,6 @@ class MetadataTemplate(qdb.base.QiitaObject):
         """
         with qdb.sql_connection.TRN:
             # Retrieve all the information from the database
-            cols = self.categories
             sql = """SELECT sample_id, sample_values
                      FROM qiita.{0}
                      WHERE sample_id != '{1}'""".format(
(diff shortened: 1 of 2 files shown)


qiita_db/metadata_template/base_metadata_template.py
```diff
@@ -1183,34 +1182,11 @@ def _common_to_dataframe_steps(self, samples=None):
...
-                # append the row to the full matrix
-                data.append(vals)
-            cols.insert(0, 'sample_id')
-            df = pd.DataFrame(data, columns=cols, dtype=str)
-            df.set_index('sample_id', inplace=True)
-
-            # Make sure that we are changing np.NaN by Nones
+            df = pd.DataFrame(qdb.sql_connection.TRN.execute_fetchindex())
+            df = pd.concat([df.drop(1, axis=1),
+                            pd.DataFrame(df[1].tolist())], axis=1)
```
wasade, line 1187: Would it be possible to do `pd.concat([df.drop(1, axis=1), df[1]])`? It not really obvious why the new `DataFrame` construction is needed or the `tolist`. Oh wait, is this supposed to be a reorder operation moving the first column to the end of the frame? If so, then i think this does it:

```python
cols = [df.columns[0]] + df.columns[2:].tolist() + [df.columns[1]]
df = df[cols]
```

antgonza: Good question. The metadata tables are basically 2 columns: sample_id, metadata (json, which is returned as a dict). In other words, the SQL will return [[sample_id, {'col_a': 'a', 'col_b': 'b'}, ...], [sample_id, {'col_a': 'a', 'col_b': 'b'}, ...], ...] so that line basically expands the dict into actual columns in the dataframe, like this:
```python
In [61]: data = ['x', {'a': 1, 'b': 2}], ['y', {'a': 1, 'c': 2}]                                                                                                                                                                                                                            

In [62]: df = pd.DataFrame(data)                                                                                                                                                                                                                                                            

In [63]: df                                                                                                                                                                                                                                                                                 
Out[63]: 
   0                 1
0  x  {'a': 1, 'b': 2}
1  y  {'a': 1, 'c': 2}

In [64]: pd.concat([df.drop(1, axis=1), pd.DataFrame(df[1].tolist())], axis=1)                                                                                                                                                                                                              
Out[64]: 
   0  a    b    c
0  x  1  2.0  NaN
1  y  1  NaN  2.0
```

wasade: Okay, thanks. What about this? It's a little less juggling and you get the index "for free"? 

```python
>>> import pandas as pd
>>> data = ['x', {'a': 1, 'b': 2}], ['y', {'a': 1, 'c': 2}]
>>> pd.DataFrame([d for _, d in data], index=[i for i, _ in data])
   a    b    c
x  1  2.0  NaN
y  1  NaN  2.0
```

antgonza: Thanks, this made the code a bit faster too. If you agree with the changes, and the tests pass, could you merge?

ElDeveloper:
These changes look good. @antgonza I'm not familiar with the pgsql json side of things but, this seems like the kind of thing that pgsql could probably do much faster than Pandas/Python. Did you happen to look into querying the data directly into a table form? Rather than querying a JSON string, and then converting it to a table?

I'm looking into it now, but just want to make sure I shared this observation.

ElDeveloper:
Following up on this @antgonza shared some benchmarks showing that the pgsql implementation was in fact slower than the changes in this PR.

ElDeveloper:
Thanks @wasade and @antgonza

PR #3080, opened 2021-02-25
fix-3064

antgonza:


.github/workflows/qiita-ci.yml
```diff
@@ -0,0 +1,207 @@
...
+          export REDBIOM_HOST="http://localhost:7379"
+
+          echo "1. Setting up redis"
+          redis-server --daemonize yes --port 7777
+          redis-server --daemonize yes --port 6379
+
+          echo "2. Starting webdis"
+          pushd webdis
+          ./webdis &
+          popd
```
ElDeveloper, line 128:
```suggestion
          $PWD/webdis/webdis &
```

.github/workflows/qiita-ci.yml
```diff
@@ -0,0 +1,207 @@
...
+
+      - name: Starting services
+        shell: bash -l {0}
+        run: |
+          conda activate qiita
+          export QIITA_SERVER_CERT=`pwd`/qiita_core/support_files/server.crt
+          export QIITA_CONFIG_FP=`pwd`/qiita_core/support_files/config_test.cfg
+          export REDBIOM_HOST="http://localhost:7379"
+
+          echo "1. Setting up redis"
```
ElDeveloper, line 121: Instead of using echo, another alternative is to put each of these sections under a `-name: Setting up redis` tag so that the UI summarizes which step failed, etc.

.github/workflows/qiita-ci.yml
```diff
@@ -0,0 +1,207 @@
...
+          kill $QIITA_PID
+          sleep 10
+          if [[ "$COVER_PACKAGE" == *"qiita_db"* ]]; then test_data_studies/commands.sh; all-qiita-cron-job; fi
+
+      - uses: codecov/codecov-action@v1
+        with:
+         file: codecov.yml
+         fail_ci_if_error: true
+
+  lint:
```
ElDeveloper, line 195: Up to you but maybe the `main` job could depend on the `lint` job so that the full test suite isn't run before you find there's a syntax error or something like that. That would also mean that the tests aren't run if there are linting errors which might be undesirable.

.github/workflows/qiita-ci.yml
```diff
@@ -0,0 +1,207 @@
...
+    - uses: codecov/codecov-action@v1
+    - name: flake8
+      uses: actions/setup-python@v2
+      with:
+        python-version: 3.6
+    - name: install dependencies
+      run: python -m pip install --upgrade pip
+    - name: lint
+      run: |
+        pip install -q flake8
```
ElDeveloper, line 207: Isn't this missing the command that lints the code?

ElDeveloper (changes requested):
Looks great, @antgonza just a few small suggestions.

.github/workflows/qiita-ci.yml
```diff
@@ -0,0 +1,207 @@
...
+
+      - name: Setup for conda
+        uses: conda-incubator/setup-miniconda@v2
+        with:
+          auto-update-conda: true
+          python-version: 3.6
+
+      - name: Basic dependencies install
+        env:
+          COVER_PACKAGE: ${{ matrix.COVER_PACKAGE }}
```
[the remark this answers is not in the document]
antgonza: Sure, note that it works as expected so my guess is case insensitive

.github/workflows/qiita-ci.yml
antgonza: I'm OK with echos! :)

.github/workflows/qiita-ci.yml
antgonza: In my own experience the webdis expects and looks for files in the folder; if you run from outside the folder the expected files are not where they should be

ElDeveloper:
Looks great, thanks @antgonza!

PR #3085, opened 2021-03-16
add Command.processing_jobs

antgonza:
It has become more common to search/get all the jobs generated by a command; this method allows for easy access to this information.

### DIFF
--- a/qiita_db/software.py
+++ b/qiita_db/software.py
@@ -761,6 +761,7 @@ class Command(qdb.base.QiitaObject):
         Returns
         -------
         list of qiita_db.processing_job.ProcessingJob
+            List of jobs that used this command.
         """
 
         with qdb.sql_connection.TRN:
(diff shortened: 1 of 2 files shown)


qiita_db/software.py
```diff
@@ -754,6 +754,25 @@ def resource_allocation(self):
 
         return result[0]
 
+    @property
+    def processing_jobs(self):
+        """All the processing_jobs that used this command
+
+        Returns
+        -------
+        list of qiita_db.processing_job.ProcessingJob
```
ElDeveloper, line 763:
```suggestion
        list of qiita_db.processing_job.ProcessingJob
            List of jobs that used this command.
```

ElDeveloper (approved):
Looks in good do you think jobs should be sorted by date? Other than that, just one small comment.

antgonza:
Thank you; I don't think is worth sorting by time, and to be fair, neither by something else ...

ElDeveloper:
Thanks @antgonza

PR #3092, opened 2021-04-13
vis-network-9.0.4

antgonza:
Updating vis-network to latests version 9.0.4 and a minor error reporting modification that was helpful while creating https://github.com/qiita-spots/qp-spades

PR #3093, opened 2021-04-16
supersede #3092

antgonza:
Extra from the changes in #3092, adding is `_count_jobs_in_scheduler`: previously we were counting "only" the jobs submitted by qiita, the new code counts all jobs in the scheduler.


ElDeveloper:
Looks good, thanks @antgonza.

PR #3094, opened 2021-04-27
fix #3079

### LINKED ISSUE #3079: Improve processing graph while adding raw data
The following pointers were created by @jdereus:
* If it says click on the graph, there should be a graph to click on.
* if there is a countdown timer to refreshing the graph, it is even more confusing when there is nothing there.
* if it says that a job is running, that job should actually be running.   Not something else that was spawned.  While most people won't have access to the cluster to actually look at the job, its just not accurate.

antgonza:
This together with the new libraries for network display (will issue a PR soon) fix #3079.

### DIFF
--- a/qiita_ware/private_plugin.py
+++ b/qiita_ware/private_plugin.py
@@ -348,7 +348,9 @@ def complete_job(job):
             artifacts = None
             error = payload['error']
         c_job = qdb.processing_job.ProcessingJob(param_vals['job_id'])
+        c_job.step = 'Completing via %s [%s]' % (job.id, job.external_id)
         try:
+
             c_job.complete(payload['success'], artifacts, error)
         except Exception:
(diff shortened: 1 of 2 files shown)


qiita_ware/private_plugin.py
```diff
@@ -348,7 +348,9 @@ def complete_job(job):
             artifacts = None
             error = payload['error']
         c_job = qdb.processing_job.ProcessingJob(param_vals['job_id'])
+        c_job.step = 'Completing via %s [%s]' % (job.id, job.external_id)
         try:
+
```
wasade, line 353:
```suggestion
```

wasade (approved):

wasade:
Good to merge?

antgonza:
yes, please -- not sure if you get a message with the 👍🏽

PR #3096, opened 2021-04-28
New network

antgonza:
There are a lot of libraries/files being updated and moved that _do not_ need review; reviewers should focus on:
- qiita_pet/static/js/networkVue.js
- qiita_pet/static/vendor/README.md
- qiita_pet/templates/sitebase.html
- qiita_pet/templates/workflows.html


qiita_pet/static/js/networkVue.js
```diff
@@ -742,79 +705,75 @@ Vue.component('processing-graph', {
...
+        zoomInIcon: 'fa fa-plus',
+        zoomOutIcon: 'fa fa-minus',
+        resetIcon: 'fa fa-expand'
       };
 
-      vm.network = new vis.Network(container, data, options);
-      vm.network.on("stabilized", function (params) {
+      // Note: we only need to sort the edges to keep the same structure of the
+      //       graph; in other words, nodes order is not important
+      vm.edges = vm.edges.sort((a, b) => (a.data.source > b.data.source) ? 1 : (a.data.source === b.data.source) ? ((a.data.target > b.data.target) ? 1 : -1) : -1 );
```
wasade, line 748: Triple nested ternary is a bit hard... any chance this could be refactored? I think it would be:

```suggestion
      var sort_order = -1;
      if ((a.data.source > b.data.source) || ((a.data.source === b.data.source) && (a.data.target > b.data.target)) {
          sort_order = 1;
      }
      
      vm.edges = vm.edges.sort(sort_order);
```

qiita_pet/static/js/networkVue.js
```diff
@@ -742,79 +705,75 @@ Vue.component('processing-graph', {
...
-        }
-        // [0] cause users can only select 1 node
-        var clickedNode = vm.nodes_ds.get(ids)[0];
-        var element_id = ids[0];
-        if (clickedNode.group == 'artifact') {
+      vm.network.on('tap', 'node', function (evt) {
+        var data = evt.target.data();
+        var element_id = data.id;
+
+        if (data.group == 'artifact') {
```
wasade, line 774: minor i think here, but possible to use `===` to be consistent with the other equalities?

qiita_pet/static/js/networkVue.js
```diff
@@ -742,79 +705,75 @@ Vue.component('processing-graph', {
...
           vm.populateContentArtifact(element_id);
-        } else if (clickedNode.group == 'deleting') {
+        } else if (data.group == 'deleting') {
```
wasade, line 776: same same

qiita_pet/templates/workflows.html
```diff
@@ -55,14 +47,25 @@
...
 }
 
+function makePopper(node) {
+  node.tippy = tippy(node.popperRef(), {
+    content: () => {
+      let content = document.createElement('div');
+      content.innerHTML = node.data('tooltip');
+      return content;
+    },
+    trigger: 'manual' // probably want manual mode
```
wasade, line 57: What would non-manual mode be?

wasade (changes requested):
Just some minor stuff, otherwise :+1:, thanks!!

qiita_pet/templates/workflows.html
antgonza: 😄 😄 😄    that's from the [example for tippy/popper](https://github.com/cytoscape/cytoscape.js-popper#usage-with-tippyjs) and looking closer there are [2 other methods](https://atomiks.github.io/tippyjs/#triggers) - however, not manual but it works ...

qiita_pet/static/js/networkVue.js
antgonza: Sounds good, just note that this is a function like in python `list.sort(key=sorting_function)`.

antgonza (commented):
Thank you @wasade!

qiita_pet/static/js/networkVue.js
wasade: oh right, so pack into a function... but main item was, if okay, to expand the nested ternaries :)

qiita_pet/templates/workflows.html
wasade: ah okay :)

PR #3097, opened 2021-04-28
fix #3089

### LINKED ISSUE #3089: Check and throw GUI error when run_prefix is duplicated in per_sample_FASTQ association

antgonza:
Extra from blocking, I added the possibility of showing a message so users have a better idea of why is blocked.

- Current: before
- This PR; note that all options are not valid but the per_sample_FASTQ shows a message with the reason:

### DIFF
--- a/qiita_pet/templates/study_ajax/artifact_file_selector.html
+++ b/qiita_pet/templates/study_ajax/artifact_file_selector.html
@@ -107,8 +107,8 @@
       // children
 
       if ((count_ok || (!is_required && num_children === 0))) {
-        if ($("#artifact-type").val() === 'per_sample_FASTQ' && num_samples !== num_children) {
-          message = "per_sample_FASTQ expects the same number of samples and run_prefix.";
+        if (($("#artifact-type").val() === 'per_sample_FASTQ') && (num_samples !== num_children)) {
+          message = "per_sample_FASTQ expects one run_prefix per sample. Check for duplicate or missing run_prefix entries.";
           mark_as_wrong(this, message);
         } else {
           mark_as_correct(this);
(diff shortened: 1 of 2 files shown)


qiita_pet/templates/study_ajax/artifact_file_selector.html
```diff
@@ -96,8 +105,14 @@
       // If the count is correct, then it is always correct. If the count is not
       // correct, it is correct if and only if it is not required and has no
       // children
-      if (count_ok || (!is_required && num_children === 0)) {
-        mark_as_correct(this);
+
+      if ((count_ok || (!is_required && num_children === 0))) {
+        if ($("#artifact-type").val() === 'per_sample_FASTQ' && num_samples !== num_children) {
+          message = "per_sample_FASTQ expects the same number of samples and run_prefix.";
```
adswafford, line 111: Maybe a slightly altered message to be clearer:
per_sample_FASTQ expects one run_prefix per sample. Check for duplicate or missing run_prefix entries.

adswafford (changes requested):
One small text update to help guide users and reduce qiita.help request, otherwise good, so when fixed I'll pop back on to approve.

qiita_pet/templates/study_ajax/artifact_file_selector.html
```diff
@@ -96,8 +105,14 @@
       // If the count is correct, then it is always correct. If the count is not
       // correct, it is correct if and only if it is not required and has no
       // children
-      if (count_ok || (!is_required && num_children === 0)) {
-        mark_as_correct(this);
+
+      if ((count_ok || (!is_required && num_children === 0))) {
+        if ($("#artifact-type").val() === 'per_sample_FASTQ' && num_samples !== num_children) {
```
wasade, line 110:
```suggestion
        if (($("#artifact-type").val() === 'per_sample_FASTQ') && (num_samples !== num_children)) {
```

wasade (approved):

PR #3098, opened 2021-04-28
Fix 3070 retry

antgonza:
Long story short: https://github.com/qiita-spots/qiita/pull/3075/files added the change for programatic connections but not for GUI - in other words, is fine to have the other changes but we also _need_ this.

To make sure that it actually works as expected, I changed a visibility via GUI and checked the warnings:
Screen Shot 2021-04-28 at 10 31 14 AM

### DIFF
--- a/qiita_pet/handlers/artifact_handlers/base_handlers.py
+++ b/qiita_pet/handlers/artifact_handlers/base_handlers.py
@@ -20,6 +20,7 @@ from qiita_db.artifact import Artifact
 from qiita_db.software import Command, Software, Parameters
 from qiita_db.processing_job import ProcessingJob
 from qiita_db.util import get_visibilities, send_email
+from qiita_db.logger import LogEntry
 
 
 PREP_TEMPLATE_KEY_FORMAT = 'prep_template_%s'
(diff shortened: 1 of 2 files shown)


qiita_pet/handlers/artifact_handlers/base_handlers.py
```diff
@@ -370,6 +371,10 @@ def artifact_patch_request(user, artifact_id, req_op, req_path, req_value=None,
                            "directly to <a href='mailto:{0}'>{0}</a>.".format(
                                email_to))
                     raise QiitaHTTPError(400, msg)
+            else:
+                msg = '%s changed artifact %s (study %d) to %s' % (
+                    user.email, artifact_id, sid, req_value)
+                LogEntry.create('Warning', msg)
```
wasade, line 377: Would it be better for the `LogEntry.create` to occur within the `property.set` on `artifact.visibility`? That way, and if I understand this correctly, a log entry would be recorded for any visibility change whereas I think the log entry here would only be recorded if done via the GUI?

wasade (commented):
Just a question, which I could see being out-of-scope...

qiita_pet/handlers/artifact_handlers/base_handlers.py
antgonza: Yes, good question and possible but that will require that [the method](https://github.com/qiita-spots/qiita/blob/master/qiita_db/artifact.py#L754) accept the user as a parameter - which might complicate things a bunch.

wasade: Ya... was worried this would be a complex item, which I think either would require changing that from a property to a method, or allowing the `Artifact` constructor to accept the user so the object knew about about them at the time of the check. Either would be complex and definitely out of scope here.

wasade (approved):

PR #3103, opened 2021-05-06
fix #3102

### LINKED ISSUE #3102: Label WoL/Rep200 Woltka artifacts with prefix or suffix to indicate the database used
Can we automatically append or prepend 'WoL' or 'rep200' to the artifacts produced from Woltka so that it's clear which db was used? I find myself having to frequently relabel them to try to keep track.

antgonza:
This should do the trick.

### DIFF
--- a/qiita_db/processing_job.py
+++ b/qiita_db/processing_job.py
@@ -1284,7 +1284,7 @@ class ProcessingJob(qdb.base.QiitaObject):
                 if naming_params:
                     params = self.parameters.values
                     art_name = "%s %s" % (
-                        out_name, ' '.join([str(params[p])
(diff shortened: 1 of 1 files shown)


adswafford (approved):

antgonza:
Deployed this in qiita-rc and tested, the new artifacts look like this:
Screen Shot 2021-05-07 at 7 41 48 AM

We could also rename all not rename artifacts for woltka (only) to follow this naming convention.

adswafford:
Looks good to me and deploying throughout for improved artifact labelling
would be a big help. Tackling this in the Analysis workflows would also be
nice but likely harder than just the path snip approach here.

antgonza:
Yup, in that case, it will be good to open an issue for the analysis objects.

PR #3108, opened 2021-05-27
fix #3101

### LINKED ISSUE #3101: limit the number of samples in a preparation
Add a new parameter in the database to limit the number of samples allowed in a single preparation. 

The current number of samples per lane within the KL are:
- Generally we run 768 per lane for metagenomics but 1536 per lane may happen soon
- 16S the max we can sequence per run is 800 since we only have 800 unique barcodes

Thus, decided to go with 800.

antgonza:
Fixes #3101 and some minor documentation things.


qiita_db/metadata_template/sample_template.py
```diff
@@ -351,3 +351,7 @@ def to_dataframe(self, add_ebi_accessions=False, samples=None):
                 lambda sid: accessions[sid])
 
         return df
+
+    @classmethod
+    def max_samples(self):
```
wasade, line 356:
```suggestion
    @staticmethod
    def max_samples():
```

qiita_db/metadata_template/prep_template.py
```diff
@@ -702,3 +709,7 @@ def modification_timestamp(self):
                      WHERE prep_template_id = %s"""
             qdb.sql_connection.TRN.add(sql, [self.id])
             return qdb.sql_connection.TRN.execute_fetchlast()
+
+    @classmethod
+    def max_samples(self):
```
wasade, line 714:
```suggestion
    @staticmethod
    def max_samples():
```

qiita_db/support_files/patches/82.sql
```diff
@@ -0,0 +1,13 @@
...
+-- Adding max samples in a single preparation
+-- we need to do it via a DO because IF NOT EXISTS in ALTER TABLE only exists
+-- in PostgreSQL 9.6 or higher and we use 9.5
+DO $do$
+BEGIN
+  IF NOT EXISTS (
+    SELECT DISTINCT table_name FROM information_schema.columns
+        WHERE table_name = 'settings' AND column_name = 'max_preparation_samples'
+     ) THEN
+     ALTER TABLE settings ADD COLUMN max_preparation_samples INT DEFAULT 800;
```
wasade, line 11: Are there existing studies that violate this limit?

wasade (changes requested):

qiita_db/support_files/patches/82.sql
wasade: okay cool thanks!

PR #3110, opened 2021-05-28
fix #3100

### LINKED ISSUE #3100: Limit the options shown in the commands/parameters-sets in the GUI
Originally from: https://github.com/qiita-spots/qiita/pull/3099#issuecomment-829340020.

Ideally, we will not show the parameters that that processing branch has, for example for raw data it will show ""Adaptor and host filtered" with all parameter-sets, then if a user processes that with "Adaptor and host filtered"/"horse", the resulting artifact will show all - "horse" for "Adaptor and host filtered", and so on.

antgonza:


qiita_pet/handlers/api_proxy/processing.py
```diff
@@ -80,9 +83,30 @@ def list_options_handler_get_req(command_id):
...
+    if artifact_id is not None:
+        artifact = Artifact(artifact_id)
+        for job in artifact.jobs(cmd=command):
+            jstatus = job.status
+            outputs = job.outputs if job.status == 'success' else None
+            # this ignore any jobs that weren't successful or are in
+            # construction, or the results have been deleted [outputs == {}]
+            if jstatus not in {'success', 'in_construction'} or outputs == {}:
+                continue
+            params = job.parameters.values
```
wasade, line 102: Is this a copy or the same object? I'm wondering if the `del` below will modify state in `job.parameters.values`

antgonza: Nop, different ... but I can easily add a .copy if you think is better, should I?

wasade: That would be defensive, thanks

wasade (changes requested):

PR #3111, opened 2021-05-28
disable "Add Files" button after click

antgonza:
cc: @sjanssen2


sjanssen2 (approved):
wow, that is a fast bug-fix :-) looks good!

PR #3115, opened 2021-06-17
fix #3054

### LINKED ISSUE #3054: Improve preparation tab in GUI
After discussing with @mestaki, we thought it would be a good idea to reformat the prep information tab to make it more helpful.

As background we currently display:
Screen Shot 2020-11-24 at 1 25 07 PM
A. If the prep info has been submitted to EBI-ENA (the logo) and how many samples were submitted; nothing if it hasn't
B. The preparation name - ID - visibility status
C. The raw artifact type and ID
D. The last children name and artifact type

What about having:
A.  If the prep info has been submitted to EBI-ENA (the logo) and how many samples were submitted; and if it hasn't, the total number of samples in the prep
B. The preparation name (ID) - visibility - (total artifacts in the preparation)
C. Initial creation time
D. Last time it was updated

@justinshaffer, @adswafford, any suggestions?

AmandaBirmingham:
Improves the display of prep info in the study page side-bar as specified in #3054, adding for each prep the total number of samples as well as the creation and modification date, which artifacts have been produced, and whether it is ready for submission.  Displays preps in ascending order by creation date.  See example below.

preps_display_w_dates_mod


antgonza (changes requested):
Looks great, thank you for putting together, some minor suggestions, see below:
- if the prep has been submitted, no need to show that is OK to submit
- the comma (,) between the number of samples and the samples submitted has a space, can you remove it?
- I don't think we need the word `Progress` in the progress line, perhaps worth removing?
Screen Shot 2021-06-17 at 3 00 46 PM

AmandaBirmingham:
@antgonza Thank you!  I will remove the word "Progress" and the space before the comma, as you described.  I need a bit more guidance to know when to put in the "ok to submit" entry: should I leave it off if the `is_submitted_to_ebi` property of the prep is True? Or?

antgonza:
Thank you; BTW there are also spaces after the commas in the Progress elements. 

That's a good question, I think we should remove the "OK to submit" for now, to avoid confusion, and add once we have this: https://github.com/qiita-spots/qiita/issues/2721 ready.

AmandaBirmingham:
@antgonza Ok, can do. How does this look?

Screen Shot 2021-06-17 at 3 26 12 PM

(I know it kinda looks like there are still spaces after the green checkmarks, but there really aren't :)

antgonza:
Looks great, could you push to your branch so the tests run and, if everything passes, I can merge? Thank you.

AmandaBirmingham:
@antgonza should be ready for you!

antgonza (approved):
Thank you so much!

PR #3116, opened 2021-06-24
Add default processing

antgonza:
First merge https://github.com/qiita-spots/qiita/pull/3110; then I'll update this PR.

This PR adds a single button to generate/populate based on our default pipelines, for example:
add-default-workflow

### DIFF
--- a/qiita_db/metadata_template/prep_template.py
+++ b/qiita_db/metadata_template/prep_template.py
@@ -713,3 +713,192 @@ class PrepTemplate(MetadataTemplate):
     @staticmethod
     def max_samples():
         return qdb.util.max_preparation_samples()
+
+    def add_default_workflow(self, user):
+        """The modification timestamp of the prep information
+
+
+        Parameters
+        ----------
+        user : The user that requested to add the default workflows
+
+        Returns
+        -------
+        ProcessingWorkflow
+            The workflow created
+
+        Raises
+        ------
+        ValueError
+            If this preparation doesn't have valid workflows
+            If this preparation has been fully processed
+        """
+        # helper functions to avoid duplication of code
+
+        def _get_node_info(node):
+            # retrieves the merging scheme of a node
+            parent = list(wk.graph.predecessors(node))
+            if parent:
+                parent = parent.pop()
+                pdp = parent.default_parameter
+                pcmd = pdp.command
+                pparams = pdp.values
+            else:
+                pcmd = None
+                pparams = {}
+
+            dp = node.default_parameter
+            cparams = dp.values
+            ccmd = dp.command
+
+            parent_cmd_name = None
+            parent_merging_scheme = None
+            if pcmd is not None:
+                parent_cmd_name = pcmd.name
+                parent_merging_scheme = pcmd.merging_scheme
+
+            return qdb.util.human_merging_scheme(
+                ccmd.name, ccmd.merging_scheme, parent_cmd_name,
+                parent_merging_scheme, cparams, [], pparams)
+
+        def _get_predecessors(node):
+            # recursive method to get predecessors of a given node
+            for pnode in wk.graph.predecessors(node):
+                pred = _get_predecessors(pnode)
+                cxns = {x[0]: x[2]
+                        for x in wk.graph.get_edge_data(
+                            pnode, node)['connections'].connections}
+                data = [pnode, node, cxns]
+                if pred is None:
+                    pred = [data]
+                else:
+                    pred.append(data)
+                return pred
+
+        # Note: we are going to use the final BIOMs to figure out which
+        #       processing is missing from the back/end to the front, as this
+        #       will prevent generating unnecessary steps (AKA already provided
+        #       by another command), like "Split Library of Demuxed",
+        #       when "Split per Sample" is alrady generated
+        #
+        # The steps to generate the default workflow are as follow:
+        # 1. retrieve all valid merging schemes from valid jobs in the
+        #    current preparation
+        # 2. retrive all the valid workflows for the preparation data type and
+        #    find the final BIOM missing from the valid available merging
+        #    schemes
+        # 3. loop over the missing merging schemes and create the commands
+        #    missing to get to those processed samples and add them to a new
+        #    workflow
+        # 4.
+
+        # 1.
+        prep_jobs = [j for c in self.artifact.descendants.nodes()
+                     for j in c.jobs(show_hidden=True)
+                     if j.command.software.type == 'artifact transformation']
+        merging_schemes = {
+            qdb.archive.Archive.get_merging_scheme_from_job(j): {
+                x: y.id for x, y in j.outputs.items()}
+            for j in prep_jobs if j.status == 'success' and not j.hidden}
+
+        # 2.
+        pt_dt = self.data_type()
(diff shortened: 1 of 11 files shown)


qiita_db/metadata_template/prep_template.py
```diff
@@ -713,3 +713,192 @@ def modification_timestamp(self):
...
+                pred = _get_predecessors(pnode)
+                cxns = {x[0]: x[2]
+                        for x in wk.graph.get_edge_data(
+                            pnode, node)['connections'].connections}
+                data = [pnode, node, cxns]
+                if pred is None:
+                    pred = [data]
+                else:
+                    pred.append(data)
+                return pred
```
wasade, line 776: I *think* that having the `return` in scope of the `for` would imply a node can only have a single predecessor. Is that correct for both interpretation and intent?

qiita_db/metadata_template/test/test_prep_template.py
```diff
@@ -1340,11 +1340,26 @@ def test_artifact_setter_error(self):
...
     def test_artifact_setter(self):
         pt = qdb.metadata_template.prep_template.PrepTemplate.create(
-            self.metadata, self.test_study, self.data_type_id)
+            self.metadata, self.test_study, '16S')
         self.assertEqual(pt.artifact, None)
         artifact = qdb.artifact.Artifact.create(
             self.filepaths, "FASTQ", prep_template=pt)
         self.assertEqual(pt.artifact, artifact)
+
+        # here we can test that we can properly create a workflow
```
wasade, line 1349: This seems like a limited test given the complexity of the function, are there edge cases beyond the `ValueError` check below that should be examined?

qiita_db/metadata_template/prep_template.py
```diff
@@ -713,3 +713,192 @@ def modification_timestamp(self):
...
+
+        Raises
+        ------
+        ValueError
+            If this preparation doesn't have valid workflows
+            If this preparation has been fully processed
+        """
+        # helper functions to avoid duplication of code
+
+        def _get_node_info(node):
```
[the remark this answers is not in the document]
antgonza: I could add but when this is called, wk is "global" within this method; do you think is better to pass?

qiita_db/metadata_template/prep_template.py
```diff
@@ -713,3 +713,192 @@ def modification_timestamp(self):
...
+                parent_cmd_name = pcmd.name
+                parent_merging_scheme = pcmd.merging_scheme
+
+            return qdb.util.human_merging_scheme(
+                ccmd.name, ccmd.merging_scheme, parent_cmd_name,
+                parent_merging_scheme, cparams, [], pparams)
+
+        def _get_predecessors(node):
+            # recursive method to get predecessors of a given node
+            for pnode in wk.graph.predecessors(node):
```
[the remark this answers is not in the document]
antgonza: Yup, same, they exists in `add_default_workflow` and the functions are defined and called within that scope so they are "global" ... in fact, we might not even need to pass node. Anyway, I'm OK changing to whatever you think is best, just let me know.

qiita_db/metadata_template/prep_template.py
antgonza: Yes!! Note that this is a recursive function (line 767 calls itself) as we need the "grandparent" information to create the "parent" information.

qiita_db/metadata_template/prep_template.py
```diff
@@ -713,3 +713,192 @@ def modification_timestamp(self):
...
+            for node, degree in wk.graph.out_degree():
+                if degree != 0:
+                    continue
+                mscheme = _get_node_info(node)
+                if mscheme not in merging_schemes:
+                    missing_artifacts[wk][mscheme] = node
+            if not missing_artifacts[wk]:
+                del missing_artifacts[wk]
+        if not missing_artifacts:
+            raise ValueError('This preparation is complete')
```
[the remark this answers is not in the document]
antgonza: You are right, I'll add it.

qiita_db/metadata_template/test/test_prep_template.py
antgonza: Good point, I'll add more tests.

antgonza (commented):
@wasade, thank you for detailed review, I have some questions; then, I'll issue fixes to your comments.

qiita_db/metadata_template/prep_template.py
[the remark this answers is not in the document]
wasade: it may be nice to either pass both or pass neither as `node` is also non-local from what I can tell from the function call

qiita_db/metadata_template/prep_template.py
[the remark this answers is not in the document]
wasade: passing both may be nice to be explicit, tracing non-local variables can get confusing quickly

qiita_db/metadata_template/prep_template.py
wasade: Ya, but just so I understand, is it correct that a node only ever has a single parent?

qiita_db/metadata_template/prep_template.py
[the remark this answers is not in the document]
wasade: Thanks!

qiita_db/metadata_template/test/test_prep_template.py
wasade: Thanks!

mestaki:
Wasn't sure if this should go here or have its own separate issue, let me know and happy create a new issue.

As per the recent email thread on this topic: the Standard Workflow option is great, but without a brief description of what it actually is/does it might be confusing for users. Adding a brief description of this would be thus very useful. Happy to contribute to this. 
Option 1: Add a short description on top of the main network processing page, 
Option 2: A brief "hover-over" text box appears over the "Standard Workflow" button,
Option 3: A detailed description in the main help files and just provide a link to that in the network processing page?
Option 4: something else?

Depending on the options the description I can make the text brief or more detailed.

antgonza:
Thank you @mestaki; I think 1 and 3 would be great. If you send me an email with that I can open a PR adding those texts so you can review there ... or you can open a PR with those changes and I can review; whatever works best. BTW no need to open a new issue.

qiita_db/metadata_template/test/test_prep_template.py
antgonza: Just FYI, I checked everything that came to mind and I think those "simple" tests actually cover all scenarios by having to create a full pipeline and then checking that full pipeline and not having anything else to create.

qiita_db/metadata_template/prep_template.py
antgonza: Thank you for double checking this; better to be safe. 

Yes, I spent sometime looking at this and this is the expected behavior; basically, by being recursive it will go down to the origin of the workflow (always a single artifact) and return (`if pred is None:`) the single only parent, then it will start coming out of the recursion and appending in order the children jobs (`else:`). In case it helps, we tested this in qiita-rc and everything seem fine.

antgonza (commented):
Thank you @wasade.

wasade:
Good to merge?

antgonza:
Yes, thank you. The tutorial/text updates will be done in a separate PR (together with the release notes).

PR #3118, opened 2021-06-24
Filter preps GUI + some minor fixes

antgonza:
Final touch to filter prep tabs via JS:
filter-preps-GUI

### DIFF
--- a/qiita_ware/private_plugin.py
+++ b/qiita_ware/private_plugin.py
@@ -402,9 +402,6 @@ def list_remote_files(job):
             job._set_error(traceback.format_exception(*exc_info()))
         else:
             job._set_status('success')
-        finally:
-            # making sure to always delete the key so Qiita never keeps it
-            remove(private_key)
 
 
 def download_remote_files(job):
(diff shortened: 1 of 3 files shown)


qiita_ware/private_plugin.py
```diff
@@ -402,9 +402,6 @@ def list_remote_files(job):
             job._set_error(traceback.format_exception(*exc_info()))
         else:
             job._set_status('success')
-        finally:
-            # making sure to always delete the key so Qiita never keeps it
```
wasade, line 406: ?

wasade (commented):
Minor question

qiita_ware/private_plugin.py
antgonza: This is the bug we found that was preventing listing files via scp for study 12294; basically `list_remote` already removes the key and this fails cause the file never exists when we reach this point.

wasade: ah, thanks!

PR #3121, opened 2021-07-06
2021.07

antgonza:
@mestaki, this is how the prep page will look like:
Screen Shot 2021-07-06 at 8 53 44 AM

... and this is the new workflow page:
Screen Shot 2021-07-06 at 8 53 21 AM

What do you think?

### DIFF
--- a/qiita_pet/static/js/networkVue.js
+++ b/qiita_pet/static/js/networkVue.js
@@ -1144,8 +1144,10 @@ Vue.component('processing-graph', {
         '<td><small>Job status (circles):</small></td>' +
         '<td>' + circle_statuses.join('') + '</td>' +
         '<td rowspan="2" width="20px">&nbsp;</td>' +
-        '<td rowspan="2">' +
+        '<td rowspan="2">&nbsp;&nbsp;&nbsp;</td>' +
+        '<td rowspan="2" align="center">' +
             '<a class="btn btn-success form-control" id="add-default-workflow"><span class="glyphicon glyphicon-flash"></span> Add Default Workflow</a>' +
+             "<br/><a href='https://qiita.ucsd.edu/workflows/' target='_blank'>What's a Default Workflow?</a>" +
         '</td>' +
       '</tr>' +
       '<tr>' +
--- a/qiita_pet/static/js/networkVue.js
+++ b/qiita_pet/static/js/networkVue.js
@@ -1147,7 +1147,14 @@ Vue.component('processing-graph', {
         '<td rowspan="2">&nbsp;&nbsp;&nbsp;</td>' +
         '<td rowspan="2" align="center">' +
             '<a class="btn btn-success form-control" id="add-default-workflow"><span class="glyphicon glyphicon-flash"></span> Add Default Workflow</a>' +
-             "<br/><a href='https://qiita.ucsd.edu/workflows/' target='_blank'>What's a Default Workflow?</a>" +
+             "<br/><br/><a href='https://qiita.ucsd.edu/workflows/' target='_blank'> "+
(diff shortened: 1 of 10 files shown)


qiita_pet/templates/workflows.html
```diff
@@ -63,8 +63,25 @@
 
 {% block content %}
   {% if workflows %}
-    <h3>Recommended Workflows</h3>
+    <h3>Recommended Default Workflows</h3>
+    <div class="row">
+      <div class="col-sm-10">
+        In this page you will find our recommendations on how to processes each data type available in Qiita and the minimum processing requirements to make your data public.
```
mestaki, line 69: *This is very useful, just some rewording below to hopefully make it more clear for novice users.

----
 In this page you will find our recommendations on how to processes each data type available in Qiita and the minimum processing requirements to make your data public.
        <br/><br/>
        For convenience, there is an "Add Default Workflow" button for each preparation that will add the recommended workflow steps for your raw data which then you can process. It is important to note however that this workflow is based on the default Earth Microbiome Project protocol and so assumes the uploaded data are multiplexed sequences with the reversed barcodes in your mapping file and index  sequence file (see [here](https://earthmicrobiome.org/protocols-and-standards/) for more details). If this protocol does not apply to your data you can still use the Default Workflow, however, you should first manually process your data using the appropriate steps until you have demultiplexed your reads. After demultiplexing the Default Workflow is safe to use with any protocol.
        <br/><br/>
        If you have already manually performed one of the processing steps in the Defaul Workflow pipeline, the "Add Default Workflow" button will not re-select those steps but instead will only select any remaining steps that have not been completed. You can also add additional workflows on top of the recommended Default Workflow at any time.
        <br/><br/>
        Note that this is not a full inclusive list of data types accepted by Qiita but only those that have a <u>defined</u> workflow.
      </div>

qiita_pet/static/js/networkVue.js
```diff
@@ -1144,8 +1144,10 @@ Vue.component('processing-graph', {
         '<td><small>Job status (circles):</small></td>' +
         '<td>' + circle_statuses.join('') + '</td>' +
         '<td rowspan="2" width="20px">&nbsp;</td>' +
-        '<td rowspan="2">' +
+        '<td rowspan="2">&nbsp;&nbsp;&nbsp;</td>' +
+        '<td rowspan="2" align="center">' +
             '<a class="btn btn-success form-control" id="add-default-workflow"><span class="glyphicon glyphicon-flash"></span> Add Default Workflow</a>' +
+             "<br/><a href='https://qiita.ucsd.edu/workflows/' target='_blank'>What's a Default Workflow?</a>" +
```
mestaki, line 1150: I was thinking that if we are not vigilant enough with explaining the EMP assumptions of the Default Workflow we may get a lot of headaches later on when people not-knowingly process their non-EMP protocols. So anything we can to draw their attention to that important aspect I see as time-saving for admin later on.

One suggestion, what about replacing "What's a Default Workflow" to "Important note about Default Workflow"? <- That way they are more likely to read that description with the important note about the demux step. Or maybe adding a :warning: logo beside the button?

mestaki (changes requested):
Just some rewording and one suggestion about adding a "notice/warning" about EMP dependency of Standard Workflow

antgonza:
Thank you; now it looks like this:

Screen Shot 2021-07-07 at 7 47 55 PM

and this:
Screen Shot 2021-07-07 at 7 48 18 PM

Note that I used most of your suggestions but tried to keep the explanation as general as possible by using the target gene example but still mention that there are multiple workflows.

mestaki (approved):

PR #3123, opened 2021-07-07
adding @AnjaliVanamala suggestion

antgonza:
Hi @AnjaliVanamala, do you think this would have help you figuring out the artifact conversions?

Note that the changes in this PR will look like this in the website:

Screen Shot 2021-07-07 at 6 53 25 AM


AnjaliVanamala:
That looks great! It would have been and probably still will be extremely helpful!

antgonza:
Thank you for your review! @AnjaliVanamala

PR #3130, opened 2021-08-05
making sure the files are copied when copy in insert_filepaths

antgonza:
@adswafford, this is the required change to correctly copy files when using qebil + tests.


adswafford (approved):
Looks good to me.

PR #3131, opened 2021-08-17
add base for new admin plugin

antgonza:


qiita_db/artifact.py
```diff
@@ -337,7 +337,7 @@ def create(cls, filepaths, artifact_type, name=None, prep_template=None,
             raise qdb.exceptions.QiitaDBArtifactCreationError(
                 "When provided, both parents and processing parameters should "
                 "be provided")
-        elif bool(analysis) != bool(data_type):
+        elif bool(analysis) and not bool(data_type):
```
wasade, line 340: `data_type` is described as a `str`. Just warning:

```python
>>> bool("True")
True
>>> bool("False")
True
>>>
```

antgonza: I remember this being discussed, at length, long time ago (when it was first implemented) and if I remember correctly, it was decided that this was fine because of this:
```python
>>> bool("")
False
>>> bool(None)
False
```

wasade: Okay. It just continues to catch me by surprise every time I encounter it, hence the warning. Since `anaylsis is not None` is used on line 479 in this PR though, should the same null handling be used here (or there)?

antgonza: Same here 😄; I have to think 🤔 about it when I see it. 

Good point! I can change it "there". 

Thank you.

unique-identifier (approved):

ElDeveloper:
@wasade, if you approve of the changes here would you mind merging?

wasade (approved):

ElDeveloper:
Thanks @wasade!

PR #3132, opened 2021-08-17
rm pt.generate_files from sample update

antgonza:


wasade (commented):
I'm not really sure I understand how this change relates to the update time noted with the WGS preps?

antgonza:
Ha! Good question: basically, when a user updates a sample information file, if a sample is present in a given preparation it will be updated. This was useful for when we actually stored in disk the qiime mapping file (sample + prep for each preparation), this update generated the new preps (still the same info than before); however, now we don't do that. The sample+prep is now generated on the fly so that code is no longer necessary.

wasade:
I’m still not sure I understand as in this case, preps were reported as updated which did not have samples reflected in the update. Or was there a side effect where updating just a single sample would trigger generation of all preps regardless of sample presence?

antgonza:
It will not update all if no new columns were added/deleted, if a new column was added, it will trigger a full update ... sadly.

qiita_pet/test/test_download.py
```diff
@@ -457,11 +455,12 @@ def test_download(self):
...
-            '1_[0-9]*-[0-9]*.txt mapping_files/4_mapping_file.txt\n'
+            '1_qiime_19700101-000000.txt mapping_files/4_mapping_file.txt\n'
             '- [0-9]* /protected/processed_data/1_study_1001_closed_'
             'reference_otu_table.biom processed_data/1_study_1001_closed_'
-            'reference_otu_table.biom\n- [0-9]* /protected/templates/1_prep_1'
-            '_[0-9]*-[0-9]*.txt mapping_files/5_mapping_file.txt\n')
+            'reference_otu_table.biom\n- [0-9]* /protected/templates/1_prep_'
+            '1_qiime_19700101-000000.txt mapping_files/5_mapping_file.txt\n')
+
         self.assertRegex(response.body.decode('ascii'), exp)
```
unique-identifier, line 464: I submitted a fix for the assertRegex failure to your @antgonza fork, Antonio.

antgonza:
This is ready for review/merge. Note that the main change was to delete the preparation generation when updating the sample info file and the tests that relied on that - mainly, the timestamp of the latest preparation and the filepath counts.

ElDeveloper:
@unique-identifier @wasade if this PR looks good to you can you go ahead and merge it?

charles-cowart:
@antgonza May I have write access on Qiita project? I can merge once I have it. Thanks!

antgonza:
@charles-cowart, just invited you; thank you.

Issue #3124, opened 2021-07-12
'NoneType' object has no attribute 'descendants'


antgonza:
This error happens when a user tries to delete a full study with empty preps (no artifacts/descendants).

PR #3136, opened 2021-08-23
fix #3122

### LINKED ISSUE #3122: Include "raw" artifacts in the analysis artifact count
Reported by one of our users, when you create an analysis the "raw" (initial) artifact is not included in the artifact count within the listing; which can be confusing.

antgonza:
The code was actually ignoring the "raw" analysis artifacts, now it doesn't.


antgonza:
@ElDeveloper By mistake I also pushed the fix for #3124 here; the changes for both issues are pretty small but let me know if you want me to split them.

ElDeveloper:
Looks great to me, thanks so much @antgonza!

PR #3152, opened 2021-10-06
Fix #3149

### LINKED ISSUE #3149: Should newlines be allowed on metadata upload
I recently pushed metadata to Qiita which unbeknownst to me contained newlines in some cells. I discovered this when downloading the pushed metadata after, and observed what appeared to be incorrect sample names in the resulting tab delimited file, coming from the contained newlines.

Should Qiita detect and error, or detect and replace, newline characters prior to accepting metadata updates?

wasade:
...route load through Qiita's template loading mechanisms

### DIFF
--- a/qiita_pet/handlers/rest/study_samples.py
+++ b/qiita_pet/handlers/rest/study_samples.py
@@ -6,6 +6,7 @@
 # The full license is in the file LICENSE, distributed with this software.
 # -----------------------------------------------------------------------------
 from collections import defaultdict
+import io
 
 from tornado.escape import json_encode, json_decode
 import pandas as pd
@@ -150,8 +151,15 @@ class StudySamplesHandler(RESTHandler):
         else:
             sample_info = study.sample_template.to_dataframe()
 
-        data = pd.DataFrame.from_dict(json_decode(self.request.body),
-                                      orient='index')
+        # convert from json into a format that qiita can validate
+        rawdata = pd.DataFrame.from_dict(json_decode(self.request.body),
+                                         orient='index')
+        buffer = io.StringIO()
+        rawdata.to_csv(buffer, sep='\t', index=True, header=True)
+        buffer.seek(0)
+        
+        # validate on load
+        data = qiita_db.metadata_template.util.load_template_to_dataframe(rawdata)
 
         if len(data.index) == 0:
             self.fail('No samples provided', 400)


qiita_pet/handlers/rest/study_samples.py
```diff
@@ -150,8 +151,15 @@ def patch(self, study_id):
         else:
             sample_info = study.sample_template.to_dataframe()
 
-        data = pd.DataFrame.from_dict(json_decode(self.request.body),
-                                      orient='index')
+        # convert from json into a format that qiita can validate
+        rawdata = pd.DataFrame.from_dict(json_decode(self.request.body),
```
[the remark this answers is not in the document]
wasade: The code looks like it opens a filehandle, so I don't think a regular string would work. The spec for the REST endpoint as well is not a str representation of a TSV, but a JSON representation derived from `DataFrame.to_json`. I may be missing something, but I think this transform is necessary

qiita_pet/handlers/rest/study_samples.py
```diff
@@ -150,8 +151,15 @@ def patch(self, study_id):
...
+                                         orient='index')
+        buffer = io.StringIO()
+        rawdata.to_csv(buffer, sep='\t', index=True, header=True)
+        buffer.seek(0)
+        
+        # validate on load
+        data = qiita_db.metadata_template.util.load_template_to_dataframe(rawdata)
```
wasade, line 162:
```suggestion
        data = load_template_to_dataframe(rawdata)
```

qiita_pet/handlers/rest/study_samples.py
```diff
@@ -150,8 +152,15 @@ def patch(self, study_id):
...
-                                      orient='index')
+        # convert from json into a format that qiita can validate
+        rawdata = pd.DataFrame.from_dict(json_decode(self.request.body),
+                                         orient='index')
+        buffer = io.StringIO()
+        rawdata.to_csv(buffer, sep='\t', index=True, header=True)
+        buffer.seek(0)
+        
+        # validate on load
+        data = load_template_to_dataframe(rawdata)
```
wasade, line 163:
```suggestion
        data = load_template_to_dataframe(buffer)
```

qiita_pet/handlers/rest/study_samples.py
```diff
@@ -150,8 +152,14 @@ def patch(self, study_id):
         else:
             sample_info = study.sample_template.to_dataframe()
 
-        data = pd.DataFrame.from_dict(json_decode(self.request.body),
-                                      orient='index')
+        # convert from json into a format that qiita can validate
+        rawdata = pd.DataFrame.from_dict(json_decode(self.request.body),
+                                         orient='index')
+        buffer = io.StringIO()
```
wasade, line 158:
```suggestion
        rawdata.index.name = 'sample_name'
        rawdata.reset_index(inplace=True)
        buffer = io.StringIO()
```

PR #3159, opened 2021-11-05
fix Overflow in download analysis metadata

antgonza:


qiita_db/handlers/analysis.py
```diff
@@ -56,15 +59,34 @@ def get(self, analysis_id):
...
+                    chunk = response[i*chunk_len:win]
+                    try:
+                        self.write(chunk)
+                        await self.flush()
+                    except StreamClosedError:
+                        break
+                    finally:
+                        del chunk
+                        # pause the coroutine so other handlers can run
+                        await gen.sleep(0.000000001)  # 1 nanosecond
```
wasade, line 90: wouldn't this be taken care of by the `await` on 84?

qiita_db/handlers/analysis.py
```diff
@@ -56,15 +59,34 @@ def get(self, analysis_id):
...
         with qdb.sql_connection.TRN:
             a = _get_analysis(analysis_id)
             mf_fp = qdb.util.get_filepath_information(
                 a.mapping_file)['fullpath']
-            response = None
             if mf_fp is not None:
                 df = qdb.metadata_template.util.load_template_to_dataframe(
                     mf_fp, index='#SampleID')
-                response = df.to_dict(orient='index')
+                response = dumps(df.to_dict(orient='index'))
```
wasade, line 77: i think all of this makes sense, but I think it should be possible to do this outside of the TRN context, and close the transaction?

qiita_db/handlers/analysis.py
```diff
@@ -56,15 +59,34 @@ def get(self, analysis_id):
...
-        self.write(response)
+                crange = range(chunk_len, len(response)+chunk_len, chunk_len)
+                for i, (win) in enumerate(crange):
+                    chunk = response[i*chunk_len:win]
+                    try:
+                        self.write(chunk)
+                        await self.flush()
+                    except StreamClosedError:
+                        break
+                    finally:
```
wasade, line 87: I _think_ the intention here is for this to be `else`?

antgonza: I literally borrow this code from the link in the notes ...

qiita_db/handlers/analysis.py
antgonza: Nop, in fact, if you check the code in the reference, the magic actually happens here; without this everything goes 🍌  ... at least based on that link.

qiita_db/handlers/analysis.py
wasade: sure but it doesn't seem necessary here as this method essentialy returns right after `StreamClosedError` is caught. i don't really care though, so up to you

qiita_db/handlers/analysis.py
wasade: for the large data case, it shouldn't be necessary, but since this code is used on small targets it seems reasonable. it's ashame the `await None` doesn't work. it's possible `yield None` does but it may depend on aspects of decoration that i dont understand

qiita_db/handlers/analysis.py
antgonza: You are right, I guess the difference is how the original code is creating the chunks vs. how it's done here ... I'll remove.

antgonza:
@wasade, finally passing; could you take another look? Thank you.

antgonza:
Thanks @wasade !

PR #3162, opened 2021-11-12
[WIP] Improve SQL study listing

antgonza:


antgonza:
After further evaluation the "easiest" would be not display this as part of the study listing; currently discussing in the qiita-admin ... thus, closing.

PR #3167, opened 2021-11-29
Wtform validators fix

charles-cowart:
@antgonza I encountered an issue when re-installing Qiita where 'required' was no longer an attribute of wtforms.validators. I believe this is the proper fix, based on https://wtforms.readthedocs.io/en/2.3.x/validators. I'm using WTForms 2.3.3. Perhaps we need to fix it to an earlier version instead of changing this code?


antgonza:
Yup, this is ["fixed" in dev](https://github.com/qiita-spots/qiita/blob/dev/setup.py#L108); please use that branch vs. master.

PR #3169, opened 2021-12-09
Python 3.9

antgonza:


.github/workflows/qiita-ci.yml
```diff
@@ -45,7 +45,7 @@ jobs:
         uses: conda-incubator/setup-miniconda@v2
         with:
           auto-update-conda: true
-          python-version: 3.6
+          python-version: '3.9'
```
charles-cowart, line 48: Before this was 3.6 as a float. Just to confirm, should it be in fact a string?

.github/workflows/qiita-ci.yml
```diff
@@ -210,7 +214,7 @@ jobs:
     - name: flake8
       uses: actions/setup-python@v2
       with:
-        python-version: 3.6
+        python-version: '3.9'
```
charles-cowart, line 217: same as above.

qiita_ware/ebi.py
```diff
@@ -6,6 +6,7 @@
 # The full license is in the file LICENSE, distributed with this software.
 # -----------------------------------------------------------------------------
 
+import hashlib
```
charles-cowart, line 9: Minor point but I would think 'from hashlib import md5' is cleaner, since we're only using the one function. Not sure if we have a preference in the lab.

setup.py
```diff
@@ -105,13 +105,13 @@
       install_requires=['psycopg2', 'click', 'bcrypt', 'pandas',
                         'biom-format', 'tornado<6.0', 'toredis', 'redis',
                         'scp', 'pyparsing', 'h5py',  'natsort', 'nose', 'pep8',
-                        'networkx', 'humanize', 'scikit-bio', 'wtforms<3.0.0',
+                        'networkx', 'humanize', 'wtforms<3.0.0',
```
charles-cowart, line 108: Just to confirm, since scikit-bio is no longer needed in install-dependencies?

charles-cowart (changes requested):

.github/workflows/qiita-ci.yml
antgonza: Yes, this is fine. Note that in [the documentation](https://github.com/actions/setup-python) it's actually a string, I guess to support non numeric versions.

.github/workflows/qiita-ci.yml
antgonza: same as above.

qiita_ware/ebi.py
antgonza: In the past, I would always do your recommendation but for whatever reason I decided to go this way. Now, reading online, it seems like there is [no actual benefit (except writing less)](https://stackoverflow.com/a/710603) to go with suggestion. Anyway, I rather leave as it is but I can change if you feel strongly about it; just let me know.

antgonza (commented):
Thank you @charles-cowart.

charles-cowart (commented):
Reviewed and accepted with no changes.

PR #3172, opened 2022-01-26
Ignore deleted outputs in default workflows

antgonza:
This is a simple change but adding tests for this is really hard in a test environment so decided to explain the reasons behind the change in the code so it can be used during this and future reviews.


charles-cowart (approved):
Looks good. I also believe this might be difficult to simulate inside a unit-test.

PR #3176, opened 2022-02-01
add backend to specify analysis metadata categories

antgonza:


antgonza:
@wasade, this is the backend to select which metadata categories should an analysis use.

qiita_db/analysis.py
```diff
@@ -804,6 +836,8 @@ def build_files(self, merge_duplicated_sample_ids):
             If the duplicated sample ids in the selected studies should be
             merged or prepended with the artifact ids. If false prepends
             the artifact id
+        categories : set of str
```
wasade, line 839:
```suggestion
        categories : set of str, optional
```

wasade (commented):
This seems reasonable given what I understand of the system

antgonza:
reviewed by @wasade, merging.

PR #3177, opened 2022-02-15
updating JS libraries 2022.02

antgonza:
Here I'm updating all possible JS libraries to their latests:
- jquery: 2.1.0 -> 3.6.0
- jquery-ui: 1.11.1 -> 1.13.1
- jquery chosen: 1.1.0 -> 1.8.7
- jquery dataTables: 1.10.13 -> 1.11.4
- jQuery Validation: v1.12.0 -> 1.19.3
- vue: 2.1.6 -> 2.6.14
- Select2: 4.0.2 -> 4.0.13
- SlickGrid: 2.2 -> 2.4.44
- jquery.event.drag: 2.2 -> 2.3.0
- Cytoscape.js: 3.18.2 -> 3.20.0
- cytoscape-dagre.js 2.3.2 -> 2.4.0
- Note that Bootstrap was kept at 3.3.7 even as there is a much newer version (5.1.3); the reason behind that is because a change like that will mean rewriting all boostrap calls, see: https://stackoverflow.com/q/68004792

Replaced:
- tag-it -> tagify 4.9.6

Removing (as they are no longer used): 
- underscore.min.js
- string.min.js


qiita_pet/templates/list_studies.html
```diff
@@ -6,7 +6,6 @@
 
 <script src="{% raw qiita_config.portal_dir %}/static/vendor/js/select2.min.js"></script>
 <script src="{% raw qiita_config.portal_dir %}/static/js/sharing.js"></script>
-<script src="{% raw qiita_config.portal_dir %}/static/vendor/js/tag-it.min.js" type="text/javascript" charset="utf-8"></script>
```
charles-cowart, line 9: just to confirm, it's okay that we're removing this reference to tag-it? I noticed it's still in the 'vendor' directory.

antgonza: I'm pretty sure it's removed ... no?

qiita_pet/static/js/qiita.js
```diff
@@ -265,7 +265,12 @@ var qiita_websocket = new function () {
         ws.onerror = on_error;
 
         ws.onmessage = function(evt) {
-            message = decode(evt.data);
+            var data = evt.data;
+            if (data === 'hello'){
```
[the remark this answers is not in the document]
antgonza: Good point, opened an issue. This change is really to remove an error in the developers console that I have noticed in the past. Anyway, opened an issue: https://github.com/qiita-spots/qiita/issues/3179

antgonza (commented):
Thank you.

qiita_pet/templates/list_studies.html
charles-cowart: Ah! You are right, sorry. I took the blank screen to mean it was another large diff, but it was in fact deleted.

charles-cowart (approved):
Approved. 'Hello' string will be described in a separate issue.

PR #3180, opened 2022-02-22
adding how to select metadata for analysis via GUI

antgonza:
The user can select via GUI which metadata categories to use. Note that overlapping categories are preselected but you can add/remove any of them. 

The GUI looks like this:
create-analysis


wasade:
This is great! Would it be possible to sort the tags?

charles-cowart (approved):
Everything looks good. No changes requested.

PR #3184, opened 2022-03-03
rm sleeps when starting plugins

antgonza:


antgonza:
This made the build a little faster and we had no errors so ready to review.

antgonza:
AFAIK that will not be a problem because they are started serially on a single workers (master). If I remember correctly those sleeps are "legacy" from when we used to use ipython workers to start things.

PR #3190, opened 2022-03-15
fix #3158

### LINKED ISSUE #3158: Issues with  "Add Default Workflow"
- [x] in some occasions the code will create a new full pathway vs using the existing one; see preparation in study 14057 to reproduce
- [x] 16S per-sample-FASTQ without demultiplexing doesn't find the correct raw data type

antgonza:


qiita_db/metadata_template/prep_template.py
```diff
@@ -874,8 +882,9 @@ def _get_predecessors(workflow, node):
                         for iname, dname in rp.items():
                             if dname not in init_artifacts:
                                 msg = (f'Missing Artifact type: "{dname}" in '
-                                       'this preparation; are you missing a '
-                                       'step to start?')
+                                       'this preparation; this might be due '
+                                       'to missing steps or not having the '
+                                       'correct raw data.')
```
charles-cowart, line 887: Raises option c seems cryptic after briefly reviewing the function. It's not part of the PR though, so I'll raise it as separate and low-priority issue.

qiita_db/metadata_template/test/test_prep_template.py
```diff
@@ -1412,13 +1412,13 @@ def test_artifact_setter(self):
...
+        # the default_workflow_id = 1
         qdb.sql_connection.perform_as_transaction(
-            'UPDATE qiita.artifact SET artifact_type_id = 10 WHERE '
-            f'artifact_id = {pt.artifact.id}')
-        with self.assertRaisesRegex(ValueError, 'Missing Artifact type: '
-                                    '"FASTQ" in this preparation; are you '
-                                    'missing a step to start?'):
+            'UPDATE qiita.default_workflow SET artifact_type_id = 10 WHERE '
+            'default_workflow_id = 1')
+        with self.assertRaisesRegex(ValueError, 'taxa_summary is not part of '
```
charles-cowart, line 1420: @antgonza just out of curiosity, why are these RaisesRegex when the expected string are just literals?

qiita_pet/handlers/software.py
```diff
@@ -72,13 +72,23 @@ def _default_parameters_parsing(node):
         #                   output_type: output_node_name}, ...}
         # for easy look up and merge of output_names
         main_nodes = dict()
-        for x, y in graph.edges:
+        for i, (x, y) in enumerate(graph.edges):
```
charles-cowart, line 75: Just to confirm, the 'i' coming from enumerate is meant to identify the very first loop iteration only? I guess if it depends on having x and y defined then it's overall the cleanest representation.

qiita_pet/handlers/software.py
```diff
@@ -72,13 +72,23 @@ def _default_parameters_parsing(node):
...
             vals_y, input_y, output_y = _default_parameters_parsing(y)
+
+            if i == 0:
+                # we are in the first element so we can specifically select
+                # the type we are looking for
+                at = w.artifact_type
+                if at in input_x[0][1]:
+                    input_x[0][1] = at
+                else:
+                    input_x[0][1] = '** WARNING, NOT DEFINED **'
```
charles-cowart, line 90: This wouldn't be better expressed as an Error of some type?

qiita_pet/test/test_software.py
```diff
@@ -206,8 +208,8 @@ def test_retrive_workflows(self):
             'qual_score_window': '0', 'disable_primers': 'False',
             'reverse_primers': 'disable', 'reverse_primer_mismatches': '0',
             'truncate_ambi_bases': 'False'}],
-        ['input_params_3_FASTA | FASTA_Sanger | SFF', 2,
-         'FASTA | FASTA_Sanger | SFF'],
+        ['input_params_3_** WARNING, NOT DEFINED **', 2,
+         '** WARNING, NOT DEFINED **'],
```
charles-cowart, line 212: see above.

charles-cowart (changes requested):
Looks good for the most part. Just want to confirm a few things before approving.

qiita_db/metadata_template/prep_template.py
antgonza: This function adds a default workflow to a preparation so IMOO the error message makes sense.

qiita_db/metadata_template/test/test_prep_template.py
antgonza: Leftover from a previous iteration - I think; happy to change if you feel strongly about it.

qiita_pet/handlers/software.py
antgonza: This is correct: i is enumerate and is just to figure out if the first loop.

qiita_pet/test/test_software.py
antgonza: ... see response above ... 😄

qiita_db/metadata_template/prep_template.py
charles-cowart: ok.

qiita_db/metadata_template/test/test_prep_template.py
charles-cowart: I'm okay with if you are.

qiita_pet/handlers/software.py
charles-cowart: Good point. It is in qiita_pet so it's in the UI layer, after all.

charles-cowart (approved):
Approved.

antgonza:
Thank you!

PR #3191, opened 2022-03-17
rm specimen_id

antgonza:


charles-cowart (approved):
Confirmed every omission is related to specimen_id. Everything appears straightforward enough and in order.

PR #3193, opened 2022-03-22
minimal fix for plugin testing

antgonza:
While developing new plugins, I noticed that once in a while some will not load correctly. After further investigation, it resulted that if the plugin relied on a plugin-type and the plugin-type had a filename that was alphanumerically higher than the plugin - the system reset will fail because the plugin-type will not be added before the plugin. 

This PR makes sure that plugin-types are always activated before regular plugins; solving the above issue.


charles-cowart (approved):
Approved. Everything looks good. No changes requested.

PR #3199, opened 2022-05-12
fixes after 2022.04

antgonza:
General improvements after deployment.

### DIFF
--- a/qiita_db/artifact.py
+++ b/qiita_db/artifact.py
@@ -626,13 +626,13 @@ class Artifact(qdb.base.QiitaObject):
 
                 # there are cases that an artifact would not be linked to a
                 # study
-                pt_ids = [pt.id for a in all_artifacts
+                pt_ids = [tuple([pt.id]) for a in all_artifacts
                           for pt in a.prep_templates]
                 if pt_ids:
                     sql = """UPDATE qiita.prep_template
                              SET artifact_id = NULL
                              WHERE prep_template_id IN %s"""
-                    qdb.sql_connection.TRN.add(sql, [tuple(x) for x in pt_ids])
+                    qdb.sql_connection.TRN.add(sql, pt_ids)
             else:
                 sql = """DELETE FROM qiita.parent_artifact
                          WHERE artifact_id IN %s"""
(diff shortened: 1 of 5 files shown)


qiita_db/artifact.py
```diff
@@ -79,17 +99,6 @@ def iter_by_visibility(cls, visibility):
             for a_id in qdb.sql_connection.TRN.execute_fetchflatten():
                 yield cls(a_id)
 
-    @classmethod
```
antgonza, line 82: This is not used anywhere in the code.

qiita_db/artifact.py
```diff
@@ -615,12 +624,15 @@ def delete(cls, artifact_id):
                 qdb.util.move_filepaths_to_upload_folder(
                     study.id, filepaths)
 
-                sql = """UPDATE qiita.prep_template
-                         SET artifact_id = NULL
-                         WHERE prep_template_id IN %s"""
-                qdb.sql_connection.TRN.add(
-                    sql, [tuple(pt.id for a in all_artifacts
-                                for pt in a.prep_templates)])
+                # there are cases that an artifact would not be linked to a
```
antgonza, line 627: Now we have artifacts (like those generated by the sequencing processing pipeline) without prep/analysis linkages.

qiita_pet/templates/sitebase.html
```diff
@@ -532,11 +532,7 @@
                   </li>
                   <li role="separator" class="divider"></li>
                   <li>
-                    {% if qiita_sha=="" %}
-                      <a href="#">
-                    {% else %}
-                      <a href="https://github.com/biocore/qiita/commit/{{qiita_sha}}">
-                    {% end %}
+                    <a href="https://github.com/qiita-spots/qiita/releases">
```
antgonza, line 535: We should point to the releases page vs. the `sha` that is too dev oriented.

qiita_ware/commands.py
```diff
@@ -43,11 +43,8 @@ def _ssh_session(p_url, private_key):
     """
     scheme = p_url.scheme
     hostname = p_url.hostname
-    # if port is '' Python 2.7.6 will raise an error
-    try:
-        port = p_url.port
-    except Exception:
-        port = 22
+
```
antgonza, line 46: The ssh connection is currently broken due to newer python, this fixes it.

charles-cowart (approved):
Everything looks good so far.

PR #3201, opened 2022-06-07
fix #3087

### LINKED ISSUE #3087: How to select multiple artifacts for analytical commands? [Discussion]
### Background

Currently, the qp-qiime2 plugin (for analyses) only adds available commands that have 1 input artifact via the GUI - to be clear these are inputs not parameters; for example the steps for a Unweighted UniFrac PCoA are:
- rarefaction: only needs the feature table
- Unweighted UniFrac: this actually requires two but is "easy" to retrieve one of them internally, the feature table and the tree (which is selected from a drop down menu)
- PCoA: distance matrix
- Emperor: the PCoA

This configuration allows us to have "simple" input -> process -> output display

### Issue to address

Let's use this example and imagine how can we add an [Empress](https://github.com/qiita-spots/qp-qiime2/issues/58) command to my analysis page. 
Empress
Note that Empress needs: a feature table, a tree and a PCoA.

- Where should we add this command to the GUI? Should we add it to the FeatureTable, to the PCoA, or both processing options? Should it be done all the time or only when both artifacts exist in the analysis? As we need both, perhaps on the "highest" level, I mean only in the PCoA
- How should a user select the other artifacts? I would like to go from "easy" to implement, like list all the valid artifact ids-names and allow the user to select from that drop-down menu, to "hardest" to implement, like allow the user to click on the artifact they want to select.

Thoughts?

antgonza:


qiita_pet/handlers/api_proxy/processing.py
```diff
@@ -104,14 +117,27 @@ def _helper_process_params(params):
                 del params[k]
             eparams.append(_helper_process_params(params))
 
+        # removing this artifact from extra_artifacts
+        if artifact.artifact_type in extra_atypes:
+            extra_atypes.remove(artifact.artifact_type)
+
+    extra_artifacts = defaultdict(list)
```
charles-cowart, line 124: @antgonza Just to confirm, you're using a defaultdict() here, but do you set the default value w/setdefault() anywhere?

qiita_pet/handlers/api_proxy/processing.py
```diff
@@ -80,18 +81,30 @@ def list_options_handler_get_req(command_id, artifact_id=None):
...
         {'status': str,
          'message': str,
-         'options': list of dicts of {'id: str', 'name': str,
-                                      'values': dict of {str: str}}}
+         'options': list of dicts of {'id: str',
+                                      'name': str,
+                                      'values': dict of {str: str}},
+         'req_options': dict,
+         'opt_options': dict,
+         'extra_artifacts': dict}
```
charles-cowart, line 89: @antgonza if I understand this right, the first artifact is passed the legacy way and any additional artifacts are stored in here?

charles-cowart (changes requested):
Everything else appears in order. Requesting clarification on the two points before approving.

qiita_pet/handlers/api_proxy/processing.py
antgonza: yup, mainly because the "legacy" is the "main" input and the "extra_artifacts" are other that are available that can be used in the same command.

qiita_pet/handlers/api_proxy/processing.py
antgonza: no need to initialize, that's the point of a default dict, see line 130.

charles-cowart: Ah I see. when 'list' is passed as the parameter to defaultdict(), the default is set to a sensible empty list, much like 'int' will set the default to 0.

qiita_pet/handlers/api_proxy/processing.py
charles-cowart: After discussing with @antgonza, I agreed that passing the extra artifacts through a separate channel and not with the main 'returned' artifact was prudent. Since they will be handled differently, it makes sense to keep them separated rather than join them only to subsequently filter them out again.

charles-cowart (approved):
Approved. No changes or clarifications needed.

PR #3202, opened 2022-06-08
archive biom: back changes

antgonza:
This adds the possibility to archive BIOM files and delete their log files.

### DIFF
--- a/qiita_db/artifact.py
+++ b/qiita_db/artifact.py
@@ -669,7 +669,7 @@ class Artifact(qdb.base.QiitaObject):
         QiitaDBOperationNotPermittedError
             If the artifact is not public
             If the artifact_type is not BIOM
-            If the artifact belowns to an analysis
+            If the artifact belongs to an analysis
             If the artifact has no parents (raw file)
         """
         artifact = cls(artifact_id)
@@ -688,7 +688,7 @@ class Artifact(qdb.base.QiitaObject):
                 'Only non raw artifacts can be archived')
 
         # let's find all ancestors that can be deleted (it has parents and no
-        # ancestors, and delete them
+        # ancestors (that have no descendants), and delete them
         to_delete = [x for x in artifact.ancestors.nodes()
                      if x.id != artifact_id and x.parents and
                      not [y for y in x.descendants.nodes()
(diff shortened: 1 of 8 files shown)


qiita_db/util.py
```diff
@@ -1950,6 +1952,10 @@ def open_file(filepath_or, *args, **kwargs):
             fh.close()
 
 
+def artifact_visibilities_to_skip():
+    return tuple([qdb.util.convert_to_id('archived', "visibility")])
```
charles-cowart, line 1956: A new function for one line of code seems a little excessive. However I appreciate the readability. Not sure I would suggest unfolding it.

qiita_db/artifact.py
```diff
@@ -655,6 +655,73 @@ def delete(cls, artifact_id):
...
+        ----------
+        artifact_id : int
+            The artifact to be archived
+
+        Raises
+        ------
+        QiitaDBOperationNotPermittedError
+            If the artifact is not public
+            If the artifact_type is not BIOM
+            If the artifact belowns to an analysis
```
charles-cowart, line 672: belowns -> belongs

qiita_db/artifact.py
```diff
@@ -655,6 +655,73 @@ def delete(cls, artifact_id):
...
+                'Only BIOM artifacts can be archived')
+        if artifact.analysis is not None:
+            raise qdb.exceptions.QiitaDBOperationNotPermittedError(
+                'Only non analysis artifacts can be archived')
+        if not artifact.parents:
+            raise qdb.exceptions.QiitaDBOperationNotPermittedError(
+                'Only non raw artifacts can be archived')
+
+        # let's find all ancestors that can be deleted (it has parents and no
+        # ancestors, and delete them
```
charles-cowart, line 691: ancestors -> descendants?

charles-cowart (changes requested):

qiita_db/util.py
antgonza: I tried adding as a global variable in the artifact.py and as a member of the Artifact object but turns out that this is not possible because the code will try to initialize those variables when the code is initiated (and there is no database) so it fails, [for example](https://github.com/qiita-spots/qiita/runs/6801800785?check_suite_focus=true).

PR #3204, opened 2022-06-15
send job emails to users

antgonza:


charles-cowart (approved):
Approved. Everything looks in order.

PR #3205, opened 2022-06-16
allow artifact summary update

antgonza:
This will allow us to update artifact summaries via: `artifact_summary_post_request(user, artifact_id, force_creation=True)` and while adding this, I realized that we were removing the current file if the file had the same name that the previous version.

### DIFF
--- a/qiita_db/artifact.py
+++ b/qiita_db/artifact.py
@@ -1131,10 +1131,12 @@ class Artifact(qdb.base.QiitaObject):
             the HTML file
         """
         with qdb.sql_connection.TRN:
-            if self.html_summary_fp:
-                # Delete the current HTML summary
+            old_summs = self.html_summary_fp
+            to_delete_fps = []
+            if old_summs:
+                # Delete from the DB current HTML summary; below we will remove
+                # files, if necessary
                 to_delete_ids = []
-                to_delete_fps = []
                 for x in self.filepaths:
                     if x['fp_type'] in ('html_summary', 'html_summary_dir'):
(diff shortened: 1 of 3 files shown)


qiita_db/artifact.py
```diff
@@ -1171,6 +1162,19 @@ def path_cleaner(fps):
...
+        if to_delete_fps:
+            for x in self.filepaths:
+                if x['fp'] in to_delete_fps:
+                    to_delete_fps.remove(x['fp'])
+
+            for fp in to_delete_fps:
+                if isfile(fp):
+                    remove(fp)
+                else:
+                    rmtree(fp)
```
charles-cowart, line 1176: rmtree() can take a callback function to handle errors or ignore failures to delete. Just to confirm, is it better to just have it raise an Error here instead of using a handler or ignoring it?

charles-cowart (changes requested):
Approved. Everything appears in order. Just want to confirm one behavior before approving.

qiita_db/artifact.py
antgonza: it has been fine so far ... but if you have specific examples of what to add and why, please let me know.

charles-cowart: True. I just realized path_cleaner() is using the same parameters. I was thinking it might be good to avoid an Error in that part of the code, but it hasn't caused issues so far, and I'd need to look deeper into it in order to provide a specific example.

charles-cowart (approved):

PR #3206, opened 2022-06-28
2022.07

antgonza:


qiita_db/processing_job.py
```diff
@@ -781,14 +784,22 @@ def _set_status(self, value):
...
-                    qdb.util.send_email(self.user.email, subject, message)
+            if value not in ('waiting'):
+                if self.user.info['receive_processing_job_emails']:
+                    # skip if software is artifact definition
+                    ignore_software = ('artifact definition', )
+                    if self.command.software.name not in ignore_software:
+                        ignore_commands = ('Validate', 'complete_job',
+                                           'release_validators')
+                        if self.command.name not in ignore_commands:
+                            subject = 'Job status change: %s (%s)' % (
```
charles-cowart, line 795: I like to use %s myself when I have the choice. Just want to confirm that it's okay to use %s and not f'{}' in new code.

charles-cowart (changes requested):
Everything else looks good. One possible change and one request for clarification.

qiita_db/processing_job.py
charles-cowart: Agreed this is a matter of preference and what works well for placating flake8.

charles-cowart (approved):
Approved!

PR #3207, opened 2022-07-14
Updated INSTALL.md Instructions

aliu104:
Hello, my name is Annie Liu, and I am an undergraduate researcher currently working with Professor Niema Moshiri. This summer, along with a few other undergraduates, I will be working on creating a Qiita plugin that integrates many tools created by Professor Tajana Rosing and Professor Rob Knight’s students. To test this plugin, we installed a Qiita server locally, using the current [INSTALL.md](https://github.com/qiita-spots/qiita/blob/master/INSTALL.md) instructions. This pull request consists of some missing steps and clarifications to the INSTALL.md that were helpful for us in installing the local Qiita server.


INSTALL.md
```diff
@@ -19,6 +19,7 @@ conda update conda
 Setup a virtual environment in conda named `qiita` by executing the following:
 
 ```bash
+conda config --add channels conda-forge
```
charles-cowart, line 22: @aliu104 Annie, can you confirm that conga-forge is needed for Qiita's environment? I know that it's a requirement for some of our plugins and other projects. However I'm not sure if it's needed for Qiita itself.

INSTALL.md
```diff
@@ -54,7 +55,7 @@ source deactivate
 Install the non-python dependencies
 -----------------------------------
 
-* [PostgreSQL](http://www.postgresql.org/download/) (minimum required version 9.5.14, we have tested most extensively with 9.5.15)
+* [PostgreSQL](http://www.postgresql.org/download/) (currently using v13)
```
charles-cowart, line 58: @aliu104 Good catch! Thank you!

INSTALL.md
```diff
@@ -64,7 +65,10 @@ There are several options to install these dependencies depending on your needs:
 - Alternatively, you could install them via conda. However, the conda repository may not have the exact versions of these dependencies that you want.
 - You could setup a full development environment with [Vagrant](https://www.vagrantup.com/), and continue using conda under it to primarily manage python dependencies. Note that we don't cover Vagrant in these instructions.
 
-### PostgreSQL installation on Mac OS X
+### PostgreSQL installation on Linux
+For Linux, you can install Postgres through these [instructions](https://computingforgeeks.com/how-to-install-postgresql-13-on-ubuntu/). These instructions were tested on Postgres v13 on Ubuntu v20.04.4.
```
charles-cowart, line 69: @aliu104 Thank you for adding this information. However, this link looks to be a blog post and as such, we can't rely on it to be available in the future. I would suggest paraphrasing their instructions and citing them as a source, or citing a webpage from say the PostgreSQL homepage.

INSTALL.md
```diff
@@ -83,6 +87,10 @@ Assuming you have [homebrew](http://www.brew.sh) installed, you can install redi
 brew update
 brew install homebrew/versions/redis28
 ```
+Alternatively, you can sudo install the latest version of redis:
```
charles-cowart, line 90: @aliu104 In the past, we've tried to encourage organizations to use 2.8, since that is/was used in production. It might be confusing to the user to specify such an old version on the one hand (2.8) and immediately after give instructions on how to install the latest version. I would encourage you to test Qiita with the latest version of Redis and if it appears to work correctly, remove the instructions for 2.8.

INSTALL.md
```diff
@@ -112,7 +120,7 @@ Install Qiita development version and its python dependencies
 Clone the git repository with the development version of Qiita into your current directory:
 
 ```bash
-git clone https://github.com/biocore/qiita.git
+git clone https://github.com/qiita-spots/qiita.git
```
charles-cowart, line 123: @aliu104 good catch. :)

INSTALL.md
```diff
@@ -121,12 +129,17 @@ Navigate to the cloned directory and ensure your conda environment is active:
 cd qiita
 source activate qiita
 ```
-
+Before we install Qiita, make sure you have a few other tools installed:
+```bash
+sudo apt install gcc
```
charles-cowart, line 134: @aliu104 That's right, ubuntu doesn't install a CC by default. I would rephrase and ask the user to type 'cc' first. If the response is 'program not found' or similar, then install either gcc or clang, depending on their preference. Gcc specifically isn't a requirement - we just need a C++ compiler. This makes the documentation appropriate for organizations using distributions other than Ubuntu, where CC might already be installed.

INSTALL.md
```diff
@@ -121,12 +129,17 @@ Navigate to the cloned directory and ensure your conda environment is active:
...
+sudo apt-get install libpq-dev
```
charles-cowart, line 135: @aliu104 This is also specific to Ubuntu. I would suggest adding some text along the lines of - Make sure development libraries and include files for PostgreSQL are available by installing libpq-dev for Ubuntu or similar for other distributions if needed.

INSTALL.md
```diff
@@ -148,7 +161,7 @@ Move the Qiita sample configuration file to a different directory by executing:
  cp ./qiita_core/support_files/config_test.cfg ~/.qiita_config_test.cfg
 ```
 
-Note that you will need to change `BASE_URL = https://localhost:8383` to `BASE_URL = https://localhost:21174` if you are not using NGINX.
+Note that you will need to change `BASE_URL = https://localhost:8383` to `BASE_URL = https://localhost:21174` in the new copy of the configuration file if you are not using NGINX. Additionally, you will also need to change all URLs that start with `/home/runner/work/qiita/qiita/...` into wherever your qiita directory is (e.g. `/home/<username>/qiita/...`).
```
charles-cowart, line 164: @aliu104 very nice. good catch.

INSTALL.md
```diff
@@ -162,7 +175,7 @@ Set your `QIITA_CONFIG_FP` environment variable to point to that file (into `.ba
 
 Update paths in the newly copied configuration file to match your settings, e.g. replace /home/travis/ with your user home directory.
 
-Next, make a test environment:
+If you are working on Linux WSL, you will need to start the redis server in a separate Ubuntu window with the command provided [here](https://github.com/qiita-spots/qiita/blob/master/INSTALL.md#start-qiita) before making a test environment. Next, make a test environment:
```
charles-cowart, line 178: @aliu104 Most people won't know what Linux WSL is. I would suggest expanding to Windows Subsystem for Linux (WSL), and then using WSL for any future mentions.

INSTALL.md
```diff
@@ -252,6 +265,23 @@ It might be necessary to restart postgresql: `sudo service postgresql restart`.
 
 Furthermore, the `pg_hba.conf` file can be modified to change authentication type for local users to trust (rather than, e.g., md5) but we haven't tested this solution.
 
+#### `FATAL: password authentification failed for user "postgres"`
+If you get a traceback similar to this one when making a test environment:
```
charles-cowart, line 269: @aliu104 I wouldn't include the traceback. It's specific to Jorge and if we started including traceback examples for every scenario the documentation would become unreasonably long. I would include just the error message encountered, and your solution. Note that your proposed solution lowers security, and that's something other users should be made aware of. I would include a reference to the PostgreSQL documentation on how to modify the pg_hba.conf file and I would change, "modifying the `pg_hba.conf`" to include the full path to the file on Ubuntu. I would also note that the path may be different for different distributions.

charles-cowart (changes requested):
A good first draft. Please see my attached comments.

INSTALL.md
aliu104: @charles-cowart Hi Charles, when I would run  `conda create -q --yes -n qiita python=3.6 pip libgfortran numpy nginx` without running  `conda config --add channels conda-forge`, it tells me that I failed with repodata from current_repodata.json, will try with next reodata source. It tells me that libgfortran is not available to download. However, after I ran the `conda config --add channels conda-forge` command, I was able to install libgfortran (see attached screenshots). This seemed to be a problem among me and 2 other students, who all had newly installed versions of Ubuntu.

Not using conda-forge:
image

Using conda-forge:
image

INSTALL.md
aliu104: @charles-cowart That makes a lot of sense! We installed the latest version of Redis, and we have not ran into any problems with running the Qiita server yet but will update if that changes.

INSTALL.md
```diff
@@ -64,7 +65,41 @@ There are several options to install these dependencies depending on your needs:
...
+```
+To make sure all changes have been reflected, restart the Postgres server:
+```bash
+sudo service postgresql restart
+```
+Installing Postgres is now complete. Note that you will need to start the Postgres server every time you start the Qiita server. You can do this with the following command:
+```bash
+sudo service postgresql start
+```
+### PostgreSQL installation on Mac OS X (Outdated)
```
charles-cowart, line 102: @aliu104 I'm confused as to why '(outdated)' has been added to this title. As far as I know, the installation instructions are still correct for MacOSX, and if they're not, they should be updated. Instructions marked as 'obsolete' doesn't make much sense, IMHO.

INSTALL.md
```diff
@@ -75,14 +110,18 @@ echo 'export PATH="$PATH:/Applications/Postgres.app/Contents/Versions/9.5/bin/"'
 source ~/.bash_profile
 ```
 
-### Redis-server installation on Mac OS X
+### Redis-server installation on Mac OS X or Linux
```
charles-cowart, line 113: @aliu104 I would suggest breaking this up into two sections, one for MacOSX and one for Linux just as you did above. The reason for this is home-brew is not a Linux thing (unless things have recently changed) and apt-get is very specifically an _Ubuntu_ thing (not even an Linux thing).

INSTALL.md
charles-cowart: @aliu104 Ah, it just occurred to me that Ubuntu is likely the core of WSL. In that case, these instructions work for WSL. I believe however that they also apply to installing on Ubuntu as well. If you can note that in the documentation it would be much appreciated.

charles-cowart (changes requested):
Thank you very much! I think draft two looks great! If you don't mind addressing the few comments I added below, I'll be happy to approve and merge this documentation Into the codebase!

INSTALL.md
aliu104: @charles-cowart We labelled these instructions as outdated temporarily since the instructions were for Postgres v9.5 and we did not have a Mac to test them. However, we are currently in the process of obtaining a Macbook and testing these instructions and will keep you posted!

INSTALL.md
aliu104: @charles-cowart I wasn't too familiar with home-brew, but after searching it up, it seems that home-brew also works for Linux now: https://brew.sh/. Should I still break up these sections?

INSTALL.md
aliu104: Done in draft two!

INSTALL.md
charles-cowart: @aliu104 let's not do temporarily, since these instructions become official as soon as I approve and merge them. Usually the time one would test such things is before issuing the PR. It's all right in this case to change v9.5 below to "v9.5, v13". All Qiita developers in Knight Lab and CMI use MacOSX and developed and tested the code against v9.5 and v13 databases using Postgres.app.

INSTALL.md
charles-cowart: @aliu104 Ooh that is interesting news. Yes, we should still break it up into sections, since the second method is specifically for Ubuntu. IMHO, The title of the first section should reflect on Homebrew and its cross-platform nature specifically. "Redis-server installation using Homebrew (Mac OS X, Linux)" would be my input.

charles-cowart (changes requested):

charles-cowart (approved):
@aliu104 Looks good! Thank you for contributing to Qiita!

charles-cowart:
@antgonza I realized today that this PR is pointed at master. Do you think it's okay to merge in this case? Or should we ask @aliu104 to cancel this PR and point to qiita-spots:dev instead?

antgonza:
Thank you @aliu104 for your contribution to Qiita.

@charles-cowart, thank you for reviewing this PR; I changed the branch to dev and merged it.

PR #3208, opened 2022-08-10
mv to slurm

antgonza:


qiita_core/configuration_manager.py
```diff
@@ -42,11 +42,11 @@ class ConfigurationManager(object):
...
     valid_upload_extension : str
         The extensions that are valid to upload, comma separated
-    trq_owner : str
-        Email address of submitter of Torque jobs
-    trq_poll_val : int
-        Interval (in seconds) to wait between calls to Torque's qstat program
-    trq_dependency_q_cnt : int
+    job_scheduler_owner : str
+        Email address of submitter of jobs
+    job_scheduler_poll_val : int
```
charles-cowart, line 47: Not really a request, but just an observation: job_scheduler sounds platform-neutral, but this really is a migration to slurm. slurm might be more descriptive because the code is slurm-specific.

charles-cowart (approved):

qiita_core/configuration_manager.py
antgonza: Glad that you pointed that out; the plan was to make it as neutral as possible so the next time we only need to change the slurm commands to whatever is new without having to change the variable names ... which was a pain!

PR #3209, opened 2022-08-10
update libs

antgonza:
There is a new scipy that break some of the dependencies so making sure to not install the latests + there is a new flake8 so there are new rules that we need to follow.


charles-cowart (approved):
Looks good. No changes requested.

PR #3213, opened 2022-09-20
add sample validation page for wetlab

sarayupai:


charles-cowart:
@sarayu @antgonza Waiting on unittests. Happy to review without them if you want the feedback.

qiita_pet/handlers/admin_processing_job.py
```diff
@@ -91,3 +92,40 @@ def get(self):
...
+    @coroutine
+    @execute_as_transaction
+    def get(self):
+        self._check_access()
+
+        self.render("sample_validation.html", input=True)
+
+    @execute_as_transaction
+    def post(self):
+
```
charles-cowart, line 107: I would consider removing this blank line at the top. Some coders begin functions and methods immediately after the declaration and some don't. However most if not all code in Qiita begins immediately after and it's good to be consistent.

qiita_pet/handlers/admin_processing_job.py
```diff
@@ -91,3 +92,40 @@ def get(self):
...
+        # Get user-inputted qiita id and sample names
+        qid = self.get_argument("qid")
+        snames = self.get_argument("snames").split()
+
+        qsnames = list(Study(qid).sample_template)
```
charles-cowart, line 112: It's often good to add one maybe two lines of comment in plain English describing what you're doing here. Something like, "Stripping leading Qiita ID from sample-names." would be good. Adding an example of what you expect to see and what you want to transform it into is even better. If something changes in the future that breaks this code, it makes it easier to detect that an expectation was changed.

qiita_pet/handlers/admin_processing_job.py
```diff
@@ -91,3 +92,40 @@ def get(self):
...
+        for i, qsname in enumerate(qsnames):
+            if qsname.startswith(qid):
+                qsnames[i] = qsname.replace(f'{qid}.', "", 1)
+
+        blank = []
+        for name in snames:
+            if name.lower().startswith('blank'):
+                blank.append(name)
+                snames.remove(name)
+
```
charles-cowart, line 122: I would consider replacing lines 117-121 with the following two lines below:
blank = [x for x in snames if x.lower().startswith('blank')]
snames = [x for x in snames if 'blank' not in x.lower()]

These are called list comprehensions and internally they are faster than anything within a 'for name in snames' loop will be. Currently you're iterating through a list and modifying it at the same time, which can lead to unpredictable behavior. The list comprehensions above create a new list based on snames without altering it. Instead the second line will replace the old snames with the new one once the loop has completed.

qiita_pet/test/test_admin_processing_job_handlers.py
```diff
@@ -45,5 +45,19 @@ def test_get_missing_argument(self):
                       response.body.decode('ascii'))
 
 
+class TestSampleValidation(BaseAdminTests):
+    def test_get(self):
```
charles-cowart, line 49: As mentioned during the meeting, a good unittest should canvas the common things that might break your method/function. For example, calling post without one or both of the post_args, or giving a value for qid or snames that is obviously bad. This confirms that error-handling exists and is working properly.

qiita_pet/test/test_admin_processing_job_handlers.py
```diff
@@ -45,5 +45,19 @@ def test_get_missing_argument(self):
...
+        response = self.get('/admin/sample_validation/')
+        self.assertEqual(response.code, 200)
+
+    def test_post(self):
+        post_args = {
+            'qid': 1,
+            'snames': 'SKB1.640202 SKB2.640194 BLANK.1A BLANK.1B'
+        }
+        response = self.post('/admin/sample_validation/', post_args)
+        self.assertEqual(response.code, 200)
```
charles-cowart, line 59: Some coders will return 200 as long as the command executed without failure on the server-side. It's good to check the response.text and confirm it does not contain error messages and contains at least some string of text you're expecting.

charles-cowart (changes requested):
Good job! Sorry it's a little wordy.

qiita_pet/handlers/admin_processing_job.py
sarayupai: I believe the blank line was recommended by flake8 - let me know if you would still like me to remove it

qiita_pet/handlers/admin_processing_job.py
sarayupai: Added comment + example

qiita_pet/handlers/admin_processing_job.py
sarayupai: Replaced lines 117-121 with code suggested above

qiita_pet/test/test_admin_processing_job_handlers.py
sarayupai: Added a unit test to check error handling of post with invalid qid

qiita_pet/test/test_admin_processing_job_handlers.py
sarayupai: Modified unit test to check that the response body contains the inputted sample names

qiita_pet/handlers/admin_processing_job.py
charles-cowart: I don't recall flake8 doing that. Yes please.

qiita_pet/test/test_admin_processing_job_handlers.py
```diff
@@ -45,5 +45,30 @@ def test_get_missing_argument(self):
...
+        snames = ['SKB1.640202', 'SKB2.640194', 'BLANK.1A', 'BLANK.1B']
+        body = response.body.decode('ascii')
+        for name in snames:
+            self.assertIn(name, body)
+
```
charles-cowart, line 64: @sarayupai Good job! But note that the test is called 'test_post' and you're testing positive and negative outcomes in this test. It's not obvious to the reader why this test should return 500. You have to compare each input to realize the only difference is 2 and perhaps that's an invalid id. It would be good to add a comment here summarizing what you're doing and what you expect to see happen.

charles-cowart (changes requested):
Good job! Please just add the one comment and we're good to go!

qiita_pet/handlers/admin_processing_job.py
sarayupai: Removed the blank line

qiita_pet/test/test_admin_processing_job_handlers.py
sarayupai: Added comments indicating the positive and negative tests

antgonza:
@charles-cowart, is this ready to be merged?

antgonza (approved):

antgonza:
Thank you @sarayupai and @charles-cowart !

PR #3217, opened 2022-10-31
fix #3216 + other smaller fixes

### LINKED ISSUE #3216: get_qiita_version should store and retrieve information from redis
Currently, every time a page gets displayed we check the current git version of the qiita code via "get_qiita_version", which uses GitPython. The issue is that it actually uses Popen to do this operation which means that it issues a new command every time, which is less than ideal.

antgonza:


qiita_core/util.py
```diff
@@ -96,6 +96,7 @@ def get_qiita_version():
     try:
         repo = Repo(git_repo_path)
         sha = repo.active_branch.commit.hexsha
+        repo.__del__()
```
charles-cowart, line 99: @antgonza this method is usually called automatically once the object falls out of scope. Is there a reason you want to explicitly call it? That seems a little odd. Maybe if you can add a comment saying why this is needed?

charles-cowart (changes requested):

qiita_core/util.py
antgonza: This: https://github.com/qiita-spots/qiita/issues/3216#issuecomment-1297345627

charles-cowart (approved):

PR #3218, opened 2022-10-31
fix #3214

### LINKED ISSUE #3214: use `sample_name` in sample information from a given prep
@justinshaffer just reported that when you download a sample information file from a given prep, the file uses sample_id as the index of the file; however, this is not a valid value for Q2. We should use `sample_name`.

antgonza:


charles-cowart (approved):

antgonza:
@charles-cowart, tests are now passing. Thank you.

PR #3220, opened 2022-11-08
Updates after deploy 2022.11

antgonza:
This is messier than what I expected because by mistake we have been merging PRs to dev and master; sorry for the confusion.


qiita_db/metadata_template/base_metadata_template.py
```diff
@@ -63,8 +63,10 @@ def _helper_get_categories(table):
                  WHERE sample_id = '{1}'""".format(table, QIITA_COLUMN_NAME)
         qdb.sql_connection.TRN.add(sql)
         results = qdb.sql_connection.TRN.execute_fetchflatten()
-        if results:
+        if results and results != [None]:
```
charles-cowart, line 66: Just to confirm, it is intentional that this is testing to make sure results isn't a list with just one element that is None, rather than an empty list []?

antgonza: Yup

charles-cowart (approved):

PR #3229, opened 2023-01-10
add filesize ArtifactHandler

antgonza:


qiita_db/handlers/tests/test_artifact.py
```diff
@@ -66,9 +66,13 @@ def test_get_artifact(self):
...
         exp_fps = {
             "raw_forward_seqs":
-                [path_builder('1_s_G1_L001_sequences.fastq.gz')],
+                [{'filepath': path_builder('1_s_G1_L001_sequences.fastq.gz'),
+                  'size': 58}],
             "raw_barcodes":
-                [path_builder('1_s_G1_L001_sequences_barcodes.fastq.gz')]}
+                [{'filepath': path_builder(
+                  '1_s_G1_L001_sequences_barcodes.fastq.gz'),
+                  'size': 58}]
```
charles-cowart, line 74: I'm wondering if it would be good to describe the nature of these two 58s and the 1093210 below with a comment. Intentionally small file size and intentionally larger file size and all that.

charles-cowart (changes requested):
Everything looks good. One possible suggestion for @antgonza to review. Will approve and merge after his reply.

qiita_db/handlers/tests/test_artifact.py
antgonza: I don't think there is anything behind those numbers, they are just what's preloaded in the database.

charles-cowart: Ok.

charles-cowart (approved):

PR #3232, opened 2023-01-21
PATCH /api/v1/study/1/samples now accepts new categories

charles-cowart:
PATCH /api/v1/study/1/samples now silently accepts metadata with new categories in them when updating a sample or adding a new one. Existing samples will be assigned a null value while samples in the metadata will be assigned the value or empty string '' if one was not given.

(This is in-line with the existing code)


charles-cowart:
Closing this PR as the branch was built against master and directed against master. The changes need to be against dev branch and directed against dev branch.

PR #3235, opened 2023-01-21
/api/v1/study/1/samples accepts new categories

charles-cowart:
calling PATCH on /api/v1/study/1/samples now silently accepts metadata with new categories in them when updating a sample or adding a new one. Existing samples will be assigned a null value while samples in the metadata will be assigned the value or empty string if one was not given.

### DIFF
--- a/qiita_pet/handlers/rest/study_samples.py
+++ b/qiita_pet/handlers/rest/study_samples.py
@@ -172,12 +172,7 @@ class StudySamplesHandler(RESTHandler):
             if set(data.columns).issubset(categories):
                 self.fail('Not all sample information categories provided',
                           400)
-            else:
-                unknown = set(data.columns) - categories
-                self.fail("Some categories do not exist in the sample "
-                          "information", 400,
-                          categories_not_found=sorted(unknown))
-            return
+                return
 
         existing_samples = set(sample_info.index)
         overlapping_ids = set(data.index).intersection(existing_samples)
@@ -186,15 +181,19 @@ class StudySamplesHandler(RESTHandler):
 
         # warnings generated are not currently caught
         # see https://github.com/biocore/qiita/issues/2096
-        if overlapping_ids:
-            to_update = data.loc[overlapping_ids]
-            study.sample_template.update(to_update)
-            status = 200
 
+        # processing new_ids first allows us to extend the sample_template
+        # w/new columns before calling update(). update() will return an
+        # error if unexpected columns are found.
         if new_ids:
             to_extend = data.loc[new_ids]
             study.sample_template.extend(to_extend)
             status = 201
+
+        if overlapping_ids:
+            to_update = data.loc[overlapping_ids]
+            study.sample_template.update(to_update)
+            status = 200
 
         self.set_status(status)
         self.finish()
--- a/qiita_pet/handlers/rest/study_samples.py
+++ b/qiita_pet/handlers/rest/study_samples.py
@@ -168,11 +168,11 @@ class StudySamplesHandler(RESTHandler):
 
         categories = set(study.sample_template.categories)
 
-        if set(data.columns) != categories:
-            if set(data.columns).issubset(categories):
-                self.fail('Not all sample information categories provided',
-                          400)
-                return
+        if set(data.columns) != categories and set(data.columns).issubset(
+                categories):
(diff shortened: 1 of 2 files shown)


charles-cowart:
singlularly, the tests complete successfully, but locally as a whole 'nosetests' fails on these tests as well. It seems like an issue with setup and teardown. Will look into it tomorrow night if I have time.

qiita_pet/handlers/rest/study_samples.py
```diff
@@ -172,12 +172,7 @@ def patch(self, study_id):
             if set(data.columns).issubset(categories):
```
antgonza, line 172: Should this and the previous if be merged?

qiita_pet/test/rest/test_study_samples.py
```diff
@@ -33,6 +33,120 @@ def _sample_creator(ids):
...
+
+        # confirm new categories are a part of the samples.
+        response = self.get('/api/v1/study/1/samples/info',
+                            headers=self.headers)
+        self.assertEqual(response.code, 200)
+        obs = json_decode(response.body)
+        self.assertIn('new_field1', obs['categories'])
+        self.assertIn('new_field2', obs['categories'])
+
+        # TODO: need a test to get metadata for 1.SKM1.999998 and 1.SKM1.999999
```
antgonza, line 134: rm?

antgonza (commented):
Looks good, thank you @charles-cowart. Just some minor comments.

antgonza:
@charles-cowart, this PR has some (minor) real errors on CI.

qiita_pet/handlers/rest/study_samples.py
charles-cowart: Good catch!

qiita_pet/test/rest/test_study_samples.py
```diff
@@ -91,21 +209,75 @@ def test_patch_sample_ids_exist_incomplete_metadata(self):
...
+        self.assertEqual(response.code, 201)
+        # successful response should be empty string
+        self.assertEqual(response.body, b'')
+
+        # If new categories are added, patch() should succeed.
+        # New and existing samples should have new categories.
+        # 640201 is an existing sample. blank.a2 is a new sample
+        body = _sample_creator(['1.SKM8.640201',
+                                'blank.a2'], categories=current)
+        # body['blank.a2']['DOES_NOT_EXIST'] will be '', not None.
```
antgonza, line 228: is this True? I think is foo, no? Same for WHAT/bar ...

antgonza (changes requested):
Thank you @charles-cowart ... one more minor comment.

qiita_pet/test/rest/test_study_samples.py
charles-cowart: Nope. Confirmed Friday that whenever you add a new category like we're doing here, all samples mentioned in the body will have their values set to the value you explicitly define or ''. All samples NOT mentioned in the body will have their values set to None.

qiita_pet/handlers/rest/study_samples.py
```diff
@@ -168,15 +168,10 @@ def patch(self, study_id):
...
-        if set(data.columns) != categories:
-            if set(data.columns).issubset(categories):
-                self.fail('Not all sample information categories provided',
-                          400)
-            else:
-                unknown = set(data.columns) - categories
-                self.fail("Some categories do not exist in the sample "
-                          "information", 400,
-                          categories_not_found=sorted(unknown))
+        if set(data.columns) != categories and set(data.columns).issubset(
```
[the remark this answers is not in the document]
charles-cowart: What 171 would do is catch all subsets of categories up to but not including an exact match for categories. Sample_template.update() handles the condition you mentioned.

Although now that you bring it up, after swapping .extend() and .update(), new categories could still be added successfully and the rest call would still fail on updates - leaving the system in an unexpected state after an error. I'll take a look.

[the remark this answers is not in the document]
charles-cowart: Resolved. Thanks for the catch! We now test for superset and raise a 400 if False.

antgonza:
Thank you @cassidysymons and @charles-cowart !

PR #3238, opened 2023-01-26
Fix #3237 Add prep ID to artifacts list included in an analysis

### LINKED ISSUE #3237: Add prep ID to list of artifacts included in an analysis
Hello,

I think it would be useful to include the prep ID for the artifacts listed as included in an analysis. This will make it easier to track back to individual artifacts.

Thank you!

charles-cowart:
Updated UI and API call to return a list of artifact.prep_template ids included in each analysis listed.

### DIFF
--- a/qiita_pet/handlers/analysis_handlers/base_handlers.py
+++ b/qiita_pet/handlers/analysis_handlers/base_handlers.py
@@ -74,9 +74,10 @@ def analysis_description_handler_get_request(analysis_id, user):
     artifacts = {}
     for aid, samples in analysis.samples.items():
         artifact = Artifact(aid)
+        foo = list(set([str(x.id) for x in artifact.prep_templates]))
         study = artifact.study
         artifacts[aid] = (
-            study.id, study.title, artifact.merging_scheme, samples)
+            study.id, study.title, artifact.merging_scheme, samples, foo)
 
     return {'analysis_name': analysis.name,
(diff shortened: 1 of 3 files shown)


qiita_pet/handlers/analysis_handlers/base_handlers.py
```diff
@@ -74,9 +74,10 @@ def analysis_description_handler_get_request(analysis_id, user):
     artifacts = {}
     for aid, samples in analysis.samples.items():
         artifact = Artifact(aid)
+        foo = list(set([str(x.id) for x in artifact.prep_templates]))
```
antgonza, line 77: Can you use a better name for the "foo" variable? Maybe prep_ids? Also, not sure why we need to convert to list, a set should be fine, no?

antgonza (changes requested):
One comment. Also, could you copy/paste a screenshot of how this looks (the new analysis list) in your local Qiita install? Thank you.

qiita_pet/handlers/analysis_handlers/base_handlers.py
charles-cowart: doh! Thanks for the catch. :)

charles-cowart:
prep_ids

antgonza:
@charles-cowart, thank you. Waiting for CI to finish.

antgonza (approved):

PR #3241, opened 2023-01-27
Fix #3222

### LINKED ISSUE #3222: Deleting artifacts should not move logs/summaries to the upload folder
Currently, when you delete a sequence artifact, the code moves the summary file (qtp-sequencing-validate-data.csv) to the upload folder, which then blocks the upload functionality.

charles-cowart:


qiita_db/artifact.py
```diff
@@ -618,13 +618,17 @@ def delete(cls, artifact_id):
                      WHERE artifact_id IN %s"""
             qdb.sql_connection.TRN.add(sql, [all_ids])
 
+            # do not move these files back to upload folder.
```
antgonza, line 621: Only one question: why add a parameter to `move_filepaths_to_upload_folder` vs. adding it directly to it? I mean there is already a `if x['fp_type'] in ('html_summary', 'html_summary_dir'):` where the `do_not_move` list could also be checked. What do you think? My concern is that now we have 2 places to define this.

charles-cowart: I'm for that. It is true we are already hard-coding one list, and the hardcode seems reasonable.

antgonza (approved):

antgonza:
Thank you.

PR #3244, opened 2023-01-31
Fix #3230

### LINKED ISSUE #3230: Additional information in job status change Qiita emails
(Not sure if this is the right place for this!)

It would be beneficial to have additional information in "Job Status Change" emails regarding the artifact that is moving through the workflow pipeline (as well as error emails). This could include prep # (& data type) and Qiita IDs affected and can be located before the "New Status" line:
Screen Shot 2023-01-18 at 4 08 41 PM

This is helpful when you are getting multiple emails from multiple jobs you are running, or if you are coming back from out of office and need a refresher on what jobs and projects are being referred to.

charles-cowart:
Status change emails now more informative

### DIFF
--- a/qiita_db/processing_job.py
+++ b/qiita_db/processing_job.py
@@ -714,19 +714,19 @@ class ProcessingJob(qdb.base.QiitaObject):
 
         if value == 'waiting':
             # notification not needed.
-            return
+            return None
 
         if not self.user.info['receive_processing_job_emails']:
             # notification not needed.
-            return
+            return None
 
         if self.command.software.name in ignored_software:
             # notification not needed.
-            return
+            return None
 
         if self.command.name in ignored_commands:
             # notification not needed.
-            return
+            return None
 
         # generate subject line
         subject = 'Job status change: %s (%s)' % (self.command.name, self.id)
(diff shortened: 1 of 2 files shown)


charles-cowart:
No existing tests on email generation to speak of. Jobs in testing appear to be Processing Jobs type. As discussed we may consider this upgrade w/out tests.

antgonza (changes requested):
@charles-cowart, I like the new `_notify_updated_status` and code looks good. However, after your changes, I think `_notify_updated_status` could become something like `_generate_notification_message` that will return a subject/message strings and if empty nothing will be sent via email (while if they are not empty an email will be sent). This separation will allow to test the subject/message generation; what do you think?

antgonza:
I think (at minimum) you can use `Artifact(1)._generate_notification_message` and `Artifact(8)._generate_notification_message`

qiita_db/processing_job.py
```diff
@@ -703,6 +703,101 @@ def status(self):
...
+            # notification not needed.
+            return
+
+        if self.command.software.name in ignored_software:
+            # notification not needed.
+            return
+
+        if self.command.name in ignored_commands:
+            # notification not needed.
+            return
```
antgonza, line 729: Why not return None explicitly (for reference: https://realpython.com/python-return-statement/#returning-none-explicitly)?

antgonza (changes requested):
Some minor comments.

antgonza:
BTW you need to pull from dev as the error you are finding has already fixed in that version.

qiita_db/processing_job.py
charles-cowart: I'm agnostic on this. Personally I see the value in being explicit but I understand pythonistas might find it redundant. I'll make the change.

PR #3247, opened 2023-02-08
fix #3246

### LINKED ISSUE #3246: make sure that the ebi-ena null values are lower case
We currently don't force ebi-ena null values to be lower case but we should. 

The code is something like this:
```
update_values = {
    'not collected': 'not collected',
    'not provided': 'not provided',
    'restricted access': 'restricted access',
    'not applicable': 'not applicable',
    'unspecified': 'not applicable',
    'not_collected': 'not collected',
    'not_provided': 'not provided',
    'restricted_access': 'restricted access',
    'not_applicable': 'not applicable',
    'missing: not collected': 'not collected',
    'missing: not provided': 'not provided',
    'missing: restricted access': 'restricted access',
    'missing: not applicable': 'not applicable',
}

df = [sample_or_prep_object].to_dataframe().fillna("").applymap(str.lower)
ddf = df[df.isin(update_values.keys()).any(axis=1)]
if ddf.shape[0] != 0:
    cols = [c for c in ddf.columns if set(update_values) & set(ddf[c].values)]
    to_replace = ddf[cols].copy()
    to_replace.replace(update_values, inplace=True)
    [sample_or_prep_object].update(to_replace)    
```

antgonza:


qiita_db/metadata_template/base_metadata_template.py
```diff
@@ -54,6 +54,22 @@
 # information
 QIITA_COLUMN_NAME = 'qiita_sample_column_names'
 
+INSDC_NULL_VALUES = {
+    'not collected': 'not collected',
+    'not provided': 'not provided',
+    'restricted access': 'restricted access',
+    'not applicable': 'not applicable',
+    'unspecified': 'not applicable',
```
charles-cowart, line 62: Just to confirm, unspecified should be mapped to non-applicable? It seems to be the only one without a unique string.

qiita_db/metadata_template/test/test_prep_template.py
```diff
@@ -311,6 +312,7 @@ def setUp(self):
                             'qiita_prep_id': 1000,
                             'instrument_model': 'Illumina MiSeq',
                             'library_construction_protocol': 'AAAA',
+                            'insdc_nulls': 'NoT applicable',
```
charles-cowart, line 315: 'NoT applicable' or 'Not applicable'?

charles-cowart (changes requested):
Everything appears to be good. Just need confirmation on the two questions.

qiita_db/metadata_template/base_metadata_template.py
antgonza: yup, long discussion on qiita.admin ...

qiita_db/metadata_template/test/test_prep_template.py
antgonza: yup, just to make sure that it works with any casing ...

charles-cowart: If it's not too much trouble, would you mind adding a comment regarding the casing issue? I feel like it would be easy for someone in the future to see that and think 'quick fix' without thinking it through.

antgonza: IMOO is an overkill but if you feel strongly about it, I'll add it ...

charles-cowart: It is in a test method so I'll defer. I'm less likely to think it's unintentional in a test method.

charles-cowart (approved):

PR #3259, opened 2023-02-28
Allow plugin prep creation

antgonza:


qiita_db/handlers/prep_template.py
```diff
@@ -126,3 +126,7 @@ def post(self):
         pt = qdb.metadata_template.prep_template.PrepTemplate.create(
             metadata, qdb.study.Study(study), data_type)
         self.write({'prep': pt.id})
+
+
+class PrepTemplateAPItestHandler(PrepTemplateAPIHandler):
```
charles-cowart, line 131: Is there are reason to continue having PrepTemplateAPItestHandler when it only contains a pass statement?

antgonza: The only reason reason I have is to keep a "logical" separation between the tests and the new endpoint but I guess they could both point to the same class ... what do you prefer?

charles-cowart: Both options have their downsides. I prefer logical separation though.

PR #3262, opened 2023-03-03
Added prep_id as an optional parameter

charles-cowart:
Added prep_id as an optional parameter to
.../study/description/<study_id>?prep_id=<prep_id>.

This will bring up the prep_id workflow page instead of summary information.

### DIFF
--- a/.github/workflows/qiita-ci.yml
+++ b/.github/workflows/qiita-ci.yml
@@ -108,7 +108,7 @@ jobs:
       - name: Install plugins
         shell: bash -l {0}
         run: |
-          wget https://data.qiime2.org/distro/core/qiime2-2022.11-py38-linux-conda.yml
+          wget --no-check-certificate https://data.qiime2.org/distro/core/qiime2-2022.11-py38-linux-conda.yml
           conda env create --quiet -n qtp-biom --file qiime2-2022.11-py38-linux-conda.yml
           rm qiime2-2022.11-py38-linux-conda.yml
           export QIITA_SERVER_CERT=`pwd`/qiita_core/support_files/server.crt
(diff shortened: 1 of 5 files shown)


charles-cowart:
@antgonza ready for review. We can remove the --no-check-certificate line in GitHub workflows whenever it's appropriate.

.github/workflows/qiita-ci.yml
```diff
@@ -108,7 +108,7 @@ jobs:
       - name: Install plugins
         shell: bash -l {0}
         run: |
-          wget https://data.qiime2.org/distro/core/qiime2-2022.11-py38-linux-conda.yml
+          wget --no-check-certificate https://data.qiime2.org/distro/core/qiime2-2022.11-py38-linux-conda.yml
```
antgonza, line 111: Please rm `--no-check-certificate`

antgonza (changes requested):
Works great, thank you! A couple of things:
1. if prep-id is not an integer, the code crashes :: expected behavior: do not change frame and display JS error like: "[prep-id] is not a valid preparation for this study".
2. if the prep-id doesn't exist in that study, the window hangs :: expected behavior: similar to point 1.

.github/workflows/qiita-ci.yml
charles-cowart: done!

charles-cowart:
WIP: Unit tests for the new error-checking need to be added, and a change needs to be made to have the error-page only show up in the main frame, but this gets the functionality we need.

antgonza:
Works as expected and code looks fine. Now, I originally thought a JS (JavaScript) error message should have been enough (see my original review) to let the user know that the prep-id doesn't exist and thought that full HTTPError might be to harsh; however, as this is already done with HTTPError, I think is fine to leave as is and if we get a user feedback we can change to a more simpler JS error + in theory these links should be computer generated.

antgonza (approved):

PR #3263, opened 2023-03-04
Fix #3249

### LINKED ISSUE #3249: Individual sample delete on sample information page not working
When individual samples are selected for deletion there is an error message returned that no samples were selected

charles-cowart:


charles-cowart:
I'm not an expert at Javascript, but it appears that the original string from linkFormatter was itself good. I was able to host the element inside an alert message and the toggle function was reached. It appears that SlickGrid did not like an attribute named 'onchanged'. In fact, it doesn't seem to like any any attribute starting with 'on'. Inside the inspector, you could see that all of the other attributes were loaded into the DOM correctly, but the value of 'onchanged' was there by itself, without a key. If I manually set a key for the value, everything worked as intended.

Eventually I opted to use SlickGrid's own event handler and that seemed to work well.

antgonza (changes requested):
Thank you @charles-cowart! Note that during testing I noticed that the check-all toggle doesn't work; see:


Could you also fix?

charles-cowart:
@antgonza  ready for review.

PR #3267, opened 2023-03-16
Updated reference genome information

justinshaffer:


antgonza:
This looks great, thank you. Just note that normally we don't push to master (only the dev branch) but due to the conflicts I'll merge to master and fix during dev deployment (which should happen really soon).

PR #3270, opened 2023-03-21
fix #3245

### LINKED ISSUE #3245: Public files archive includes "bytes" prefix
Just downloaded the public archive, and noticed that the filename is: `b'QIITA-public.tgz'`. All with quotations and the `b` prefix. File seems to open just fine otherwise.

antgonza:


charles-cowart (approved):
Everything appears in order. Approved.

PR #3276, opened 2023-04-11
fix #3219

### LINKED ISSUE #3219: Add new prep page should not allow to submit an empty file path.
Currently users can submit an empty file selection and the GUI will report that the "empty" file can't be found:
image.

charles-cowart:
To my knowledge, this appears to fix the outstanding issue. Validation of the two required fields is handled by bootstrap and jinja2. Unselected pulldowns will now display a default tool-tip bubble.

### DIFF
--- a/qiita_pet/templates/study_ajax/add_prep_template.html
+++ b/qiita_pet/templates/study_ajax/add_prep_template.html
@@ -6,6 +6,7 @@ $(document).ready(function () {
   // Hide the user and new ontology divs
   $("#user-ena-info").hide();
   $("#new-ena-info").hide();
+
   // If the ena-ontology select changes, check if we need to show the user list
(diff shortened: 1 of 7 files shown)


qiita_pet/templates/study_ajax/add_prep_template.html
```diff
@@ -26,18 +29,17 @@
     }
   });
 
-  // Make sure that the form gets validated prior submission
-  $("#create-prep-form").validate();
-
   // change the submit behavior
```
charles-cowart, line 32: @charles-cowart Removing this line is the actual proposed fix for 3219.

charles-cowart:
Superseded by https://github.com/qiita-spots/qiita/pull/3278

PR #3283, opened 2023-05-02
Updated INSTALL.md with apple silicon messages

CatFish47:


INSTALL.md
```diff
@@ -20,7 +20,7 @@ Setup a virtual environment in conda named `qiita` by executing the following:
 
 ```bash
 conda config --add channels conda-forge
-conda create -q --yes -n qiita python=3.9 pip libgfortran numpy nginx cython redis
+conda create -q --yes -n qiita python=3.9 pip libgfortran numpy nginx
```
charles-cowart, line 23: I would include cython and redis packages as well, just like we do in the GitHub workflows configuration file.

INSTALL.md
```diff
@@ -101,7 +101,7 @@ sudo service postgresql start
 ```
 ### PostgreSQL installation on Mac OS X
 
-For Mac OS X, you can either install postgres through the [Postgres.app](https://postgresapp.com/downloads.html). These instructions were tested with the Postgres.app v9.5, v13.
+For Mac OS X, you can install postgres through the [Postgres.app](https://postgresapp.com/downloads.html). These instructions were tested with the Postgres.app v9.5 and v13.
```
charles-cowart, line 104: I would go a little further and reword it to be, "For Mac OS X it is strongly encouraged to install Postgres through the Postgres.app."

IMOO, the second line implies that there are separate versions of the Postgres.app, namely 9.5 and 13. This is not the case. The Postgres.app allows you to dynamically create and run one or more instances of Postgres, each of which can be a different version. We should state instead that Qiita is supported on versions 9.5 and 13 of PostgreSQL itself, not the Postgres.app.

INSTALL.md
```diff
@@ -334,6 +334,13 @@ sudo apt-get install postgresql-contrib
 # or: sudo apt-get install postgresql-contrib-9.3 depending on your OS and apt repository versions
 ```
 
+### Apple Silicon Mac (M1/M2)
+
+#### `no such file or directory` or `fatal error: file not found`
+
+M1 and M2 macs have a new feature for homebrew where homebrew is not installed to the path `usr/local/bin` like Intel Macs are, but to `opt/homebrew/bin`. Since some old code likely hasn't been updated yet, this error could possibly be from the code looking into the old Intel Mac path. Make that homebrew libraries are being searched for in the `opt/homebrew/lib` path.
```
charles-cowart, line 341: Be assertive. "Since some old code likely hasn't been updated yet, this error could possibly be from the code looking into the old Intel Mac path." is so cautiously worded that it doesn't inspire faith in the installation document. 

It's good to attribute sources to buttress your assertions, but in this case I think it would be better if you paraphrased the relevant information from the URL and placed it here as well as attributing the source.

In part this is because this URL is to a third-party source that could easily disappear and important information would then be lost. Secondly, we should provide enough relevant information to the user so that traversing the link and processing an entire additional document is optional, rather than a requirement.

charles-cowart (changes requested):
A good first place. Please review my comments and make changes as needed. Thanks!

CatFish47:
@charles-cowart I've added in the above changes suggested. Let me know how it looks now!

charles-cowart (approved):
Approved! Looks good. Although attributing the website where you obtained the information would have been nice.

INSTALL.md
charles-cowart: Note that these were not added.

PR #3287, opened 2023-05-12
add name to PrepTemplateAPIHandler

antgonza:


antgonza:
@charles-cowart, this change is so the SPP could add the prep name (Run ID + lane number) via this endpoint.

antgonza:
BTW as this was a pretty small PR and we also needed to record the job-id that was creating any given prep, I just added that functionality here.

PR #3290, opened 2023-05-16
expose no_public to sample_template_checks

antgonza:


antgonza:
Before this change, the "delete samples" function will create a new job that will fail like this:
before

Now, it will fail with this **before** creating the job:
after

wasade:
Looks good!

PR #3294, opened 2023-06-01
fix #3227

### LINKED ISSUE #3227: Admin Sample Validation fails on non-existing study-id
If you pass a non-existing study to the sample validation the page brakes, it should tell you that the study doesn't exist.

antgonza:


qiita_pet/handlers/admin_processing_job.py
```diff
@@ -101,32 +102,39 @@ class SampleValidation(AdminProcessingJobBaseClass):
...
+
+        if error is None:
+            for i, qsname in enumerate(qsnames):
+                if qsname.startswith(qid):
+                    qsnames[i] = qsname.replace(f'{qid}.', "", 1)
+
+            # Remove blank samples from sample names
+            blank = [x for x in snames if x.lower().startswith('blank')]
+            snames = [x for x in snames if 'blank' not in x.lower()]
+
```
charles-cowart, line 131: Not really an issue for me but snames = set(names) - set(blank) would be a stronger choice, especially since snames is converted to a set down below.

qiita_pet/templates/sample_validation.html
```diff
@@ -14,9 +14,13 @@
 </style>
 {% end %}
 {% block content %}
-<h1>Sample Validation</h1><br>
-{% if input %}
-<form action="{% raw qiita_config.portal_dir %}/admin/sample_validation/" method="post" id="sample_validation_form"> 
+{% if input or error is not None %}
+  {% if error is not None %}
+    <div class="alert alert-danger" role="alert">
```
charles-cowart, line 19: Is there a better way to phrase this? Perhaps making them entirely separate if - elif? It seems odd to check for both and then immediately check for the latter and if so bail early, and if not do something entirely different?

qiita_pet/test/test_admin_processing_job_handlers.py
```diff
@@ -62,13 +62,15 @@ def test_post(self):
...
             self.assertIn(name, body)
+
         # Check failure: invalid qiita id
         post_args = {
             'qid': 2,
             'snames': 'SKB1.640202 SKB2.640194 BLANK.1A BLANK.1B'
         }
         response = self.post('/admin/sample_validation/', post_args)
-        self.assertEqual(response.code, 500)
+        self.assertEqual(response.code, 200)
```
charles-cowart, line 72: Just to confirm, you're changing the response returned from 500 to 200. This is okay, right?

charles-cowart (changes requested):
Looks good - just want to confirm a few things.

qiita_pet/templates/sample_validation.html
antgonza: Thought about it but that will require a full rewrite of the method AKA IMOO not worth it.

qiita_pet/test/test_admin_processing_job_handlers.py
antgonza: Yup, that's basically what this PR is solving a 500 error is not actually reported to the user - this is based on the configuration we have in nginx and to force that the errors are actually nicely catch/parse.

charles-cowart (approved):

PR #3295, opened 2023-06-02
fix-3275

antgonza:


qiita_pet/handlers/admin_processing_job.py
```diff
@@ -109,21 +110,43 @@ def post(self):
...
         qid = self.get_argument("qid")
         snames = self.get_argument("snames").split()
-        error, matching, missing, extra, blank = [None]*5
+        error, matching, missing, extra, blank, duplicates = [None]*6
 
         # Stripping leading qiita id from sample names
         # Example: 1.SKB1.640202 -> SKB1.640202
         try:
-            qsnames = list(Study(qid).sample_template)
+            st = Study(qid).sample_template
```
charles-cowart, line 118: IMHO st should be changed to something more substantial. I made the same mistake thinking it was a tmp variable but it's actually used farther down in the code.

qiita_pet/handlers/admin_processing_job.py
```diff
@@ -109,21 +110,43 @@ def post(self):
...
+                    if v in (None, 'None', ''):
+                        continue
+                    if k.startswith(qid):
+                        k = k.replace(f'{qid}.', "", 1)
+                    tube_ids[k] = v
+
             for i, qsname in enumerate(qsnames):
                 if qsname.startswith(qid):
-                    qsnames[i] = qsname.replace(f'{qid}.', "", 1)
+                    qsname = qsname.replace(f'{qid}.', "", 1)
```
charles-cowart, line 139: IMHO, I'm not a fan of big loops with lots of conditionals inside. I think the sanitizing of qsname and tube_ids could be handled better by list comprehensions before entering the loops. This would make the loops easier to read. However I understand this wasn't your code.

qiita_pet/test/test_admin_processing_job_handlers.py
```diff
@@ -63,6 +65,21 @@ def test_post(self):
         for name in snames:
             self.assertIn(name, body)
 
+        # Check succes with tube_id
```
charles-cowart, line 68: succes -> success

charles-cowart (changes requested):

qiita_pet/handlers/admin_processing_job.py
antgonza: I don't like multiple loops as things get slower and slower as things grow (remember we are dealing with GUI, users and lots of samples); thus, I prefer a single loop with a few conditionals inside.

qiita_pet/handlers/admin_processing_job.py
```diff
@@ -109,21 +110,43 @@ def post(self):
...
         except QiitaDBUnknownIDError:
             error = f'Study {qid} does not exist'
 
         if error is None:
+            # if tube_id is present then this should take precedence in qsnames
+            tube_ids = dict()
+            if "tube_id" in st.categories:
+                for k, v in st.get_category("tube_id").items():
+                    # ignoring empty values
+                    if v in (None, 'None', ''):
```
wasade, line 131: Should this be normalized on create/update of the sample template?

antgonza: To be honest, not sure but I think this is something we can improve depending on usage. I decided to use those values as just doing [this](https://github.com/qiita-spots/qiita/pull/3295/files#diff-bcfe36e4beb62086a028b012a89b4432f0812e664ff2214d1d9b7559a26c700bR69); basically returned 'None' in the not updated samples AKA I'm trying to be extra careful.

PR #3298, opened 2023-06-04
fix 3222 via fp_type

antgonza:
The [original fix](https://github.com/qiita-spots/qiita/pull/3241/files) relied on filenames, which meant that every new file (which depends on the plugins) needed to be added. This PR does the same but via the `fp_type`, which should substantially should reduce the need of adding new names.


charles-cowart (approved):
Everything appears in order!

PR #3299, opened 2023-06-07
2023.06-fixes-rc

antgonza:


qiita_pet/handlers/admin_processing_job.py
```diff
@@ -57,35 +59,44 @@ def get(self):
...
 
-        jobs = []
-        for ps in self._get_private_software():
-            for cmd in ps.commands:
-                if cmd.id != command_id:
-                    continue
+        with TRN:
+            # different versions of the same plugin will have different
+            # command_id, this will make sure to get them all (commands)
+            sql = """SELECT processing_job_id FROM qiita.processing_job
```
charles-cowart, line 65: I like subqueries better than JOINs as well, but I believe JOINs are still more performant in PostgreSQL. If this is a heavily-used chunk of code, you may consider revising.

charles-cowart (changes requested):
Everything looks good. I have just one suggestion, and it's just that.

qiita_pet/handlers/admin_processing_job.py
wasade: If this is performance sensitive, then I recommend evaluating directly with `EXPLAIN ANALYZE` rather than guessing as to ensure the query meets needs.

antgonza: @wasade, exactly what I also thought:
```
                                                                                    QUERY PLAN                                                                                    
----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
 Nested Loop  (cost=643.02..4050.96 rows=191 width=16) (actual time=1.169..2.582 rows=26 loops=1)
   ->  HashAggregate  (cost=642.59..642.68 rows=9 width=8) (actual time=0.928..0.931 rows=2 loops=1)
         Group Key: software_command.command_id
         Batches: 1  Memory Usage: 24kB
         ->  Hash Join  (cost=8.30..642.57 rows=9 width=8) (actual time=0.041..0.925 rows=2 loops=1)
               Hash Cond: ((software_command.name)::text = (software_command_1.name)::text)
               ->  Seq Scan on software_command  (cost=0.00..632.26 rows=726 width=55) (actual time=0.005..0.711 rows=726 loops=1)
               ->  Hash  (cost=8.29..8.29 rows=1 width=47) (actual time=0.010..0.010 rows=1 loops=1)
                     Buckets: 1024  Batches: 1  Memory Usage: 9kB
                     ->  Index Scan using pk_software_command on software_command software_command_1  (cost=0.28..8.29 rows=1 width=47) (actual time=0.006..0.007 rows=1 loops=1)
                           Index Cond: (command_id = 595)
   ->  Index Scan using idx_processing_job_command_id on processing_job  (cost=0.42..378.33 rows=37 width=24) (actual time=0.167..0.822 rows=13 loops=2)
         Index Cond: (command_id = software_command.command_id)
         Filter: ((NOT hidden) AND ((heartbeat > (CURRENT_DATE - '14 days'::interval day)) OR (heartbeat IS NULL)))
         Rows Removed by Filter: 347
 Planning Time: 0.771 ms
 Execution Time: 2.634 ms
```

wasade: Seems fast enough for right now?

antgonza: Agreed and IMOO the subquery is super fast for this case of uses: just give me all the command_ids with the same name -- as the table is small and using a JOIN might not work well.

charles-cowart: I'm convinced. If everyone agrees, may we consider this issue resolved and I can approve?

charles-cowart (approved):

PR #3304, opened 2023-07-12
adding script and notebook for resource allocations

antgonza:


.github/workflows/qiita-ci.yml
```diff
@@ -234,4 +234,4 @@ jobs:
     - name: lint
       run: |
         pip install -q flake8
-        flake8 qiita_* setup.py scripts/qiita*
+        flake8 qiita_* setup.py scripts/qiita* notebooks/*/*.py
```
charles-cowart, line 237: Just to confirm, the notebooks directory is a new directory that will contain just the .py file below? Just checking to confirm that your local branch doesn't have additional files you want added. Also, this will not check any ipynb notebooks. This is all right yes?

notebooks/resource-allocation/generate-allocation-summary.py
```diff
@@ -0,0 +1,107 @@
...
+    extra_info = ''
+    rvals = StringIO(check_output(sacct + [eid]).decode('ascii'))
+    _d = pd.read_csv(rvals, sep='|')
+    _d['QiitaID'] = j.id
+    cmd = j.command
+    s = j.command.software
+    try:
+        samples, columns, input_size = j.shape
+    except QiitaDBUnknownIDError:
+        continue
```
charles-cowart, line 36: I think it might be good to add a comment on why it's okay to ignore the QiitaDBUnknownIDError and continue, although I believe I understand why.

notebooks/resource-allocation/generate-allocation-summary.py
```diff
@@ -0,0 +1,107 @@
...
+    _d['columns'] = columns
+    _d['input_size'] = input_size
+    _d['extra_info'] = extra_info
+    _d.drop(columns=['Unnamed: 10'], inplace=True)
+    data.append(_d)
+
+df = pd.concat(data)
+
+df.to_csv(
+    f'/panfs/qiita/job_stats_{datetime.now().strftime("%m%d%y")}.tsv.gz',
```
charles-cowart, line 106: It might be good to use os.path.join() since we usually use it elsewhere.

charles-cowart (changes requested):

charles-cowart:
@antgonza I looked at the notebook and I think Sections 1 and 2 could use a little more description like Section 3 has. For example there are the following comments which don't really have an explanation:
# ['JobID', 'ElapsedRaw', 'MaxRSS', 'QiitaID', 'external_id', 'sId',
#  'sName', 'sVersion', 'cId', 'cName', 'samples', 'columns', 'input_size',
#  'MaxRSSRaw'],

Everything else appears okay. Not sure if your preference is to clear all of the outputs or leave them. I suppose it works better with the walkthrough to leave them.

.github/workflows/qiita-ci.yml
antgonza: Correct.

charles-cowart (approved):

PR #3307, opened 2023-07-19
add default-workflow to future artifact - SPP

antgonza:


qiita_db/handlers/tests/test_artifact.py
```diff
@@ -411,6 +411,58 @@ def test_post(self):
...
+        job = qdb.processing_job.ProcessingJob(jid)
+        children = [c.command.name for c in job.children]
+        grandchildren = [gc.command.name for c in job.children
+                         for gc in c.children]
+        self.assertEqual('Validate', job.command.name)
+        self.assertEqual(['Split libraries FASTQ'], children)
+        self.assertEqual(['Pick closed-reference OTUs'], grandchildren)
+
+        # just to avoid any tentative issues, let's wait for the main job to
+        # finish
```
charles-cowart, line 462: A solid idea, but just out of curiosity - is there any possible way this loop could never exit? If we have a job that's still registered in Qiita as running when it's errored out, will this still be spinning?

qiita_db/metadata_template/prep_template.py
```diff
@@ -726,13 +726,15 @@ def modification_timestamp(self):
     def max_samples():
         return qdb.util.max_preparation_samples()
 
-    def add_default_workflow(self, user):
+    def add_default_workflow(self, user, workflow=None):
```
charles-cowart, line 729: Sorry to ask this, but why would we ever call a function add_default_workflow() and not specify a workflow? Having a default of None seems very strange.

charles-cowart (changes requested):
Just a few questions....

qiita_db/handlers/tests/test_artifact.py
antgonza: If it errors, it will respond with that ... AKA it should not loop forever.

qiita_db/metadata_template/prep_template.py
antgonza: I see what you are saying, my guess is missing some context: basically this will add the commands in a default workflow (definition) to a preparation (thus, it lives in qiita_db/metadata_template/prep_template.py) and before these changes it would just create a new workflow (object) and add the commands of the default workflow (definition); and now if a workflow (object) is passed it will not create a new one but use the object passed. Does this make sense?

qiita_db/handlers/tests/test_artifact.py
charles-cowart: Okay.

qiita_db/metadata_template/prep_template.py
charles-cowart: Thank you for the explanation! I see what you're saying. I think adding this as a comment would be very helpful to future readers of the code. Perhaps changing the name to update_default_workflow() might be good too, maybe?

antgonza: I don't agree with that name cause it doesn't only update but adds/creates a new workflow; any other recommendation?

charles-cowart: maybe just having comment is best, yes?

charles-cowart (approved):

PR #3309, opened 2023-08-05
fix #3300

### LINKED ISSUE #3300: Select investigation type when prep is added with SPP
On the prep interface under summary - select WGS, RNA-Seq, amplicon or other depending on sequence/prep type

antgonza:


qiita_db/metadata_template/prep_template.py
```diff
@@ -136,6 +131,18 @@ def create(cls, md_template, study, data_type, investigation_type=None,
                 data_type_id = qdb.util.convert_to_id(data_type, "data_type")
                 data_type_str = data_type
 
+            # If the investigation_type is None let's add it based on the
+            # data_type being created - if possible
+            if investigation_type is None:
+                if data_type_str in {'16S', '18S', 'ITS'}:
```
charles-cowart, line 137: It might be good to do this as an enumerated type like we do in metapool and whatnot, but I can go either way.

charles-cowart (changes requested):
Looks good! One possible suggestion.

qiita_db/metadata_template/prep_template.py
antgonza: Do you mean like this: https://docs.python.org/3/library/enum.html ? If not, what do you mean? Also, I'm not sure of the benefits of going that route, could you expand?

wasade: Under the [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself) principle, unless this type of test is truly only done once ever

antgonza: Got it! ... and that comment made me remember TARGET_GENE_DATA_TYPES; changing for that.

PR #3312, opened 2023-08-11
install guide updates

AmandaBirmingham:
Some very small changes to clarify bits of the install guide, based on a trial usage :) ... now directed to the correct branch!

BTW, this PR's build fails, but so did the commit before it, so I don't think this markdown-only change is the cause ;)

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -1,5 +1,5 @@
 
-Qiita installation
+Qiita installation 
 ==================
 
 Qiita is pip installable, but depends on specific versions of python and non-python packages that must be installed first. We strongly recommend using virtual environments; a popular solution to manage them is [miniconda](http://conda.pydata.org/miniconda.html), a lightweight version of the virtual environment, python distribution, and package manager anaconda. These instructions will be based on miniconda.
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -181,7 +181,7 @@ github organization. Each of the plugins have their own installation instruction
 suggest looking at each individual .travis.yml file to see detailed installation
 instructions. Note that the most common plugins are:
 - [qtp-biom](https://github.com/qiita-spots/qtp-biom)
-- [qtp-target-gene](https://github.com/qiita-spots/qtp-target-gene)
+- [qtp-sequencing](https://github.com/qiita-spots/qtp-sequencing)
 - [qp-target-gene](https://github.com/qiita-spots/qp-target-gene)
 
 ## Configure Qiita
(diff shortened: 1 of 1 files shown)


INSTALL.md
```diff
@@ -181,7 +181,7 @@ github organization. Each of the plugins have their own installation instruction
 suggest looking at each individual .travis.yml file to see detailed installation
 instructions. Note that the most common plugins are:
 - [qtp-biom](https://github.com/qiita-spots/qtp-biom)
-- [qtp-target-gene](https://github.com/qiita-spots/qtp-target-gene)
+- [qtp-sequencing](https://github.com/qiita-spots/qtp-sequencing)
```
AmandaBirmingham, line 184: qtp-target-gene has been renamed to qtp-sequencing. The old link still works, but it is surprising (and potentially worrying :) to the reader to end up at a different URL than the one they expected.

INSTALL.md
```diff
@@ -206,9 +206,7 @@ Set your `QIITA_CONFIG_FP` environment variable to point to that file (into `.ba
   source activate qiita
 ```
 
-Update paths in the newly copied configuration file to match your settings, e.g. replace /home/travis/ with your user home directory.
```
AmandaBirmingham, line 209: There are no travis paths in the config anymore.

INSTALL.md
```diff
@@ -206,9 +206,7 @@ Set your `QIITA_CONFIG_FP` environment variable to point to that file (into `.ba
...
-
-If you are working on WSL, you will need to start the redis server with the following command before making a test environment:
+If you are working on Mac OS X or WSL, you will need to start the redis server with the following command before making a test environment:
```
AmandaBirmingham, line 209: Installing on mac (not just WSL) fails if one doesn't run the redis-server start ... ask me how I know :)

antgonza:
Thank you @AmandaBirmingham ! Could you pull from dev (`git pull upstream dev`) and push back (`git push origin dev`)? This should solve the build errors. 

BTW if you don't have upstream you can do `git remote add upstream https://github.com/qiita-spots/qiita`.

AmandaBirmingham:
Hi, @antgonza Unfortunately I do not think the build issue, whatever it is, is as simple as that. I synced my fork with the qiita-spots:dev branch before beginning this PR ... but just in case I did it wrong :), I have followed the directions you provided, and the message I got was `Everything up to date` (so the push did nothing).  I have also verified by manually looking into some of the actual files that your latest commit to qiita-spots:dev *has* been integrated into this branch, so it isn't an issue of the repos being out of sync.  

The build failure occurs early on, in the "Install plugins" stage, where it says:
`ERROR conda.core.link:_execute(952): An error occurred while installing package 'bioconda::bioconductor-genomeinfodbdata-1.2.9-r42hdfd78af_0'.`
... and this same failure occurred for my fork *before* I made any changes (see screenshot).
Screenshot 2023-08-14 at 12 11 31 PM

Maybe it is some sort of transient problem with bioconductor?  I will bump this with a small change and see if I get the same result today ...

AmandaBirmingham:
@antgonza Happily, it does look like it was some sort of intermittent bioconductor problem .. building again today with only an added space in the markdown gets everything passing :D

antgonza:
Thank you!

PR #3320, opened 2023-09-27
initial fixes after qiita-rc 09.2023

antgonza:


qiita_db/handlers/artifact.py
```diff
@@ -315,9 +315,9 @@ def post(self):
         values['template'] = prep_id
         cmd = qdb.software.Command.get_validator(atype)
         params = qdb.software.Parameters.load(cmd, values_dict=values)
-        if add_default_workflow:
+        if add_default_workflow or add_default_workflow == 'True':
```
charles-cowart, line 318: just to confirm, we expect add_default_workflow to be the string 'True' in that exact case? It seems like it should be add_default_workflow is True instead. Or add_default_workflow.lower() == 'true'.

charles-cowart (changes requested):

qiita_db/handlers/artifact.py
antgonza: We expect to either be True (python boolean) or it's text: 'True'

charles-cowart: gotcha!

charles-cowart (approved):

PR #3328, opened 2023-11-14
fix #3326

### LINKED ISSUE #3326: Display the last update timestamp for the sample and prep info files
Currently the full name of the sample/prep info files are not displayed (they contain the last time it was updated it) and it will be good to have it. Note that in the current display, you can see the older version (pressing the button on the image). 
Screen Shot 2023-11-07 at 4 53 41 PM

antgonza:
Now the sample info page looks like:
sample_info

------------------------
and the prep:
prep_info


charles-cowart (approved):
Everything appears in order.

PR #3330, opened 2023-11-16
Backend changes to fix 3323

antgonza:


qiita_db/support_files/patches/89.sql
```diff
@@ -0,0 +1,8 @@
+-- Nov 1, 2023
+-- add creation_job_id to qiita.prep_template
+ALTER TABLE qiita.analysis ADD slurm_reservation VARCHAR DEFAULT '' NOT NULL;
+ALTER TABLE qiita.user_level ADD slurm_parameters VARCHAR DEFAULT '--nice 10000' NOT NULL;
+
+UPDATE qiita.user_level SET slurm_parameters = '--nice 5000' WHERE name = 'admin';
+
+UPDATE qiita.user_level SET slurm_parameters = '' WHERE name = 'wet-lab admin';
```
charles-cowart, line 8: wet-lab admins are between regular users and admins, but it seems here that wet-lab admins will have the same priority as normal users. This is okay if slurm_parameters is only used for things like SPP jobs, but in general wet-lab admins should have higher priority than normal users right?

qiita_db/test/test_analysis.py
```diff
@@ -685,6 +685,14 @@ def test_is_public_make_public(self):
 
         self.assertTrue(analysis.is_public)
 
+    def test_slurm_reservation(self):
+        analysis = qdb.analysis.Analysis(1)
+        self.assertIsNone(analysis.slurm_reservation)
+        text = 'this is a tests!'
```
charles-cowart, line 691: technically 'tests!' -> 'test!'

qiita_db/test/test_processing_job.py
```diff
@@ -873,28 +876,28 @@ def _set_allocation(memory):
...
-            job_not_changed.get_resource_allocation_info(),
-            '-p qiita -N 1 -n 5 --mem 120gb --time 130:00:00')
-        self.assertEqual(job_changed.get_resource_allocation_info(),
-                         '-p qiita --mem 2G')
+            job_not_changed.resource_allocation_info,
+            '-p qiita -N 1 -n 5 --mem 120gb --time 130:00:00 --nice 10000')
+        self.assertEqual(job_changed.resource_allocation_info,
+                         '-p qiita --mem 2G --nice 10000')
 
         # restore allocation
```
charles-cowart, line 902: I haven't seen a test that confirms a return of niceness level 5000 for admins. Is this an easy test to add? It might be good to know that niceness can be something other than 10000.

charles-cowart (changes requested):
Everything appears in order. Just a few comments.

qiita_db/support_files/patches/89.sql
antgonza: The other way around, all users have a --nice 10000 value, which wet-lab admin have no value - this means that they will have the highest priority. In other words, the higher nice value the lower priority it has.

qiita_db/test/test_processing_job.py
antgonza: See https://github.com/qiita-spots/qiita/pull/3330/files#diff-99af21f52020eab43de2a74b4ddeb3b48475c35fcc8c07e3a08cef55e8cc3c61R519

charles-cowart (commented):
Approved.

PR #3333, opened 2023-11-27
Support parameters in workflows

antgonza:


qiita_db/metadata_template/prep_template.py
```diff
@@ -850,13 +850,35 @@ def _get_predecessors(workflow, node):
...
+                wk_params = wk.parameters
+                reqs_satisfied = True
+
+                if wk_params['sample']:
+                    df = ST(self.study_id).to_dataframe(samples=list(self))
+                    for k, v in wk_params['sample'].items():
+                        if k not in df.columns or v not in df[k].unique():
+                            reqs_satisfied = False
+
+                if wk_params['prep']:
```
charles-cowart, line 867: Just to confirm, wk_params['prep'] doesn't need to be an elif instead of an independent if? I believe it's correct now but just to be sure.

charles-cowart (changes requested):
Everything looks good - I can't find anything that looks off. Just one confirmation is needed.

qiita_db/metadata_template/prep_template.py
antgonza: That's correct, there can be sample, prep or both kinds of parameters.

charles-cowart (approved):

PR #3341, opened 2023-12-18
some fixes

antgonza:


CHANGELOG.md
```diff
@@ -1,5 +1,20 @@
 # Qiita changelog
 
+Version 2023.12
+---------------
+
+* The sample and preparation information pages will display the timestamp of their last update.
+* Added a ProcessingJob.complete_processing_job method to retrieve the job that is completing the current job.
+* Added a ProcessingJob.complete_processing_job method to retrieve the job that is completing the current job.
```
charles-cowart, line 8: This line looks like a duplicate of the one above.

qiita_db/metadata_template/prep_template.py
```diff
@@ -851,20 +851,30 @@ def _get_predecessors(workflow, node):
...
+        check_requirements = False
+        default_parameters = {'prep': {}, 'sample': {}}
+        if [wk for wk in all_workflows if wk.parameters != default_parameters]:
+            check_requirements = True
         ST = qdb.metadata_template.sample_template.SampleTemplate
-        for wk in qdb.software.DefaultWorkflow.iter():
+        workflows = []
+        for wk in all_workflows:
             if wk.artifact_type == pt_artifact and pt_dt in wk.data_type:
+                if check_requirements and wk.parameters == default_parameters:
```
charles-cowart, line 864: Just to make sure, '==' has higher precedence than 'and', so the above two statements are:
if ((wk.artifact_type == pt_artifact) and (pt_dt)) in wk.data_type:
                 if ((check_requirements) and (wk.parameters == default_parameters)):

Is this how you want it?

charles-cowart (changes requested):
Just want to confirm a few things.

qiita_db/metadata_template/prep_template.py
antgonza: Yes, took me a minute to remember but yes - if we need to check for requirements and the `wk.parameters` are `==` to `default_parameters`, then continue. In other words, when `check_requirements` is False do not skip anything but if is True check the values in wk.parameters to see if we should skip things below. 

Anyway, you comment actually made me realize that we don't need the lower code `check_requirements = True`; I'll remove them.

charles-cowart (approved):
Everything looks good!

PR #3343, opened 2024-01-03
escape(error)

antgonza:


jdereus:
XSS vulnerability was discovered in qiita url.  added escapes to prevent in the future.  additional testing/scanning will be requested going forward to isolate any further issues.

antgonza:
Thank you @jdereus, merging.

PR #3344, opened 2024-01-05
quick_mounts_purge

antgonza:


qiita_db/test/test_util.py
```diff
@@ -1291,6 +1291,12 @@ def test_purge_filepaths_test(self):
         fps_viewed = self._get_current_filepaths()
         self.assertCountEqual(fps_expected, fps_viewed)
 
+    def test_quick_mounts_purge(self):
+        # one of the tests creates a conflicting artifact_type so this test
```
charles-cowart, line 1295: Just to confirm, you're relying on state modified by another test for this test to work properly. That does seem to go against the idea of each test being independent and state being reset between tests. I'm okay with it, but how do we ensure that the test that creates the conflicting artifact_type is always called before this test?

qiita_db/util.py
```diff
@@ -896,6 +897,74 @@ def purge_filepaths(delete_files=True):
             qdb.sql_connection.TRN.execute()
 
 
+def quick_mounts_purge():
+    r"""This is a quick mount purge as it only slightly relies on the database
+
+    Notes
+    -----
+        Currenlty we delete anything older than 30 days that is not linked
```
charles-cowart, line 905: Currenlty -> Currently

qiita_db/util.py
```diff
@@ -896,6 +897,74 @@ def purge_filepaths(delete_files=True):
...
+        to the database. This number is intentionally hardcoded in the code.
+        At time of writting this number seem high but keeping it this way to be
```
charles-cowart, line 907: At time of writting -> At the time of this writing?

charles-cowart: changes -> changed.

qiita_db/util.py
```diff
@@ -896,6 +897,74 @@ def purge_filepaths(delete_files=True):
...
+    for f, s in stats.items():
+        report.append(f'{f}\t{naturalsize(s)}')
+    report.append(
+        f'Total files {len(to_delete)} {naturalsize(sum(stats.values()))}')
+    report.append('----------------------')
+
+    for td in list(to_delete):
+        if not exists(td):
+            continue
+        rmtree(td)
```
charles-cowart, line 963: if exists(td):
    rmtree(td)

might be more straightforward, no?

charles-cowart (changes requested):
Looks good! Just a few minor suggestions.

qiita_db/test/test_util.py
antgonza: Kind of, the issue is that other tests start and restart the database multiple times and some of them leave behind some artifacts so the test folder is populated with multiple versions of the artifacts - which we had no idea about. In specific, some leftover artifacts have the same id (2) in different mounts (in specific Demultiplexed and per-sample-FASTQ). I really don't have a strong opinion about fixing it but please open an issue to address that if you feel like we should fix.

charles-cowart: I think as long as it passes testing that's good enough for me.

charles-cowart:
Approved.

PR #3346, opened 2024-01-09
fix #3327

### LINKED ISSUE #3327: Limit the number of jobs or artifacts that a user can add to a workflow.
Currently there is no limit, which will result in users adding multiple jobs with multiple resulting artifacts and creating a cascading effect. For example, core diversity analysis jobs will create around 17 output artifacts; which means 3 jobs (main, validator, complete) for each artifact; thus, if a user creates 5 core diversity jobs, we will have: 5+3+(5*17*3) = 263 jobs.

antgonza:


charles-cowart (approved):
Approved! Everything looks good.

PR #3359, opened 2024-02-15
fix environment_script for private plugins

sjanssen2:
I found that patch 54.sql for the test database uses an old conda activate mechanism for travis. We might want to fix this to the latest github action method of choice?


antgonza:
This looks like a broken change; are you planning on fixing or should we close it?

Issue #3361, opened 2024-02-15
nginx missing websocket proxy


sjanssen2:
https://github.com/qiita-spots/qiita/blob/57d1b146729223da9daa286a1cc75db6f2923746/qiita_pet/nginx_example.conf#L59-L66
The provided example nginx configuration seems to not forward websocket HTTP traffic.
```
W 240215 18:58:08 web:2161] 400 GET /study/list/socket/ (127.0.0.1) 4.57ms
[D 240215 18:58:08 websocket:164] Can "Upgrade" only to "WebSocket".
```
Should we add a working configuration to this example, such that user do not have to solve this issue by themselves? Btw. I am not sure what the matching configuration would look like, can we follow: https://www.serverlab.ca/tutorials/linux/web-servers-linux/how-to-configure-nginx-for-websockets/ ?

antgonza:
Fixed here: https://github.com/qiita-spots/qiita/pull/3362

PR #3362, opened 2024-02-16
enable websocket communication

sjanssen2:
Addressing #3361 
extend the nginx example configuration file such that communication through WebSockets is automatically enabled.
I am not 100% sure if
  - [ ] the location definition is comprehensive. There might be more endpoints that need WebSocket communication
  - [x] the current settings cause unintended side effects

### DIFF
--- a/qiita_pet/nginx_example.conf
+++ b/qiita_pet/nginx_example.conf
@@ -13,6 +13,12 @@ http {
         server localhost:21177;
     }
 
+    # define variables for the actions that shall be taken for websocket handshake
+    map $http_upgrade $connection_upgrade {
+        default upgrade;
+        ''      close;
+    }
+
     # listening to 8080 and redirecting to https
     server {
            listen         8080;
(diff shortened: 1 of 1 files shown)


antgonza:
Thank you @sjanssen2; if we can't confirm the list of things in your last comment would it be better to add these changes and details to the INSTALL.md document? That way is available and admins/users can decide if they need to add these changes or not.

sjanssen2:
I @antgonza, last week, I have checked the added configuration to nginx with our local admins. They couldn't find any security related misconfiguration. Thus, I would say the second point is checked, while the first one about the locations is something you can answer best.

antgonza:
To be honest, I don't recall having to add those lines (or anything similar) to our nginx configuration but it sounds like it worked for you and your sys-admin. Thus, I think is fine to merge. Thank you.

sjanssen2:
Seems to depend on the actual routing of the traffic. Admins told me, we are running nginx behind an HA proxy which routes external HTTP**S** requests internally via HTTP. In this case, the additional lines are also probably not necessary - but wouldn't hurt either.

Issue #3363, opened 2024-02-16
no commands listed


sjanssen2:
I came across the situation that I configured and started plugins on top of the qiita_test system. Adding a new study, with new 16S prep and uploaded per_sample_fastq files led to an empty combo box when processing the per_sample_fastq artifact:

![image](https://github.com/qiita-spots/qiita/assets/11960616/069ec060-899b-436d-8370-87b27099ea38)

If tracked this down to the fact that the set of active commands `cids` and the commands from default workflows for 16S `cmds` had zero overlap. This might be an edge case, but would it be worth warning the user about this situation, i.e. print to the log that subsetting results in an empty command list?

https://github.com/qiita-spots/qiita/blob/58e15a470919b1183c6caf5af5f36f990f37f9ce/qiita_db/artifact.py#L1676

sjanssen2:
am I right that `qiita-env clean-test` does not clear the database from test entries, but rather restores the test data should they have been messed up via some test? If so, how could I best remove / update the defined "recommended workflows" such that their command_ids match those provided from the active plugins?

sjanssen2:
I found qp_target_gene/support_files/patches/171029_QIIME_v191_to_QIIMEq2_v191.sql which I guess is the source of the renaming of the plugin name (QIIME -> QIIMEq2), but this patch does not cover workflows. Probably because they did not exists at this time?

antgonza:
Just to answer your questions:
1. Yes, the `qiita-env clean-test` cleans/restores the qiita-test database so it can be in the same state for new tests to run. This is helpful when you need to tests something and the cleaning of the database is not possible. Note that the decorator `@qiita_test_checker` runs that code too at the end of each of the qiita-test objects; and that the system needs to be in tests mode so this is applied. 
2. QIIME->QIIMEq2 was done many (many ...) moons before the workflow idea was applied.

Now, in general plugin testing should happen via the testing framework (you can see any plugin for examples); which means that they are not normally shown in the GUI - mainly because at the end of the testing the database is reset.

Hope this helps.

antgonza:
Closing for now, please reopen if you have more questions.

sjanssen2:
I am coming back to this filtering as I do not fully understand the rationale here. I set up a fresh Qiita instance in test mode and installed the following plugins:

![Image](https://github.com/user-attachments/assets/c02f7ab5-053a-4074-9ebc-9df18195f093)

[...]

PR #3364, opened 2024-02-18
Fix cmdname qp target gene

sjanssen2:
Addressing #3363 by renaming the qp-target-gene command name to "QIIMEq2"


sjanssen2:
Yes, the command are available now. However, I am not sure if this rather cosmetic fixing is the most elegant way. For future testing, I feel this PR can help avoid tedious debugging.

PR #3371, opened 2024-02-23
Multiple inputs add default workflow

antgonza:


qiita_db/metadata_template/prep_template.py
```diff
@@ -894,8 +903,9 @@ def _get_predecessors(workflow, node):
         # let's just keep one, let's give it preference to the one with the
         # most total_conditions_satisfied
         _, wk = sorted(workflows, key=lambda x: x[0], reverse=True)[0]
+        GH = wk.graph
```
charles-cowart, line 906: It's a small point, but I'm not sure what GH is supposed to mean - it might not be a bad idea to expand it or add a comment.

charles-cowart (approved):
Everything looks good - just one small suggestion. I can approve either way.

qiita_db/metadata_template/prep_template.py
antgonza: graph, happy to change to whatever you recommend.

PR #3382, opened 2024-03-14
PR: Resource allocation plots in util.py

Gossty:


qiita_db/util.py
```diff
@@ -2311,3 +2317,337 @@ def send_email(to, subject, body):
...
+    sname: str, optional
+        Specified job sub type.
+
+    Returns
+    ----------
+    matplotlib.pyplot object
+        Returns a matplotlib object with a plot
+    """
+    # Constants
+    COL_NAME = 'samples * columns'
```
antgonza, line 2340: Can we make this a variable passed to resource_allocation_plot as an argument ?

qiita_db/util.py
```diff
@@ -2311,3 +2317,337 @@ def send_email(to, subject, body):
...
+
+    df = pd.read_csv(file, sep='\t', dtype={'extra_info': str})
+    df['ElapsedRawTime'] = pd.to_timedelta(df.ElapsedRawTime)
+    if sname == "all":
+        _df = df[(df.cName == cname)].copy()
+    else:
```
antgonza, line 2346: I'm not sure why an "all" plot would be useful, what do you think?

qiita_db/util.py
```diff
@@ -2311,3 +2317,337 @@ def send_email(to, subject, body):
...
+        _df = df[(df.cName == cname) & (df.sName == sname)].copy()
```
antgonza, line 2347: `df` is created and used in this block so I don't think we need to create anther variable; this was only useful in the notebook.

qiita_db/util.py
```diff
@@ -2311,3 +2317,337 @@ def send_email(to, subject, body):
...
+
+    fig, axs = plt.subplots(ncols=2, figsize=(10, 4), sharey=False)
+    _df.dropna(subset=['samples', 'columns'], inplace=True)
+    _df[COL_NAME] = _df.samples * _df['columns']
+    ax = axs[0]
+    # k a b
+    # bmin = _df['MaxRSSRaw'].min()
```
antgonza, line 2354: Are these comments needed?

qiita_db/util.py
```diff
@@ -2311,3 +2317,337 @@ def send_email(to, subject, body):
...
+    models = [mem_model1, mem_model2, mem_model3, mem_model4]
```
antgonza, line 2355: models should be a variable on top of the script?

qiita_db/util.py
```diff
@@ -2311,3 +2317,337 @@ def send_email(to, subject, body):
...
+    # bmin = _df['ElapsedRaw'].min()
+    ax = axs[1]
+    models = [time_model1, time_model2, time_model3, time_model4]
+    _resource_allocation_plot_helper(_df, ax, cname, sname, "ElapsedRaw",
+                                     models)
+    return fig, axs
+
+
+def _resource_allocation_plot_helper(_df, ax, cname, sname, curr, models):
+    COL_NAME = 'samples * columns'
```
antgonza, line 2368: This should also be an argument.

qiita_db/util.py
```diff
@@ -2311,3 +2317,337 @@ def send_email(to, subject, body):
...
+    M1G = 2**30
```
antgonza, line 2369: Please move at the top of the script.

qiita_db/util.py
```diff
@@ -2311,3 +2317,337 @@ def send_email(to, subject, body):
...
+    models: list, required
+        List of functions that will be used for visualization
+
+    """
+
+    x_data, y_data = _df[COL_NAME], _df[curr]
+    ax.scatter(x_data, y_data, s=2, label="data")
+    d = dict()
+    for index, row in _df.iterrows():
+        x_value = int(row[COL_NAME])
```
antgonza, line 2394: why not cast via pandas the full column vs. doing it on each iteration?

qiita_db/util.py
```diff
@@ -2311,3 +2317,337 @@ def send_email(to, subject, body):
...
+        y_value = row[curr]
+        if x_value not in d:
+            d[x_value] = []
+        d[x_value].append(y_value)
+
+    for key in d.keys():
+        d[key].sort()
+        d[key] = [d[key][-1] * 1.05]
```
antgonza, line 2402: can you add a small comment explaining why 1.05?

qiita_db/util.py
```diff
@@ -2311,3 +2317,337 @@ def send_email(to, subject, body):
...
+    x_plot = np.array(sorted(_df[COL_NAME].unique()))
+
+    if curr == "MaxRSSRaw":
+        y_plot = M1G*1.5 + (np.log(x_plot)*M1G*1.5) + (x_plot*3500)
+        ax.plot(x_plot, y_plot, linewidth=1, color='red',
+                label="predefined curve")
+    if curr == "ElapsedRaw":
+        y_plot = 600 + (np.log(x_plot)*3000)
+        ax.plot(x_plot, y_plot, linewidth=1, color='red',
+                label="predefined curve")
```
antgonza, line 2435: remove the predefined curves, they were nice for the presentation but not needed on the real code

qiita_db/util.py
```diff
@@ -2311,3 +2317,337 @@ def send_email(to, subject, body):
...
+    best_result = None
+    best_failures = np.inf
+    best_max = np.inf
+    counter = 0
+    for model in models:
+        counter += 1
+        sl = 2000
+        sr = 100000
+        left = sl
+        right = sr
```
antgonza, line 2498: Can you add a small comment explaining what are these?

qiita_db/util.py
```diff
@@ -2311,3 +2317,337 @@ def send_email(to, subject, body):
...
+                elif cmax < min_max:
+                    min_max = cmax
+                    res = options
+                    right = middle
+                else:
+                    right = middle
+
+            if left >= right and cnt < depth:
+                sl += 10000
+                sr += 10000
```
antgonza, line 2537: could you add a small comment explaining the increments?

qiita_db/util.py
```diff
@@ -2311,3 +2317,337 @@ def send_email(to, subject, body):
...
+    type_: str, required
+        Specifies for which type we're getting failures (e.g. MaxRSSRaw)
+
+    Returns
+    ----------
+    pandas.Dataframe
+        Dataframe containing failures for current type.
+    """
+
+    COL_NAME = 'samples * columns'
```
antgonza, line 2617: argument.

qiita_db/util.py
```diff
@@ -2311,3 +2317,337 @@ def send_email(to, subject, body):
...
+
+def time_model2(x, k, a, b):
+    return a + b*x + np.log(x) * k
+
+
+def time_model3(x, k, a, b):
+    return a + b*np.log(x)**2 + np.log(x) * k
+
+
+def time_model4(x, k, a, b):
```
antgonza, line 2652: can you add a small docstring explaining each of these?

antgonza (changes requested):
Looking good, a few comments/requests.

qiita_db/util.py
Gossty: Not sure what this comment means. On top of the script like a global variable?

antgonza: Sorry for not being clear, rm this line and put it at the top of util.py (well under the library imports) as:
```
MODELS = [mem_model1, mem_model2, mem_model3, mem_model4]
```
so it's easy to find and modify. 

Thanks.

qiita_db/util.py
Gossty: These are initialization constants for binary search, where left is 2000 and right is 100000 at the start. I define them separately from left and right because I run the search on different models. Binary Search tries to find smallest penalty weight for which the curve will have least errors.

qiita_db/util.py
Gossty: I added these because I don't have a formal proof why binary search would always result in the optimal constant, hence, I do the search in several windows up searching for the best constant until i reach the total number of iterations.

qiita_db/util.py
Gossty: I'm not sure how to create a dict of lists using pandas

qiita_db/util.py
Gossty: Cannot add MODELS at the top because the functions are defined at the end of the file.

antgonza (approved):
Approved but it will be great if @charles-cowart can also do a pass. Thank you.

qiita_db/util.py
```diff
@@ -2311,3 +2317,476 @@ def send_email(to, subject, body):
         raise RuntimeError("Can't send email!")
     finally:
         smtp.close()
+
+
+def resource_allocation_plot(file, cname, sname, COL_NAME):
+    """Builds resource allocation plot for given filename and jobs
```
charles-cowart, line 2323: Similar to \_df, why is COL_NAME in all caps?

qiita_db/util.py
```diff
@@ -2311,3 +2317,476 @@ def send_email(to, subject, body):
...
+    init = [1, 1, 1]
+    best_model = None
+    best_result = None
+    best_failures = np.inf
+    best_max = np.inf
+    counter = 0
+    for model in models:
+        counter += 1
+        # start values for binary search, where sl is left, sr is right
+        sl = 2000
```
charles-cowart, line 2487: The comment is appreciated but these still seem like magic numbers to me. A longer description of sl and sr might be nice to have.

qiita_db/util.py
```diff
@@ -2311,3 +2317,476 @@ def send_email(to, subject, body):
...
+    """
+
+    init = [1, 1, 1]
+    best_model = None
+    best_result = None
+    best_failures = np.inf
+    best_max = np.inf
+    counter = 0
+    for model in models:
+        counter += 1
```
charles-cowart, line 2485: I don't see where counter is being used.

qiita_db/util.py
```diff
@@ -2311,3 +2317,476 @@ def send_email(to, subject, body):
...
+            if left >= right and cnt < depth:
+                sl += 10000
+                sr += 10000
+                left = sl
+                right = sr
+
+            cnt += 1
+
+        # it's ok if we have a couple failures
+        failure_tolerance = 3
```
charles-cowart, line 2538: this seems like it should be a parameter, not a obfuscated local variable.

qiita_db/util.py
```diff
@@ -2311,3 +2317,476 @@ def send_email(to, subject, body):
...
+        Dataframe containing failures for current type.
+    """
+
+    x_plot = np.array(_df[COL_NAME])
+    _df[f'c{type_}'] = model(x_plot, k, a, b)
+    failures_df = _df[_df[type_] > _df[f'c{type_}']]
+    return failures_df
+
+
+def mem_model1(x, k, a, b):
```
charles-cowart, line 2619: It seems odd to implement these different memory and time models as one line functions. Why incur the overhead of a function call, when these can simply be defined as consts within the one function that uses them?

charles-cowart (changes requested):
Some questions.

charles-cowart:
@Gossty it might be good to change the name of this PR if it's no longer a WIP and no longer just an initial commit.

qiita_db/test/test_util.py
```diff
@@ -1303,6 +1307,70 @@ def test_quick_mounts_purge(self):
...
+        self.assertIsInstance(
+            fig1, Figure,
+            "Returned object fig1 is not a Matplotlib Figure")
+        for ax in axs1:
+            self.assertIsInstance(
+                ax, Axes,
+                "Returned object axs1 is not a single Matplotlib Axes object")
+
+    def test_minimize_const(self):
+        _df = self.df[(self.df.cName == self.CNAME)
```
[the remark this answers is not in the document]
Gossty: I had these dataframes created in notebooks initially and didn't change _df to self.df. Changed now.

qiita_db/util.py
Gossty: COL_NAME was defined as global constant before being added as an argument to functions. Changed to col_name

qiita_db/util.py
```diff
@@ -2311,3 +2317,318 @@ def send_email(to, subject, body):
...
+    df = df[(df.cName == cname) & (df.sName == sname)]
+    df.dropna(subset=['samples', 'columns'], inplace=True)
+    df[col_name] = df.samples * df['columns']
+    df[col_name] = df[col_name].astype(int)
+
+    fig, axs = plt.subplots(ncols=2, figsize=(10, 4), sharey=False)
+
+    ax = axs[0]
+    # models for memory
+    mem_model1 = (lambda x, k, a, b: k * np.log(x) + x * a + b)
```
charles-cowart, line 2353: It's not desirable to have these defined once in utils and once in tests. If you need them for tests, I would suggest moving them out of this function and defining them at the top of util.py. Then you can still use them in this function, but you can also import them into test_util.py (actually I think you just need to import the model_mem and model_time arrays) and use them there. In that way you're treating util.py as the place to go to define and get that information. Currently you have to know that this is defined in two places and make changes to both as needed.

charles-cowart (approved):
Looks good! Thanks for all your hard work!

Issue #3375, opened 2024-02-25
hard coded emails should be part of the configuration


sjanssen2:
https://github.com/qiita-spots/qiita/blob/57d1b146729223da9daa286a1cc75db6f2923746/qiita_db/processing_job.py#L902-L906

antgonza:
There at least a couple of scenarios where we send emails to alert people (based on previous discussions and request) but I agree that they might be better in a configuration file. I'll change the title of this issue.

antgonza:
Fixed here: https://github.com/qiita-spots/qiita/pull/3384

PR #3384, opened 2024-03-21
make help email address configurable instead of hard coding

sjanssen2:
Addresses issue #3375 
When we want to run multiple Qiita instances, it would make sense to use different email addresses a user can get help from. Therefore, I extended the configuration manager to parse two new strings for the help_email address and Jeff's sys-admin address.
Once you merge this PR, don't forget to update your configuration files!

### DIFF
--- a/qiita_core/configuration_manager.py
+++ b/qiita_core/configuration_manager.py
@@ -117,6 +117,12 @@ class ConfigurationManager(object):
         The script used to start the plugins
     plugin_dir : str
         The path to the directory containing the plugin configuration files
+    help_email : str
+        The email address a user should write to when asking for help
+        Defaults to qiita.help@gmail.com
+    sysadmin_email : str
+        The email address, Qiita sends internal notifications to a sys admin
+        Defaults to jdereus@health.ucsd.edu
 
     Raises
     ------
@@ -233,6 +239,14 @@ class ConfigurationManager(object):
         if not self.key_file:
             self.key_file = join(install_dir, 'qiita_core', 'support_files',
                                  'ci_server.key')
+
+        self.help_email = config.get('main', 'HELP_EMAIL')
+        if not self.help_email:
+            self.help_email = 'qiita.help@gmail.com'
+
+        self.sysadmin_email = config.get('main', 'SYSADMIN_EMAIL')
+        if not self.sysadmin_email:
+            self.sysadmin_email = 'jdereus@health.ucsd.edu'
 
     def _get_job_scheduler(self, config):
         """Get the configuration of the job_scheduler section"""
--- a/qiita_core/support_files/config_test.cfg
+++ b/qiita_core/support_files/config_test.cfg
@@ -68,6 +68,12 @@ COOKIE_SECRET = SECRET
 # The value used to secure JWTs for delegated permission artifact download.
 JWT_SECRET = SUPER_SECRET
 
+# Address a user should write to when asking for help, default to qiita.help@gmail.com
+HELP_EMAIL =
+
+# The email address, Qiita sends internal notifications to a sys admin
+SYSADMIN_EMAIL =
+
 # ----------------------------- SMTP settings -----------------------------
 [smtp]
 # The hostname to connect to
--- a/qiita_core/tests/test_configuration_manager.py
+++ b/qiita_core/tests/test_configuration_manager.py
@@ -180,6 +180,24 @@ class ConfigurationManagerTests(TestCase):
 
         self.assertEqual(obs.qiita_env, "")
 
+    def test_help_email(self):
+        obs = ConfigurationManager()
+
+        obs._get_main(self.conf)
+        self.assertEqual(obs.help_email, 'foo@bar.com')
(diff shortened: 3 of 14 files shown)


qiita_core/configuration_manager.py
```diff
@@ -117,6 +117,12 @@ class ConfigurationManager(object):
         The script used to start the plugins
     plugin_dir : str
         The path to the directory containing the plugin configuration files
+    help_email : str
+        The email address a user should write to when asking for help
+        Defaults to qiita.help@gmail.com
+    sysadmin_email : str
```
antgonza, line 123: Should these accept lists? I'm not sure but it might be helpful. BTW it's fine to leave like this.

qiita_core/configuration_manager.py
```diff
@@ -234,6 +240,14 @@ def _get_main(self, config):
             self.key_file = join(install_dir, 'qiita_core', 'support_files',
                                  'ci_server.key')
 
+        self.help_email = config.get('main', 'HELP_EMAIL')
+        if not self.help_email:
+            self.help_email = 'qiita.help@gmail.com'
+
+        self.sysadmin_email = config.get('main', 'SYSADMIN_EMAIL')
+        if not self.sysadmin_email:
+            self.sysadmin_email = 'jdereus@health.ucsd.edu'
```
antgonza, line 249: I'm on the fence about this being better than actually raising an error if those are not defined; what do you think?

antgonza (commented):
Beautiful! Thank you @sjanssen2. Just a couple of questions.

qiita_core/configuration_manager.py
sjanssen2: hm, the help_email is used to prompt the user via e.g. mailto: to one specific email address. I wouldn't know how an HTML element should deal with multiple addresses. And it would also confuse the user.

A list for the systadmin_email makes more sense to me. But I think the best way to inform multiple people about issues would be a mailing list or a shared mail account - which would boil down to one address here

qiita_core/configuration_manager.py
sjanssen2: agree! Thus, we remove "personalized" emails completely. What type of error do you suggest?

antgonza: I guess something like: `raise ValueError('You need to have a value in the HELP_EMAIL to continue')` but anything that lets the person configuring/starting the system that they are missing that value should work fine. Thank you.

sjanssen2: Done. I was wondering about the class of the error, not the message ;-) Sorry for being unspecific.

antgonza:
BTW just noticed that this PR was issued against master vs. dev; I just changed it.

qiita_core/configuration_manager.py
```diff
@@ -117,6 +117,12 @@ class ConfigurationManager(object):
         The script used to start the plugins
     plugin_dir : str
         The path to the directory containing the plugin configuration files
+    help_email : str
+        The email address a user should write to when asking for help
+        Defaults to qiita.help@gmail.com
```
antgonza, line 122: This is no longer true, right? Could you remove this and the one below?

qiita_core/support_files/config_test.cfg
```diff
@@ -68,6 +68,12 @@ COOKIE_SECRET = SECRET
 # The value used to secure JWTs for delegated permission artifact download.
 JWT_SECRET = SUPER_SECRET
 
+# Address a user should write to when asking for help
+HELP_EMAIL = qiita.help@gmail.com
+
+# The email address, Qiita sends internal notifications to a sys admin
+SYSADMIN_EMAIL = jdereus@health.ucsd.edu
+
```
antgonza, line 76: Should these match the ones used in the actual test? 
```
# Address a user should write to when asking for help
HELP_EMAIL = foo@bar.com
# The email address, Qiita sends internal notifications to a sys admin
SYSADMIN_EMAIL = jeff@bar.com
```
and perhaps changing the error (which should never happen now) to a warning when the emails match those.

qiita_core/configuration_manager.py
```diff
@@ -234,6 +240,22 @@ def _get_main(self, config):
             self.key_file = join(install_dir, 'qiita_core', 'support_files',
                                  'ci_server.key')
 
+        self.help_email = config.get('main', 'HELP_EMAIL')
+        if not self.help_email:
+            raise ValueError(
```
antgonza, line 245: Maybe change this to something like:
```
if self.help_email == 'foo@bar.com':
    warnings.warn("Using the github fake email for HELP_EMAIL, are you sure this is OK?")
```

antgonza (commented):
A few minor comments.

qiita_core/configuration_manager.py
```diff
@@ -247,6 +245,11 @@ def _get_main(self, config):
                 "section of Qiita's config file. This address is essential "
                 "for users to ask for help as it is displayed at various "
                 "location throughout Qiita's web pages.")
+        if (self.help_email == 'foo@bar.com') and \
+           (self.test_environment is False):
```
antgonza, line 249: Nice! Good idea!

PR #3388, opened 2024-04-04
Config statsmap lat long

sjanssen2:
It would be nice, if the Stats world map view could be centered by the user, e.g. I'd like to showcase European samples:
image
Therefore, I suggest to have latitude and longitude as parameters in Qiita's configuration file. I thought the portal section might be appropriate, because different portals also might want to show different locations?
The defaults are Boulder, CO, i.e. if a user gives the parameters without values or skips these parameters altogether


antgonza:
Thank you, this looks great!

PR #3390, opened 2024-04-16
Pac bio smrt new instruments

antgonza:


qiita_ware/ebi.py
```diff
@@ -117,6 +117,9 @@ class EBISubmission(object):
                        'PACBIO_SMRT': ['PACBIO RS',
                                        'PACBIO RS II',
                                        'SEQUEL',
+                                       'ONSO',
+                                       'REVIO',
+                                       'SEQUEL IIE',
                                        'SEQUEL II']}
```
charles-cowart, line 124: @antgonza just a quick sanity check - I noticed the columns in the .rst file go 'pacbio rs', 'pacbio rs ii', 'sequel', 'sequel ii', 'sequel iie', 'revio', and 'onso'. However the change in ebi.py is a list with a different order. Since we are talking about a list and not a set, is the different ordering okay?

charles-cowart (changes requested):
quick question

qiita_ware/ebi.py
antgonza: Yes, it's OK, thank you for checking.

charles-cowart (approved):
Approved!

PR #3391, opened 2024-04-16
qiita.slurm_resource_allocations

### LINKED ISSUE #3368: add settings table to DB schema
The Qiita DB Schema file qiita_db/support_files/qiita-db.html is extremely helpful to understand the design of the database. I just found that there are additional tables (outside the qiita.* schema) that are also used by qiita, e.g. `settings` https://github.com/qiita-spots/qiita/blob/57d1b146729223da9daa286a1cc75db6f2923746/qiita_db/environment_manager.py#L351
Could you add these tables to the HTML overview?

antgonza:
This adds the basic changes to add a new table to keep the slurm_resource_allocations info and fix #3368.


charles-cowart (approved):
Everything appears in order.

PR #3396, opened 2024-04-23
add reprocess_job_id to PrepTemplate

antgonza:


qiita_db/support_files/qiita-db.dbs
```diff
@@ -1001,9 +1001,12 @@
 			<column name="creation_job_id" type="uuid" jt="1111" >
 				<defo><![CDATA[NULL]]></defo>
 			</column>
-			<column name="current_human_filtering" type="boolean" jt="1111" >
```
charles-cowart, line 1004: What's the significance of -7?

charles-cowart (changes requested):
Everything looks good! Just one question.

qiita_db/support_files/qiita-db.dbs
antgonza: I think that the order of the objects in the view; however, note that this file is autogenerated by DbSchema.

charles-cowart (approved):

PR #3401, opened 2024-04-30
Update to DB qiita.slurm_resource_allocations

Gossty:
* Added a Python script "upload_df.py" in resource allocation notebooks.
* Updated "populate_test_db.sql"


antgonza:
@Gossty, thank you for this PR. Can you check the errors and fix them? Also, could you modify [these tests](https://github.com/qiita-spots/qiita/blob/dev/qiita_db/test/test_util.py#L1312-L1319) so the data is pulled from the database vs. the tgz file.

antgonza:
Superseded by https://github.com/qiita-spots/qiita/pull/3406

PR #3403, opened 2024-05-06
Clean plugins

antgonza:


charles-cowart:
Confirmed database schema and test database are both loaded automatically. User will need to manually remove the test data and reinstall the database manually to have a clean database.

PR #3405, opened 2024-05-07
adding test_db_sql

antgonza:


CONTRIBUTING.md
```diff
@@ -94,6 +94,18 @@ After the initial production release of Qiita, changes to the database schema wi
...
+### Patch 91.sql
+
+In May 2024 we decided to:
+* Merge all patches into the main database, this means that there are no patches younger than 92.sql.
+* Not allow INSERT/UPDATE queries within the main *.sql patches, to minimize the possibility of unnexpected changes.
+* Added a new folder `patches/test_db_sql/` where we can store sql files that will only be applied for the test environment.
+
+Note that these changes mean:
+1. 92.sql is the current first sql file to patch the database.
+2. If you need to make changes to both the production and tests database you will need to use the *.py.
```
charles-cowart, line 106: sounds kludgy. What if a change was part of the test_db_sql AND independently in a .py file for production? I would think I'd want to have all the test data in one place and this would go against that.

charles-cowart (changes requested):
Everything but one comment in the documentation looks good. I'm wondering if we can make it sound more palatable.

charles-cowart (approved):
Approved!

PR #3406, opened 2024-05-08
Test util db

Gossty:
Updated test_util.py for ResourceAllocationPlotTests by connecting the tests to database.


qiita_db/test/test_meta_util.py
```diff
@@ -299,6 +299,8 @@ def _get_daily_stats():
 
         for k, exp, f in vals:
             redis_key = '%s:stats:%s' % (portal, k)
+            print("here's redis", f(redis_key))
```
antgonza, line 302: can you remove these prints ?

qiita_db/test/test_user.py
```diff
@@ -491,17 +491,10 @@ def test_jobs(self):
             limit=1, ignore_status=ignore_status), [
                 PJ('b72369f9-a886-4193-8d3d-f7b504168e75')])
 
-        # no jobs
```
antgonza, line 494: 👍🏽

qiita_db/test/test_user.py
```diff
@@ -491,17 +491,10 @@ def test_jobs(self):
...
-        self.assertEqual(qdb.user.User('admin@foo.bar').jobs(
-            ignore_status=ignore_status), [])
-
         # generates expected jobs
         jobs = qdb.user.User('shared@foo.bar').jobs()
         self.assertEqual(jobs, [])
 
-        # no jobs
```
antgonza, line 502: 👍🏽

qiita_db/util.py
```diff
@@ -2361,9 +2360,7 @@ def resource_allocation_plot(file, cname, sname, col_name):
         Returns a matplotlib object with a plot
     """
 
-    df = pd.read_csv(file, sep='\t', dtype={'extra_info': str})
-    df['ElapsedRawTime'] = pd.to_timedelta(df.ElapsedRawTime)
-    df = df[(df.cName == cname) & (df.sName == sname)]
+    # df = df[(df.cName == cname) & (df.sName == sname)]
```
antgonza, line 2363: rm this line?

qiita_db/test/test_util.py
```diff
@@ -1309,19 +1309,21 @@ def test_quick_mounts_purge(self):
...
         self.col_name = 'samples * columns'
-        self.df = pd.read_csv(self.PATH_TO_DATA, sep='\t',
-                              dtype={'extra_info': str})
+        self.columns = [
+                "sName", "sVersion", "cID", "cName", "processing_job_id",
+                "parameters", "samples", "columns", "input_size", "extra_info",
+                "MaxRSSRaw", "ElapsedRaw"]
+        self.df = pd.DataFrame(
+            qdb.util._retrieve_resource_data(self.CNAME, self.SNAME),
+            columns=self.columns)
```
antgonza, line 1321: IMOO _retrieve_resource_data should just assign the column names and create the dataframe so they are always the same. The current separations feels clunky as the column. In other words, this should only be:
`self.df = qdb.util._retrieve_resource_data(self.CNAME, self.SNAME)`.

antgonza (commented):
Looks good, just some minor comments. @charles-cowart, could you also take a look? Thank you,

qiita_db/test/test_meta_util.py
charles-cowart: I've been guilty of doing this myself. One alternative that can work better is to use nosetests' variant of pdb; by adding the following line: from nose.tools import set_trace; set_trace(), you can begin watching variables, stepping through functionality, etc.

qiita_db/support_files/patches/test_db_sql/92.sql
```diff
@@ -1,4 +1,928 @@
--- .. date .. 
+-- 05/07/24 
 -- tmp file for @Gossty
 
-SELECT 1;
+
```
charles-cowart, line 4: A comment describing the reasoning for this large number of dummy jobs might be useful to those not so closely associated with this PR.

qiita_db/test/test_software.py
```diff
@@ -494,8 +494,11 @@ def test_processing_jobs(self):
                     '6ad4d590-4fa3-44d3-9a8f-ddbb472b1b5f',
                     '063e553b-327c-4818-ab4a-adfe58e49860',
                     'ac653cb5-76a6-4a45-929e-eb9b2dee6b63']
-        exp = [qdb.processing_job.ProcessingJob(j) for j in exp_jids]
-        self.assertCountEqual(qdb.software.Command(1).processing_jobs, exp)
+
+        exp = set([qdb.processing_job.ProcessingJob(j) for j in exp_jids])
+        set_jobs = set(qdb.software.Command(1).processing_jobs)
```
charles-cowart, line 499: Sorting both the observed and expected output before comparing for equality would be a better choice. The disadvantage of taking sets() of both is that you implicitly remove any duplicate entries that might be present in qdb.software.Command(1).processing_jobs for whatever bug or other reason.

qiita_db/test/test_util.py
```diff
@@ -1309,19 +1309,20 @@ def test_quick_mounts_purge(self):
...
+        self.CNAME = "Split libraries FASTQ"
+        self.SNAME = "QIIMEq2"
         self.col_name = 'samples * columns'
-        self.df = pd.read_csv(self.PATH_TO_DATA, sep='\t',
-                              dtype={'extra_info': str})
+        self.columns = [
+                "sName", "sVersion", "cID", "cName", "processing_job_id",
+                "parameters", "samples", "columns", "input_size", "extra_info",
+                "MaxRSSRaw", "ElapsedRaw"]
+        self.df = qdb.util._retrieve_resource_data(
```
charles-cowart, line 1319: self.df doesn't seem very descriptive.

charles-cowart (changes requested):
Looks good. Just a few comments.

qiita_db/test/test_software.py
Gossty: We don't want there to be duplicates in qdb.software.Command(1).processing_jobs, right? 

Could I create a set for qdb.software.Command(1).processing_jobs and compare its length to the length of qdb.software.Command(1).processing_jobs, thus, checking for duplicates?

charles-cowart: Right, we don't want there to be but since this is test code I wouldn't assume that a recent change couldn't cause such an error to occur; it's just defensive programming.

IMHO comparing two ordered lists checks both the content and the lengths more elegantly, but I'll accept your alternative if you include a comment as to why you're checking the length.

PR #3415, opened 2024-06-05
qiita.slurm_resource_allocations insertion

Gossty:


qiita_db/util.py
```diff
@@ -2673,3 +2678,216 @@ def _resource_allocation_failures(df, k, a, b, model, col_name, type_):
...
+    with qdb.sql_connection.TRN:
+        sql = sql_command
+        qdb.sql_connection.TRN.add(sql)
+        res = qdb.sql_connection.TRN.execute_fetchindex()
+        columns = ["processing_job_id"]
+        df = pd.DataFrame(res, columns=columns)
+
+    data = []
+    sacct = ['sacct', '-p', '--format=JobID,ElapsedRaw,MaxRSS,Submit,Start,'
+             'CPUTimeRAW,ReqMem,AllocCPUs,AveVMSize', '-j']
```
wasade, line 2750: Rather than what could be a very large number of calls to `sacct`, why not batch jobs and query `','.join(list_of_job_ids)`? The primary concern is creating a high burden for the resource manager, and secondary is the overhead of system calls

qiita_db/util.py
```diff
@@ -2673,3 +2678,216 @@ def _resource_allocation_failures(df, k, a, b, model, col_name, type_):
...
+
+    # Generating new columns
+    df['MaxRSSRaw'] = df.MaxRSS.apply(lambda x: MaxRSS_helper(str(x)))
+    df['ElapsedRawTime'] = df.ElapsedRaw.apply(
+        lambda x: timedelta(seconds=float(x)))
+
+    for index, row in df.iterrows():
+        with qdb.sql_connection.TRN:
+            sql = """
+                INSERT INTO qiita.slurm_resource_allocations (
```
wasade, line 2877: The resources used are sensitive to extrinsic factors like other work on the system or cluster. Scoping to include that type of information carries a reasonable amount of complexity, so I don't think we want to attempt that. However, if the date the job was started on was tracked, then it would at least allow for excluding periods of known high burden that artificially increase resource use.

There are also intrinsic factors though that are important for interpretation of resource use, such as the node which ran the job, and ideally the model of processor (`/proc/cpuinfo`) on the node (as hostnames are not assured to be unique over time). It would be great to have information about the memory of the system but that may require sudo (see e.g., `dmidecode`) so could be out of scope. These values are important as our compute cluster is heterogeneous

qiita_db/util.py
antgonza: I actually discussed this with @Gossty and this decision was made based on my previous experience with `sacct` and `pgsql`. As far as I was able to tell, `sacct` didn't create any burden to the resource manager - I think is because it is an actual database, which is not actually directly connected to the resource manager. Then in the past we have had issues with large inserts (vs. multiple single inserts) so I think basic inserts are the way to go here.

qiita_db/util.py
antgonza: @wasade, if I understand the comment correctly basically you are suggesting storing: 
a. the node name
b. the submission timestamp
correct?

qiita_db/util.py
wasade: See the warning [here](https://jhpce.jhu.edu/slurm/tips-sacct/)

qiita_db/util.py
wasade: (a) yes (b) the job start time stamp, not submission and ideally (c) the actual CPU model on the node

antgonza: Got it, thank you for the clarification. BTW why explicitly (c) vs. only (a)? Also, will `sacct` return (c)?

qiita_db/util.py
wasade: "That database needs to be modified constantly as jobs start and complete, so we don't want it tied up answering sacct queries." 

As you know, SQL engines are really good at bulk queries as there is a lot of overhead for a single query.

qiita_db/util.py
wasade: "... hostnames are not assured to be unique over time..." so just getting hte name isn't reliable over time. I'm not sure what information is or isn't easily available from slurm

qiita_db/util.py
```diff
@@ -2673,3 +2680,238 @@ def _resource_allocation_failures(df, k, a, b, model, col_name, type_):
...
+    """
+
+    # retrieve the most recent timestamp
+    sql_timestamp = """
+        SELECT
+            sra.job_start
+        FROM
+            qiita.slurm_resource_allocations sra
+        ORDER BY job_start DESC;
+    """
```
antgonza, line 2720: You could add LIMIT 1 so you only get 1 result; this will make the query faster.

qiita_db/util.py
```diff
@@ -2673,3 +2680,238 @@ def _resource_allocation_failures(df, k, a, b, model, col_name, type_):
...
+        res = qdb.sql_connection.TRN.execute_fetchindex()
+        columns = ['timestamp']
+        timestamps = pd.DataFrame(res, columns=columns)
+        if len(timestamps['timestamp']) == 0:
+            dates[0] = datetime.strptime('2023-04-28', '%Y-%m-%d')
+        else:
+            dates[0] = str(timestamps['timestamp'].iloc[0]).split(' ')[0]
+        date0 = datetime.strptime(dates[0], '%Y-%m-%d')
+        date1 = date0 + timedelta(weeks=1)
+        dates[1] = date1.strftime('%Y-%m-%d')
```
antgonza, line 2736: Why convert to a dataframe? I mean, the database should already return a timestamp object that can be used

qiita_db/util.py
```diff
@@ -2673,3 +2680,238 @@ def _resource_allocation_failures(df, k, a, b, model, col_name, type_):
...
+        qdb.sql_connection.TRN.add(sql)
+        res = qdb.sql_connection.TRN.execute_fetchindex()
+        columns = ['timestamp']
+        timestamps = pd.DataFrame(res, columns=columns)
+        if len(timestamps['timestamp']) == 0:
+            dates[0] = datetime.strptime('2023-04-28', '%Y-%m-%d')
+        else:
+            dates[0] = str(timestamps['timestamp'].iloc[0]).split(' ')[0]
+        date0 = datetime.strptime(dates[0], '%Y-%m-%d')
+        date1 = date0 + timedelta(weeks=1)
```
antgonza, line 2735: This is a good idea, could you expose this parameters in the function and make the default 1 week?

qiita_db/util.py
```diff
@@ -2673,3 +2680,238 @@ def _resource_allocation_failures(df, k, a, b, model, col_name, type_):
...
+        ----------
+        test: pandas.DataFrame, optional
+            Represents dataframe containing slurm data from 2023-04-28. Used
+            for testing only.
+    """
+
+    # retrieve the most recent timestamp
+    sql_timestamp = """
+        SELECT
+            sra.job_start
```
antgonza, line 2716: Can you select also the job id, maybe external if available?

qiita_db/util.py
```diff
@@ -2673,3 +2680,238 @@ def _resource_allocation_failures(df, k, a, b, model, col_name, type_):
...
+            LEFT JOIN
+                qiita.slurm_resource_allocations sra
+                ON pj.processing_job_id = sra.processing_job_id
+            WHERE
+                pjs.processing_job_status = 'success'
+            AND
+                pj.external_job_id ~ '^[0-9]+$'
+            AND
+                sra.processing_job_id IS NULL;
+        """
```
antgonza, line 2758: Note that this query will return _all_ the rows, which will be a lot so it will be good to limit based on the job_id of the latests result in the previous block (what you are using to extract the time). Could you limit to just return anything newer than that job?

antgonza (changes requested):
A few requests. Please let me know if you would like to chat to clarify any of the requests.

qiita_db/util.py
Gossty: Would we determine whether the job is newer through external_id? I don't think qiita.processing_job stores information about when the job was submitted.

antgonza: That's correct, the external id is an incremental integer so it can be used to find what's newer.

antgonza (approved):
Looks great, thank you!

Issue #3378, opened 2024-03-01
track time a new user account was requested


sjanssen2:
Thinking about performance indicators, it would be nice to show an increase in validated users over time. Is there any information about **when** a user account was created or validated? If not, are there arguments against tracking this in the database?

I am not sure what your mechanism for validating a newly registered user is, I'd love to have a simple admin page with all registered but not validated accounts with a tickbox and one submit button which would than issue the validating action of all ticked users. I figure it would also be useful to sort this list according on registration date.

antgonza:
This will be great. I think also having a button to purge non-validated users (all or one by one) will be really useful - sometimes this happens when the user inputs their email incorrectly. 

Anyway, FWIW, we normally track this information using the https://qiita.ucsd.edu/stats/ information; in summary: We have a [nightly cron-job](https://github.com/qiita-spots/qiita/blob/master/scripts/qiita-cron-job#L47) that updates the information in redis and inserts that version [into SQL](https://github.com/qiita-spots/qiita/blob/master/qiita_db/meta_util.py#L337-L340); then we can see how many new users, studies, samples, etc have been created in those dates.

sjanssen2:
If I read the SQL statement correctly, you simply count the number of rows of table `qiita.qiita_user` https://github.com/qiita-spots/qiita/blob/13eafdf1a2eb0c8b5d4019ad294b7490cd396139/qiita_db/util.py#L1265-L1268 regardless of users being verified or not?

Isn't redis an in memory DB? What if the machine fails? Do you have backups on disk or would you loose the user growth information? Would you mind if I add another column to `qiita.qiita_user` to log the date of user activation? Having this date should always provide means to reconstruct this growth. I am asking, because our consortium currently discusses performance indicators and number of user is one of the key candidates.

antgonza:
- regardless of users being verified or not?: 
   - That sound about right; fine to change to verified users if you think it's better.
- Isn't redis an in memory DB?: 
   - yes, it is.
- What if the machine fails?: 
   - We used it for two things: (1) errors/warnings/etc messages are not stored so these will be lost - which I still think is not a bit issue; and (2) user profiles for faster retrieval - this one is populated at the start
- Would you mind if I add another column to qiita.qiita_user to log the date of user activation?: 
   - Sounds great! Thanks

sjanssen2:
addressed by #3416 and #3417

PR #3416, opened 2024-06-19
log user creation timestamp

sjanssen2:
Addressing #3378 this PR will use SQL's `NOW()` function to log creation of a new user account (validated or not). This is handy for:
a) reconstructing user growth
b) pruning accounts that have been created but not validated for X days

I've also added this information to the user's profile page, if the value is not None:
image


qiita_db/support_files/qiita-db-unpatched.sql
```diff
@@ -1891,7 +1891,8 @@ CREATE TABLE qiita.qiita_user (
     receive_processing_job_emails boolean DEFAULT false,
     social_orcid character varying DEFAULT NULL,
     social_researchgate character varying DEFAULT NULL,
-    social_googlescholar character varying DEFAULT NULL
+    social_googlescholar character varying DEFAULT NULL,
```
antgonza, line 1894: A few requests:

1. Could you undo the changes on this file and add them [as a SQL patch](https://github.com/qiita-spots/qiita/blob/master/CONTRIBUTING.md#developer-workflow)? Do not worry about changing the DBSchema, I'll do that with this and other minor changes it has happened before the release. As a reminder,  the reason those changes were made in this file was because I merged all changes into a single file a few PRs ago. 
2. Could you make the default of creation_timestamp NOW()? That way it will be controlled by the database during creation. 

Thank you.

qiita_db/user.py
```diff
@@ -234,6 +234,12 @@ def create(cls, email, password, info=None):
                 cls._table, ','.join(columns), ','.join(['%s'] * len(values)))
             qdb.sql_connection.TRN.add(sql, values)
 
+            # log timestamp of user creation
+            sql = """UPDATE qiita.{0}
+                     SET creation_timestamp = NOW()
+                     WHERE email = %s""".format(cls._table)
+            qdb.sql_connection.perform_as_transaction(sql, [email])
+
             # Add system messages to user
```
antgonza, line 243: If the default in the database is now(), then this is not required.

qiita_db/test/test_user.py
```diff
@@ -75,7 +75,8 @@ def setUp(self):
             'receive_processing_job_emails': True,
```
antgonza, line 75: The issue with this test and my suggestion about using now() in the database is that there is going to be a small difference on the tests so they will need to be adjusted.

antgonza (changes requested):
@sjanssen2, this is great; really appreciate the contribution.

qiita_db/support_files/qiita-db-unpatched.sql
sjanssen2: @1 oh, I thought the patching mechanism is dead as I didn't see patch files anymore :-) I've moved the change into `93.sql`. Changes are much cleaner this way.
@2 love to :-)

antgonza: 1. in fact, I think you should use 92.sql which is the patch for this release; fine to do that on the other PR if you want me to merge this one first
2. 👍🏽

sjanssen2: @1: I moved the alterations to `92.sql`. I didn't know that you collect changes for a release. However, tests fail. I might overlook something, but I have the impression that no patch is currently applied in the github actions?!

qiita_db/support_files/populate_test_db.sql
```diff
@@ -50,7 +50,7 @@ INSERT INTO qiita.user_level VALUES (7, 'wet-lab admin', 'Can access the private
 -- Data for Name: qiita_user; Type: TABLE DATA; Schema: qiita; Owner: antoniog
 --
 
-INSERT INTO qiita.qiita_user VALUES ('test@foo.bar', 4, '$2a$12$gnUi8Qg.0tvW243v889BhOBhWLIHyIJjjgaG6dxuRJkUM8nXG9Efe', 'Dude', 'Nowhere University', '123 fake st, Apt 0, Faketown, CO 80302', '111-222-3344', NULL, NULL, NULL, false, '0000-0002-0975-9019', 'Rob-Knight', '_e3QL94AAAAJ');
+INSERT INTO qiita.qiita_user VALUES ('test@foo.bar', 4, '$2a$12$gnUi8Qg.0tvW243v889BhOBhWLIHyIJjjgaG6dxuRJkUM8nXG9Efe', 'Dude', 'Nowhere University', '123 fake st, Apt 0, Faketown, CO 80302', '111-222-3344', NULL, NULL, NULL, false, '0000-0002-0975-9019', 'Rob-Knight', '_e3QL94AAAAJ', '2015-12-03 13:52:42.751331-07');
```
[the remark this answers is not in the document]
sjanssen2: hm, I thought that the new patches should be free of test data. I've moved this addition of the creation_timestamp info into the 92.sql patch. Let's wait for the tests.

antgonza (approved):
Looks great, thank you.

qiita_db/support_files/qiita-db-unpatched.sql
sjanssen2: >     2. Could you make the default of creation_timestamp NOW()? That way it will be controlled by the database during creation.

As a follow up: I've just made the DB changes in my "productive" instance and found that all existing users (which did not have the `creation_timestamp` column of cause) now have the timestamp of DB change as their `creation_timestamp` information, even though they existed for many month in the past. 

Is that really the behaviour we want to see in a productive Qiita, @antgonza ?

antgonza: I think that's fine.

PR #3417, opened 2024-06-19
Purge non-verified user

sjanssen2:
Handle #3416 first!

Addressing #3378 as well.
This PR is based on user creation_timestamp logging and offers a new Admin only page that lists those users that 
a) are non yet verifier (i.e. user_status_id==5) and 
b) where registers more than 30 days ago (give a new user sufficient time to react)
Individual or all users can be selected in the website and removed by one click.

image


antgonza:
@sjanssen2, could you pull from dev and update this branch? Thank you.

qiita_db/support_files/patches/92.sql
```diff
@@ -43,3 +43,10 @@ COMMENT ON COLUMN qiita.qiita_user.creation_timestamp IS 'The date the user acco
 -- for testing: provide creation date for one of the existing users
 
 UPDATE qiita.qiita_user SET creation_timestamp = '2015-12-03 13:52:42.751331-07' WHERE email = 'test@foo.bar';
+
+-- Jun 20, 2024
+-- Add some non-verified users to the test DB to test new admin page: /admin/purge_users/
+
+INSERT INTO qiita.qiita_user VALUES ('justnow@nonvalidat.ed', 5, '$2a$12$gnUi8Qg.0tvW243v889BhOBhWLIHyIJjjgaG6dxuRJkUM8nXG9Efe', 'JustNow', 'NonVeriUser', '1634 Edgemont Avenue', '303-492-1984', NULL, NULL, NULL, false, NULL, NULL, NULL, NOW());
+INSERT INTO qiita.qiita_user VALUES ('ayearago@nonvalidat.ed', 5, '$2a$12$gnUi8Qg.0tvW243v889BhOBhWLIHyIJjjgaG6dxuRJkUM8nXG9Efe', 'Oldie', 'NonVeriUser', '172 New Lane', '102-111-1984', NULL, NULL, NULL, false, NULL, NULL, NULL, NOW() - INTERVAL '1 YEAR');
+INSERT INTO qiita.qiita_user VALUES ('3Xdays@nonvalidat.ed', 5, '$2a$12$gnUi8Qg.0tvW243v889BhOBhWLIHyIJjjgaG6dxuRJkUM8nXG9Efe', 'TooLate', 'NonVeriUser', '564 C Street', '508-492-222', NULL, NULL, NULL, false, NULL, NULL, NULL, NOW() - INTERVAL '30 DAY');
```
antgonza, line 52: Sorry for not catching this before but could you move these lines to [patches/test_db_sql/92.sql](https://github.com/qiita-spots/qiita/blob/dev/CONTRIBUTING.md#patch-91sql). The issue is that if we leave them here they will also be inserted in the main deployment but in `patches/test_db_sql` they will only be applied to the test environment.

antgonza (commented):
One comment that I completely missed in my original review. Thanks.

qiita_db/support_files/patches/92.sql
antgonza: FWIW, I think this is the source of the error.

sjanssen2:
I am a bit puzzled about the current DB error raised in the tests. It looks to me like the portals test does not clean up properly for tearDown. In my attempt to reproduce this issue locally, all tests passed when executed once, but re-running those caused the same errors as in github. However, I wonder why this becomes an issue now and not already earlier. Any ideas @antgonza ?

sjanssen2:
> I am a bit puzzled about the current DB error raised in the tests. It looks to me like the portals test does not clean up properly for tearDown. In my attempt to reproduce this issue locally, all tests passed when executed once, but re-running those caused the same errors as in github. However, I wonder why this becomes an issue now and not already earlier. Any ideas @antgonza ?

I might got the problem. Adding three new test users caused more iterations of this loop https://github.com/qiita-spots/qiita/blob/8b3f717ac9d6717a5bc08ed2ddb57a61eca2e9fd/qiita_db/portal.py#L89-L94 when creating a portal. Thus the assertion https://github.com/qiita-spots/qiita/blob/8b3f717ac9d6717a5bc08ed2ddb57a61eca2e9fd/qiita_db/test/test_portal.py#L51 failed and subsequent code to clean up the new portal was never executed, leading to follow up fails.

PR #3418, opened 2024-06-20
add missing template

sjanssen2:
a bit embarrassing, but I forgot to add the actual html template for the new admin page in PR #3417 ... and no test did catch this mistake :-/


antgonza:
Well, this wasn't caught during review either. I'll merge when tests finish. Thanks.

PR #3419, opened 2024-06-21
User tracking

sjanssen2:
This PR adds the ability to define some JavaScript code for visitor tracking of Qiita. Raw code, i.e. contents of variable `JS_CODE` in config section `user_tracking` is injected as top element in `<body>` of the `sitebase.html` template. 

We, i.e. NFDI4Microbiota will use this mechanism to measure performance indicators.


sjanssen2:
sorry, I've used the wrong fork (and branch) for this PR :-/

PR #3426, opened 2024-07-23
minor changes after deployment

antgonza:


charles-cowart (approved):
Everything appears in order. At first it appeared like I didn't have enough context to review the biggest change in artifact.py, but after realizing the code was simply moved downward in the code it made more sense.

PR #3428, opened 2024-08-26
fix #3427

### LINKED ISSUE #3427: default workflows only work when they have 2+ steps
If a workflow only has one step, the GUI brakes and it can't be applied to the prep template. This is inconvenient.

antgonza:


qiita_pet/handlers/software.py
```diff
@@ -144,6 +149,37 @@ def _default_parameters_parsing(node):
...
+                'Error, please check your workflow configuration')
+
+        # note that this block is similar but not identical to adding connected
+        # nodes
+        for i, (_, x) in enumerate(not_used_nodes.items()):
+            vals_x, input_x, output_x = _default_parameters_parsing(x)
+            if at in input_x[0][1]:
+                input_x[0][1] = at
+            else:
+                input_x[0][1] = '** WARNING, NOT DEFINED **'
```
charles-cowart, line 166: Are we sure that it wouldn't be better to raise an Error if something was undefined instead of encoding a warning message?

charles-cowart (changes requested):
Looks good! just one comment to confirm.

qiita_pet/handlers/software.py
antgonza: I'm 100% sure about it.

charles-cowart: ok.

charles-cowart (approved):
Approved!

PR #3430, opened 2024-08-28
Resource Allocation data in Redis

Gossty:


qiita_db/meta_util.py
```diff
@@ -37,11 +37,18 @@
 from re import sub
 from json import loads, dump, dumps
 
-from qiita_db.util import create_nested_path
+from qiita_db.util import create_nested_path, _retrieve_resource_data
```
antgonza, line 40: If `_retrieve_resource_data` is going to be use outside of util, then it should not start with _, could you change?

qiita_db/meta_util.py
```diff
@@ -546,3 +553,102 @@ def generate_plugin_releases():
...
+    for software in software_list:
+        sname = software.name
+        sversion = software.version
+        commands = software.commands
+        if sname not in software_commands:
+            software_commands[sname] = {}
+        if sversion not in software_commands[sname]:
+            software_commands[sname][sversion] = []
+        for command in commands:
+            software_commands[sname][sversion].append(command.name)
```
antgonza, line 570: This could be simplified with a defaultdict, could you check that?

antgonza (changes requested):
Thank you @Gossty, this looks real good; just some minor comments and suggestions.

antgonza (approved):

PR #3431, opened 2024-09-10
minor fixes after qiita-rc deploy

antgonza:


charles-cowart (approved):
Approved. Looks good.

PR #3445, opened 2024-12-10
some improvements

antgonza:


qiita_db/analysis.py
```diff
@@ -215,6 +215,22 @@ def create(cls, owner, name, description, from_default=False,
         job.submit()
         return instance
 
+    @classmethod
+    def delete_analysis_artifacts(cls, _id):
```
charles-cowart, line 219: Does it make sense to have this be a classmethod? It looks like the first thing it does is instantiate an instance of the class using _id as the parameter. Wouldn't it be more straightforward to create an instance of said class with _id and then call this method?

qiita_db/util.py
```diff
@@ -2334,7 +2334,7 @@ def send_email(to, subject, body):
     msg = MIMEMultipart()
     msg['From'] = qiita_config.smtp_email
     msg['To'] = to
-    msg['Subject'] = subject
+    msg['Subject'] = subject.replace('\n', '')
```
charles-cowart, line 2337: subject.strip() or subject.rstrip() w/out parameters will also strip the newline from strings and is more concise. It also has the side benefit of removing '\r\n' for cases when you're reading in from a Windows text file that has carriage return as well as line-feed.

charles-cowart (changes requested):
A few questions. Everything else appears solid.

qiita_db/util.py
charles-cowart: strip() removes leading whitespace from both the left and right sides of a string, while rstrip() only cleans up the end of the string. It's perhaps the more conservative choice to use for this situation.

qiita_db/analysis.py
antgonza: That's how it was originally, see: https://github.com/qiita-spots/qiita/pull/3445/files#diff-e921e32f252a38ad15c8af69792ae15dfa5ecc2e81937b9511d5511117b530dcL295 ... but IMOO it was one too many steps. What do you think?

charles-cowart: I'm okay with it if you're okay with it. I agree it looks cleaner when using it.

charles-cowart (approved):

PR #3446, opened 2024-12-12
fix add_default_workflow

antgonza:


antgonza:
FWIW, this was tested in qiita via a terminal and it works as expected.

qiita_db/metadata_template/prep_template.py
```diff
@@ -989,7 +999,7 @@ def _get_predecessors(workflow, node):
                     init_artifacts = {
                         wkartifact_type: f'{starting_job.id}:'}
                 else:
-                    init_artifacts = {wkartifact_type: self.artifact.id}
+                    init_artifacts = {wkartifact_type: str(self.artifact.id)}
```
charles-cowart, line 1002: I hate to say it but having to cast an internal id to string is a little bit concerning to me. Internal code should use the numeric form of the artifact_id and it should also be easy to pass the numeric version for any REST call. May I ask why this is needed? I feel like even if it's needed, it would be much clearer to cast it right before it's used to make it obvious why it's needed. Sorry if that's not clear.

qiita_db/metadata_template/prep_template.py
```diff
@@ -871,7 +881,7 @@ def _get_predecessors(workflow, node):
                          'artifact transformation']
         merging_schemes = {
             qdb.archive.Archive.get_merging_scheme_from_job(j): {
-                x: y.id for x, y in j.outputs.items()}
+                x: str(y.id) for x, y in j.outputs.items()}
```
charles-cowart, line 884: The conversion to string for y.id is also not obvious here (see below). Perhaps a comment as to why it's necessary would be appropriate?

charles-cowart (changes requested):
Ty for your patience! I have one concern.

qiita_db/metadata_template/prep_template.py
antgonza: Workflows expect input artifacts to be strings as they normally are in the for of: "job_id:artifact_name" - let's call this a "promise-input"; however, when a processing pipeline has some steps already done, the init artifact will be an int (basically just the artifact-id). FWIW (1) the code already checks if the input is in the form of a "promised input" or "artifact-id", and (2) the reason we never caught this issue before was because our workflows only had 2-3 steps and now they have more.

charles-cowart (approved):
Approved.

PR #3453, opened 2025-01-07
Resource allocation colormap

Gossty:


notebooks/resource-allocation/upload_df.py
antgonza, line 1: Do we need this file? Can it be deleted?

antgonza (changes requested):
Looks good to me; do you think we can move the equations to the database in this PR?

notebooks/resource-allocation/upload_df.py
Gossty: We can delete this file.

Gossty:
> Looks good to me; do you think we can move the equations to the database in this PR?

We could do that. I think the best approach would be to have a function that would convert Python expression (e.g. np.log(x) * k...) into a string and upload it to DB similar to  resource allocation table.

antgonza:
To be honest I think a direct insert would be easier - at the end, only admins with DB access can add them.

qiita_db/util.py
```diff
@@ -2390,19 +2369,63 @@ def resource_allocation_plot(df, cname, sname, col_name):
...
     ax = axs[1]
     # models for time
     _resource_allocation_plot_helper(
-        df, ax, cname, sname, "ElapsedRaw", MODELS_TIME, col_name)
+        df, ax, "ElapsedRaw",  time_models, col_name)
 
     return fig, axs
 
 
+def _retrieve_equations():
```
antgonza, line 2385: Can you make this public by removing the _?

qiita_db/util.py
```diff
@@ -2390,19 +2369,63 @@ def resource_allocation_plot(df, cname, sname, col_name):
...
+    with qdb.sql_connection.TRN:
+        sql = ''' SELECT * FROM qiita.allocation_equations; '''
+        qdb.sql_connection.TRN.add(sql)
+        res = qdb.sql_connection.TRN.execute_fetchindex()
+        for models in res:
+            if 'mem' in models[1]:
+                memory_models[models[1]] = lambda x, k, a, b: eval(models[2])
+            else:
+                time_models[models[1]] = lambda x, k, a, b: eval(models[2])
+        return (memory_models, time_models)
```
antgonza, line 2409: This loop can be outside the TRN, right?

qiita_db/util.py
```diff
@@ -74,47 +74,30 @@
...
-time_model1 = (lambda x, k, a, b: a + b + np.log(x) * k)
-time_model2 = (lambda x, k, a, b: a + b * x + np.log(x) * k)
-time_model3 = (lambda x, k, a, b: a + b * np.log(x)**2 + np.log(x) * k)
-time_model4 = (lambda x, k, a, b: a * np.log(x)**3 + b * np.log(x)**2
-               + np.log(x) * k)
-
-MODELS_TIME = [time_model1, time_model2, time_model3, time_model4]
-
 
 def get_model_name(model):
```
antgonza, line 85: This method can really be replaced by _retrieve_equations, no? I mean, I don't understand why we need to have this here vs. using _retrieve_equations.

notebooks/resource-allocation/upload_df.py
antgonza: OK, thank you; then please rm.

antgonza (changes requested):
A few comments.

qiita_db/support_files/patches/test_db_sql/93.sql
```diff
@@ -0,0 +1,10 @@
+INSERT INTO qiita.allocation_equations(equation_name, expression)
```
antgonza, line 1: Same as the other sql: these lines need to be moved to 94.sql.

qiita_db/test/test_util.py
```diff
@@ -1344,46 +1343,46 @@ def test_minimize_const(self):
...
                          msg=f"""Best time model
                                 doesn't match
                                 Coefficients:{k} {a} {b}
-                                 {qdb.util.time_model1}, "qdb.util.time_model1"
-                                 {qdb.util.time_model2}, "qdb.util.time_model2"
-                                 {qdb.util.time_model3}, "qdb.util.time_model3"
-                                 {qdb.util.time_model4}, "qdb.util.time_model4"
                                 """)
-        self.assertEqual(failures, 1, "Number of failures must be 1")
+        self.assertEqual(failures, 0, "Number of failures must be 0")
```
antgonza, line 1385: Just wondering why this changed from 1 to 0?

qiita_db/util.py
```diff
@@ -2390,19 +2348,87 @@ def resource_allocation_plot(df, cname, sname, col_name):
...
+            model_name = "a + b * x + log(x) * k"
+        elif models[1] == 'time_model3':
+            model_name = "a + b * log(x)^2 + log(x) * k"
+        elif models[1] == 'time_model4':
+            model_name = "a * log(x)^3 + b * log(x)^2 + log(x) * k"
+        if 'mem' in models[1]:
+            memory_models[models[1]] = {
+                "equation_name": model_name,
+                "equation": lambda x, k, a, b: eval(models[2])
+            }
```
antgonza, line 2406: Why do we need this? I mean the model_name is the expression column in the database, no?

antgonza (changes requested):

qiita_db/test/test_util.py
Gossty: Now the code chooses 4th model instead of 1st, which has 0 failures.

qiita_db/util.py
Gossty: We have this for readability on the frontend. It was hard to read "np.log10(x)...". Maybe there's a way to automatically convert those from lambda function to a readable equation?

antgonza: I see, I think leaving as np.log or doing just log in the database will be fine. Note that if you replace np.log for log in the database you would need to change the `import numpy as np` to `from numpy import log`.

Gossty: We previously also talked about adding parenthesis to ensure correct order of operations between systems, which makes the readability worse. Should we still keep the parenthesis or should we assume the regular order of operations? For example:
(k * (np.log(x))) + (b * ((np.log(x))**2)) + (a * ((np.log(x))**3))

Could just be k * log(x) + b * log(x)**2 + a * log(x)**3

antgonza: I like the extra parenthesis so we don't assume the parenthesis order.

antgonza (approved):

PR #3457, opened 2025-02-11
Added column_name variability to resource_allocation

Gossty:


qiita_db/meta_util.py
```diff
@@ -555,20 +555,44 @@ def generate_plugin_releases():
         f(redis_key, v)
 
 
-def get_software_commands(active):
+def _build_software_commands_col_names_object(active):
```
antgonza, line 558: This is only called once and is not directly tested, why create a separate method for this?

qiita_db/meta_util.py
```diff
@@ -555,20 +555,44 @@ def generate_plugin_releases():
...
+    '''
+    Helper function for update_resource_allocation_redis
+
+    Constructs a nested dictionary containing software commands and associated
+    column names from the database.
+    '''
+
+    col_name_list = []
+    with qdb.sql_connection.TRN:
+        sql = ''' SELECT * FROM qiita.resource_allocation_column_names; '''
```
antgonza, line 568: rm extra spaces from sql.

qiita_db/meta_util.py
```diff
@@ -555,20 +555,44 @@ def generate_plugin_releases():
...
+        qdb.sql_connection.TRN.add(sql)
+        res = qdb.sql_connection.TRN.execute_fetchindex()
+
+    for col_name in res:
+        col_name_list.append(col_name[1])
+
```
antgonza, line 574: The sql above selects all columns but you only use [1], could you make sure to just select the column you want and use fetch_flatten to get only the values you need?

qiita_db/meta_util.py
```diff
@@ -555,20 +555,44 @@ def generate_plugin_releases():
...
         sversion = software.version
-        commands = software.commands
 
-        for command in commands:
-            software_commands[sname][sversion].append(command.name)
-        software_commands[sname] = dict(software_commands[sname])
+        for command in software.commands:
+            cmd_name = command.name
+            for col in col_name_list:
+                software_commands[sname][sversion][cmd_name].append(col)
```
antgonza, line 586: This could actually be replaced by:
```software_commands[sname][sversion][cmd_name] = col_name_list```
no?

File: qiita_db/meta_util.py [hunk 5]
```diff
@@ -555,20 +555,44 @@ def generate_plugin_releases():
...
 
-    return dict(software_commands)
+    final_obj = {
+        sname: {
+            sversion: dict(commands)
+            for sversion, commands in dict(versions).items()
+        }
+        for sname, versions in dict(software_commands).items()
+    }
+    return final_obj
```
antgonza, line 595: Could you make sure that you create this object in this format so you don't have to redo it?

qiita_db/meta_util.py
```diff
@@ -555,20 +555,44 @@ def generate_plugin_releases():
...
 
 
 def update_resource_allocation_redis(active=True):
```
antgonza, line 598: can you add a verbose parameter that is by default set to False that is use to print/not-print the messages above?

qiita_db/meta_util.py
```diff
@@ -581,70 +605,106 @@ def update_resource_allocation_redis(active=True):
...
-                ]
-
-                for k, v, f in values:
-                    redis_key = 'resources$#%s$#%s$#%s$#%s:%s' % (
-                                cname, sname, version, col_name, k)
-                    r_client.delete(redis_key)
-                    f(redis_key, v)
+                # column_name_str looks like col1*col2*col3, etc
+                for col_name_str in col_name_list:
+                    df_copy = df.copy()
```
antgonza, line 630: why do we need a copy?

qiita_db/meta_util.py
```diff
@@ -581,70 +605,106 @@ def update_resource_allocation_redis(active=True):
...
+                    new_column = None
+                    col_name_split = col_name_str.split('*')
+                    df_copy.dropna(subset=col_name_split, inplace=True)
```
antgonza, line 633: I think you need the copy cause here you are changing the dataframe, maybe just do:
df_copy = df.dropna(subset=col_name_split)

qiita_db/util.py
```diff
@@ -2326,43 +2326,43 @@ def send_email(to, subject, body):
...
+    print("\tCalculating best model for time")
     _resource_allocation_plot_helper(
-        df, ax, "ElapsedRaw",  time_models, col_name)
-
+        df, ax, "ElapsedRaw",  time_models, col_name_str)
     return fig, axs
 
 
-def retrieve_equations():
+def _retrieve_equations():
```
antgonza, line 2365: why the name change?

qiita_db/util.py
```diff
@@ -2517,9 +2517,15 @@ def _resource_allocation_plot_helper(
...
 
+    print(f"\t\tFitting best model for {curr}; column {col_name}")
     # 50 - number of maximum iterations, 3 - number of failures we tolerate
     best_model_name, best_model, options = _resource_allocation_calculate(
         df, x_data, y_data, models, curr, col_name, 50, 3)
+    if options is None:
+        print(df)
+        print(x_data)
+        print(y_data)
+    print(f"\t\tSuccessfully chose best model for {curr}; column {col_name}")
```
antgonza, line 2528: This is not needed, right?

antgonza (changes requested):
Thank you @Gossty; a few comments, I'll do another pass once these are fixed.

qiita_db/meta_util.py
Gossty: This can be done, we would just need to avoid using defaultdict in the lines above

qiita_db/meta_util.py
Gossty: By messages above you mean print statements within this function and function it calls within?

qiita_db/util.py
Gossty: We only use this function within resource_allocation_plot, making it an internal helper

qiita_db/meta_util.py
antgonza: yup, thanks

qiita_db/util.py
antgonza: Then similar to `_build_software_commands_col_names_object `: This is only called once and is not directly tested, why create a separate method for this?

qiita_db/meta_util.py
```diff
@@ -555,96 +555,144 @@ def generate_plugin_releases():
...
+def update_resource_allocation_redis(active=True, verbose=False):
     """Updates redis with plots and information about current software.
 
     Parameters
     ----------
     active: boolean, optional
         Defaults to True. Should only be False when testing.
 
+    verbose: boolean, optional
+        Defaults to False. Prints status on what functin
```
antgonza, line 567: functin -> function

qiita_db/meta_util.py
```diff
@@ -555,96 +555,144 @@ def generate_plugin_releases():
...
+
     """
     time = datetime.now().strftime('%m-%d-%y')
-    scommands = get_software_commands(active)
+
+    # Retreave available col_name for commands
+    with qdb.sql_connection.TRN:
+        sql = 'SELECT col_name FROM qiita.resource_allocation_column_names;'
+        qdb.sql_connection.TRN.add(sql)
+        col_names_list = qdb.sql_connection.TRN.execute_fetchflatten()
```
antgonza, line 576: do not add to the variable name, the type ... this could be just col_names; however, I can see why you call it list ...

qiita_db/meta_util.py
```diff
@@ -555,96 +555,144 @@ def generate_plugin_releases():
...
-                    ("title_time", titles[1], r_client.set)
-                ]
-
-                for k, v, f in values:
-                    redis_key = 'resources$#%s$#%s$#%s$#%s:%s' % (
-                                cname, sname, version, col_name, k)
-                    r_client.delete(redis_key)
-                    f(redis_key, v)
+                # column_name_str looks like col1*col2*col3, etc
+                for col_name_str in col_name_list:
```
antgonza, line 615: col_name_str -> col_name to make it `for col_name in col_names`

qiita_db/meta_util.py
```diff
@@ -555,96 +555,144 @@ def generate_plugin_releases():
...
+                    new_column = None
+                    col_name_split = col_name_str.split('*')
+                    df_copy = df.dropna(subset=col_name_split)
+
+                    # Create a column with the desired columns
+                    for curr_column in col_name_split:
+                        if new_column is None:
+                            new_column = df_copy[curr_column]
+                        else:
+                            new_column *= df_copy[curr_column]
```
antgonza, line 625: I don't follow this line, is this multiplying the values?

antgonza (changes requested):
Great @Gossty, thank you for the updates. Some minor comments/questions.

qiita_db/meta_util.py
Gossty: If you have "samples" as input, it would just have 1 column. If you have "samples*columns" it would have samples*columns in the new column.

antgonza: Shouldn't this be an eval? I mean, what if the formula is "samples+columns" or "samples/columns"

Gossty: I'm not sure if eval would work here because to get the operation to work we need to have columns as indexes to dataframe:
 df_copy['samples'] +  df_copy['columns']

antgonza: OK, good to know; then let's leave it like that

Issue #3379, opened 2024-03-01
concurrency with deblur using `PLUGIN_LAUNCHER = qiita-plugin-launcher`


sjanssen2:
I seem to constantly running into an concurrency issue. I've set up a local Qiita instance on a Ubuntu virtual machine, based on the example nginx configuration, i.e. with one master and three workers, and installed all plugins necessary for 16S processing. 

Creating a test study with just 3 samples let me process up to the timming:
![image](https://github.com/qiita-spots/qiita/assets/11960616/cbd1b21b-2936-452b-bea5-512f87ff5bca)
However, deblur fails to finish. It executes the 4 deblur step with additional 4 SEPP steps but to my surprise, it seems to trigger two posts to `/qiita_db/jobs/JOBID/complete/` from two different workers:

```
==> qiita_21177.log <==
[I 240301 15:55:36 web:2161] 200 POST /qiita_db/jobs/00e919d2-b9b6-458d-bc72-3bfb3cb3bb89/complete/ (127.0.0.1) 68.18ms

==> qiita_21175.log <==
[I 240301 15:55:40 web:2161] 200 POST /qiita_db/jobs/00e919d2-b9b6-458d-bc72-3bfb3cb3bb89/complete/ (127.0.0.1) 603.42ms
```
The admin error table lists `Cannot change the step of a job whose status is not 'running', jobid=00e919d2-b9b6-458d-bc72-3bfb3cb3bb89`

This might be due to the fact, that deblur generats two artifacts?! Referring to the different processes that are triggered https://github.com/qiita-spots/qiita/blob/13eafdf1a2eb0c8b5d4019ad294b7490cd396139/scripts/qiita-recover-jobs#L144-L155 I see  that the jobid `00e919d2-b9b6-458d-bc72-3bfb3cb3bb89` belongs to the `validate` job of the reference-hit.biom artifact 

`SELECT * FROM qiita.processing_job WHERE processing_job_id = '00e919d2-b9b6-458d-bc72-3bfb3cb3bb89';`:
```
          processing_job_id           |     email      | command_id |                                                                                                                                                                                                                                                                                                                                                                                                   command_parameters                                                                                                                                                                                                                                                                                                                                                                                                    | processing_job_status_id | logging_id |         heartbeat          |                            step                             | pending | hidden | external_job_id 
[...]

PR #3465, opened 2025-03-11
Fix multiple validation jobs

sjanssen2:
Addressing #3379: 
With the proposed changes, both validation jobs of a deblur command successfully run through in my local setup. I suspect two mistakes of the current code:
1. maybe a copy&paste error of lines 1108ff as they seem to be identical to 1091ff
https://github.com/qiita-spots/qiita/blob/5ab3ebbca79e66441f2df2b73d879ec136450048/qiita_db/processing_job.py#L1091-L1114
2. even though the dependent jobs where "submitted" as child processes, they were not "registered" by Qiita's DB, such that Qiita did not "poll" for successful execution and thus could never set status to "success" of these job(s) even though the actual process did run through and finished with exit 0


antgonza:
Got it, thank you for the fix. Do you think it will be possible to add a retroactive test that brakes without the fix and passes now with it?

sjanssen2:
We would need to have a plugin that generates multiple artifacts as a test case. Adding the environment, registering all plugins to be "active", executing the workflow up to this point runs approx 15min on my laptop. Not sure if we want to invest that much compute in github actions.
Maybe we can create a mock plugin that just "touches" files for two ore more artifacts?!

antgonza:
That's right and after doing a quick search in the code base I think this is happening in `test_processing_job.py` `ProcessingJobTest.test_complete_multiple_outputs`. Maybe the only thing missing for the test is changing the value of qiita_config.job_scheduler_dependency_q_cn from 1 to 2; what do you think?

sjanssen2:
I think the problem with the existing test is that it manually sets the state of the dependent job here to "running": https://github.com/qiita-spots/qiita/blob/5ab3ebbca79e66441f2df2b73d879ec136450048/qiita_db/test/test_processing_job.py#L347 which in a running qiita instance should indirectly be done by the submit() function. But the code could be a primer?!

antgonza:
OK, thank you for checking. I agree with your assessment.

PR #3466, opened 2025-04-10
propagate error from resource_allocation_info

antgonza:


AmandaBirmingham (changes requested):
Hi, @antgonza -- If I am reading this right, the code here calls `resource_params = self.resource_allocation_info` *twice*, once on line 1052 outside the `try`, and once on line 1053 inside the `try`.  Is this intentional?  I can't figure out how it does what we want.

antgonza:
Good catch, thank you; forgot to remove the line above.

PR #3474, opened 2025-05-15
fix #3461

### LINKED ISSUE #3461: unused field in configuration file
I might miss something here, but from reading the code of https://github.com/qiita-spots/qiita/blob/fdad618cc6283430decc6511c0cc6b9b477c1fb6/qiita_core/configuration_manager.py#L21 I cannot spot where the field `UPLOAD_DATA_DIR` of the `[main]` section is ever used. Maybe you want to remove it?

https://github.com/qiita-spots/qiita/blob/fdad618cc6283430decc6511c0cc6b9b477c1fb6/qiita_core/support_files/config_test.cfg#L28-L30

also in the test

https://github.com/qiita-spots/qiita/blob/fdad618cc6283430decc6511c0cc6b9b477c1fb6/qiita_core/tests/test_configuration_manager.py#L309-L311

antgonza:


.github/workflows/qiita-ci.yml
```diff
@@ -169,17 +169,20 @@ jobs:
           conda deactivate
 
           echo "8. Setting up SSH"
-          ssh-keygen -t rsa -b 4096 -N '' -f $PWD/qiita_ware/test/test_data/test_key
+          ssh-keygen -t ed25519 -a 200 -N '' -f $PWD/qiita_ware/test/test_data/test_key
```
AmandaBirmingham, line 172: Curious: why was this switch necessary?

.github/workflows/qiita-ci.yml
```diff
@@ -169,17 +169,20 @@ jobs:
...
           mkdir ~/.ssh/
           cp $PWD/qiita_ware/test/test_data/test_key* ~/.ssh/
           cat ~/.ssh/test_key.pub > ~/.ssh/authorized_keys
+          cat ~/.ssh/test_key.pub > ~/.ssh/authorized_keys2
           chmod 600 $PWD/qiita_ware/test/test_data/test_key*
```
AmandaBirmingham, line 177: Where does `authorized_keys2` get used?

.github/workflows/qiita-ci.yml
```diff
@@ -169,17 +169,20 @@ jobs:
...
           chmod 600 ~/.ssh/*
           chmod 700 ~/.ssh/
           echo "Connecting as $USER@localhost"
           # this line (and the -o StrictHostKeyChecking=no) is so the server
           # is added to the list of known servers
-          scp -O -o StrictHostKeyChecking=no -i $PWD/qiita_ware/test/test_data/test_key $USER@localhost:/home/runner/work/qiita/qiita/qiita_ware/test/test_data/random_key /home/runner/work/qiita/qiita/qiita_ware/test/test_data/random_key_copy_1
+
+          # 05/22/25: commenting this line out as github actions is not allowing this step
+          # scp -O -o StrictHostKeyChecking=no -i $PWD/qiita_ware/test/test_data/test_key $USER@localhost:/home/runner/work/qiita/qiita/qiita_ware/test/test_data/random_key /home/runner/work/qiita/qiita/qiita_ware/test/test_data/random_key_copy_1
```
AmandaBirmingham, line 185: What is the effect of commenting this out?  I see the comment above that says "this line is so the server is added to the list of known servers", so I assume that now it isn't added to the list of known servers :) ... but what effect does that have on behavior?

qiita_ware/test/test_commands.py
```diff
@@ -56,26 +56,31 @@ def test_list_scp_wrong_key(self):
             list_remote('scp://runner@localhost:'+self.remote_dir_path,
                         self.test_wrong_key)
 
-    def test_list_scp_nonexist_key(self):
+    def test_download_remote_nonexist_key(self):
```
AmandaBirmingham, line 59: Not clear to me why this switched from a test of `list_remote` to a test of `download_remote` ... but if it is now for `download_remote`, might be nice to move it down next to the other test for `download_remote` (even though that one is currently commented out).

AmandaBirmingham (commented):
A couple of questions in line, and one larger one: does your investigation suggest that the ssh/scp functionality is going to come back to github actions, or is it gone for good?

.github/workflows/qiita-ci.yml
antgonza: Reading online they say that ed25519 vs rsa (just different encoding and in theory more secure) is necessary for github actions, which it wasn't in the past; thus decided to change it as part of the many trials to make it work. My thought was to leave it as this as it supposed to be the newer way

.github/workflows/qiita-ci.yml
antgonza: I think this should help: https://serverfault.com/a/116193. Similarly to the previous change, it's just to be sure that we have all the possible things to make this work.

.github/workflows/qiita-ci.yml
antgonza: This is the main culprit of the failings tests as it sets all the internal parameters for SCP to work by doing a single file scp; however, in all the builds it failed.

AmandaBirmingham (approved):
Thanks!

PR #3483, opened 2025-09-17
Tornado fetch file from central handler

sjanssen2:
This the accompanying PR to https://github.com/qiita-spots/qiita_client/pull/60

**In short:** I've extended fetch and push from only files to also directories. ~~In test mode, objects can also be deleted.~~ Depending on the route of API calls (through nginx =  8383 OR directly to tornado = 2117x), file delivery is handled differently to make us of nginx's much faster transfer capability

**More elaborate**:
1. Resetting the DB of qiita is done by triggering the `/apitest/reset/` endpoint (and also others). However, this only works properly if qiita master is addressed. Tests therefore do **not** speak against nginx (port 8383) but directly with the master instance (port 21174). However, when sending files, I relied on delivery through nginx. If requests go directly to master, the file content will be empty - and tests fail. With PR https://github.com/qiita-spots/qiita/pull/3483 the according endpoint checks if a request was made through nginx (and then uses it's fast mechanism to deliver the file) or directly to the tornado instance (which now can also send the file in a much slower fashion). Should be fast enough for testing.
2. Similarly, clients can now fetch directories. As I fear that misbehaving clients might trigger fetching the whole QIITA_BASE_DIR, I am limiting this to those directories that are "managed" by Qiita's DB as filetype directory.
3. ~~I realized that https://github.com/qiita-spots/qtp-sequencing/blob/4a325e7a75ca989eea9d9029acd02be2384a601e/qtp_sequencing/validate.py#L81 is directly deleting files in QIITA_BASE_DIR. Therefore, I added an according function to qiita_client and qiita, but only delete the file when using "filesystem" protocol OR qiita is run in test mode as I am too skeptical about providing an API endpoint to delete arbitrary files.~~


sjanssen2:
Hi @antgonza it took some time for me to come back to this PR. I feel we agree on removing the ability to delete files in qiita main is the way forward, as it currently would only have effects for the upload directory, but not on BASE_DATA_DIR.

I've updated the changes accordingly. If you find some time, can you please give this PR another review?
Thanks
Stefan

qiita_pet/handlers/cloud_handlers/tests/test_file_transfer_handlers.py
```diff
@@ -1,89 +1,266 @@
...
 from qiita_db.handlers.tests.oauthbase import OauthTestingBase
+from qiita_db.sql_connection import TRN
+from qiita_pet.handlers.cloud_handlers.file_transfer_handlers import \
+    is_directory
 
 
 class FetchFileFromCentralHandlerTests(OauthTestingBase):
     def setUp(self):
         super(FetchFileFromCentralHandlerTests, self).setUp()
+        self.endpoint = '/cloud/fetch_file_from_central/'
```
antgonza, line 17: Do we expect this endpoint to change? If no, then this is fine; if yes, maybe adding it to the configuration file would be better.

qiita_pet/handlers/cloud_handlers/file_transfer_handlers.py
```diff
@@ -1,11 +1,77 @@
...
+        #   c) whose mountpoint matches the provided parent_directory
+        sql = """SELECT filepath_id
+                FROM qiita.filepath
+                    JOIN qiita.filepath_type USING (filepath_type_id)
+                    JOIN qiita.data_directory USING (data_directory_id)
+                WHERE filepath_type IN ('directory', 'html_summary_dir') AND
+                    filepath=%s AND
+                    position(%s in mountpoint)>0;"""
+        qdb.sql_connection.TRN.add(sql, [dirname, mount_dirname])
+        hits = qdb.sql_connection.TRN.execute_fetchflatten()
```
antgonza, line 73: Do you think this should go to `qiita_db.util.py` as a method to find a specific `filepath_type`? Then you could call it with `directory` and `html_summary_dir`?

qiita_pet/handlers/cloud_handlers/file_transfer_handlers.py
```diff
@@ -29,34 +95,103 @@ def get(self, requested_filepath):
...
-        self.set_header("Expires", "0")
-        self.set_header("Cache-Control", "no-cache")
-        self.set_header(
-            "Content-Disposition",
-            "attachment; filename=%s" % os.path.basename(protected_filepath),
-        )
+            raise HTTPError(403, reason=(
+                "The requested file is not present in Qiita's BASE_DATA_DIR!"))
+
+        filename_directory = "qiita-main-data.zip"
```
antgonza, line 106: What's this? FWIW, remember that mod_zip does _not_ reduce the size of the files but creates a zip without compression; see: https://github.com/qiita-spots/qiita/issues/2920

antgonza (commented):
Thank you @sjanssen2. A few comments and suggestions.

qiita_pet/handlers/cloud_handlers/tests/test_file_transfer_handlers.py
sjanssen2: I don't foresee this to change. It is for testing only. I am not sure if there is a clean way to extract the endpoint URL from here: https://github.com/jlab/qiita/blob/e0e7480a1b04740c24ac550574b636d7693f1c85/qiita_pet/handlers/cloud_handlers/__init__.py#L9 Therefore, I suggest to keep as is

qiita_pet/handlers/cloud_handlers/file_transfer_handlers.py
sjanssen2: I like the idea to move this function into qiita_db.util.py but am not sure if it aids understanding if we split functionality:

The lines you marked are for the SQL statement, but they require parameters `dirname` and `mount_dirname`, which I compute in the lines above. They depend on classical filepath operations, which are probably more complicated to realize as SQL statements?! How do you feel about moving the whole function, i.e lines 16 to 74, into qiita_db.util.py and make the `filepath_type`  a second parameter besides `filepath`?

Or, to avoid two calls to the DB: make filepath_type a list of str or a set to ask for logical OR between these types?

qiita_pet/handlers/cloud_handlers/file_transfer_handlers.py
sjanssen2: If I understand POST correctly, you can only transfer **one** file per request. If a user request a whole directory (e.g. for job-output-folder or summarize) we need to transfer **multiple** files.

If qiita operated behind nginx, I can re-use existing functions, i.e. `BaseHandlerDownload._write_nginx_file_list`, but for testing, qiita's tornado server is directly addressed. In this case, I need to "fuse" individual files into one Archive, which is then transferred to the client. 

I assume most files in qiita main are already compressed (biom, fastq.gz, ...) therefore an additional compression would incur much compute but probably gains not much - and it is mainly for testing. Thus, no further compression level should be fine here.

antgonza: Sounds reasonable, thank you for the explanation. FWIW that's correct and why is fine (not good or great) to use mod_zip.

qiita_pet/handlers/cloud_handlers/file_transfer_handlers.py
antgonza: I see, thank you for the explanation. I think moving the full block/method to `util.py` makes sense. Now, after a little more thought, this is basically defining if a `filepath` is a dir or not but a side consequence is that is actually checking if that `filepath` exists in the database; however, is not checking if the data is actually available for that user or cloud environment; right? I mean this assumes full access to everything in `central`, correct? If this is correct, I think we should limit this to only public data? Also, maybe splitting the functionality in: `def get_filepath_id_from_filepath(filepath)` which will validate the filtepath and get the filepath id from the database, then we can use that to validate access (public, private, sandbox); what do you think?

sjanssen2: The function checks if a filepath (that is send from the DB to a qiita plugin) is actually managed as a directory in the DB (and thus is a collection of files and subdirs) or is a regular single file. Since the DB does NOT know which files and subdirs the folder contains, it is hard to check if they are actually physically present :-)

Users of this function are not end-user through their browser but uncoupled q*p-xxx plugins. These plugins should only be able to access the endpoint of qiita central if `client_secret` and `client_id` is in qiita central's DB (via `@authenticate_oauth` decorators).

But you are pointing to a very important aspect! Should plugins have full access to ALL qiita central data or only to parts of it. If I get your architecture right, currently "trusted (=oauth registered)" plugins always have unlimited access :-/ Can we change this in the future? Maybe attaching a user cookie to the plugin job request? But how to then compute batch jobs?!

antgonza: I see, FWIW, plugins only work via `ProcessingJob` - a plugin can not be executed outside that object - and they are linked to a user. A user can only create a new job on a given artifact if they have access to that artifact. Thus, I think we should be fine; no?

sjanssen2: That is true from the view of Qiita central. But, in principle, if plugin code changes, I don't see why the once submitted plugin should not be able to access arbitrary artifact data via e.g. https://github.com/qiita-spots/qp-deblur/blob/80467b7c76b6177abb577e19c83266ee54dbcd79/qp_deblur/deblur.py#L424 and simply iterating the `artifact_id` variable :-/

But this should only hurt, once we distribute compute through a cloud with plugins that we as developers not fully trust. So for now, we are fine.

sjanssen2:
Hi @antgonza, is there anything left to discuss in this PR? Would be amazing to get merged :-)

antgonza:
Done, sorry for the delay.

